PR #23, opened 2023-08-29
18: front end setup

BCerki:
card: https://app.zenhub.com/workspaces/climate-action-secretariat-60ca4121764d710011481ca2/issues/gh/bcgov/cas-registration/18


.tool-versions
```diff
@@ -0,0 +1 @@
+yarn 1.22.19
```
andrea-williams, line 1: also need to set a node version here

andrea-williams (approved):

PR #121, opened 2023-09-19
Chore: 26 facilities components

BCerki:
Front end components to support https://app.zenhub.com/workspaces/team-moose-sprint-board-60d216175a2689000ec774d0/issues/gh/bcgov/cas-registration/26 and other frontend work

I put the components on the page with the counter so reviewers can see how they look, but I think we'll probably end up moving them around later


client/app/layout.tsx
```diff
@@ -5,16 +5,22 @@ import { Nav } from "./components/Nav";
...
 
 export default function RootLayout(props: React.PropsWithChildren) {
   return (
     <ReduxProvider>
       <html lang="en">
         <body>
           <section className={styles.container}>
-            <Nav />
+            <Navigation user="Fred Penner" />
+            <BCGovTypography />
```
BCerki, line 20: This was inconsistent on my machine--sometimes the Navigation component would show up with the proper formatting, sometimes not

BCerki:
No longer using Button components, closing. Header/footer will be part of another card

PR #124, opened 2023-09-20
chore: 26 create fixtures

BCerki:
Fixture creation to support https://app.zenhub.com/workspaces/team-moose-sprint-board-60d216175a2689000ec774d0/issues/gh/bcgov/cas-registration/26

### DIFF
--- a/bc_obps/registration/fixtures/operator.json
+++ b/bc_obps/registration/fixtures/operator.json
@@ -11,8 +11,26 @@
       "business_structure": "ladedah",
       "mailing_address": "ladedah",
       "bceid": "ladedah",
-      "parent_operator": 1,
+      "parent_operator": 2,
       "relationship_with_parent_operator": "ladedah",
+      "compliance_obligee": 2,
+      "date_aso_became_responsible_for_operator": "2023-09-20T15:26:43.284Z"
+    }
+  },
+  {
+    "model": "registration.operator",
+    "pk": 1,
+    "fields": {
+      "id": 2,
+      "legal_name": "zipadeedoodah",
+      "trade_name": "zipadeedoodah",
+      "cra_business_number": "zipadeedoodah",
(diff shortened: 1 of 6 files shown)


bc_obps/registration/fixtures/operator.json
```diff
@@ -0,0 +1,20 @@
...
+    "fields": {
+      "id": 1,
+      "legal_name": "ladedah",
+      "trade_name": "ladedah",
+      "cra_business_number": "ladedah",
+      "bc_corporate_registry_number": "ladedah",
+      "business_structure": "ladedah",
+      "mailing_address": "ladedah",
+      "bceid": "ladedah",
+      "parent_operator": 1,
```
Sepehr-Sobhani, line 14: I know this is dummy data, but it would make more sense not to set this instance as its own parent and compliance obligee.

bc_obps/registration/fixtures/user.json
Sepehr-Sobhani, line 1: Should we include the `UserAndContactCommonInfo` fields in the user object? They appear to be non-nullable.

Sepehr-Sobhani (changes requested):
This looks great!👍 Only a few minor changes need to be made.

bc_obps/registration/fixtures/user.json
BCerki: True... I'm surprised I didn't get an error. Added now

Sepehr-Sobhani: I think we don't get errors when loading fixtures:https://docs.djangoproject.com/en/3.2/ref/django-admin/#loaddata:~:text=When%20fixture%20files%20are%20processed%2C%20the%20data%20is%20saved%20to%20the%20database%20as%20is.%20Model%20defined%20save()%20methods%20are%20not%20called%2C

Sepehr-Sobhani (approved):

Issue #90, opened 2023-09-08
Write Tests for Django Models


BCerki:
#### Describe the task

This task is to write tests for the existing Django models that include:
- all columns exist
- create record??
- suggest pytest

#### Acceptance Criteria

- [ ] first
- [ ] second
- [ ] third

#### Additional context

- Add any other context about the task here.
- Or here

Sepehr-Sobhani:
Might be more convenient to use Django specific plugin for this: https://pytest-django.readthedocs.io/en/latest/index.html#

PR #145, opened 2023-09-27
Test/90 model tests

joshgamache:
Addresses #90 [Zenhub](https://app.zenhub.com/workspaces/team-moose-sprint-board-60d216175a2689000ec774d0/issues/gh/bcgov/cas-registration/90). Adds Automated testing of the Django models. And further suggested tests would be appreciated!

## Changes 🚧

- Added tests for all existing models 

## To test 🔬

1. Navigate to the `~/bc_obps` directory, and ensure you are in the Poetry shell. Start Postgres with `make start_pg`. Follow the [developer environment setup](https://github.com/bcgov/cas-registration/blob/develop/docs/developer-environment-setup.md#backend-environment-setup) if you need to create the DB locally.
2. Run `python manage.py test`. It should find and run 106 tests. 

### Notes 📝

These tests might not be complete. Following the paradigms outlined by the [Django Project](https://docs.djangoproject.com/en/4.2/topics/testing/overview/) and  [Mozilla](https://developer.mozilla.org/en-US/docs/Learn/Server-side/Django/Testing), I've included tests for code defined by us rather than Django. These tests will need to be extended if methods are ever added to the model classes.

### DIFF
--- a/bc_obps/registration/tests.py
+++ b/bc_obps/registration/tests.py
@@ -1,3 +1,69 @@
 from django.test import TestCase
 
-# Create your tests here.
+from .models import User
+
+
+class UserModelTest(TestCase):
+    @classmethod
+    def setUpTestData(cls):
+        # Set up non-modified objects used by all test methods
+        User.objects.create(
+            user_guid="00000000-0000-0000-0000-000000000000",
+            first_name="fname-test",
+            last_name="lname-test",
+            mailing_address="12 34 Street, Test",
+            email="test@example.com",
+            phone_number="12345678900",
+            business_guid="11111111-1111-1111-1111-111111111111",
+            position_title="test",
+        )
+
+    def test_first_name_label(self):
+        testUser = User.objects.get(user_guid="00000000-0000-0000-0000-000000000000")
+        field_label = testUser._meta.get_field("first_name").verbose_name
+        self.assertEqual(field_label, "first name")
+
+    def test_first_name_max_length(self):
+        testUser = User.objects.get(user_guid="00000000-0000-0000-0000-000000000000")
+        max_length = testUser._meta.get_field("first_name").max_length
+        self.assertEqual(max_length, 1000)
+
+    def test_last_name_label(self):
+        testUser = User.objects.get(user_guid="00000000-0000-0000-0000-000000000000")
+        field_label = testUser._meta.get_field("last_name").verbose_name
+        self.assertEqual(field_label, "last name")
+
+    def test_last_name_max_length(self):
+        testUser = User.objects.get(user_guid="00000000-0000-0000-0000-000000000000")
+        max_length = testUser._meta.get_field("last_name").max_length
+        self.assertEqual(max_length, 1000)
+
+    def test_mailing_address_label(self):
+        testUser = User.objects.get(user_guid="00000000-0000-0000-0000-000000000000")
+        field_label = testUser._meta.get_field("mailing_address").verbose_name
+        self.assertEqual(field_label, "mailing address")
+
+    def test_mailing_address_max_length(self):
+        testUser = User.objects.get(user_guid="00000000-0000-0000-0000-000000000000")
+        max_length = testUser._meta.get_field("mailing_address").max_length
+        self.assertEqual(max_length, 1000)
+
+    def test_email_label(self):
+        testUser = User.objects.get(user_guid="00000000-0000-0000-0000-000000000000")
+        field_label = testUser._meta.get_field("email").verbose_name
+        self.assertEqual(field_label, "email")
+
+    def test_email_max_length(self):
+        testUser = User.objects.get(user_guid="00000000-0000-0000-0000-000000000000")
+        max_length = testUser._meta.get_field("email").max_length
+        self.assertEqual(max_length, 254)
+
+    def test_phone_number_label(self):
+        testUser = User.objects.get(user_guid="00000000-0000-0000-0000-000000000000")
+        field_label = testUser._meta.get_field("phone_number").verbose_name
+        self.assertEqual(field_label, "phone number")
+
+    def test_position_title_label(self):
+        testUser = User.objects.get(user_guid="00000000-0000-0000-0000-000000000000")
+        field_label = testUser._meta.get_field("position_title").verbose_name
+        self.assertEqual(field_label, "position title")
--- a/bc_obps/registration/tests.py
+++ b/bc_obps/registration/tests.py
@@ -1,6 +1,6 @@
 from django.test import TestCase
 
-from .models import User
+from .models import User, NaicsCode
 
 
 class UserModelTest(TestCase):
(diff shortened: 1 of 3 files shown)


shon-button:
Hi Josh,

Test steps worked for me and ran 106 tests.
Related to the [developer environment setup](https://github.com/bcgov/cas-registration/blob/develop/docs/developer-environment-setup.md#backend-environment-setup); perhaps, you might update the setup instructions to include the steps I had to complete to accomplish "Backend Environment Setup" (suggestion below).
Also, I suspect, adding your test step would also be valuable.

**developer environment setup suggestion**

## Backend Environment Setup

1. Navigate to folder: `cas-registration/bc_obps`.
2. Copy/Paste the `.env.example` file and rename it `.env`.
3. Complete the .env file values reflecting the 1Password vault document `OBPS backend ENV`.
4. Download 1Password vault file `OBPS GCS json` and store it in folder location detailed in .env\GOOGLE_APPLICATION_CREDENTIALS.
5. From the terminal, cd into directory `cas-registration/bc_obps`.
6. Run `make install_dev_tools`. This will install asdf plugins, poetry and activate the poetry virtual environment (to get into the environment again after setup, run `poetry shell`).
7. Run `make install_poetry_deps` to install all python dependencies.
8. Run `make start_pg` to start the postgres server if it is not already running.
9. Run `make create_db` to create the database.
10. Run `make migrate` to run all database migrations.
11. Run `make run`, which will start running the development server locally (default port is :8000; terminal output will indicate what localhost address to use to access the backend. server).
   - to test it out, navigate to the `/api/docs` endpoint in your browser, you should see documentation for the /add endpoint
   - navigate to the `api/add?a=4&b=2` endpoint in your browser, which should return as a result the sum of the specified values for a and b.
12. Optional: to test the Django server's connection to your database, run `python3 manage.py check --database default`.
13. Optional: to load mock data via fixtures, run `make loadfixtures`.

bc_obps/registration/tests.py
Sepehr-Sobhani, line 1: Overall looks great! I have some suggestions to improve it and make sure it follows the [Python style guide](https://peps.python.org/pep-0008/) To start, consider using snake_case variable names 🐍. Also, try using the [`subTest` context manager](https://docs.python.org/3/library/unittest.html#distinguishing-test-iterations-using-subtests), which is shown in the code below. This will provide more information if an assertion fails, making it easier to identify and fix issues. Finally, I made the first test (as an example) more concise, but you can expand it into other tests. Thanks for your hard work!

```python
class UserModelTest(TestCase):
    @classmethod
    def setUpTestData(cls):
        cls.test_user = User.objects.create(
            first_name="fname-test",
            last_name="lname-test",
            mailing_address="12 34 Street, Test",
            email="test@example.com",
            user_guid="00000000-0000-0000-0000-000000000000",
            business_guid="11111111-1111-1111-1111-111111111111",
            position_title="test",
        )

        cls.test_user.documents.set(
            [
                Document.objects.create(file="test1.tst", document_type="test", description="test"),
                Document.objects.create(file="test2.tst", document_type="test", description="test"),
            ]
        )

    def assertFieldLabel(self, field_name, expected_label):
        field = self.test_user._meta.get_field(field_name)
        self.assertEqual(field.verbose_name, expected_label)

    def assertFieldMaxLength(self, field_name, expected_max_length):
        field = self.test_user._meta.get_field(field_name)
        self.assertEqual(field.max_length, expected_max_length)

    def test_field_labels_and_max_lengths(self):
        field_data = [
            ("first_name", "first name", 1000),
            ("last_name", "last name", 1000),
            ("mailing_address", "mailing address", 1000),
            ("email", "email", 254),
            ("phone_number", "phone number", None),  # Replace None with the actual max length if available
            ("user_guid", "user guid", None),
            ("business_guid", "business guid", None),
            ("position_title", "position title", 1000),
            ("documents", "documents", None),
        ]

        for field_name, expected_label, expected_max_length in field_data:
            with self.subTest(field_name=field_name):
                if expected_label:
                    self.assertFieldLabel(field_name, expected_label)
                if expected_max_length is not None:
                    self.assertFieldMaxLength(field_name, expected_max_length)

    def test_user_has_multiple_documents(self):
        self.assertEqual(self.test_user.documents.count(), 2)
```

Sepehr-Sobhani (changes requested):

bc_obps/registration/tests.py
joshgamache: Thanks for the links and the feedback Sepehr. I've addressed it and updated with c1a470f!

Sepehr-Sobhani, line 1: This looks great, Thank you! After most recent database changes, I think you might need to update your test cause I'm getting an error when running tests:
```
TypeError: Field 'id' expected a number but got <Operator: Operator object (1)>
```
Also, I'd like to suggest two more modifications:
- We can make this code more DRY by having a base test class and inheriting that class in other classes to minimize the need to repeat those field assertion functions, something like:
```python
class BaseTestCase(TestCase):
    def assertFieldLabel(self, instance, field_name, expected_label):
        field = instance._meta.get_field(field_name)
        self.assertEqual(field.verbose_name, expected_label)

    def assertFieldMaxLength(self, instance, field_name, expected_max_length):
        field = instance._meta.get_field(field_name)
        self.assertEqual(field.max_length, expected_max_length)
```
and then using that class in other test cases:
```python
class UserModelTest(BaseTestCase):
...

class DocumentModelTest(BaseTestCase):
...
```
This way we only need to maintain `assertFieldLabel` and `assertFieldMaxLength` functions in one place.

- I would also suggest having two tests regarding unique constraints, something like this:
```python
def test_unique_user_guid_and_business_guid_constraint(self):
        # Attempt to create two users with the same user_guid and business_guid
        user1 = User.objects.create(
            first_name="fname-test1",
            last_name="lname-test1",
            mailing_address="12 34 Street, Test1",
            email="test1@example.com",
            user_guid="00000000-0000-0000-0000-000000000000",
            business_guid="11111111-1111-1111-1111-111111111111",
            position_title="test1",
        )

        # Attempt to create another user with the same user_guid and business_guid
        user2 = User(
            first_name="fname-test2",
            last_name="lname-test2",
            mailing_address="12 34 Street, Test2",
            email="test2@example.com",
            user_guid="00000000-0000-0000-0000-000000000000",
            business_guid="11111111-1111-1111-1111-111111111111",
            position_title="test2",
        )

        with self.assertRaises(IntegrityError):
            user2.save()
```
and
```python
def test_unique_email_constraint(self):
        # Attempt to create two contacts with the same email address
        contact1 = Contact.objects.create(
            first_name="fname-test1",
            last_name="lname-test1",
            mailing_address="12 34 Street, Test1",
            email="test@example.com",
            phone_number="12345678900",
            is_operational_representative=False,
        )

        # Attempt to create another contact with the same email address
        contact2 = Contact(
            first_name="fname-test2",
            last_name="lname-test2",
            mailing_address="12 34 Street, Test2",
            email="test@example.com",
            phone_number="9876543210",
            is_operational_representative=True,
        )

        with self.assertRaises(IntegrityError):
            contact2.save()
```

joshgamache: Great suggestions!
- Addressed the migrated `operator_id` ➡️ `operator` field in one of the models to cause tests to pass with 5648305e.
- Refactored methods into a reusable base class with 6605b1eb.
- Uniqueness tests added in 04507ab1.

Thanks for the feedback, definitely improves the test structure even more. Should help future test writing be much cleaner (these refactoring passes dropped the line count by something like 60% 🎉!).

Sepehr-Sobhani: Excellent work! Thank you for applying those suggestions. 🙏

Sepehr-Sobhani (approved):

PR #146, opened 2023-09-27
chore: add k8s liveness and readiness middleware

dleard:


bc_obps/registration/kubernetesMiddleware.py
Sepehr-Sobhani, line 1: nit: I suggest putting this middleware into a folder called middleware to have a place for all middleware (if any down the road) and use the Python folder naming convention for this module.(https://peps.python.org/pep-0008/#package-and-module-names)
Also, I've written a unit test to make sure this middleware works as intended:
```python
from django.http import HttpRequest
from django.test import TestCase
from .middleware.kubernetes_middleware import KubernetesHealthCheckMiddleware


class MiddlewareTestCase(TestCase):
    def test_liveness_endpoint(self):
        middleware = KubernetesHealthCheckMiddleware(None)
        request = HttpRequest()
        request.method = "GET"
        request.path = "/liveness"
        response = middleware(request)

        self.assertEqual(response.status_code, 200)
        self.assertEqual(response.content, b'OK. Server is running.')

    def test_readiness_endpoint(self):
        middleware = KubernetesHealthCheckMiddleware(None)
        request = HttpRequest()
        request.method = "GET"
        request.path = "/readiness"
        response = middleware(request)

        self.assertEqual(response.status_code, 200)
        self.assertEqual(response.content, b'OK. Server is ready.')
`

File: bc_obps/bc_obps/settings.py [hunk 1]
```diff
@@ -47,6 +47,7 @@
 ]
 
 MIDDLEWARE = [
+    'registration.kubernetesMiddleware.KubernetesHealthCheckMiddleware',
```
Sepehr-Sobhani, line 50: We might also need to change this if we decide to change the name of the module and add a middleware folder.

Sepehr-Sobhani (changes requested):
Looks great! just some nits :)

bc_obps/registration/kubernetesMiddleware.py
dleard: Would this go in tests.py?

dleard: Created #156 to handle test directory structure & adding this test

Sepehr-Sobhani (approved):

Issue #139, opened 2023-09-26
Test the operation forms


BCerki:
###Description of the Tech Debt###

Typically, jest and e2e tests are included in the DOD for user stories. However, the add/edit user story (#26) was getting so large that we decided to split out testing. This involves:
- jest tests (but not an operations form one because the fields will be changed in https://app.zenhub.com/workspaces/team-moose-sprint-board-60d216175a2689000ec774d0/issues/gh/bcgov/cas-registration/200)
- e2e playwright tests
- python tests

###Tech Debt Triage#
The purpose of our technical debt triage process is to analyze technical debt to determine risk level of the technical debt and the value in tackling that technical debt.

Risk Value Scoring:

| Level  | Value |
| ------ | ----- |
| High   | 3     |
| Medium | 2     |
| Low    | 1     |

| Technical Debt - Risk Types                                                                                                   | Level | Value |
| ----------------------------------------------------------------------------------------------------------------------------- | ----- | ----- |
| Business Area Risk - Risk of business area visibility / damage to user experience                                             |       |       |
| Developer Fault Risk - How likely will this tech debt cause a future error related to coding on top of it                     |       |       |
| System Fault Risk - Risk of system errors or application downtime                                                             |       |       |
| Time Scale Risk - Compound risk effect if left alone. How much more difficult to fix or dangerous will this become over time? |       |       |
| Time Sink Risk - How much will this tech debt slow the development process down                                               |       |       |
| TOTAL SCORE:                                                                                                                  |       |       |

PR #199, opened 2023-10-16
Test: create operations e2e test

BCerki:
Part of #139. See readme for test-running instructions

I don't think the sonarcloud fail is a problem since the test endpoint will never exist anywhere but dev, but if reviewers disagree just let me know!


shon-button:
(WIP)
image

bc_obps/registration/api.py
```diff
@@ -19,10 +19,32 @@
...
+from django.http import HttpResponse
+import os
 
 router = Router()
 
+# testing endpoint
+@router.get("/test-setup")
+def setup(request):
+    print(settings.ENVIRONMENT)
+    print(type(settings.ENVIRONMENT))
```
andrea-williams, line 37: do we still need these print statements? It's not a big deal now, but we'll want to make sure they're removed before we go to production

client/e2e/operations.spec.ts
```diff
@@ -0,0 +1,101 @@
+import { test, expect } from "@playwright/test";
+
+// Annotate entire file as serial.
+test.describe.configure({ mode: "serial" });
+
+test.beforeEach(async ({ page }, testInfo) => {
+  // eslint-disable-next-line no-console
+  console.log(`Running ${testInfo.title}`);
+  // reset db and confirm there are no errors setting up the test data
+  await page.goto("http://localhost:8000/api/registration/test-setup");
```
andrea-williams, line 10: I think we'll need to parameterize these URLs in order for them to work in Openshift environments...?

BCerki: ... maybe? Do we normally run e2e tests in Openshift, or just locally/in github actions?

andrea-williams: fair point. I don't know if we do on CAS (haven't been paying attention). I assumed so but that's entirely baseless. Any of the devopsy people (e.g., @dleard  , @Sepehr-Sobhani  or @mikevespi) able to shed light on this?

mikevespi (approved):
👍

client/e2e/operations.spec.ts
dleard: only in gh actions

PR #208, opened 2023-10-20
Chore/backend tests

BCerki:
Part of https://app.zenhub.com/workspaces/team-moose-sprint-board-60d216175a2689000ec774d0/issues/gh/bcgov/cas-registration/139

To run the python tests, see the readme instructions

I made a card to adjust the thresholds for sonar's scan for test files: https://app.zenhub.com/workspaces/team-moose-sprint-board-60d216175a2689000ec774d0/issues/gh/bcgov/cas-registration/212


bc_obps/registration/api.py
```diff
@@ -101,6 +121,7 @@ def get_operation(request, operation_id: int):
 
 @router.post("/operations")
 def create_operation(request, payload: OperationIn):
+    print(payload)
```
andrea-williams, line 124: delete (I know I inserted this - sorry!)

bc_obps/registration/api.py
```diff
@@ -58,12 +57,33 @@ class Config:
         model_fields = "__all__"
 
 
-class OperationIn(OperationSchema):
-    # temporarily setting a default operator since we don't have login yet
-    operator: int = 1
-    # Converting types
-    start_of_commercial_operation: date = None
-    verified_at: date = None
+# class OperationIn(OperationSchema):
```
andrea-williams, line 60: I think this block can be deleted...?

bc_obps/registration/tests/test_endpoints.py
```diff
@@ -0,0 +1,187 @@
...
+    endpoint = base_endpoint + 'operations/1'
+
+    def test_put_nonexistant_operation(self, client):
+        response = client.get(self.endpoint)
+        assert response.status_code == 404
+
+    def test_get_method_for_200_status(self, client):
+        naics_code = baker.make(NaicsCode)
+        naics_category = baker.make(NaicsCategory)
+        operator = baker.make(Operator)
```
andrea-williams, line 115: this works now?! 🤯

bc_obps/registration/tests/test_endpoints.py
```diff
@@ -0,0 +1,187 @@
...
+            physical_province='BC',
+            physical_postal_code='V1V 1V1',
+            legal_land_description='It\'s legal',
+            latitude=90,
+            longitude=-120,
+            petrinex_ids=['123', '124'],
+            regulated_products=[1, 2],
+            documents=[document.id],
+            contacts=[contact.id],
+            status='Pending',
```
andrea-williams, line 92: I think that the status field should not be part of the OperationIn class - it introduces the opportunity for an applicant to set the status to 'Approved' automatically, which we don't want.

Instead we should submit the mock_operation body with no set status, then once it's been added, fetch that same operation with the get-by-id operation endpoint, and check then that the returned operation now has a status of Pending

(This will also affect your unit tests for the PUT request)

andrea-williams (commented):
Nearly there - I'm requesting one change re: the `status` field in the `OperationIn` class. And then there's some junk to delete that I created 😈

bc_obps/registration/tests/test_endpoints.py
BCerki: The put request doesn't return the full operation, just name, so I didn't add it there (but maybe put should be returning everything?)

bc_obps/registration/tests/test_endpoints.py
BCerki: once we removed compliance entity it started working, yes

bc_obps/registration/tests/test_endpoints.py
andrea-williams: oh I just meant removing the status field from the OperationIn object used for the put test. You're correct that we may need to change the response sent from the PUT request later, but until we know for sure, let's leave it as is for now :)

andrea-williams (approved):
LGTM! 🚀

PR #213, opened 2023-10-24
chore: 162 - install release-it & configure

dleard:


docs/release.md
```diff
@@ -0,0 +1,33 @@
+## Release Process
+
+Before releasing our application to our `test` and `prod` environments, an essential step is to add a tag to our sqitch plan, to identify which database mutations are released to prod and should be immutable.
```
BCerki, line 3: cas-registration doesn't have a sqitch plan

also, once this PR is in, do we have to be careful about axing django migrations on every new PR?

BCerki (commented):
In my unqualified devops opinion, this looks good except for a reference to sqitch in the readme

docs/release.md
dleard: we run release-it manually. Once we've created a release, if we change any previous migrations the change won't be applied (because django knows it already applied the migration). We can probably come up with something to automate an alert if something gets changed in a file that's already been migrated.

Sepehr-Sobhani (approved):

PR #250, opened 2023-11-02
Feat: 103 Create Operator

BCerki:
Card: https://app.zenhub.com/workspaces/team-moose-sprint-board-60d216175a2689000ec774d0/issues/gh/bcgov/cas-registration/103

Work completed so far:
- create page


shon-button:
Re:
```
create page for adding an operation
create app/components/routes/select-operator/form/AddOperator.tsx (I don't remember why we do this in addition to a page, ask Shon)
```

The page can contain all the code...
If the page has await data, a strategy for better time to display is to, display static text and nest the await the async content using <Suspense> with loading fallback.
In this scenario I might opt **not** to have `app/components/routes/select-operator/form/AddOperator.tsx` and put its code directly into page:

e.g.
client/app/(authenticated)/dashboard/select-operator/(request-access)/create/page.tsx

```
import UserOperatorForm from "@/app/components/form/UserOperatorForm";
import { userOperatorSchema } from "@/app/utils/jsonSchema/userOperator";

export default  function Page() {
  return (
    <section className="text-center my-20 text-2xl flex flex-col gap-3">
      <UserOperatorForm schema={userOperatorSchema} />
    </section>
  );
}
```

Sepehr-Sobhani:
@BCerki To keep you informed about this PR, in case you want to work on it while I'm away on Monday.

### Done:
- Refactored endpoints
- Added a new endpoint to handle the workflow in this feature
- Added tests for some existing endpoints
- Added test for utils
- Moved Schemas into separate files
- Made some small tweaks here and there

The whole feature is working now.

### TODO:
- Add tests for all `select-operator` endpoints.
- Resolve conflicts with Develop

andrea-williams:
The "Business Structure" field is a freeform text box - I would've expected a dropdown?
Screenshot 2023-11-22 at 2 08 01 PM

andrea-williams (changes requested):
See attached screenshot re: freeform text box for Business Structure field. I recommend fixing this issue in this PR.

The other issue I encountered is that if I try to Submit the form when I have 1 or more invalid data entries (e.g., my BC Corporate Registry number has letters in it), then at the bottom of the form the fields show up for the Operator's mailing address, even when the "Is the mailing address the same as the physical address?" radio button is selected "Yes". I then can't submit until I've filled out duplicate address info. (I also tried clicking the "No" button and then switching back to "Yes" - that didn't fix it either). Depending on how complicated the fix for this is (is it an RJSF issue?) I'm fine if you choose to not fix this in this PR and just create a bug ticket instead.

Sepehr-Sobhani:
> See attached screenshot re: freeform text box for Business Structure field. I recommend fixing this issue in this PR.
> 
> The other issue I encountered is that if I try to Submit the form when I have 1 or more invalid data entries (e.g., my BC Corporate Registry number has letters in it), then at the bottom of the form the fields show up for the Operator's mailing address, even when the "Is the mailing address the same as the physical address?" radio button is selected "Yes". I then can't submit until I've filled out duplicate address info. (I also tried clicking the "No" button and then switching back to "Yes" - that didn't fix it either). Depending on how complicated the fix for this is (is it an RJSF issue?) I'm fine if you choose to not fix this in this PR and just create a bug ticket instead.

Thank you for giving your feedback on this Pull Request, @andrea-williams ! I have a couple of questions in relation to it. First, do we have any predefined value for the `Business Structure` field? I noticed that it is a `Charfield` in the database without any initial value or enum. Second, with regards to the `RJSF` issue, I will add it to ticket `299` if that is alright with you. We plan to change the schema and make the form multistep on that ticket.

PR #264, opened 2023-11-08
Feat: approve prime admin request

JoshLarouche:
[zenhub](https://app.zenhub.com/workspaces/team-moose-sprint-board-60d216175a2689000ec774d0/issues/gh/bcgov/cas-registration/25)

### DIFF
--- a/bc_obps/registration/api/user_operator.py
+++ b/bc_obps/registration/api/user_operator.py
@@ -94,6 +94,7 @@ def list_user_operators(request):
         operator_related_fields_dict = model_to_dict(
             operator,
             fields=[
+                "status",
                 "legal_name",
             ],
         )
--- a/bc_obps/registration/api/user_operator.py
+++ b/bc_obps/registration/api/user_operator.py
@@ -60,22 +60,12 @@ def get_user_operator(request, user_operator_id: int):
     }
 
 
-# @router.get("/user-operators", response=List[UserOperatorOut])
-# def list_operations(request):
-#     qs = UserOperator.objects.all()
-#     print(qs)
-#     return qs
-
-
 @router.get("/user-operators", response=List[UserOperatorListOut])
 def list_user_operators(request):
     qs = UserOperator.objects.all()
-    print(qs)
     user_operator_list = []
 
     for user_operator in qs:
-        # user_operator = get_object_or_404(UserOperator)
-        print("user_operator: ", user_operator)
         user_operator_related_fields_dict = model_to_dict(
             user_operator,
             fields=[
@@ -83,7 +73,6 @@ def list_user_operators(request):
                 "status",
             ],
         )
-        print("user_operator_related_fields_dict: ", user_operator_related_fields_dict)
         user = user_operator.user
         user_related_fields_dict = model_to_dict(
             user,
@@ -93,7 +82,6 @@ def list_user_operators(request):
                 "email",
             ],
         )
-        print("user_related_fields_dict: ", user_related_fields_dict)
         operator = user_operator.operator
         operator_related_fields_dict = model_to_dict(
             operator,
@@ -101,8 +89,6 @@ def list_user_operators(request):
                 "legal_name",
             ],
         )
-        print("operator_related_fields_dict: ", operator_related_fields_dict)
-        print("type: ", type(operator_related_fields_dict))  # dict or json
 
         user_operator_list.append(
             {
@@ -111,8 +97,6 @@ def list_user_operators(request):
                 **operator_related_fields_dict,
             }
         )
-        print("user_operator_list: ", user_operator_list)
-        print("length: ", len(user_operator_list))
 
     return user_operator_list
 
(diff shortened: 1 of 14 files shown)


client/app/data/dashboard/admin.json
```diff
@@ -10,7 +10,7 @@
   {
     "title": "User Access Management",
     "content": "Slightly more to say than what Card A has to say",
-    "links": [{ "title": "Users", "href": "/dashboard/tbd" }]
+    "links": [{ "title": "Users", "href": "/dashboard/access-requests" }]
```
BCerki, line 13: I think we're supposed to get to access-requests via an Operator tile: 
image
I think someone might have created it already and once you rebase you'll have it

client/app/(authenticated)/dashboard/access-requests/page.tsx
```diff
@@ -0,0 +1,19 @@
...
+  return (
+    <>
+      <h1>Operator Admin Access Requests List</h1>
+      {/* <Link href="/dashboard/operations/create">
+        <Button variant="contained">Add Operation</Button>
+      </Link> */}
+      <Suspense fallback={<Loading />}>
+        <AccessRequests />
+      </Suspense>
+    </>
```
BCerki, line 17: I think the wireframe might have changed since this went up for review; now there's no header and there's a note

client/app/components/routes/access-requests/AccessRequests.tsx
```diff
@@ -0,0 +1,77 @@
...
+  return (
+    <>
+      <DataGrid
+        cntxt="user-operators"
+        rows={rows}
+        columns={[
+          { field: "id", headerName: "Request ID", width: 150 },
+          { field: "first_name", headerName: "First Name", width: 150 },
+          { field: "last_name", headerName: "Last Name", width: 150 },
+          { field: "email", headerName: "Email", width: 150 },
```
BCerki, line 64: wireframe now has a business bceid column too

client/app/components/routes/access-requests/form/Review.tsx
```diff
@@ -0,0 +1,108 @@
...
+      },
+    );
+    if (response.error) {
+      setErrorList([{ message: response.error }]);
+    } else if (response.ok) {
+      setSuccessMessageList([
+        { message: "You have approved the request for prime admin access." },
+      ]);
+
+      // Add a delay of 2000 milliseconds (2 seconds) before clearing the success message
```
BCerki, line 37: I don't see the success/reject message when I click the button (though I do see the change in the db): https://www.loom.com/share/86988548fef04422bbed9da5bb65c36a
Have I missed a step somewhere?
Also, I think this may be another case of changing wireframes, because it looks like the message is supposed to stay on the page now

bc_obps/registration/api/user_operator.py
```diff
@@ -52,12 +58,55 @@ def get_user_operator(request, user_operator_id: int):
...
+        )
+
+        user_operator_list.append(
+            {
+                **user_operator_related_fields_dict,
+                **user_related_fields_dict,
+                **operator_related_fields_dict,
+            }
+        )
+    return user_operator_list
```
BCerki, line 108: If I'm understanding this python code correctly (don't trust me, lol), this code is about transforming the data to give the frontend everything it needs. Probably the same thing could be done with django ninja: https://django-ninja.dev/guides/response/#nested-objects. I don't know which is better, though. I've got another comment about different approaches below

BCerki (commented):
Looking awesome! This was a huge feature!

A few overall notes:
- you'll probably want to rebase before addressing any comments because I think the latest develop is going to affect a lot of this work
- this AC: `API endpoint to update the user who submitted the request (Prime Gertrude) for the org to associate them with the operator (with no permissions to begin with)` was already completed in other PRs, correct? I shouldn't see any of this work in your PR?
- what should I be seeing for this AC: `if the request was approved, the database record for the user who submitted the request (Prime Gertrude) should be updated to mark them as an admin for the operator`? I don't see a change to Samantha Garcia in the user table after clicking accept
- the styling doesn't look exactly like figma; I assume that's fine at this stage and I haven't commented on it
- the breadcrumb component doesn't appear to working correctly;. I think someone fixed that and you'll get it once you rebase. If not, if we don't have a task card already, let's make one.

bc_obps/registration/api/user_operator.py
```diff
@@ -203,4 +252,24 @@ def create_user_operator_request(request, user_operator_id: int, payload: UserOp
     return 200, {"operator_id": operator.id}
 
 
+@router.put("/user-operators/{int:user_operator_id}/update-status")
```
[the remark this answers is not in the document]
andrea-williams: for security reasons, we should keep the endpoint for updating statuses separate. The reason being that the PUT requests to update most of the fields in user-operator should only be accessible to external users, but the PUT request to set the status of the record should only be accessible to internal users. If we were to merge these 2 functionalities into one endpoint the role-based protections get a lot more complicated.

client/app/(authenticated)/dashboard/access-requests/page.tsx
JoshLarouche: updated

client/app/components/routes/access-requests/form/Review.tsx
JoshLarouche: I don't know what change I made caused it not to persist.

Sepehr-Sobhani:
Addressed in: https://github.com/bcgov/cas-registration/pull/383

PR #287, opened 2023-11-16
Chore: add role table

dleard:
card: 281

### DIFF
--- a/bc_obps/registration/models.py
+++ b/bc_obps/registration/models.py
@@ -159,7 +159,7 @@ class User(UserAndContactCommonInfo):
     app_role = models.ForeignKey(
         AppRole,
         on_delete=models.DO_NOTHING,
-        related_name="user_app_role",
+        related_name="user_app_roles",
         db_comment="The role assigned to this user which defines the permissions the use has.",
     )
 
--- a/bc_obps/registration/models.py
+++ b/bc_obps/registration/models.py
@@ -7,12 +7,18 @@ from localflavor.ca.models import CAPostalCodeField, CAProvinceField
 class AppRole(models.Model):
     """AppRole model"""
 
-    role_name = models.CharField(primary_key=True, serialize=False, db_comment='The name identifying the role assigned to a user. This role defines their permissions within the app. Also acts as the primary key.', max_length=100)
+    role_name = models.CharField(
+        primary_key=True,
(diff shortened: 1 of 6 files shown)


bc_obps/registration/models.py
```diff
@@ -140,6 +156,12 @@ class User(UserAndContactCommonInfo):
         blank=True,
         related_name="user_documents",
     )
+    app_role = models.ForeignKey(
+        AppRole,
+        on_delete=models.DO_NOTHING,
+        related_name="user_app_role",
```
Sepehr-Sobhani, line 162: I suggest using plural names for related objects, as it retrieves multiple records from an `AppRole` instance rather than just one.

bc_obps/registration/models.py
Sepehr-Sobhani, line 1: I suggest that we populate some data in `AppRole` table, either by using a fixture or providing a function to the migration file. We can then use the ID of the created `AppRole` instance as a foreign key in the user fixture for the user instances. This will ensure that the data is properly associated and avoids any issues with data integrity.

Sepehr-Sobhani (changes requested):
Looks great! a couple of minor changes are required.

bc_obps/registration/models.py
dleard: There is a function in the migration file to populate initial data for the app_role table. The call is [here](https://github.com/bcgov/cas-registration/pull/287/files#diff-ef16f115cd68f3e0b2e49884c2b5e89844efff57b1430f32838d08b7f4ec004bR60)

Sepehr-Sobhani: Sorry, I must have missed that. This looks great, we just need to use the PK of these roles in the user fixture file.

Sepehr-Sobhani (approved):
Looks great! We just need to use the PK of roles in the user fixture file.

bc_obps/registration/models.py
dleard: I think I did that too? [here](https://github.com/bcgov/cas-registration/pull/287/files#diff-3d01cf820f7873ded05eefa3e55965df3ab06a063bae1d2ff0af1f3e457696eeR17).

Sepehr-Sobhani: Ok, I think I'm tired and should leave ASAP :D

PR #296, opened 2023-11-18
chore: create secret & add env vars for auth

dleard:


joshgamache (approved):
Looks good 🌮!

Nice simplification on the NextAuth secret too. Makes more sense that way.

PR #305, opened 2023-11-22
Chore: 301 generic action handler

dleard:
- Consolidate how we're interacting with the backend into a single action function: actionHandler()
- Pass user_guid to the backend via an Authentication header
- Remove User data fetching hardcoding in endpoints (Users.object.first())


shon-button:
Nice, and DRY and working a-ok  :)

I suspect the use of  JSON.stringify instead of FormData to submit a form might be problematic for future file upload requirements; so, perhaps implementing a FormData condition in actionHandler function might prove useful for use cases requiring FormData objects.


Something like:

```
"use server";
import { revalidatePath } from "next/cache";
import { getServerSession } from "next-auth";
import { authOptions } from "@/app/api/auth/[...nextauth]/route";

/**
 * Generic action handler that sends a request to the specified API endpoint
 * and returns the response as a JSON object.
 * @param apiUrl The relative URL of the API endpoint to send the request to.
 * @param method The HTTP method to use for the request (GET, POST, PUT, DELETE, PATCH).
 * @param pathToRevalidate The path of the data to revalidate after the request is complete.
 * @param options Optional data to include in the request body (example: body for POST, PUT, and PATCH requests, overriding cache control).
 * @param options Optional data to include in the request body (example: body for POST, PUT, and PATCH requests, overriding cache control).
 * @returns A Promise that resolves to the JSON response from the API endpoint, or an error object if the request fails.
 */
export async function actionHandler(
  endpoint: string,
  method: "GET" | "POST" | "PUT" | "DELETE" | "PATCH",
  pathToRevalidate: string,
  options: RequestInit = {},
) {
  try {

    const defaultOptions: RequestInit = {
      cache: "no-store", // Default cache option
      method,
      headers: new Headers({
        Authorization: JSON.stringify({
          user_guid: session?.user?.user_guid,
        }),
      }),
    };

    let mergedOptions: RequestInit;

    if (method === "POST" && options.body instanceof FormData) {
      // If it's a POST request and the body is FormData, handle it separately
      mergedOptions = {
        ...defaultOptions,
        ...options,
        body: options.body,
      };
    } else {
      // For other methods or non-FormData POST requests, merge options normally
      mergedOptions = {
        ...defaultOptions,
        ...options,
      };
    }

    const response = await fetch(
      `${process.env.API_URL}${endpoint}`,
      mergedOptions,
    );

    if (!response.ok) {
      // Handle HTTP errors, e.g., response.status is not in the 200-299 range
      return { error: `HTTP error! Status: ${response.status}` };
    }

    const data = await response.json();
    revalidatePath(pathToRevalidate);
    return data;
  } catch (error: unknown) {
    // Handle any errors, including network issues
    if (error instanceof Error) {
      // eslint-disable-next-line no-console
      console.error(`An error occurred while fetching ${endpoint}:`, error);
      return {
        // eslint-disable-next-line no-console
        error: `An error occurred while fetching ${endpoint}: ${error.message}`,
      };
    } else {
      // eslint-disable-next-line no-console
      console.error(`An unknown error occurred while fetching ${endpoint}`);
      return {
        error: `An unknown error occurred while fetching ${endpoint}`,
      };
    }
  }
}
```

shon-button (approved):

shon-button:
Is there a card for CSRF checks to ensure a valid token `request.headers.get('Authorization'))["user_guid"]` in each request?  
Perhaps a Django middleware?

dleard:
@shon-button good points. I'm going to merge this as-is since it has significant effects on development workflows. Then we can assess if we need follow up cards to address the potential file upload issue & CSRF check

PR #309, opened 2023-11-23
232 update css theme

marcellmueller:
A a lot of this was already completed in previous PRs and without a proper design system implemented there wasn't too much to do for this ticket, though it's something we've flagged with the designers.

I've tried to simplify the colour names to be a little bit less verbose as the old ones made for some long tailwind classes. Let me know if you have other naming suggestions though I think we should keep them fairly short.

I've also setup documentation linking to our global css file, Tailwind config and Mui theme.


shon-button:
💫 magic!

shon-button (approved):

Issue #316, opened 2023-11-23
RJSF URL and Email field validation


marcellmueller:
#### Describe the task

We need to validate URL and email fields and trigger an error in RJSF if they are not valid. Looks like we can just set `format: 'url` or `format: 'email'` in the schema next to the type to trigger an error. We may want to look into transforming custom error messages.

#### Acceptance Criteria

- [x] URL field validation
- [x] Email field validation
- [x] Update `InlineFieldTemplate` to display error message?
- [x] Update documentation

#### Additional context

- Add any other context about the task here.
- Or here

marcellmueller:
@BCerki Looks like this is blocked though I was testing formats and tried using the `format: 'url` built in to RJSF and I think it's too strict. For example it seems to require `http://` and a basic url like `www.google.com` triggers the validation.

We can easily get around this by finding a safe regex and making our own format in `FormBase` `customFormats`.

hannavovk:
@marcellmueller Can we not requite the "http" for the URL to work? cc @dleard

PR #324, opened 2023-11-23
173 number widget

marcellmueller:
In this I refactored and moved some of the shared widget styles to the MUI theme that didn't need to be reactive before I figured out that we might not need a specific `NumberWidget`. Setting the `type` as I have in the `URLWidget` for example doesn't actually get passed in RJSF to the widget, even if I extend the widget props.  As such will be getting rid of that and `EmailWidget` in #316 using the `format` option.

Now in a fields schema if  `type: 'number'` is set then the `TextWidget` will be set to `type="number"`, the input won't accept letters and the value is saved as a number in the form data.  I looked into it and we can limit the decimal places in the future easily if needed.

The arrow increment buttons are disabled globally for number inputs via CSS in the `globals.css` file.

Let me know if you think we need a specific `NumberWidget`. I like this solution though maybe there is something I missed.

### DIFF
--- a/client/app/components/form/widgets/ComboBox.tsx
+++ b/client/app/components/form/widgets/ComboBox.tsx
@@ -3,11 +3,7 @@
 import { Autocomplete, TextField } from "@mui/material";
 import { WidgetProps } from "@rjsf/utils/lib/types";
 import { useCallback } from "react";
-import {
-  BC_GOV_LINKS_COLOR,
-  DARK_GREY_BG_COLOR,
-  BC_GOV_SEMANTICS_RED,
-} from "@/app/styles/colors";
+import { DARK_GREY_BG_COLOR, BC_GOV_SEMANTICS_RED } from "@/app/styles/colors";
 
 const ComboBox: React.FC<WidgetProps> = (props) => {
   const { id, onChange, rawErrors, schema, value } = props;
@@ -37,13 +33,6 @@ const ComboBox: React.FC<WidgetProps> = (props) => {
       borderColor: DARK_GREY_BG_COLOR,
     },
     "& .MuiOutlinedInput-notchedOutline": {
-      borderColor: borderColor,
-    },
-    "&:hover .MuiOutlinedInput-notchedOutline": {
-      borderColor: BC_GOV_LINKS_COLOR,
-      borderWidth: "1px",
-    },
-    "&.Mui-focused .MuiOutlinedInput-notchedOutline": {
       borderColor: borderColor,
     },
   };
--- a/client/app/components/form/widgets/SelectWidget.tsx
+++ b/client/app/components/form/widgets/SelectWidget.tsx
@@ -5,11 +5,7 @@ import FormControl from "@mui/material/FormControl";
 import InputLabel from "@mui/material/InputLabel";
 import Select, { SelectChangeEvent } from "@mui/material/Select";
 import { WidgetProps } from "@rjsf/utils/lib/types";
-import {
-  BC_GOV_LINKS_COLOR,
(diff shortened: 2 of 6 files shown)


andrea-williams (changes requested):
Was playing around with this widget. For the new Operation form I entered an unrealistically large number for 2022 emissions and then hit Submit. On the frontend I just got a "HTTP error! Status: 500" message at the bottom of the screen, and no error is displayed next to the specific field. But looking at my logs I see this error:
"psycopg.errors.NumericValueOutOfRange: numeric field overflow
DETAIL:  A field with precision 10, scale 5 must round to an absolute value less than 10^5"
Because our numeric columns in the database have a max precision 10, scale 5, I'd recommend that we add a max limit to the number widget so that values exceeding that display an error on the frontend.

marcellmueller:
> Was playing around with this widget. For the new Operation form I entered an unrealistically large number for 2022 emissions and then hit Submit. On the frontend I just got a "HTTP error! Status: 500" message at the bottom of the screen, and no error is displayed next to the specific field. But looking at my logs I see this error: "psycopg.errors.NumericValueOutOfRange: numeric field overflow DETAIL: A field with precision 10, scale 5 must round to an absolute value less than 10^5" Because our numeric columns in the database have a max precision 10, scale 5, I'd recommend that we add a max limit to the number widget so that values exceeding that display an error on the frontend.

We can add max length easily but if I understand correctly `10^5` is `100000`? We might have to add per field max length as hardcoding in a max of 100k doesn't seem right.

andrea-williams:
> > Was playing around with this widget. For the new Operation form I entered an unrealistically large number for 2022 emissions and then hit Submit. On the frontend I just got a "HTTP error! Status: 500" message at the bottom of the screen, and no error is displayed next to the specific field. But looking at my logs I see this error: "psycopg.errors.NumericValueOutOfRange: numeric field overflow DETAIL: A field with precision 10, scale 5 must round to an absolute value less than 10^5" Because our numeric columns in the database have a max precision 10, scale 5, I'd recommend that we add a max limit to the number widget so that values exceeding that display an error on the frontend.
> 
> We can add max number easily but if I understand correctly `10^5` is `100000`? We might have to add per field max as hardcoding in a max of 100k doesn't seem right.

ahh yes you make a good point. I think there is a problem with our current definition of the previous_year_emissions column in our operation table - that's a separate issue outside the scope of this PR. 

For this PR, could we set a max number of 2147483647 (the maximum integer value that Postgres supports) please?

marcellmueller:
> ahh yes you make a good point. I think there is a problem with our current definition of the previous_year_emissions column in our operation table - that's a separate issue outside the scope of this PR.
> 
> For this PR, could we set a max number of 2147483647 (the maximum integer value that Postgres supports) please?

Sounds good, I've added a max number and it will be easy to set an option to do a per field max in the future. I think we will need to go over the number fields and add a max number, I will try to remember to make a ticket tomorrow morning.

andrea-williams:
> > ahh yes you make a good point. I think there is a problem with our current definition of the previous_year_emissions column in our operation table - that's a separate issue outside the scope of this PR.
> > For this PR, could we set a max number of 2147483647 (the maximum integer value that Postgres supports) please?
> 
> Sounds good, I've added a max number and it will be easy to set an option to do a per field max in the future. I think we will need to go over the number fields and add a max number, I will try to remember to make a ticket tomorrow morning.

I just checked with Hanna - she agrees that our configuration for that DB column is off. I've created a new ticket to make changes to the DB. Setting the max value of the number widget to be equal to the max integer value in Postgres should be good enough for now.

PR #343, opened 2023-11-24
Chore: terraform deploys with shipit

dleard:
a couple fixes to scripts/makefile to allow shipit to deploy after terraform was added


joshgamache (approved):
All looks good! 🌮

PR #347, opened 2023-11-28
fix: header colour regression from mui theme update

marcellmueller:
Fix for the homepage header colour. This was caused by the MUI theme update PR. I also added a small change to typography so by default it will use the text colour from the bc gov design system.


andrea-williams:
There's an issue with the styling of the header title:

Screenshot 2023-11-28 at 9 33 06 AM

marcellmueller:
> There's an issue with the styling of the header title:
> 
> Screenshot 2023-11-28 at 9 33 06 AM

Great catch, thank you 🙇‍♂️ 

This is fixed though really looking forward to getting Happo running in another sprint or two so we can avoid these kind of regressions. Appreciate it!!

andrea-williams (approved):

Issue #319, opened 2023-11-23
RJSF postal code widget


marcellmueller:
#### Describe the task

We need a postal code widget to format and validate postal codes

#### Acceptance Criteria

- [x] Postal code widget
- [x] Validation (to be confirmed but likely just standard canadian postal codes)
- [x] Documentation

PR #366, opened 2023-12-02
319 postal code widget

marcellmueller:
Implements #319


Sepehr-Sobhani (approved):
LGTM 🤩 Just a suggestion and also out of the scope of this ticket so feel free to ignore it :) 
Would it be helpful to display a generic error message below the field for an acceptable postal code format?

marcellmueller:
> LGTM 🤩 Just a suggestion and also out of the scope of this ticket so feel free to ignore it :) Would it be helpful to display a generic error message below the field for an acceptable postal code format?

Thanks! That is a good point and something we need to keep track of. I believe adding proper support for error messages in the `InlineFieldTemplate` will be handled in #316. However I have created another issue to track all of the custom error messages we will need here: 

https://github.com/bcgov/cas-registration/issues/380

Issue #318, opened 2023-11-23
RJSF MultiSelectWidget


marcellmueller:
#### Describe the task

MultiSelectWidget using MUI autocomplete

https://mui.com/material-ui/react-autocomplete/

#### Acceptance Criteria

- [ ] first
- [ ] second
- [ ] third

#### Additional context

- Add any other context about the task here.
- Or here

nanyangpro:
Thanks for confirming, @BCerki! 
From the Registration Fields spreadsheet (shared doc with our SMEs), the type of the "NAICS Code" is indicated as a **number** instead of a string. So to your question, I believe it's the code instead of the description.
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBejRpQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--0db0710c156773f1dd9e533c1cafbcc38eb33f71/image.png)

(I wasn't directly involved in the conversation if there was. Feel free to correct me if I'm wrong, @hannavovk or @suhafa. Thanks!)

nanyangpro:
Thanks for confirming, @BCerki! 
From the Registration Fields spreadsheet (shared doc with our SMEs), the type of the "NAICS Code" is indicated as a **number** instead of a string. So to your question, I believe it's the code instead of the description.
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBejRpQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--0db0710c156773f1dd9e533c1cafbcc38eb33f71/image.png)

(I wasn't directly involved in the conversation if there was. Feel free to correct me if I'm wrong, @hannavovk or @suhafa. Thanks!)

PR #367, opened 2023-12-04
318 multiselect widget

marcellmueller:
Implements #318 

This one took a lot of experimentation to do and not trigger RJSF validation issues. I initially tried to do it using `anyOf` as we use for many select fields using the `ComboBox` widget though storing the values as an array triggered a lot of validation errors. Rather than removing those errors on the front end I wanted to use a way that was supported by RJSF.

I ended up using `enum` as these fields initially were so the values are saved the same as before this PR as an array of the selected `enum` values. I ended up using `enumNames` to set custom labels for each selection though that might be depreciated eventually. Open to other suggestions if you have any other ideas.

While there was talk of just adding onto the functionality of `ComboBox` I decided to make `MultiSelectWidget` to keep the logic clean and separate as there is a fair amount of differences in how we update and store the data.

Also fixed the `naics_code` field so it is showing the codes not the id.

### DIFF
--- a/client/app/components/form/widgets/MultiSelectWidget.tsx
+++ b/client/app/components/form/widgets/MultiSelectWidget.tsx
@@ -57,6 +57,9 @@ const MultiSelectWidget: React.FC<WidgetProps> = ({
       filterSelectedOptions
       autoHighlight
       options={options}
+      value={value.map((val: string | number) => {
+        return options.find((option: Option) => option.const === val);
+      })}
       sx={styles}
       isOptionEqualToValue={(option: Option, val: Option) => {
         return option.const === val.const;
--- a/client/app/components/form/widgets/MultiSelectWidget.tsx
+++ b/client/app/components/form/widgets/MultiSelectWidget.tsx
@@ -4,6 +4,11 @@ import { TextField, Autocomplete, MenuItem } from "@mui/material";
 import { WidgetProps } from "@rjsf/utils/lib/types";
 import { DARK_GREY_BG_COLOR, BC_GOV_SEMANTICS_RED } from "@/app/styles/colors";
 
+interface Option {
+  const: string | number;
+  title: string;
+}
+
 const ComboBox: React.FC<WidgetProps> = ({
   id,
   onChange,
@@ -19,7 +24,7 @@ const ComboBox: React.FC<WidgetProps> = ({
   };
   const enumValues = fieldSchema?.enum as any;
   const enumNames = fieldSchema?.enumNames as any;
-  const options = enumValues.map((val, index) => ({
+  const options = enumValues.map((val: string | number, index: number) => ({
     const: val,
     title: enumNames[index] || val,
   }));
@@ -28,15 +33,14 @@ const ComboBox: React.FC<WidgetProps> = ({
     e: React.ChangeEvent<{}>,
     option: Array<{ const: string | number; title: string }>,
   ) => {
-    onChange(option.map((o) => o.const));
+    onChange(option.map((o: Option) => o.const));
   };
 
   const placeholder = uiSchema?.["ui:placeholder"]
     ? `${uiSchema["ui:placeholder"]}...`
     : "";
 
-  const displayPlaceholder =
-    typeof value === "undefined" || value?.length === 0;
+  const displayPlaceholder = value?.length === 0;
 
   const isError = rawErrors && rawErrors.length > 0;
   const borderColor = isError ? BC_GOV_SEMANTICS_RED : DARK_GREY_BG_COLOR;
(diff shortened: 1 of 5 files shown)


client/app/components/form/widgets/MultiSelectWidget.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+  const handleChange = (e: React.ChangeEvent<{}>, option: Array<Option>) => {
+    onChange(option.map((o: Option) => o.const));
+  };
+
+  const placeholder = uiSchema?.["ui:placeholder"]
+    ? `${uiSchema["ui:placeholder"]}...`
+    : "";
+
+  const displayPlaceholder = value?.length === 0;
```
BCerki, line 43: I'm surprised the mui component doesn't have this check built-in

client/app/components/form/widgets/MultiSelectWidget.tsx
```diff
@@ -0,0 +1,103 @@
+"use client";
+
+import { Autocomplete, Chip, MenuItem, TextField } from "@mui/material";
+import { WidgetProps } from "@rjsf/utils/lib/types";
+import { DARK_GREY_BG_COLOR, BC_GOV_SEMANTICS_RED } from "@/app/styles/colors";
+
+interface Option {
+  const: string | number;
+  title: string | number;
```
BCerki, line 9: `const` and `title` are both key names in RJSF schemas, so I'd suggest changing these to something else to avoid mix-ups (also I don't know what they are just from the key name--const is the id maybe, and title is the text we want to show?)

client/app/components/form/widgets/MultiSelectWidget.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+  return (
+    <Autocomplete
+      disablePortal
+      id={id}
+      multiple
+      filterSelectedOptions
+      autoHighlight
+      options={options}
+      value={value.map((val: string | number) => {
```
BCerki, line 63: I think `val` makes sense as a map variable name here because the RJSF prop really is called `value`. I see `val` used elsewhere too--if in any of those cases it's not RJSF's `value`, then I'd suggest changing it to something else to avoid confusion

client/app/components/form/widgets/MultiSelectWidget.tsx
```diff
@@ -0,0 +1,103 @@
...
+      borderColor: borderColor,
+    },
+  };
+
+  return (
+    <Autocomplete
+      disablePortal
+      id={id}
+      multiple
+      filterSelectedOptions
```
BCerki, line 60: This looks really nice in the app

BCerki (commented):
Looking good!

This AC is incomplete: `reporting activities dropdown should show name of activity conditionally based on what's applicable to operation type (see applicable_to column in db)`. We have an example of conditional dropdowns in CIF (rfp widget or parent widget I think), but there might be a nicer/easier way to do it now that we're using mui instead of the toolkit.

client/app/components/form/widgets/MultiSelectWidget.tsx
marcellmueller: No by default the placeholder stays. I made the decision to conditionally hide it once the user has selected something as depending on the chip length it would hide some of it as shown here:

Screenshot 2023-12-05 at 3 23 43 PM


However I did find an example of always putting the placeholder on a new line. There is a demo here though this might be a question for the designers if they want further refinement.
https://codesandbox.io/p/sandbox/67090943how-can-i-set-the-placeholder-of-autocomplete-in-a-new-line-u8vbz

client/app/components/form/widgets/MultiSelectWidget.tsx
marcellmueller: Thanks, I have change the values to `id` and `label` which seems more clear.

client/app/components/form/widgets/MultiSelectWidget.tsx
marcellmueller: Sounds good, I have changed most of the other `val` names with one exception, right below this is the prop `isOptionEqualToValue` and I thought it made sense to keep `val` there as well:

```
  isOptionEqualToValue={(option: Option, val: Option) => {
        return option.id === val.id;
      }}
```

Let me know if you think otherwise.

client/app/components/form/widgets/MultiSelectWidget.tsx
marcellmueller: Thanks!

marcellmueller:
> Looking good!
> 
> This AC is incomplete: `reporting activities dropdown should show name of activity conditionally based on what's applicable to operation type (see applicable_to column in db)`. We have an example of conditional dropdowns in CIF (rfp widget or parent widget I think), but there might be a nicer/easier way to do it now that we're using mui instead of the toolkit.

Great catch, I have made an issue to handle this missing acceptance criteria here: 
https://github.com/bcgov/cas-registration/issues/374

I will try and complete that issue in the near future.

Issue #291, opened 2023-11-17
Improve handling of Statuses for Operations & Operators


andrea-williams:
#### Describe the task

There's 2 goals to this ticket: (1) fix a bad-practice coding habit (that greatly increases the chance of errors); and (2) improve the readability of application status labels (e.g., `not_registered` -> `Not Registered`)

#### Acceptance Criteria

- [x] remove all manual assignments of status in code with enums (e.g., we don't want to use `operation.status = 'Pending'` bc it's bad practice. Replace with `operation.status = OperationStatus.PENDING`)
- [x] remove hacky string fixes to status strings in FE (such as calling `capitalizeString()` in Operations table)
- [x] rework the Status labels in Django models. Sepehr's suggestion is to use
```python
class Statuses(models.TextChoices):

        NOT_REGISTERED = "Not Registered"

        PENDING = "Pending"

        APPROVED = "Approved"

        REJECTED = "Rejected"
```

#### Additional context

- Add any other context about the task here.
- Or here

andrea-williams:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Team-Moose-Sprint-Board-60d216175a2689000ec774d0/issues/bcgov/cas-registration/291?planning-poker) with Zenhub @andrea-williams @dleard @joshgamache @JoshLarouche @marcellmueller @Sepehr-Sobhani @shon-button

PR #377, opened 2023-12-06
291 improve status handling

marcellmueller:
Implements #291 

Some of the issues in this PR have already been fixed so there wasn't too much to do here.


shon-button:
Occurrence of "manual assignments of status in code" in `client/app/components/form/OperationsForm.tsx`
`formData?.status === "Registered" || formData?.status === "Pending"`

Suggest adding REGISTERED to export enum Status and changing code:

` formData?.status ===  Status.REGISTERED || formData?.status ===  Status.PENDING`

marcellmueller:
> Occurrence of "manual assignments of status in code" in `client/app/components/form/OperationsForm.tsx` `formData?.status === "Registered" || formData?.status === "Pending"`

It's not an assignment but great catch! However I'm wondering if `formData?.status === "Registered"` isn't correct as we don't have a `Registered` status, only `Not Registered`. I wonder if this check for read only is supposed to be `Approved`? 🤔

shon-button (approved):

Issue #268, opened 2023-11-10
Clean up <Review> component in Operations view


andrea-williams:
###Description of the Tech Debt###

A `<Review />` component has been added to `forms/Operation` for internal users to approve/reject an Operation's request for carbon tax exemption. It is functional but there is outstanding front-end work to polish up the styling and fix a small bug. 

To see the Review component, click "View Details" on an Operation whose status is 'Pending'. 

- [x] make whitespace in component more aesthetic & functional
- [x] I think according to the wireframes the buttons should be moved over to the right side of the page (?)
- [x] check with designers: should users be prompted to confirm they want to Approve/Reject before the change happens? (i.e., "Are you sure?"). At the moment there's no route to undo a status change once it's been made
- [x] the `Alert` component that should appear when an Operation is approved or rejected used to show up and stay there until the user dismissed it, but since the version upgrade of Next.js, it only flashes very briefly and then disappears. Would like to get that either staying until user manually dismisses it, or timeout after ~5 seconds

marcellmueller:
Putting this back in the sprint backlog for now as I am blocked waiting for a design sync.

PR #378, opened 2023-12-06
feat/268 clean up review component

marcellmueller:
Implements #268 

We now have a confirmation modal to approve/deny the operation application. I made a generic `<Modal>` component using MUI that we can reuse in the future.

Refactored the `Review` component to be reusable since we have the same feature in the user operator form. 

The bug that was causing the `Alert` components to disappear right away is because the `<Review>` component was conditionally rendered based on the status. So when that changed the entire component would disappear. 

Note: I also snuck in two fixes for `ComboBox` and `MultiSelectWidget` as I noticed while working on this that they did not have a disabled state (ie: `status === PENDING`).


JoshLarouche (approved):
Looks good! Nice generalization of the component!

PR #401, opened 2023-12-09
Fix column type of previous_year_attributable_emissions

andrea-williams:
Chore: Changed type of db col for previous_year_attributable_emissions from DecimalField to IntegerField


Sepehr-Sobhani (approved):
🤩

Issue #280, opened 2023-11-15
IO can provide operator data when operation is owned by multiple operators


dleard:
#### Description:

In the operation admin form, the user is asked if the operation has multiple operators. If they answer yes, they are presented with a form to enter data about that operator. This operator data should be treated differently from the data in the operator table. It is purely metadata, whereas the operator table is for registered operators that are the primary operator for their dependent operations. The data entered in this form should not mutate the data in the operator table, it should live in its own place.

wireframe of multiple operator form:
https://www.figma.com/file/LsDmLDCdnJqI8UahyqvOD8/OBPS-App-Designs?type=design&node-id=1312-1134&mode=design&t=Vh3QQdLInvXMO7Gu-0

#### Acceptance Criteria:

Given my operation has multiple operators
When I am presented with the form
Then I can enter the data about the extra operator

Given I need to enter multiple operator data
When I enter that data
Then the data is not saved to the operator table
And the data is saved to the multiple_operator table as metadata

#### Development Checklist:

- [x] a new model `multiple_operator` is created
- [x] the `multiple_operator` table contains all fields in the wireframe (minus the radio button, which is just a formatting thing)
- [x] when data is entered in the multiple operator section of the operation admin form, it is saved / updated in the `multiple_operator` table
- [ ] Meets the DOD

**Definition of Ready** (Note: If any of these points are not applicable, mark N/A)

- [x] User story is included
- [x] User role and type are identified
- [x] Acceptance criteria are included
- [x] Wireframes are included (if required)
- [ ] Design / Solution is accepted by Product Owner
- [ ] Dependencies are identified (technical, business, regulatory/policy)
- [ ] Story has been estimated (under 13 pts)

·**Definition of Done** (Note: If any of these points are not applicable, mark N/A)

- [ ] Acceptance criteria are tested by the CI pipeline
- [ ] UI meets accessibility requirements
- [ ] Configuration changes are documented, documentation and designs are updated
- [ ] Passes code peer-review
- [ ] Passes QA of Acceptance Criteria with verification in Dev and Test
- [ ] Ticket is ready to be merged to main branch
- [ ] Can be demoed in Sprint Review
- [ ] Bugs or future work cards are identified and created
- [ ] Reviewed and approved by Product Owner

#### Notes:

-

dleard:
[...]

PR #405, opened 2023-12-11
feat/280 business structure field

marcellmueller:
Follow up PR which completes issue #280 

When I was working on the main #280 PR the `business_structure` field was being handled for the user operator form in a separate PR so I didn't properly implement it. This PR updates business structure field to have the dropdown populated with data stored in the db as well as changes the field to a foreign key.

Also has a fix for the `multiple_operators_array` form field which would not correctly open the form if the user selected `Yes` for the `operator_has_multiple_operators` field after the form was previously saved with a null or empty array in the field.


bc_obps/registration/api/operation.py
```diff
@@ -43,7 +52,7 @@ def save_multiple_operators(multiple_operators_array, operation):
         new_operator = {}
         new_operator["operation_id"] = operation.id
         new_operator["operator_index"] = idx + 1
-
+        print(operator)
```
JoshLarouche, line 55: Leftover print statement

client/app/components/routes/operations/form/Operation.tsx
```diff
@@ -12,7 +13,7 @@ async function getNaicsCodes() {
     return await actionHandler(
       "registration/naics_codes",
       "GET",
-      "/dashboard/operations",
```
JoshLarouche, line 15: Looks like your prettier CI action is failing because it wants these commas back in there but I agree I don't like them either.

bc_obps/registration/api/operation.py
marcellmueller: Fixed, thank you 🙇‍♂️

client/app/components/routes/operations/form/Operation.tsx
marcellmueller: Thanks, looks like this is passing now.

JoshLarouche (approved):
Looks good to go now assuming the needed CI check pass!

PR #407, opened 2023-12-11
BC Corporate field format

BCerki:
card: https://app.zenhub.com/workspaces/team-moose-sprint-board-60d216175a2689000ec774d0/issues/gh/bcgov/cas-registration/357


bc_obps/registration/models.py
```diff
@@ -221,7 +222,15 @@ class Statuses(models.TextChoices):
     legal_name = models.CharField(max_length=1000, db_comment="The legal name of an operator")
     trade_name = models.CharField(max_length=1000, blank=True, db_comment="The trade name of an operator")
     cra_business_number = models.IntegerField(db_comment="The CRA business number of an operator")
-    bc_corporate_registry_number = models.IntegerField(db_comment="The BC corporate registry number of an operator")
+    bc_corporate_registry_number = models.CharField(
```
BCerki, line 225: not sure if this is necessary since I've already added validation to the schema

client/app/components/form/FormBase.tsx
```diff
@@ -7,6 +7,7 @@ const customFormats = {
   phone: /\(?\d{3}\)?[\s-]?\d{3}[\s-]?\d{4}$/,
   "postal-code":
     /^[ABCEGHJ-NPRSTVXY]\d[ABCEGHJ-NPRSTV-Z][ -]?\d[ABCEGHJ-NPRSTV-Z]\d$/i,
+  bc_corporate_registry_number: "^[A-Za-z]{1,3}\\d{7}$",
```
BCerki, line 10: To manually test the ninja schema validation, comment out this line and then try to fill and submit the form with bad data

bc_obps/registration/models.py
marcellmueller: No I think backend validation is a great idea as Django validates most of our fields already. Nice work, this is really cool and will be a good example if we need other types of custom backend validation going forward 🙇‍♂️

bc_obps/registration/schema/multiple_operator.py
```diff
@@ -11,7 +10,7 @@ class MultipleOperatorIn(Schema):
     mo_legal_name: str
     mo_trade_name: str
     mo_cra_business_number: int
-    mo_bc_corporate_registry_number: int
+    mo_bc_corporate_registry_number: str = Field(regex=r"^[A-Za-z]{1,3}\d{7}$")
```
marcellmueller, line 13: This is neat as well 😮

docs/rjsf-forms.md
```diff
@@ -17,6 +17,22 @@ number_field {
 
 The `defaultTheme` includes many custom widgets which can be set in the forms `uiSchema` using the `ui:widget` option.
 
+## Custom validation
```
marcellmueller, line 20: Thanks for adding to the docs!

marcellmueller (commented):
Great work on this, the frontend and backend validation seems to be working really well.

One thing I wonder if we should add is a `maxLength` option (of 10 for these fields) to the `ui:options` and set that in the `TextWidget` input if it exists. I think it could come handy in other fields but let me know if you think it's unnecessary.

bc_obps/registration/migrations/0003_alter_historicaloperator_bc_corporate_registry_number_and_more.py
```diff
@@ -0,0 +1,52 @@
+# Generated by Django 4.2.8 on 2023-12-12 19:17
```
marcellmueller, line 1: I assume I must have missed something as there are now three migrations files. I was under the impression that we were deleting the `001` and regenerating it until we release to prod?

bc_obps/registration/fixtures/mock/operation.json
```diff
@@ -12,7 +12,7 @@
       "naics_code": 19,
       "previous_year_attributable_emissions": "10001",
       "swrs_facility_id": 1001,
-      "bcghg_id": 546,
```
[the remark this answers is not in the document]
BCerki: I just misread the ticket at first and changed the bcghg id instead of the bc corporate id. When I put it back I didn't remember what the original numbers were, and it doesn't matter for the test, so I just left it

BCerki:
> Great work on this, the frontend and backend validation seems to be working really well.
> 
> One thing I wonder if we should add is a `maxLength` option (of 10 for these fields) to the `ui:options` and set that in the `TextWidget` input if it exists. I think it could come handy in other fields but let me know if you think it's unnecessary.

I don't think this one needs a `maxLength` because the regex already ensures that, but could be useful for other fields, yes

bc_obps/registration/migrations/0003_alter_historicaloperator_bc_corporate_registry_number_and_more.py
BCerki: Yup, I'll delete and regenerate. Just had to check in teams that that was still the plan now that we have history tables

marcellmueller:
> I don't think this one needs a `maxLength` because the regex already ensures that, but could be useful for other fields, yes

Well I meant more from a UX perspective, the validation is working great though it's nice to limit fields if we know exactly how long a field show be. I suppose it might be out of scope for this PR.

marcellmueller (approved):
Great work! 🥇 

One last very minor thing you could fix I noticed when referencing your docs for my formats PR is I think the `Custom validation` section should go outside of the `Custom widgets` section and not wedged in it (or at least make the header smaller if you think it belongs in there). Not a big deal though.

PR #412, opened 2023-12-12
316 rjsf url email validation

marcellmueller:
Implements #316 

We were going to use a custom validator for URLs since the RJSF `uri` format is strict and requires `https://` though Nan said that's okay for now and to just updated the error messaging.

### DIFF
--- a/client/app/components/form/FormBase.tsx
+++ b/client/app/components/form/FormBase.tsx
@@ -18,7 +18,7 @@ const customFormatsErrorMessages = {
   "postal-code": "Format should be A1A 1A1",
   phone: "Format should be ###-###-####",
   email: "Please enter a valid email address, e.g. mail@example.com",
-  website: "Please enter a valid website, e.g. www.website.com",
+  uri: "Please enter a valid website link, e.g. http://www.website.com, https://www.website.com",
 };
 
 const transformErrors = (errors: RJSFValidationError[]) => {
--- a/client/app/components/form/FormBase.tsx
+++ b/client/app/components/form/FormBase.tsx
@@ -16,7 +16,7 @@ const customFormatsErrorMessages = {
   bc_corporate_registry_number:
     "BC Corporate Registry number should be 1-3 letters followed by 7 digits",
   "postal-code": "Format should be A1A 1A1",
-  phone: "Format should be ###-###-####",
+  phone: "Format should be ### ### ####",
   email: "Please enter a valid email address, e.g. mail@example.com",
   uri: "Please enter a valid website link, e.g. http://www.website.com, https://www.website.com",
 };
--- a/client/app/components/form/FormBase.tsx
+++ b/client/app/components/form/FormBase.tsx
@@ -42,10 +42,9 @@ const FormBase: React.FC<FormPropsWithTheme<any>> = (props) => {
       formData={formData ?? {}}
       noHtml5Validate
       omitExtraData={omitExtraData ?? true}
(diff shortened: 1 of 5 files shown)


client/app/styles/rjsf/InlineFieldTemplate.tsx
```diff
@@ -73,15 +75,27 @@ function InlineFieldTemplate({
...
+            display: "flex",
+            alignItems: "center",
+            color: "error.main",
+            marginLeft: {
+              xs: "0",
+              md: "16px",
+            },
+          }}
+        >
+          <AlertIcon classNames="hidden md:block mr-4" />
```
BCerki, line 95: The icon changes size depending on the length of the error message: 
image

client/app/components/form/FormBase.tsx
```diff
@@ -18,7 +18,7 @@ const customFormatsErrorMessages = {
   "postal-code": "Format should be A1A 1A1",
   phone: "Format should be ###-###-####",
```
BCerki, line 19:
```suggestion
  phone: "Format should be ### ### ####",
```

client/app/utils/jsonSchema/operations.ts
```diff
@@ -457,9 +459,7 @@ export const operationUiSchema = {
     "ui:widget": "RadioWidget",
   },
   is_application_lead_external: {
-    "Would you like to add an exemption ID application lead?": {
```
BCerki, line 460: Why do we need to remove this?

BCerki (commented):
I'm no longer able to submit the operator form: 
image
image

I think this was probably introduced in my PR, not yours, so I'll make a bug card

client/app/utils/jsonSchema/operations.ts
marcellmueller: Apologies, I should have put a comment in this PR as it is admittedly not related to this PR but a bug that needed to be fixed. I believe this is a mistake as the field used to be called that and was changed to `is_application_lead_external`. This is basically two fields wrapping each other and as a result the `RadioWidget` wasn't used and it's just incorrectly displayed as a checkbox.

client/app/styles/rjsf/InlineFieldTemplate.tsx
marcellmueller: Nice catch! Fixed this as well as another breakpoint issue where the icon wasn't hiding correctly in small breakpoint range after the breakpoint changed to vertical layout. Thanks 🙇‍♂️

client/app/components/form/FormBase.tsx
marcellmueller: Fixed, thanks

client/app/utils/jsonSchema/operations.ts
BCerki: I think that solves the first screenshot in this bug ticket then? https://app.zenhub.com/workspaces/team-moose-sprint-board-60d216175a2689000ec774d0/issues/gh/bcgov/cas-registration/409

client/app/styles/rjsf/InlineFieldTemplate.tsx
BCerki: Looks good at all sizes

BCerki (approved):

client/app/utils/jsonSchema/operations.ts
marcellmueller: Yes that is it, I will remove it from the ticket thanks!

PR #432, opened 2023-12-15
refactor: change new operation '0' param to 'create'

marcellmueller:
Small PR to fix using `0` as the query param value for creating a new operation form back to `create`. 

Shon and I decided to fix this as part of the breadcrumb work.


shon-button:
@marcellmueller 

I added the code in `client/app/components/navigation/Bread.tsx` to implement:
- "create" parameter to be part of the crumbs
- remove the "formSection" from crumbs (assumes formSection is last index of path)

image

marcellmueller:
@shon-button nice addition, that looks great!

shon-button (approved):

Issue #419, opened 2023-12-13
Senior officer checkbox has no text [7]


dleard:
Describe the Bug:

The checkbox shown during prime admin request access to declare whether the user is the senior officer of the organisation is not showing any text. 

Probability (how likely the bug is to happen, scored from 1-5): 5

Effect (how bad the bug is when it does happen, scored from 1-5): 2

Steps to reproduce the behaviour:

Log in as an industry user with no permissions
Select an operator that does not have an admin & confirm it is the correct operator
See the checkbox at the top of the screen with no text

Screenshots:
<img width="967" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeFpjQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--d5eefe4c748524527a837759b8d98d4c932c7375/Screenshot%20from%202023-12-13%2014-40-50.png" alt="Screenshot from 2023-12-13 14-40-50.png" />

marcellmueller:
Wow nice timing @dleard, I was just about to make a ticket for this. I was looking into this and I think we need to make a `CheckBoxWidget` using MUI checkbox component that correctly handles the labels as well as have styling which aligns with the rest of our forms. 

`label: false` is set in the `ui:options` since the label was displaying twice, both in the `inlineFieldTemplate`  and also in the default RJSF checkbox widget.

PR #436, opened 2023-12-18
419 senior officer checkbox has no text

marcellmueller:
Implements #419 

Created a MUI checkbox widget so the styling would match the rest of our widgets. `label: false` now just hides the label in the `InlineFieldTemplate` instead of both the widget and the template.


joshgamache (approved):
Text is showing up where expected with the checkbox. Dropped it into another spot to test it based on the docs, all looks good!

PR #439, opened 2023-12-18
395 pending page

BCerki:
card: https://app.zenhub.com/workspaces/team-moose-sprint-board-60d216175a2689000ec774d0/issues/gh/bcgov/cas-registration/395

### DIFF
--- a/bc_obps/registration/api/user_operator.py
+++ b/bc_obps/registration/api/user_operator.py
@@ -11,6 +11,7 @@ from registration.schema import (
     UserOperatorContactIn,
     IsApprovedUserOperator,
     UserOperatorOperatorIdOut,
+    UserOperatorStatus,
 )
 from registration.schema.user_operator import SelectUserOperatorOperatorsOut
 from typing import List
@@ -41,6 +42,12 @@ import pytz
 
 
 ##### GET #####
+@router.get("/user-operator-status-from-user", response={200: UserOperatorStatus, codes_4xx: Message})
+def get_user_operator_operator_id(request):
+    user_operator = get_object_or_404(UserOperator, user_id=request.current_user.user_guid)
+    return 200, {"status": user_operator.status}
+
+
 @router.get("/is-approved-admin-user-operator/{user_guid}", response={200: IsApprovedUserOperator, codes_4xx: Message})
 def is_approved_admin_user_operator(request, user_guid: str):
     approved_user_operator: bool = UserOperator.objects.filter(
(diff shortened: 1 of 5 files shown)


bc_obps/registration/api/user_operator.py
```diff
@@ -41,6 +42,12 @@
 
 
 ##### GET #####
+@router.get("/user-operator-status-from-user", response={200: UserOperatorStatus, codes_4xx: Message})
```
BCerki, line 45: We could probably refactor and reuse an existing endpoint, but since this is a temporary solution, I create a new one for ease of cleanup later.

client/app/components/layout/Dashboard.tsx
```diff
@@ -0,0 +1,156 @@
...
+            contentsModule = await import(
+              "@/app/data/dashboard/cas_analyst.json"
+            );
+            break;
+          case Roles.INDUSTRY_USER_ADMIN:
+            contentsModule = await import(
+              "@/app/data/dashboard/industry_user_admin.json"
+            );
+            contentsModule = updateContentsModule(
+              // @ts-ignore
```
BCerki, line 65: I had trouble with the typing--when I get the response from the api, it's in a different shape than what ends up being passed through the props here. Thoughts?

shon-button: Since the `userOperator: { userOperator: Status }` props informatiojn is required for only 2 roles I would opt not to create a component with "passing through props" that are not required by all dashboard users.
To that point, I would only make the call to `const getUserOperatorStatus = async () =>` for the cases that require that information (case Roles.INDUSTRY_USER_ADMIN:   case Roles.INDUSTRY_USER)
Finally, the `const updateContentsModule` is a good concept to dynamically translate the static json object but, 
as it seems the need for the message is based only on IF the user routes to `select-operator` and the user-operator is in Status.PENDING, I would opt to display the message in `client/app/components/routes/select-operator/Page.tsx` based on the result from calling `const updateContentsModule` in that way there is no need for `const updateContentsModule`  translations and  `const getUserOperatorStatus = async () =>` is called only when it is needed

BCerki: Yup, that's a way better idea, thanks for this!

client/app/components/routes/select-operator/Page.tsx
```diff
@@ -5,10 +5,28 @@ import { BC_GOV_LINKS_COLOR } from "@/app/styles/colors";
...
+      `registration/user-operator-status-from-user`,
+      "GET",
+      "",
+    );
+  } catch (error) {
+    throw error;
+  }
+};
+
+export default async function MyOperatorPage() {
```
BCerki, line 23: I made a tech debt card to consider renaming the route now that it does more than select operator: https://app.zenhub.com/workspaces/team-moose-sprint-board-60d216175a2689000ec774d0/issues/gh/bcgov/cas-registration/444 (I expect the approved/rejected workflows will go through this page too)

shon-button (approved):

PR #440, opened 2023-12-18
fix/437 user profile

marcellmueller:
- fix: use phone widget in user profile form
- fix: oversized submit buttons


BCerki (commented):
The phone number problem is fixed, but I'm seeing some big or misaligned buttons still:
image
image

Also, why is it better practice to wrap the button in a div where you're using it?

marcellmueller:
@BCerki thanks for being thorough, I should have checked other places it was used. Will fix. 

> Also, why is it better practice to wrap the button in a div where you're using it?

Considering that `classNames` prop wasn't being used anywhere it was adding an unneeded element to the DOM and creating visual bugs elsewhere. 

I think especially with small `Button` components and the like it generally makes more sense to keep them simple and just export the button. If a div with a margin or other styling is needed then the component can be wrapped in another div easily where it is being used.

client/app/components/form/SubmitButton.tsx
```diff
@@ -5,28 +5,23 @@ import { useFormStatus } from "react-dom";
...
 }) => {
   const { pending } = useFormStatus();
   return (
-    <div className={classNames}>
-      <Button
-        variant="contained"
-        type="submit"
-        aria-disabled={disabled || pending}
-        className="h-full"
-        sx={{ marginBottom: 10 }}
```
marcellmueller, line 25: @BCerki I think this `sx={{ marginBottom: 10 }}` is another great example of why basic reusable components like buttons shouldn't be wrapped in divs and have margins applied to them. This `marginBottom` was the cause of all those big buttons here and the `select operator` page and caused other weird issues.

Of course this isn't a hard rule and a best practice. There are exceptions such as something like a `Header` component where you may want the same margins throughout the page for consistency.

BCerki (approved):
:cartwheeling: :cartwheeling:

PR #365, opened 2023-12-01
BCEID protection

BCerki:
card: https://app.zenhub.com/workspaces/team-moose-sprint-board-60d216175a2689000ec774d0/issues/gh/bcgov/cas-registration/216

PR #442, opened 2023-12-18
421 ops table

BCerki:
card: https://app.zenhub.com/workspaces/team-moose-sprint-board-60d216175a2689000ec774d0/issues/gh/bcgov/cas-registration/421

Note: endpoint tests for this one will be done/updated in https://github.com/bcgov/cas-registration/pull/365/files#diff-7005d218482c3e6376907505440ee05d887e6364f099fff3f4a728d8309841f1

### DIFF
--- a/client/app/components/datagrid/Note.tsx
+++ b/client/app/components/datagrid/Note.tsx
@@ -7,10 +7,10 @@ interface Props {
 
 const Note = ({ message }: Props) => {
   return (
-    <>
+    <div>
       <strong>Note: </strong>
       {message}
-    </>
+    </div>
   );
 };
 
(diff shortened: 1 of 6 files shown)


client/app/components/datagrid/Note.tsx
```diff
@@ -0,0 +1,17 @@
...
+interface Props {
+  message: string;
+}
+
+const Note = ({ message }: Props) => {
+  return (
+    <>
+      <strong>Note: </strong>
+      {message}
+    </>
```
marcellmueller, line 13: Nice job, I think this PR is almost good to go though I think we could consider wrapping this in a div (likely need to use a flex box on it and the text might need to be wrapped as well). 

This works well on the operations page though using just a fragment leaves it at the mercy of the `display`/styling of the parent component and could cause issues in the future.

marcellmueller (approved):
🥇

Issue #422, opened 2023-12-13
first_name and last_name are swapped when saving IDIR users to the database [5]


dleard:
Describe the Bug:

When an idir user is saved to the database, the last name is entered in the first_name column (and includes the comma) and the first name is entered in the last_name column.

Probability (how likely the bug is to happen, scored from 1-5): 3

Effect (how bad the bug is when it does happen, scored from 1-5): 2

Steps to reproduce the behaviour:
Log in with your idir
use psql or pgadmin to view the database record
the first name and last name values are swapped (and the last name includes a comma)

Screenshots:
<img width="578" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeVJjQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--c94b05a692f6eb7655dac5a76139c6df0155ef74/Screenshot%20from%202023-12-13%2014-49-36.png" alt="Screenshot from 2023-12-13 14-49-36.png" />

PR #443, opened 2023-12-18
fix: idir names entered into the db incorrectly

marcellmueller:
Fix for #422 

This seems like something that should be unified in SSO app though I imagine they'd risk breaking a bunch of projects if they changed it. IDIR user names come in the format `lastname, firstname` instead of `firstname lastname` like `BCeiD` users so I added some logic to fix it.

### DIFF
--- a/client/app/components/routes/profile/User.tsx
+++ b/client/app/components/routes/profile/User.tsx
@@ -3,6 +3,7 @@ import UserForm from "@/app/components/routes/profile/form/UserForm";
 import { UserProfileFormData } from "@/app/components/form/formDataTypes";
 import { getServerSession } from "next-auth";
 import { authOptions } from "@/app/api/auth/[...nextauth]/route";
+import { IDP } from "@/app/types/types";
 
 // 🚀 API call: GET user's data
 async function getUserFormData(): Promise<
@@ -28,7 +29,8 @@ export default async function User() {
        * getServerSession requires passing the same object you would pass to NextAuth
        */
       const session = await getServerSession(authOptions);
-      const isIdir = session?.identity_provider === "idir";
+      const isIdir = session?.identity_provider === IDP.IDIR;
+
       // IDIR names come in the format "LASTNAME, FIRSTNAME" so we will split on the comma
       // and reverse so they don't go into the wrong fields
       const idirName = session?.user?.name?.split(", ").reverse();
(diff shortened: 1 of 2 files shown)


client/app/components/routes/profile/User.tsx
```diff
@@ -28,7 +28,13 @@ export default async function User() {
        * getServerSession requires passing the same object you would pass to NextAuth
        */
       const session = await getServerSession(authOptions);
-      const names = session?.user?.name?.split(" ");
+      const isIdir = session?.identity_provider === "idir";
+      // IDIR names come in the format "LASTNAME, FIRSTNAME" so we will split on the comma
+      // and reverse so they don't go into the wrong fields
+      const idirName = session?.user?.name?.split(", ").reverse();
```
BCerki, line 34: My IDIR includes my middle initial and ENV:INXT apparently: 
image
But I think that's ~probably something to do with the way mine is set up specifically

client/app/components/routes/profile/User.tsx
```diff
@@ -28,7 +28,13 @@ export default async function User() {
        * getServerSession requires passing the same object you would pass to NextAuth
        */
       const session = await getServerSession(authOptions);
-      const names = session?.user?.name?.split(" ");
+      const isIdir = session?.identity_provider === "idir";
```
BCerki, line 31: We have some enums to help with this check: /bc_obps/registration/enums/enums.py, e.g. of use: /bc_obps/registration/api/user.py

client/app/components/routes/profile/User.tsx
marcellmueller: Mine does as well, though mine is `ENV:EX` though I decided it was out of scope for this.

client/app/components/routes/profile/User.tsx
marcellmueller: Sounds good, though I don't think we should be importing anything from the backend even if we are a monorepo. I created an `IDP` type with the other types like we do for `Status` in `client/app/types/types.ts`.

client/app/components/routes/profile/User.tsx
BCerki: I think the point of this ticket was to stop bad data from going into the db, so if this is out of scope for this ticket, I think we need to check with Dylan about a second ticket. I'll make a placeholder tech debt one so we don't forget: https://app.zenhub.com/workspaces/team-moose-sprint-board-60d216175a2689000ec774d0/issues/gh/bcgov/cas-registration/447

BCerki (approved):

Issue #441, opened 2023-12-18
As an IO user, when I click the My Operator tile, I see the appropriate page depending on my status


BCerki:
#### Description:

Currently, whenever an industry user clicks on the "My Operator" tile, they're sent to the select-your-operator workflow. They should see a different message or start a different workflow depending on their status.

#### Acceptance Criteria:

Given I am an IO user
When my access to my operator is approved and I click the "My Operator" tile
Then I see a readonly version of the user operator form
And I see an Edit button
And when I click the edit button, I can edit the user operator form
(wireframe: https://www.figma.com/file/LsDmLDCdnJqI8UahyqvOD8/OBPS-App-Designs?type=design&node-id=2414%3A1918&mode=design&t=xCpgH6PJIretcBQL-1)

Given I am an IO user
When my access to my operator is draft or rejected and I click the "My Operator" tile
Then I see the select-your-operator workflow (already done, just recording here for clarity and to ensure no bugs are introduced)

Given I am an IO user
When my access to my operator is pending and I click the "My Operator" tile
Then I see a pending message (covered in https://github.com/bcgov/cas-registration/pull/439)

#### Development Checklist:

- [x] Add conditional logic to `dashboard/select-operator` to show the user operator form for approved users
- [ ] Meets the DOD

**Definition of Ready** (Note: If any of these points are not applicable, mark N/A)

- [ ] User story is included
- [ ] User role and type are identified
- [ ] Acceptance criteria are included
- [ ] Wireframes are included (if required)
- [ ] Design / Solution is accepted by Product Owner
- [ ] Dependencies are identified (technical, business, regulatory/policy)
- [ ] Story has been estimated (under 13 pts)

·**Definition of Done** (Note: If any of these points are not applicable, mark N/A)

- [ ] Acceptance criteria are tested by the CI pipeline
- [ ] UI meets accessibility requirements
- [ ] Configuration changes are documented, documentation and designs are updated
- [ ] Passes code peer-review
- [ ] Passes QA of Acceptance Criteria with verification in Dev and Test
- [ ] Ticket is ready to be merged to main branch
- [ ] Can be demoed in Sprint Review
- [ ] Bugs or future work cards are identified and created
- [ ] Reviewed and approved by Product Owner

#### Notes:

- Blocked by #285 
- Blocking #325

hannavovk:
[...]

Issue #285, opened 2023-11-16
Applications with a "Pending" status should not be editable by industrial users


andrea-williams:
#### Description:

Applications (for both operators and operations) should not be editable when their status = "Pending". PR in progress you can continue if it's helpful: https://github.com/bcgov/cas-registration/pull/423

#### Acceptance Criteria:

Given an operation has a status of "Pending"
When an industrial user views the form data for their pending operation
Then all the data fields are view-only (can't be edited)

Given an operator has a status of "Pending"
When an industrial user views the form data for their pending operator
Then all the data fields are view-only (can't be edited)

NOTE: I'm guessing that CAS staff should never be able to edit operators/operations directly - needs to be confirmed. UPDATE Nov 27: Hanna Confirmed with business area: CAS staff should not be able to edit operators/operations directly 

#### Dependencies
- Blocked by #25 

#### Development Checklist:

- [ ] readOnly flag on Operator and Operations forms is set depending on (a) assigned role of user AND (b) whether the application's `status == PENDING`. 
- [ ] if status == "Not Registered", form should be editable by industrial users with the appropriate permissions.
- [ ] if Operation or Operator status == "Pending", it is not editable by industrial users
- [ ] if Operation or Operator status == "Approved", it is editable
- [ ] if Operation or Operator status == "Rejected", it is editable by industrial users
- [ ] if Operation or Operator status == "Changes Requested", it is editable by industrial users
- [ ] Meets the DOD

**Definition of Ready** (Note: If any of these points are not applicable, mark N/A)

- [ ] User story is included
- [ ] User role and type are identified
- [ ] Acceptance criteria are included
- [ ] Wireframes are included (if required)
- [ ] Design / Solution is accepted by Product Owner
- [ ] Dependencies are identified (technical, business, regulatory/policy)
- [ ] Story has been estimated (under 13 pts)

·**Definition of Done** (Note: If any of these points are not applicable, mark N/A)

- [ ] Acceptance criteria are tested by the CI pipeline
- [ ] UI meets accessibility requirements
- [ ] Configuration changes are documented, documentation and designs are updated
- [ ] Passes code peer-review
- [ ] Passes QA of Acceptance Criteria with verification in Dev and Test
- [ ] Ticket is ready to be merged to main branch
- [ ] Can be demoed in Sprint Review
- [ ] Bugs or future work cards are identified and created
- [ ] Reviewed and approved by Product Owner

#### Notes:

-

Dianadec:
[...]

PR #478, opened 2023-12-22
285 disable form pending status

marcellmueller:
Edit: waiting until #441 is complete so we can easily test disabled pending status for industry users

Implements #285 

The requirements in the ticket look more complex than it is, I believe as of now we just need to disable the form for CAS admin that doesn't have `cas_pending` as role and also for anyone if the form status is pending.

To whoever reviews this I don't think we need to add disabled state to the single page `UserOperatorForm`. I believe that when that form is submitted and viewed again as an industry user (once #441 is complete) or as a CAS admin/analyst then they will view the multistep form. 

Let me know if I am missing something and I will update that form as well!

### DIFF
--- a/client/app/components/form/UserOperatorMultiStepForm.tsx
+++ b/client/app/components/form/UserOperatorMultiStepForm.tsx
@@ -10,7 +10,7 @@ import UserOperatorReview from "@/app/components/routes/access-requests/form/Use
 import MultiStepFormBase from "@/app/components/form/MultiStepFormBase";
 import { UserOperatorFormData } from "@/app/components/form/formDataTypes";
 import Note from "../datagrid/Note";
-import { Status } from "@/app/types/types";
+import { Status } from "@/app/utils/enums";
 
 interface UserOperatorFormProps {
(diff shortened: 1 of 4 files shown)


client/app/components/form/UserOperatorMultiStepForm.tsx
```diff
@@ -95,9 +98,12 @@ export default function UserOperatorMultiStepForm({
         )}
         <MultiStepFormBase
           baseUrl={`/dashboard/operators/user-operator/${userOperatorId}`}
-          cancelUrl="/dashboard/operators"
+          // TODO: change cancelUrl for industry user depending on if they are creating or editing an operator
+          cancelUrl={`/dashboard/${
+            isCasInternal ? "operators" : "select-operator"
```
marcellmueller, line 103: This fix is unrelated to this PR though I noticed the `Cancel` button was broken for industrial users. Once #441 is in  we will likely need to make the link go elsewhere (maybe just the dashboard?) depending on form status.

Issue #474, opened 2023-12-21
Reject a prime admin request if an operator is rejected


Dianadec:
#### Describe the task

[#376] separates the operator/user approval into a two-step process. ~~From [#376], it looks like if the user rejects the operator then they should not be given the option to approve the user. Do we have designs for this? We could disable the 'next' button on the multistep form, but the user approval page is still accessible via url.~~

#### Acceptance Criteria

When a CAS user rejects a new operator request:
- [x] the prime admin request is rejected too (useroperator status set to rejected when we hit `@router.put("/operators/{operator_id}"`), including setting the metadata to capture which user rejected the request and timestamp
- [x] the approve/deny buttons are hidden on all of the multistep form pages
- [x] the multistep form stays readonly
- [x] implement a check on the useroperator's update-status endpoints to prevent updating the status of a useroperator if is_new is set to True

#### Additional context

- blocked by #376 
- blocked by #600 due to potential merge conflicts.

Issue #376, opened 2023-12-06
Flag when a new Operator has been added to the database


andrea-williams:
#### Describe the task

When a user creates a brand new Operator (i.e., not one that already existed in our database), we want to add a boolean flag to the record in the database indicating that it's new. This "is_new" flag will indicate whether CAS internal staff need to explicitly approve the Operator at the same time that they're approving the request for prime admin access.

(Operators imported from CIIP will have `is_new` flag set to `false`, ~~and will automatically have their `status` set to `Approved`~~ when the data is imported).

#### Acceptance Criteria

- [x] when an internal user is reviewing a prime admin access request, if the Operator info that they're viewing at the same time (https://www.figma.com/file/LsDmLDCdnJqI8UahyqvOD8/OBPS-App-Designs?type=design&node-id=413%3A48&mode=design&t=b0MX1lfLnGp4Rch3-1) has the `is_new = True` flag, the internal user will have to explicitly Approve the Operator before the prime admin request can be approved. If the Operator is not new (i.e., `is_new = False`, the internal user will not need to change the Operator's status.
- [x] once an Operator's status has been set to Approved, the `is_new` flag can be set to `False`
- [x] if the internal user Requests Changes for the new Operator, the `is_new` flag will remain as `True` until the Operator's status is changed to Approved
- [x] add a very descriptive comment to the `is_new` column on the Operator table to avoid developer confusion (most specifically, what the difference is between the `is_new` flag and the `status` column)

#### Additional context

- Blocked by #25

andrea-williams:
@nanyangpro  do you plan to update the wireframes with some sort of marker so that internal users can tell when an Operator is brand new?

nanyangpro:
Yes, @andrea-williams. I'll add the link once done. Before I proceed, I've got a [question](https://www.figma.com/file/LsDmLDCdnJqI8UahyqvOD8?node-id=413:48&mode=design#637079516) for you in Figma, could you please review and confirm there? Thanks!

Dianadec:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Team-Moose-Sprint-Board-60d216175a2689000ec774d0/issues/bcgov/cas-registration/376?planning-poker) with Zenhub @andrea-williams @BCerki @dleard @JoshLarouche @marcellmueller @Sepehr-Sobhani @shon-button

dleard:
[...]

PR #489, opened 2024-01-03
376 new operator approval

BCerki:
https://github.com/bcgov/cas-registration/issues/376

Notes:
- there are comments in the ticket about approach and scope
- re this part of the AC: 'the internal user will have to explicitly Approve the Operator before the prime admin request can be approved.' This PR allows separate approval of the operator, which is on the first step of the form, before the admin request approval. However, this PR doesn't stop an IRC user from continuing on if they reject the operator. Whatever that will look like will be handled here: https://github.com/bcgov/cas-registration/issues/474


bc_obps/registration/api/operator.py
```diff
@@ -81,4 +87,21 @@ def list_user_operators_status_of_operator(request, operator_id: int):
 ##### PUT #####
 
 
+@router.put("/operators/{operator_id}", response={200: OperatorOut, codes_4xx: Message})
+def update_operator(request, operator_id: int, payload: OperatorIn):
+    raise_401_if_role_not_authorized(request, ["cas_admin", "cas_analyst"])
+    operator = get_object_or_404(Operator, id=operator_id)
+    user: User = request.current_user
+    operator.status = payload.status
+    if operator.status in [Operator.Statuses.APPROVED, Operator.Statuses.REJECTED]:
```
BCerki, line 96: Re this AC: if the internal user Requests Changes for the new Operator, the is_new flag will remain as True until the Operator's status is changed to Approved

`is_new` defaults to true, so we don't have to do anything besides not include the CHANGES status to implement this AC

bc_obps/registration/models.py
```diff
@@ -343,6 +344,10 @@ class Statuses(models.TextChoices):
         null=True,
         related_name="operators_verified_by",
     )
+    is_new = models.BooleanField(
```
BCerki, line 347: Is this clear enough?

client/app/components/form/UserOperatorMultiStepForm.tsx
```diff
@@ -83,14 +81,26 @@ export default function UserOperatorMultiStepForm({
...
+              operatorId={formData?.id}
+            />
+          </>
+        )}
+        {isCasInternal && formSection === 1 && (
           <UserOperatorReview
             userOperator={formData as UserOperatorFormData}
             userOperatorId={Number(userOperatorId)}
+            isOperatorNew={formData?.is_new}
+            operatorId={formData?.id}
```
Sepehr-Sobhani, line 103: nit: suggest using `operator_id` rather than `id` as it would be more meaningful.

client/app/components/form/UserOperatorMultiStepForm.tsx
```diff
@@ -83,14 +81,26 @@ export default function UserOperatorMultiStepForm({
...
+            <Note message="This is a new operator. You must approve this operator before approving its admin." />
+            <UserOperatorReview
+              userOperator={formData as UserOperatorFormData}
+              userOperatorId={Number(userOperatorId)}
+              isOperatorNew={formData?.is_new}
+              operatorId={formData?.id}
+            />
+          </>
+        )}
+        {isCasInternal && formSection === 1 && (
```
Sepehr-Sobhani, line 98: Just to confirm, should we only display review buttons to approve an admin request on the second page?

Sepehr-Sobhani (commented):
Nice work! 
I have a couple of suggestions and a question. Additionally, I'd suggest implementing a check on the useroperator's `update-status` endpoints to prevent updating the status of a useroperator if `is_new` is set to True. (just an extra check)

client/app/components/form/UserOperatorMultiStepForm.tsx
BCerki: If it's a existing operator, yes, we should only display the buttons on the second page.

bc_obps/registration/api/operator.py
```diff
@@ -81,4 +87,21 @@ def list_user_operators_status_of_operator(request, operator_id: int):
...
+        operator.is_new = False
+        operator.verified_at = datetime.now(pytz.utc)
+        operator.verified_by_id = user.user_guid
```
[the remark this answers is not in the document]
BCerki: changing it caused this: django.core.exceptions.ValidationError: ['“User object (4da70f32-65fd-4137-87c1-111f2daba3dd)” is not a valid UUID.']

BCerki:
> Nice work! I have a couple of suggestions and a question. Additionally, I'd suggest implementing a check on the useroperator's `update-status` endpoints to prevent updating the status of a useroperator if `is_new` is set to True. (just an extra check)

I'm going to add this to the ticket about handling a rejected operator

client/app/components/button/Review.tsx
```diff
@@ -46,6 +46,7 @@ const Review = ({
   onApprove,
   onReject,
 }: Readonly<Props>) => {
+  console.log("am i in review");
```
andrea-williams, line 49: delete?

client/app/components/button/Review.tsx
```diff
@@ -89,11 +90,13 @@ const Review = ({
     setDismissAlert(true);
   };
 
+  console.log("isStatusPending", isStatusPending);
```
andrea-williams, line 93: delete?

client/app/components/button/Review.tsx
```diff
@@ -89,11 +90,13 @@ const Review = ({
...
   const isReviewButtons =
     isStatusPending &&
     errorList.length === 0 &&
     successMessageList.length === 0;
 
+  console.log("isReviewButtons", isReviewButtons);
```
andrea-williams, line 99: delete?

client/app/components/routes/access-requests/form/UserOperatorReview.tsx
```diff
@@ -8,13 +8,50 @@ import { UserOperatorFormData } from "@/app/components/form/formDataTypes";
...
 }
 
 export default function UserOperatorReview({
   userOperator,
   userOperatorId,
+  operatorId,
+  isOperatorNew,
 }: Props) {
-  async function approveRequest() {
+  console.log("where are my buttons");
```
andrea-williams, line 21: 🤭
delete please

andrea-williams (commented):
Aside from some surplus console logs, code looks good. Will pull and run the code locally now

BCerki:
Note to B: check the mock data--senior officer is missing for user operator 4

andrea-williams:
Ran the code locally - confirmed that internal users can now approve an Operator and can independently approve/reject a prime admin request. 

Some notes for discussion next week (probably separate tickets outside the scope of this one):

- the first section of the Operator form has Approve/Reject buttons that appear when it's a new Operator that needs to be reviewed. IMO it's not super clear that when an internal user is clicking those buttons, they're approving the Operator. Similarly, the second section of the form has Approve/Reject buttons for reviewing the prime admin access request, but it's especially unclear in that case because the form only shows information about the Senior Officer for the Operator, who might be different from the user requesting prime admin. There's no information about the user requesting prime admin shown on that page, it's only shown in the table of all Operators. I think we should be doing something like adding a third section that displays info about the user requesting prime admin and have the Review component there.
- there doesn't appear to be any checks on the backend for missing fields on the Senior Officer data - I was able to "approve" even though the data was empty.
- when the Operator is_new=True, our UI says that internal staff must approve the operator before they can approve the prime admin request, but actually an internal user can click the "Next" button in the form and approve/reject prime admin without approving the new Operator. We really shouldn't allow this - if an operator has an assigned admin, we need the operator to be approved.

client/app/components/routes/access-requests/form/UserOperatorReview.tsx
BCerki: (sorry about this)

andrea-williams (approved):

Issue #491, opened 2024-01-03
Request access as an administrator link reflects a non existent route


shon-button:
Describe the Bug:

Link `Request access as an administrator` routes to `/dashboard/select-operator/user-operator/operator/${response.user_operator_id},` but the app router folder structure does not have folder `(authenticated)/bceidbusiness/industry_user_admin/dashboard/select-operator/user-operator/operator/`

Probability (how likely the bug is to happen, scored from 1-5): 5

#### Dev Checklist:
- [ ] In `client/app/components/button/RequestAccessButton.tsx` 
**CHANGE**
```
if (isAdminRequest)
      push(
        `/dashboard/select-operator/user-operator/operator/${response.user_operator_id}`,
      );
```
**TO**
```
if (isAdminRequest)
      push(
        `/dashboard/select-operator/user-operator/${response.user_operator_id}`,
      );
```

#### Acceptance Criteria:

Given I am an external user logged in to my Business BCeID account
When I click my dashboard navigation link "My Operator" and NO prime admin/user operator request exists
Then I should be directed to the "Select Operator" screen
And then I should be able to select an Operator without a Prim Admin from the input search field
And then I should be able to confirm my Operator select
And then I should be able to click link "Request access as an administrator"
So that I can fill out the request form

Issue #490, opened 2024-01-03
Can't navigate to second page of operations form as role:  cas_admin


shon-button:
Describe the Bug:

**Issue:** User with role `cas_admin` is authorized to view `client/app/components/form/OperationsForm.tsx` but, cannot navigate to the next form section using the `Next` button
**Root Cause:** `Next` button submits the form which calls endpoint `@router.get("/user-operator-operator-id"` which has role-based authorization decorator not including `cas_admin` `@authorize(AppRole.get_industry_roles())` returning `Error: 401` 
**Notes:**
- The next form section can be accessed using browser url navigation e.g. `http://localhost:3000/dashboard/operations/3/2` as this bypasses the form submit and call to `/user-operator-operator-id`

![image](https://github.com/bcgov/cas-registration/assets/120038448/d8eff083-ece9-4df0-a499-357e0d29e244)

PR #492, opened 2024-01-03
fix: all user roles can navigate the multistep forms

BCerki:
https://github.com/bcgov/cas-registration/issues/490

I manually tested submitting and viewing all the forms (except this currently broken one: https://github.com/bcgov/cas-registration/issues/491) since changing the multistep component affects a lot of the app.

Backend tests are failing because it's 2024 now. They're fixed in https://github.com/bcgov/cas-registration/pull/489 (and possibly other PRs) that will be merged in before this one, so they'll start passing once I rebase.

### DIFF
--- a/client/app/components/form/MultiStepButtons.tsx
+++ b/client/app/components/form/MultiStepButtons.tsx
@@ -11,7 +11,6 @@ interface SubmitButtonProps {
   classNames?: string;
   step: number;
   steps: string[];
-  submitEveryStep?: boolean;
   allowBackNavigation?: boolean;
 }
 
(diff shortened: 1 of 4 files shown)


client/app/components/form/MultiStepButtons.tsx
```diff
@@ -54,33 +51,25 @@ const SubmitButton: React.FunctionComponent<SubmitButtonProps> = ({
...
           ))}
-        {!isFinalStep && !submitEveryStep && (
-          <Link href={`${baseUrl}/${step + 2}`}>
-            <Button variant="contained" type="button" disabled={isFinalStep}>
-              Next
-            </Button>
-          </Link>
-        )}
-        {isFinalStep && !submitEveryStep && (
+        {/* When the form is editable, the form should be submitted when navigating between steps */}
```
BCerki, line 54: I removed the `submitEveryStep` prop because I can't think of a case where we'd want to not submit. If we didn't submit the first step then the operation wouldn't be created, and submitting the second step would break.

marcellmueller: Makes sense 👍  In the future if we need a reason to prevent form submission I think we could add something like a `disableSubmit` prop.

marcellmueller (approved):
Nice job, this is a solid improvement 🥇

PR #499, opened 2024-01-04
Fix/fixtures do not load

dleard:
After splitting the fixtures into the subdirectories 'mock' and 'real', call_command('load_fixtures') was broken because it's only looking for fixtures directly in the fixtures/ directory. 

The fix: ~Updated the call_command to use 'loaddata' instead of 'load_fixtures' and passed a set of glob patterns to read all fixtures in the subdirectories.~ Updated the pathing in the load_fixtures.py function

### DIFF
--- a/bc_obps/registration/api/api_base.py
+++ b/bc_obps/registration/api/api_base.py
@@ -15,11 +15,7 @@ def setup(request):
     if settings.ENVIRONMENT == "develop":
         try:
             call_command('truncate_all_tables')
-            call_command(
-                'loaddata',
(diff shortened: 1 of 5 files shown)


bc_obps/registration/api/api_base.py
Sepehr-Sobhani, line 1: I think we should fix the `load_fixtures.py` command instead of this. we need to change: `cmd_args = list(pathlib.Path().glob('./registration/fixtures/*'))` ====> `cmd_args = list(pathlib.Path().glob('./registration/fixtures/*/*.json'))`

Sepehr-Sobhani (approved):
Thanks for the fix! Just a quick change is required otherwise looks great!

Issue #332, opened 2023-11-24
Exploration Task: Document Uploads


andrea-williams:
#### Describe the task

We want to investigate how we can handle document upload and storage to our Googley cloud bucket considering our combined tech stack of: Django, Django-Ninja, and RJSF.

#### Acceptance Criteria

- [ ] build a Proof-of-Concept RJSF form with a single field that accepts a file upload (we only care about PDF format), on Submit button click passes the doc to our backend API, which then stores the file in our bucket and updates our database (local DB is good enough) with the identifier needed to retrieve the doc from Google.

#### Additional context

- As this is simply an exploration task, don't need to implement document upload into our existing forms for Operations, Operators, etc. Those are separate tickets.
- Blocks #325

andrea-williams:
confirming with @hannavovk - we only need to handle PDF files for document upload, correct?

shon-button:
FYI:
I have implemented an upload to Google Cloud Storage feature using Next.js API, and then updating the PostgreSQL database with the file's bucket URL.

andrea-williams:
Advice received from a DDS dev on another team (who has experience with RJSF):
```
in general i would avoid trying to get rjsf to do anything other than the usual fields on form + validation.  just asking for pain.  i would wrap the form in a component so it just has the job of being form, and then add another component to the page to handle getting the file from the user into state for an api call

you can use css to make it match the form style thats going to be less headache than trying to get rjsf to do anything to do with file upload and api calls

do the api call to upload the file to s3 or wherever, get a media key, and if you need to include that in the form blob do so otherwise just link to the form record in the file upload call (whos job would be upload to s3, get media key, put key in db somewhere linked to your record)
```

Dianadec:
Hanna is clarifying the extent of document uploads with business area

hannavovk:
Let's hold back on actually implementing document uploads. It's looking like we potentially may be able to do without it for now.

PR #507, opened 2024-01-05
332 document exploration

BCerki:
https://github.com/bcgov/cas-registration/issues/332

At this point, we can attach a file in a simple form, send it to the backend, and save it into the db and google bucket. We can also retrieve the file on the front-end and download it. Here's a loom of the exploration: https://www.loom.com/share/5e778beb18b64a73a92399614894653b

Learnings on file upload:
- the rjsf file widget stores a [dataurl](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URLs) in the `formData`
- Django's `models.FileField` requires a file object. When you Document.objects.create(file=the ContentFile I made from the dataurl), django will put the file into the google bucket
- I decided to do the dataurl-to-file conversion in the endpoint and django ninja schemas. In this POC the form is very simple, but in the real app we're going to have multiple files in each form as well as form data, so the backend ~seems to be the most flexible. (I played with middleware but with full form data it would be a lot more complicated, and I'm not sure I successfully transformed the file anyway)

To test this PR:
- add GS_BUCKET_NAME and GOOGLE_APPLICATION_CREDENTIALS to your env (see 1password_
**To mimic an industry user:**
- login to the app with your idir
- navigate to `http://localhost:3000/brianna`
- upload a file and submit the form (doesn't matter if you fill out the extra random field)
- psql into the db and see that your file is in the Document table
- go into google storage and see that your file is in the bucket
**To mimic an IRC user who is reviewing the form:**
- go back to the form in the browser and refresh. your file is still there and you can

### DIFF
--- a/bc_obps/registration/api/naics.py
+++ b/bc_obps/registration/api/naics.py
@@ -1,7 +1,8 @@
 from registration.decorators import authorize
 from .api_base import router
+from django.core.files.base import ContentFile
 from typing import List
-from registration.models import NaicsCode, AppRole, Document
+from registration.models import NaicsCode, AppRole, Document,  DocumentType
 from registration.schema import (
     NaicsCodeSchema,
 )
@@ -10,6 +11,11 @@ from ninja.files import UploadedFile
 from ninja import ModelSchema
 from ninja.responses import codes_4xx, codes_5xx
 from registration.schema import Message, OperatorOut, SelectUserOperatorStatus
+from ninja import Schema
+import base64
+import io
+import json
+import re
 class DocumentOut(ModelSchema):
     """
     Schema for the Operator model
@@ -19,15 +25,38 @@ class DocumentOut(ModelSchema):
         model = Document
         model_fields = '__all__'
 
-@router.post("/upload", response={201: DocumentOut, codes_4xx: Message})
-async def create_document(request, file: UploadedFile = File(...)):
-    print('made it into the upload endpoint')
-    # i think is not actually the file
-    print('file',file)
-    data = file.read()
-    print('data',data)
-    breakpoint()
-    document = Document.objects.create(**data)
+# could use middleware for the transformation too? harder when it's part of a form
+class DocumentSchema(Schema):
+    boundary_map: str
+
+
+# @router.post("/upload", response={201: DocumentOut, codes_4xx: Message})
+# async def create_document(request, file: UploadedFile = File(...)):
+#     print('made it into the upload endpoint')
+#     # i think is not actually the file
+#     print('file',file)
+#     data = file.read()
+#     print('data',data)
+#     breakpoint()
+#     document = Document.objects.create(**data)
+
+
+#     return 201, document
+
+@router.post("/handle-file", response={201: DocumentOut, codes_4xx: Message})
+def create_document(request, payload: DocumentSchema):
+
+# for the real endpoint, we'll probably have to do some sort of loop because there can be multiple types of documents in each form
+
+    file_name =  re.search(r'name=([^;]+)', payload.boundary_map).group(1)
+    _, encoded_data = payload.boundary_map.split(',')
+
+    # Decode the base64-encoded data
+    file_data = base64.b64decode(encoded_data)
+    file = ContentFile(file_data, file_name)
+
+ 
+    document = Document.objects.create(file=file,  type=DocumentType.objects.get(name='boundary_map'), description='what was this meant to be? maybe it would make more sense to put the description in the DocumentType table?')
 
 
     return 201, document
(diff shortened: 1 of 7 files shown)


bc_obps/registration/api/naics.py
```diff
@@ -1,10 +1,66 @@
 from registration.decorators import authorize
```
BCerki, line 1: did this in the naics file because i forgot how to set up a brand-new endpoint

bc_obps/registration/api/naics.py
```diff
@@ -1,10 +1,97 @@
...
+    file_name = re.search(r'name=([^;]+)', payload.boundary_map).group(1)
+    _, encoded_data = payload.boundary_map.split(',')
+
+    # Decode the base64-encoded data
+    file_data = base64.b64decode(encoded_data)
+    file = ContentFile(file_data, file_name)
+
+    document = Document.objects.create(
+        file=file,
+        type=DocumentType.objects.get(name='boundary_map'),
```
BCerki, line 89: brianna will need to attach the operation, remember to add the foreign keys

BCerki:
Some more considerations:

[Yesterday 12:10] Marcel Mueller
I may have missed something but it doesn't look like this actually uploads a file like the acceptance criteria asks for, just saves the file metadata in the database? I think we need to save some kind of uuid or key as well so we know which file to download again.
 
I still think we should upload the file to the bucket as soon as they attach it to the file widget, then do an RJSF onchange and update the formdata once the file has uploaded and after that save the file metadata in the database on form submission. This will result in some orphan/dangling files in the bucket, though will mitigate issues on form submission especially if there are multiple upload fields.
 
If there are multiple uploads and one fails then the form won't submit and could cause a lot of headaches. Just my 2 cents though.
[Yesterday 12:20] Brianna Cerkiewicz
> it doesn't look like this actually uploads a file like the acceptance criteria asks for, just saves the file metadata in the database?
The file field in the Document tables is a Django FileField, and django takes care of putting the file into the google bucket for us. There are instructions in the PR about how to confirm this is working (check the bucket, or download the file after it's been uploaded).
 like 1
[Yesterday 12:27] Marcel Mueller
Cool, thanks for the explanation I just had a quick look  I still think we should consider sending the file to the API then when it is attached to the file widget for reasons I mentioned above. Then on submission we can confirm that the document is intended to be attached to that form.
[08:26] Brianna Cerkiewicz
> I still think we should upload the file to the bucket as soon as they attach it to the file widget, then do an RJSF onchange and update the formdata once the file has uploaded and after that save the file metadata in the database on form submission. This will result in some orphan/dangling files in the bucket, though will mitigate issues on form submission especially if there are multiple upload fields.
We're using Django to upload files to the bucket, so a file has to go into our Document table in order to be uploaded. (We could find another way to put files in the bucket, but that would require another exploration; Django was already halfway set up for file upload months ago). We could still put the file into the db onchange instead of onsubmit, but I have some UX concerns there:
we're not telling users we're doing it (vs them choosing to click the "submit" button after they've had a chance to redownload the file and make sure they didn't accidentally upload their bank statement)
we don't have hard delete, so if they do upload the wrong thing, they can't do anything about it
[08:33] Marcel Mueller (Guest)
Well no matter which way we go we would have to add some kind of loading spinner to the button (either the file widget or the submit button) to let the user know the page isn't frozen if they are on a slow connection.
 
I just have horror stories from ccbc of failed uploads and I don't think designing a form assuming everyone has great internet connections is necessarily good practice.
 
Anyways I would think we could archive/delete any associated files with that form once it was submitted and get rid of orphans in the database/maybe the bucket if the middleware support it.
 
Either way I don't feel super strongly about it, just my opinion.
[09:24] Marcel Mueller (Guest)
https://github.com/bcgov/cas-registration/issues/528
 
I just made a bug fix ticket that I think will be very important if are uploading all the files on form submission since I noticed the form state seems to clear if there is an API error. Should be an easy fix at least.
[13:23] Marcel Mueller (Guest)
I tested this and it seems to be working very well, nice work! I don't know how to log into google to verify but everything else seems to be working very well and I downloaded the file I had uploaded. Not sure if I need to approve the actual PR since it's not being merged and you know my thoughts on when I think we should upload so I will just say 'Approved' 
[13:27] Brianna Cerkiewicz
Cool, thanks. If you want google powers I think Gamache, Joshua ENV:EX is the one to talk to. I'm still thinking on when to upload, hoping for a couple more people to chime in too
 heart 1
[13:29] Marcel Mueller (Guest)
Hey one question I have looking at my backend server logs I noticed the very lengthy file url it saves has an 'Expires=' parameter. Is that going to be an issue? Or will Google send us back a valid URL to download from if we try and access the file in a year or something.
[13:32] Marcel Mueller (Guest)
Looks like it's some kind of signed url it sends back when I try to download it and not actually saved in the database. Likely I am just missing something about how this all works though if anyone knows the answer that would be helpful thanks.

BCerki:
Closing, exploration complete. A link to this PR is in the document upload ticket for reference

Issue #413, opened 2023-12-12
Logout function doesn't allow logging back in with another account


dleard:
Describe the Bug:

Logging in to the app with one account, logging out and then logging back in with another account throws an error from keycloak saying that you are already signed in.

This will be a problem for all contractors that report on behalf of multiple operators as they need to sign in/out of different accounts depending on who they are reporting for. There are a couple workarounds, though they aren't a great look if we have to tell our users who have multiple accounts to either:
1. Use a private browser and close/reopen the browser between each session
2. Manually remove the site cookies between each session 

Steps to reproduce the behaviour:

- Log in with one bceid account
- Log out
- Log in with a different bceid account
- See error message from keycloak
- 

#### Investigations

**Validating that current federated authentication works:**
- Log in with one bceid account
- Verify Keycloak login status: https://dev.loginproxy.gov.bc.ca/auth/realms/standard/account/#/
![Image](https://github.com/bcgov/cas-registration/assets/120038448/689f918d-04b3-4c61-bcb6-d663ca9eca21)

- Log out
- Verify Keycloak logout status: https://dev.loginproxy.gov.bc.ca/auth/realms/standard/account/#/
![Image](https://github.com/bcgov/cas-registration/assets/120038448/0542c068-ff11-425b-8ce7-6f4491b75825)

**Troubleshooting**

- Log in with a different bceid account
- See error message from keycloak
```
An unexpected error has occurred or you are already logged into IDIR or BCeID and you need to logout or use a fresh browser to continue. Visit here to learn more.
```
Following the link in the error message
https://github.com/bcgov/sso-keycloak/wiki/Our-Partners-and-Useful-Information#common-login-errors

```
IDIR and BCeID in the same browser
As we partner with the BC Gov Identity Partners of IDIR and BCeID please note in the same browser, you cannot have one tab logged in with IDIR and another with BCeID.

Please use a private browser by either using incognito or clearing your cache.

Other issues
Please ensure you have tested with an incognito browser as mentioned above. If it is still an issue, reachout to use on rocketchat
```

**Workarounds,**
1. Use a private browser and close/reopen the browser between each session 
**SH:** works if there is only 1 anonymous browser open
2. Manually remove the site cookies between each session 
[...]

PR #515, opened 2024-01-08
Fix: Logout bug/413

andrea-williams:
Card: #413 

Fix: fixed logout bug by chaining logout urls; updated client/.env.example

Can now log in & out of multiple BCeID/IDIR accounts without having to open Incognito windows :)


dleard (approved):
Huge bug fix! :100:

PR #516, opened 2024-01-08
Fix 498 prime admin request when operator exists

Sepehr-Sobhani:
[ZH CARD 498](https://app.zenhub.com/workspaces/team-moose-sprint-board-60d216175a2689000ec774d0/issues/gh/bcgov/cas-registration/498)

### DIFF
--- a/client/app/components/form/UserOperatorForm.tsx
+++ b/client/app/components/form/UserOperatorForm.tsx
@@ -2,7 +2,7 @@
 
 import { RJSFSchema } from "@rjsf/utils";
 import { useState } from "react";
-import { useParams, useRouter, useSearchParams } from "next/navigation";
+import { useParams, useRouter } from "next/navigation";
 import { userOperatorUiSchema } from "@/app/utils/jsonSchema/userOperator";
 import { actionHandler } from "@/app/utils/actions";
 import {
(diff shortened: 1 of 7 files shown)


client/app/components/form/UserOperatorMultiStepForm.tsx
```diff
@@ -84,7 +86,7 @@ export default function UserOperatorMultiStepForm({
     session?.user.app_role?.includes("cas") &&
     !session?.user.app_role?.includes("pending");
 
-  if (isCasInternal || !userOperatorId || formSection) {
+  if (isCasInternal || !userOperatorId) {
```
marcellmueller, line 89: I was testing this and the `formSection` is required or else page 2 of the multistep form shows the `UserOperatorContact` form. This will also be problematic when `CAS` users review the forms.

I think what we might be able to do is keep this logic the same and make a separate `userOperatorIdSearchParam` or something to use in page 2 of the form. I did something similar in my incoming PR here: 

https://github.com/bcgov/cas-registration/blob/e0a731e8f39a7a71faea7024ff0915d6ccd90a81/client/app/components/form/UserOperatorMultiStepForm.tsx

marcellmueller: Or on second thought maybe keep the param code exclusive to the `UserOperatorContactForm` ?

client/app/components/form/UserOperatorForm.tsx
```diff
@@ -2,7 +2,7 @@
 
 import { RJSFSchema } from "@rjsf/utils";
```
marcellmueller, line 3: Nice idea renaming this. Recommend renaming this file to `UserOperatorContactForm` to make it easier to find.

client/app/components/form/UserOperatorMultiStepForm.tsx
Sepehr-Sobhani: Thank you for capturing this before it becomes a new bug :)

client/app/components/routes/select-operator/form/UserOperator.tsx
```diff
@@ -25,8 +29,8 @@ async function getBusinessStructures() {
   );
 }
 
-export async function getUserOperatorFormData(id: number | undefined) {
-  if (!id) return {};
+export async function getUserOperatorFormData(id: number | string | undefined) {
+  if (!id || id === "request-access") return {};
```
marcellmueller, line 33: Nice work on this, I think it will be a lot safer. This is the only thing I'm thinking about for when my pr merges as I am doing the same but `id === 'create'`. I think I will approve this and might just refactor it and do a NaN(Number(id)) or some similar check to simplify 👍

Sepehr-Sobhani: Sounds good, Thanks 🙏

PR #518, opened 2024-01-08
Chore: 517 roll back dev migrations

dleard:
This will roll back all the migrations before the backend pod is run. This way any changes to the migrations made during development will be deployed properly.


Sepehr-Sobhani (approved):
🤩

Issue #325, opened 2023-11-23
As an IO I want to upload documents to forms


BCerki:
#### Description:

We decided to split document upload out of the forms tickets. This ticket is about adding it. As of January 2024, only one document will be uploaded, the Statutory Declaration.

#### Acceptance Criteria:
Wireframe: https://www.figma.com/file/LsDmLDCdnJqI8UahyqvOD8/OBPS-App-Designs?type=design&node-id=4903%3A986&mode=design&t=rhr96fvLC4QihbUA-1

Given I am logged in to the Registration app with my Business BCeID account
When I navigate to the Operation form
Then I can add the statutory declaration document to the form

Given I am logged in to the Registration app with my IDIR account
When I navigate to the page for a specific Operation
Then I can view the file that has been uploaded for statutory declaration

~~ Given I am logged in to the Registration app with my Business BCeID account and I have been granted "admin" role for my operator
When I navigate to the Operation (facility) form
Then I can add documents to the form (both when creating and editing)

Given I am logged in to the Registration app with my Business BCeID account and I have been granted "admin" role for my operator
When I navigate to the  OperatOR form
Then I can add documents to the form (both when creating and editing)

Given I am logged in to the Registration app with my IDIR account and I have been granted the role of a CAS internal staff member
When I navigate to the page for a specific Operation/Operator
Then I can view the files that have been uploaded for that Operation/Operator ~~

#### Development Checklist:

- [ ] Add document handling to the existing operation forms (django ninja has docs about document handling but it  seems to be if you're using views instead of React)
- [ ] Documents fields are already in the models, but may need to made mandatory (remove blank=True) if they are required (see spreadsheet Required column)
- [ ] Check if the files expires
- [ ] Determine when is best to upload (as soon as the user uploads, or on submit of the form)
- [ ] Meets the DOD

**Definition of Ready** (Note: If any of these points are not applicable, mark N/A)

- [ ] User story is included
- [ ] User role and type are identified
- [ ] Acceptance criteria are included
- [ ] Wireframes are included (if required)
- [ ] Design / Solution is accepted by Product Owner
- [ ] Dependencies are identified (technical, business, regulatory/policy)
- [ ] Story has been estimated (under 13 pts)

·**Definition of Done** (Note: If any of these points are not applicable, mark N/A)

- [ ] Acceptance criteria are tested by the CI pipeline
- [ ] UI meets accessibility requirements
- [ ] Configuration changes are documented, documentation and designs are updated
[...]

PR #532, opened 2024-01-09
325 document upload

BCerki:
https://github.com/bcgov/cas-registration/issues/325

It was decided that document upload isn't needed for MVP, so pausing this work. What's completed in this PR:

- helper functions and resolvers to transform between dataurls and django ContentFile
- users can upload a boundary map document

Brianna todos:
- file fields were never added to the useroperator json schema
- do we have a limit on filesize? next/vercel (something, check what it was for cif blocks at 1mb)


BCerki:
Closing, new branch 325-statutory-declaration

PR #533, opened 2024-01-09
fix: handle form state externally to prevent data loss on error

### LINKED ISSUE #528: Form data resets if there is an error during form submission
Describe the Bug:

If there is an error during form submission such as a bad API call then the formData seems to be lost in the form. Most of the form fields still appear to be filled but upon resubmission the form thinks they are empty.

One sure way to fix this is handle form state externally in `FormBase` eg:

`const [formState, setFormState] = useState(formData)`

and then set state in an onChange:

`onChange={(data) => setFormState(data.formData)}`

There may be a better solution while keeping the form state internal to RJSF so it might be worth exploring. Though this fix would be a huge improvement, especially if we upload files on form submission.

Probability (how likely the bug is to happen, scored from 1-5): 3
(For example, probability of 5 is something like "it happens to all users every time they log in."
Probability of 1 "only happens to certain users when a really specific and unlikely path is followed.")

Effect (how bad the bug is when it does happen, scored from 1-5): 3
(For example, effect of 5 is "the entire app crashes and makes it unusable for all users" or "the bug causes the wrong data to be saved, with critical information (e.g. payment) being affected."
Effect of 1 is "It makes some styling look a little bit weird.")

Steps to reproduce the behaviour:

Edit the `apiUrl` in the `submitHandler` in one of our forms eg: `UserOperatorMultistepForm` and change the url to some junk string. Then fill out the associated form and submit it. 

After the submission fails if you try and submit again RJSF will think the form fields are empty even though you can still see text in most of them.

Screenshots:

Additional information:

- Browser: [e.g. chrome, safari]
- Version: [e.g. 22]
- Additional context

marcellmueller:
Fixes #528 

More explanation and steps to reproduce the bug/check that it's fixed in the issue. I couldn't find anything in RJSF form props that looked like it would handle this: 

https://rjsf-team.github.io/react-jsonschema-form/docs/api-reference/form-props/

The easy fix is handling form state externally though let me know if you have other ideas!

### DIFF
--- a/client/app/components/form/FormBase.tsx
+++ b/client/app/components/form/FormBase.tsx
@@ -1,9 +1,10 @@
 import defaultTheme from "./defaultTheme";
-import { useMemo } from "react";
+import { useMemo, useState } from "react";
 import { customizeValidator } from "@rjsf/validator-ajv8";
 import { FormProps, withTheme, ThemeProps } from "@rjsf/core";
 import { customTransformErrors } from "@/app/utils/customTransformErrors";
 import { RJSFValidationError } from "@rjsf/utils";
+import { IChangeEvent } from "@rjsf/core";
 
 const customFormats = {
   phone: /\(?\d{3}\)?[\s-]?\d{3}[\s-]?\d{4}$/,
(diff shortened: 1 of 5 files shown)


Sepehr-Sobhani (approved):
This looks great! Thanks.
I'm approving this but I have 1 question and 1 suggestion.

**Question**: Does this modification imply that we can confidently eliminate instances where we utilize `setFormState(newFormData)` elsewhere?

**Suggestion**: As an improvement, consider passing the `setError` dispatcher down to the `formBase` and setting it to undefined in the event of an API error. When a user modifies any field, we can subsequently remove that error, enhancing the interactivity of the form.

Issue #510, opened 2024-01-06
Selecting the 'no' radio button on the exemption lead form causes an error [6]


dleard:
Describe the Bug:

When registering an operation as an IO, when you are on the `add exemption application lead` page, if you select 'no' in the radio buttons and click submit, the page errors out. Selecting 'yes' and filling out information works as expected.

Probability (how likely the bug is to happen, scored from 1-5): 3

Effect (how bad the bug is when it does happen, scored from 1-5): 3

Steps to reproduce the behaviour:
- As an IO, go to register an operation
- Fill out the operation info & click Next to get to the exemption lead page
- Select 'no' at the top of the page & click submit
- Page errors out

Dianadec:
Refinement notes:
Change radio button to checkbox
Change wording about senior officer
We do need the address for the senior officer, but we don't need it for the application lead

PR #534, opened 2024-01-09
510: is_application_lead_external 'no' error

marcellmueller:
Implements #510 

We were designing this as I was doing it so take a close look and let me know if I missed or misunderstood something 🙇

### DIFF
--- a/bc_obps/registration/api/operation.py
+++ b/bc_obps/registration/api/operation.py
@@ -207,18 +207,21 @@ def update_operation(request, operation_id: int, submit: str, payload: Operation
         application_lead = Contact.objects.get(id=application_lead_id)
         application_lead_address_id = application_lead.address.id if application_lead.address else None
 
-    is_application_lead_external = payload.is_application_lead_external
-    if is_application_lead_external is True:  # the user has assigned a contact
-        address, _ = Address.objects.update_or_create(
-            id=application_lead_address_id,
-            defaults={
-                "street_address": payload.street_address,
-                "municipality": payload.municipality,
-                "province": payload.province,
-                "postal_code": payload.postal_code,
-            },
-        )
-        eal, _ = Contact.objects.update_or_create(
+    is_user_application_lead = payload.is_user_application_lead
+
+    # create or update the address for either user or external application lead
+    address, _ = Address.objects.update_or_create(
+        id=application_lead_address_id,
+        defaults={
+            "street_address": payload.street_address,
+            "municipality": payload.municipality,
+            "province": payload.province,
+            "postal_code": payload.postal_code,
+        },
+    )
+
+    if is_user_application_lead is True:  # the application lead is the user
+        al, _ = Contact.objects.update_or_create(
             id=application_lead_id,
             defaults={
                 "first_name": payload.first_name,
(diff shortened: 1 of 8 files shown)


bc_obps/registration/api/operation.py
```diff
@@ -245,7 +245,6 @@ def update_operation(request, operation_id: int, submit: str, payload: Operation
                 "email": user.email,
                 "phone_number": user.phone_number,
                 "business_role": BusinessRole.objects.get(role_name="Operation Registration Lead"),
-                "address": user.address,
```
marcellmueller, line 248: `address` doesn't exist in the user model and was causing the error.

client/app/utils/jsonSchema/operations.ts
```diff
@@ -297,72 +297,114 @@ const operationPage1: RJSFSchema = {
...
+    "municipality",
+    "province",
+    "postal_code",
+  ],
   properties: {
-    is_application_lead_external: {
+    is_user_application_lead: {
       type: "boolean",
-      title: "Would you like to add an exemption ID application lead?",
+      title: "Are you the application lead?",
```
marcellmueller, line 310: Not sure if the `exemption ID` part of this copy was important so flagging here

client/app/components/routes/operations/form/Operation.tsx
```diff
@@ -194,6 +202,54 @@ export default async function Operation({ numRow }: { numRow?: number }) {
...
+    ...formData?.application_lead?.address,
+    "Did you submit a GHG emissions report for reporting year 2022?":
+      formData?.previous_year_attributable_emissions ? true : false,
+    is_user_application_lead: isUserApplicationLead,
+    // fix for null values not opening the multiple operators form if loading a previously saved form
+    multiple_operators_array: isMultipleOperatorsArray
+      ? formData?.multiple_operators_array
+      : [{}],
+  };
+
```
marcellmueller, line 252: I moved this one file up to keep `OperationsForm` cleaner as we do the same for the `UserOperatorMultistepForm`. I think it makes sense to fetch and transform all data in the server components when possible, then pass it do the client components.

client/app/components/routes/operations/form/Operation.tsx
```diff
@@ -194,6 +202,54 @@ export default async function Operation({ numRow }: { numRow?: number }) {
...
+    ...userProfileFormData,
+    ...operation,
+  };
+
+  const userEmail = (userProfileFormData as UserProfileFormData)?.email;
+  const applicationLeadEmail = formData?.application_lead?.email;
+  // If the current user is the application lead, we want to show the application lead fields
+  const isUserApplicationLead =
+    userEmail === applicationLeadEmail && applicationLeadEmail !== undefined;
+
```
marcellmueller, line 218: Wasn't entirely sure if this was enough for the check. I figured that if they didn't click the button but entered the same email there was just a misunderstanding so we would check this off. 

Also I tried to do this with a function similar to the `def user_is_senior_officer` in the `UserOperator` model and then a resolver in the schema but since we don't have access to `User` unlike in that table I decided to take care of this on the front end.

dleard (commented):
Thanks Marcel, this looks great! I tested it and it's working as intended. I do have a quick question around the isUserApplicationLead check.

client/app/components/routes/operations/form/Operation.tsx
dleard: Totally works for now, we probably want to review this implementation later to see if there is a best practice or standard that we can follow for future conditionals. I've added a tech debt card #602 to review this and start a conversation.

dleard (approved):

PR #558, opened 2024-01-12
fix: set external form state on submit

marcellmueller:
Fix for performance issues caused by handling form state externally `onChange`. Now setting the form state on form submission.


Sepehr-Sobhani (approved):
Great stuff 🤩

PR #562, opened 2024-01-12
🪪 Allow multiple logins within the same browser

shon-button:
Addresses [413](https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=48609974)

## 🚧 Changes    
- modified:  client/app/components/navigation/Profile.tsx
- deleted:  client/app/api/auth/logout/route.ts
- modified:   client/.env.local.example
- modified:   helm/cas-registration/templates/frontend  

  

## 🚀 Impact:
   -  `client/app/components/navigation/Profile.tsx` opens NEXT_PUBLIC_KEYCLOAK_LOGOUT_URL in "_self" so that Keycloak clears the cookies to allow re-signin with multiple logins within the initial tab OR within new tabs created after logout
   -  Cleanup of reverted code
 
##  🔬 Testing:

**Pre-reqs** 

1. From file `client/.env.local.example` create `client/.env.local` with values reflecting instructions in file `client/.env.local.example`

2.  From terminal command, start the api server:
 ```
cd bc_obps
make reset_db
make run
```  
3.  From terminal command, start the app development server:
 ```
cd client && yarn build && yarn start
```  

4. In your web browser, navigate to `http://localhost:3000`
5. You should be directed to the anonymous landing page
image
6. Click "Program Administrator Log In"
7. You should be directed to IDIR login screen
image
8. Login using your IDIR account
9. You should be redirected to the dashboard or profile associated with your user
10. Right hand header should display your IDIR name and a "Log out" button
image
11. Click "Log out"
12. You should be directed to the "You are logged out" Keycloak page
image
13. In same tab's location bar enter to `http://localhost:3000`
14. You should be directed to the anonymous landing page
image
15. Click "Industrial  Operator Log In"
16. You should be directed to BCeid login screen
image
17. Login in with a BCeid credentials as found in 1password
18. You should be redirected to the dashboard associated with the BCeid
19. Right hand header should display BCeid name and a "Log out" button
image
20. Click "Log out"
21. You should be directed to the "You are logged out" Keycloak page
image

 **Optional:** Repeat steps using various scenarios such as new browser tabs and/or different BCeid logins


client/app/components/navigation/Profile.tsx
```diff
@@ -3,17 +3,14 @@ import Link from "@mui/material/Link";
...
-    }
     // call nextauth logout
-    signOut();
+    await signOut();
+    // redirect to Keycloak logout
+    open(process.env.NEXT_PUBLIC_KEYCLOAK_LOGOUT_URL, "_self");
+    return false;
   } catch (err) {
     console.error(err);
   }
```
andrea-williams, line 16: nit: further down in this file when the function is called, there's a comment stating that Keycloak logout happens before nextauth logout, when the reverse is now true. Just trying to avoid future dev confusion :)

client/app/components/navigation/Profile.tsx
```diff
@@ -3,17 +3,14 @@ import Link from "@mui/material/Link";
...
-    if (!response.ok) {
-      // redirect to keycloak basic logout SSO page
-      window.open(process.env.NEXT_PUBLIC_KEYCLOAK_LOGOUT_URL, "_blank");
-    }
     // call nextauth logout
-    signOut();
+    await signOut();
+    // redirect to Keycloak logout
+    open(process.env.NEXT_PUBLIC_KEYCLOAK_LOGOUT_URL, "_self");
+    return false;
```
andrea-williams, line 13: why do we return `false` here?

andrea-williams (approved):
Ran it locally and it worked!!! Even when I tried to break it! YAYYYYY!

I'm keeping every single one of my appendages crossed that this also works without complaint in dev.

client/app/components/navigation/Profile.tsx
shon-button: tests mess- deleted!

Issue #468, opened 2023-12-21
Make operation table match wireframe


BCerki:
###Description of the Tech Debt###

This ticket is about making the currently implemented operations table match the wireframe: https://www.figma.com/file/LsDmLDCdnJqI8UahyqvOD8/OBPS-App-Designs?type=design&node-id=9-10&mode=design&t=RTQrkxl7Upj6pRkn-0

Edit: Data is currently not being shown in the table's cells. As part of this card, ensure that the data is properly displayed in the table.
- Todo:
- Add a submission_date timestamp column to the operation table
- The submission_date timestamp should be populated when the IO user submits their operation registration request
- The BORO ID should be displayed
- The bcghgid should be displayed

AC:
- [x] the operation table has a submission_date timestamp column
- [x] the submission_date column is populated when IO user submits their operation registration request
- [x] once approved, the BORO ID is displayed properly in the table for both IO & internal users
- [ ] the bcghgid is displayed properly in the table for both IO & internal users
- [x] headers and columns are the same as the wireframe
- [x] if there's no bghg id, show a placeholder
- [x] remove operations list header
- [x] if quick and easy, match the styling to what's currently implemented in the users table. If not quick and easy, make a new tech ticket for design updates

###Tech Debt Triage#
The purpose of our technical debt triage process is to analyze technical debt to determine risk level of the technical debt and the value in tackling that technical debt.

Risk Value Scoring:

| Level  | Value |
| ------ | ----- |
| High   | 3     |
| Medium | 2     |
| Low    | 1     |

| Technical Debt - Risk Types                                                                                                   | Level | Value |
| ----------------------------------------------------------------------------------------------------------------------------- | ----- | ----- |
| Business Area Risk - Risk of business area visibility / damage to user experience                                             |       |       |
| Developer Fault Risk - How likely will this tech debt cause a future error related to coding on top of it                     |       |       |
| System Fault Risk - Risk of system errors or application downtime                                                             |       |       |
| Time Scale Risk - Compound risk effect if left alone. How much more difficult to fix or dangerous will this become over time? |       |       |
| Time Sink Risk - How much will this tech debt slow the development process down                                               |       |       |
| TOTAL SCORE:                                                                                                                  |       |       |

dleard:
Updated this card to include making sure that the necessary content of the table is properly displayed (current implementation is missing data) cc @hannavovk

marcellmueller:
[...]

PR #607, opened 2024-01-17
468 make operation table match wireframe

marcellmueller:
Implements #468 

This one was a lot more involved than the `Operator` dashboard which was just a design update. 

Some features to review in this PR:
- Save `submission_date` when user submits the form and format and display it in a new column
- Conditionally show new `Operator` column for `CAS Internal` users
- Table supports line breaking text instead of truncating long strings with an `elipsis...`. Currently none of the examples have strings long enough to break though Nan asked me to add that incase some Operators/Operations have long names.
- Updated fixtures so we have an example `Approved` Operator with `BORO ID` and `submission_date`
- Some other minor design fixes following the Operator PR. I noticed during the demo that the `Action` column on Andreas screen was slightly broken so that is fixed by setting a `flex: 1` on one of the columns in the tables.

Also we are keeping the same `Operation ID` column since not all applications have a BCGHD id. Hannah said we could revisit that column post MVP.

Screenshot 2024-01-18 at 3 20 51 PM


andrea-williams (approved):
Looks great! Great work Marcel!! 🚀

(Also, I super duper appreciate the comments you've added explaining the styling stuff - very very helpful)

PR #616, opened 2024-01-17
chore: update model constraints

dleard:
Need to update some constraints in order for the import from ciip/swrs to work. Some fields that were currently not nullable are empty in the swrs/ciip datasets & the constraints needed to be relaxed in those cases.
Two fields needed to be set as unique: operator.cra_business_number & operation.swrs_facility_id. These should be unique anyway, but the import job requires them to be unique in order to add an ON CONFLICT clause & make the function idempotent.


bc_obps/registration/models.py
```diff
@@ -188,13 +188,15 @@ class Meta:
...
+    municipality = models.CharField(max_length=1000, null=True, db_comment="Municipality of relevant location")
     province = CAProvinceField(
-        db_comment="Province of the relevant location, restricted to two-letter province postal abbreviations"
+        db_comment="Province of the relevant location, restricted to two-letter province postal abbreviations",
+        null=True
     )
     postal_code = CAPostalCodeField(
-        db_comment="Postal code of relevant location, limited to valid Canadian postal codes"
+        db_comment="Postal code of relevant location, limited to valid Canadian postal codes",
+        null=True
```
Sepehr-Sobhani, line 199: We are using Address as a foreign key in other tables, so I think we can make those fields nullable instead of making all fields in the Address data model nullable.

bc_obps/registration/models.py
```diff
@@ -402,6 +407,12 @@ class Statuses(models.TextChoices):
     history = HistoricalRecords(table_name='erc_history"."operator_history', m2m_fields=[documents, contacts])
 
     class Meta:
+        constraints = [
+            models.UniqueConstraint(
+                fields=["cra_business_number"],
+                name="cra_business_number_unique_constraint",
+            )
+        ]
```
Sepehr-Sobhani, line 415: Could we add a test into this model to prevent unintentional removal of this constraint in the future?

bc_obps/registration/models.py
```diff
@@ -614,6 +626,12 @@ class Statuses(models.TextChoices):
     class Meta:
         db_table_comment = "Operations (also called facilities)"
         db_table = 'erc"."operation'
+        constraints = [
+            models.UniqueConstraint(
+                fields=["swrs_facility_id"],
+                name="swrs_facility_id_unique_constraint",
+            )
+        ]
```
Sepehr-Sobhani, line 634: Also a test for this.

Sepehr-Sobhani (approved):
I'm approving this one to not block if I'm away but we need to add two tests (or we could make them a tech debt card) and fix existing tests.

PR #627, opened 2024-01-18
fix: allow custom redirect

shon-button:
WIP:

**Issue:** https://cas-reg-frontend-dev.apps.silver.devops.gov.bc.ca/ Log Out flow stays on same page instead of  the expected `xx/openid-connect/logout` page and SiteMinder-Keycloak logout is not completed (as seen in cookie values)
image

**Troubleshooting**
Potentially, the open function fails to set the location href  because it is being prevented by security ensuring the window is in  "active interaction" state (activating or focusing a browser window in response to a user gesture).

**Possible Resolution**
Remove dependency on window.open and leverage nextauth signOut function's callbackUrl parameter in combination with customising nextauth route.ts redirect function to allow callbacks to the identity server.

### DIFF
--- a/client/app/api/auth/[...nextauth]/route.ts
+++ b/client/app/api/auth/[...nextauth]/route.ts
@@ -70,7 +70,7 @@ export const authOptions: NextAuthOptions = {
           // 🚀 API call: Get user app_role by user_guid from user table
           const responseRole = await actionHandler(
             `registration/user-app-role/${token.user_guid}`,
-            "GET"
+            "GET",
           );
           if (responseRole?.role_name) {
             // user found in table, assign role to token (note: all industry users have the same app role of `industry_user`, and their permissions are further defined by their role in the UserOperator model)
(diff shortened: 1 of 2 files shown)


dleard (approved):

shon-button:
Alternative Paths:
- [ ]  api route url=${KEYCLOAK_LOGOUT_URL}+
      `?id_token_hint=${
        token.id_token
      }&post_logout_redirect_uri=${encodeURIComponent(
        process.env.NEXTAUTH_URL as string,
      )}` const response = await fetch(url, { method: "GET" });
- [ ]  api route url= ${SITEMINDER_LOGOUT_URL}?retnow=1&returl=${KEYCLOAK_LOGOUT_URL}+
      `?id_token_hint=${
        token.id_token
      }&post_logout_redirect_uri=${encodeURIComponent(
        process.env.NEXTAUTH_URL as string,
      )}` const response = await fetch(url, { method: "GET" });
- [ ]  api route url= ${SITEMINDER_LOGOUT_URL}?retnow=1&returl=${KEYCLOAK_LOGOUT_URL} const response = await fetch(url, { method: "GET" });
- [ ] client:  window.open(process.env.NEXT_PUBLIC_KEYCLOAK_LOGOUT_URL, "_blank");
- [ ]  client: window.open(process.env.NEXT_PUBLIC_KEYCLOAK_LOGOUT_URL, "_self");
- [ ]  signOut({ callbackUrl: process.env.NEXT_PUBLIC_KEYCLOAK_LOGOUT_URL,}); with customized  nextauth callback:     async redirect({ url, baseUrl }) 
- [ ] api using response.redirect
- [ ] client location.href

shon-button:
Documenting:

One time I did receive the "You are logged off" screen and from there I was able to login to the alternated ID type where a flow of log out log in would flip between the two id without the need to provide credentials:

image

image
image
image


image

Issue #399, opened 2023-12-08
Report a Problem Link to ghgregulator email


Dianadec:
#### Description:

The "Report a Problem" tile needs to link to a contact option which can simply be the same as the "Contact Us" link in the footer which takes users to emailing the ghgregulator@gov.bc.ca email.

https://www.figma.com/file/LsDmLDCdnJqI8UahyqvOD8/OBPS-App-Designs?type=design&node-id=9%3A8&mode=design&t=AocKNZY25UVOb32s-1

#### Acceptance Criteria:

Given I am any user of the Registration app (external or internal)
When I click on the "Report a Problem" tile in my dashboard
Then an email template opens so I can write an email to the appropriate contact reporting my problem.

#### Development Checklist:

- [x] clicking the button in the "Report a Problem" tile will have the same logic as what's currently in the "Contact Us" button in the page footer (should be a simple copy-and-paste of the code)
- [ ] Checklist item
- [ ] Checklist item
- [ ] Meets the DOD

**Definition of Ready** (Note: If any of these points are not applicable, mark N/A)

- [ ] User story is included
- [ ] User role and type are identified
- [ ] Acceptance criteria are included
- [ ] Wireframes are included (if required)
- [ ] Design / Solution is accepted by Product Owner
- [ ] Dependencies are identified (technical, business, regulatory/policy)
- [ ] Story has been estimated (under 13 pts)

·**Definition of Done** (Note: If any of these points are not applicable, mark N/A)

- [ ] Acceptance criteria are tested by the CI pipeline
- [ ] UI meets accessibility requirements
- [ ] Configuration changes are documented, documentation and designs are updated
- [ ] Passes code peer-review
- [ ] Passes QA of Acceptance Criteria with verification in Dev and Test
- [ ] Ticket is ready to be merged to main branch
- [ ] Can be demoed in Sprint Review
- [ ] Bugs or future work cards are identified and created
- [ ] Reviewed and approved by Product Owner

#### Notes:

-

Dianadec:
@andrea-williams Could you add in the reference about copying code over from what was used for the "Contact Us" link in the footer?

Dianadec:
@marcellmueller  is that change suggestion you had going to be part of this card? Or a diff card. Just wondering if we can approve this based on what's in the original ac. Thanks! cc @hannavovk 

> are you okay with us changing Report a problem to the team to Report problems to GHGRegulator@gov.bc.ca  ? 
>  
[...]

PR #653, opened 2024-01-22
399 - Report a Problem mailto link

andrea-williams:
Chore: replace /problem page with mailto link
Card: https://github.com/bcgov/cas-registration/issues/399

### DIFF
--- a/client/app/(authenticated)/dashboard/problem/page.tsx
+++ b/client/app/(authenticated)/dashboard/problem/page.tsx
@@ -1,3 +1,0 @@
-export default function Page() {
-  return <h1>TO DO: Problem Page</h1>;
-}
--- a/client/app/data/dashboard/cas_admin.json
+++ b/client/app/data/dashboard/cas_admin.json
@@ -16,7 +16,12 @@
   },
   {
     "title": "Report a Problem",
-    "content": "Something wrong? Report a problem to the team",
-    "links": [{ "title": "Tell the team ➤", "href": "/dashboard/problem" }]
+    "content": "Something wrong?",
+    "links": [
+      {
+        "title": "Report a problem to the team ➤",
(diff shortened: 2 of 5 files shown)


marcellmueller (approved):
Nice work! Approved though I have two thoughts:

I wonder if we should have a tech debt card to refactor these into `.ts` files that return json so we can avoid repeating the same shared tiles 4 times?

Also one other thing that would be a nice to have is a default `subject` ie:
`href": "mailto:GHGRegulator@gov.bc.ca?subject=HELP ME"`

andrea-williams:
> Nice work! Approved though I have two thoughts:
> 
> I wonder if we should have a tech debt card to refactor these into `.ts` files that return json so we can avoid repeating the same shared tiles 4 times?
> 
> Also one other thing that would be a nice to have is a default `subject` ie: `href": "mailto:GHGRegulator@gov.bc.ca?subject=HELP ME"`

I can definitely see the value in setting a default subject title once we have multiple apps in our OBPS system. Right now, I can't think of a default subject that would add any value to users. Do you have any suggestions?

marcellmueller:
> > Nice work! Approved though I have two thoughts:
> > I wonder if we should have a tech debt card to refactor these into `.ts` files that return json so we can avoid repeating the same shared tiles 4 times?
> > Also one other thing that would be a nice to have is a default `subject` ie: `href": "mailto:GHGRegulator@gov.bc.ca?subject=HELP ME"`
> 
> I can definitely see the value in setting a default subject title once we have multiple apps in our OBPS system. Right now, I can't think of a default subject that would add any value to users. Do you have any suggestions?

No not really, it was just a nice to have. I was thinking something like `Report a problem with B.C. Industrial Emissions Reporting System`. Anyways feel free to merge this in!

PR #658, opened 2024-01-22
441 - operator tile

marcellmueller:
Implements #441 

Previous PR for reference. It was old and getting unmanageable: https://github.com/bcgov/cas-registration/pull/487

### DIFF
--- a/bc_obps/registration/api/user_operator.py
+++ b/bc_obps/registration/api/user_operator.py
@@ -358,7 +358,6 @@ def create_user_operator_contact(request, payload: UserOperatorContactIn):
     except Exception as e:
         return 400, {"message": str(e)}
 
-    # TODO: approved status may have to be changed to pending if the user is not an admin
     if user_operator_instance.status != UserOperator.Statuses.APPROVED:
         user_operator_instance.status = UserOperator.Statuses.PENDING
     user_operator_instance.save(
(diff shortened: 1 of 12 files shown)


Sepehr-Sobhani (approved):
Thank you for your dedication to this PR. Here are some notes for future reference:

- To prevent orphaned addresses, it's advisable to update an operator's existing physical and mailing addresses rather than creating new ones during the update process.

- When modifying an operator, it's better to archive all current parent_operators and generate new ones to avoid duplicate instances of parent_operator.

- Double rendering occurs when visiting the select-operator page while a user already has an admin request; this issue needs attention.

marcellmueller:
> Thank you for your dedication to this PR. Here are some notes for future reference:
> 
> * To prevent orphaned addresses, it's advisable to update an operator's existing physical and mailing addresses rather than creating new ones during the update process.
> * When modifying an operator, it's better to archive all current parent_operators and generate new ones to avoid duplicate instances of parent_operator.
> * Double rendering occurs when visiting the select-operator page while a user already has an admin request; this issue needs attention.

Thank you so much! I've created two new issues from this comment. One task for the first two and a bug issue for the double rendering. 

https://github.com/bcgov/cas-registration/issues/661
https://github.com/bcgov/cas-registration/issues/662

PR #664, opened 2024-01-23
559 data model level validation

Sepehr-Sobhani:
[ISSUE 559](https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=49845989)


bc_obps/registration/api/user_operator.py
```diff
@@ -281,7 +283,7 @@ def request_access(request, payload: SelectOperatorIn):
 
     # Making a draft UserOperator instance if one doesn't exist
     user_operator, created = UserOperator.objects.get_or_create(
-        user=user, operator=operator, status=UserOperator.Statuses.PENDING
+        user=user, operator=operator, status=UserOperator.Statuses.PENDING, role=UserOperator.Roles.REPORTER
```
dleard, line 286: I remember talking about this before but I don't remember a resolution. Should we be pre-populating the row when creating a user_operator with the REPORTER role? I had figured the role would be null until assigned by an operator admin, since we don't know what role (if any) the admin is going to assign to the user.

dleard (changes requested):
Looks great! This will really help us ensure data integrity. Just one question around setting a default role for the user_operator table on create.

dleard (approved):

bc_obps/registration/api/user_operator.py
Sepehr-Sobhani: @dleard Further to our conversation, I have captured the required changes in: https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=51804153

PR #666, opened 2024-01-23
325 statuatory declaration

BCerki:


client/app/components/form/widgets/FileWidget.tsx
```diff
@@ -109,11 +163,20 @@ const FileWidget = ({
   const handleChange = useCallback(
     (event: ChangeEvent<HTMLInputElement>) => {
       event.preventDefault();
-      if (!event.target.files) {
+      const maxSize = 1024 * 1024; // 1 MB in bytes
```
BCerki, line 166: Next uses 1MB as the default, so I've gone with that until I hear otherwise from Nan. We need to change it here and in the next config: By default body size limit is 1MB, we can change it here: https://nextjs.org/docs/app/api-reference/next-config-js/serverActions#bodysizelimit

client/app/components/form/widgets/FileWidget.tsx
```diff
@@ -109,11 +163,20 @@ const FileWidget = ({
...
+      if (!files) {
         return;
       }
-      processFiles(event.target.files).then((filesInfoEvent) => {
-        const newValue = filesInfoEvent.map((fileInfo) => fileInfo.dataURL);
+
+      processFiles(files).then((filesInfoEvent) => {
+        const newValue = filesInfoEvent.map((fileInfo) => {
+          if (fileInfo.size > maxSize) {
+            alert("File size must be less than 1MB");
```
BCerki, line 175: Error handling wasn't part of this ticket and we don't have a design for it, so in the meantime I added this so we don't get unhandled errors

bc_obps/registration/api/operation.py
```diff
@@ -287,6 +288,14 @@ def update_operation(request, operation_id: int, submit: str, payload: Operation
         for operator in operation_multiple_operators:
             operator.set_archive(modifier=user)
 
+    if payload.statutory_declaration:
+        document = Document.objects.create(
+            file=payload.statutory_declaration,
+            type=DocumentType.objects.get(name="signed_statutory_declaration"),
+            description='what was this meant to be? maybe it would make more sense to put the description in the DocumentType table?',
```
BCerki, line 295: asked in Teams

shon-button:
image
image

BCerki:
To answer the question above, pending operations should be readonly: https://github.com/orgs/bcgov/projects/122/views/2?filterQuery=pending&pane=issue&itemId=48608801

BCerki:
Re the editing workflow, it seems that when we submit edits the Operation is duplicated. This happens on the `develop` branch too, so I'm made a bug ticket to handle it separately: https://github.com/bcgov/cas-registration/issues/694. If you test the attachments and go into the one that's pending (instead of the duplicated one that's draft), you ~should be able to see the saved attachment.

client/app/components/form/widgets/FileWidget.tsx
BCerki: 20MB

shon-button:
⭐ very good implementation of a tricky requirement ⭐

Are there any concerns about multiple uploads of same named file existing in bucket only once in terms of a lack of audit trail or potential unintended overwrite?

shon-button (approved):

client/app/components/form/widgets/FileWidget.tsx
BCerki: Went over this with Nan, it's fine for now

bc_obps/registration/api/operation.py
```diff
@@ -287,10 +287,16 @@ def update_operation(request, operation_id: int, submit: str, payload: Operation
             operator.set_archive(modifier=user)
 
     if payload.statutory_declaration:
+        operation_document = Operation.documents.through.objects.filter(
+            operation_id=operation.id
+        ).first()  # there will only ever be 1 statutory declaration
+        if operation_document:
+            Document.objects.get(
+                id=operation_document.document_id
+            ).delete()  # we delete here rather than update because delete removes the file from google cloud storage
```
Sepehr-Sobhani, line 296: To simplify this a little we could do:
```python
operation.documents.filter(type=DocumentType.objects.get(name="signed_statutory_declaration")).delete()
```
If there will only ever be 1 statutory declaration.

bc_obps/registration/models.py
```diff
@@ -142,6 +143,13 @@ class Meta:
         models.Index(fields=["type"], name="document_type_idx"),
     ]
 
+    def delete(self, *args, **kwargs):
+        # Delete the file from Google Cloud Storage before deleting the model instance
+        if self.file:
+            default_storage.delete(self.file.name)
+
+        super().delete(*args, **kwargs)
+
```
Sepehr-Sobhani, line 152: Let's confirm this with @dleard . I feel that hard deleting is not what we should do here since the `Document` data model is a `TimeStampedModel` and I assume we should track its trail. Sorry for the confusion earlier.

bc_obps/registration/api/operation.py
```diff
@@ -289,9 +289,10 @@ def update_operation(request, operation_id: int, submit: str, payload: Operation
...
-        operation_document = Operation.documents.through.objects.filter(
-            operation_id=operation.id
-        ).first()  # there will only ever be 1 statutory declaration
+        operation_document = operation.documents.filter(
+            type=DocumentType.objects.get(name="signed_statutory_declaration")
+        ).delete()  # there will only ever be 1 statutory declaration
+
         if operation_document:
             Document.objects.get(
                 id=operation_document.document_id
```
Sepehr-Sobhani, line 298: @BCerki You don't need if condition and the code below it.

PR #669, opened 2024-01-23
change: get client-side env var value from server

shon-button:
WIP:

**Background:** 
App\Log Out (nextauth SignOut()) is not functioning as expected because the callbackUrl (siteminder\keycloak log out) is not correct...

[Yesterday 19:17] Dylan Leard (Guest)
It still looks like the issue is that the url being passed from Profile.tsx is undefined
[Yesterday 19:24] Dylan Leard (Guest)
I don't think the frontend app has process.env context

[Yesterday 20:11] Hogan, Shon ENV:EX
If the NEXT_PUBLIC_ prefix, used to expose environment variables to the client-side components via process.env, is not working as expected; perhaps, we could use a server action...

**Possible solution**
Obtain value of callback url from a server-side .env value


client/app/utils/actions.ts
```diff
@@ -8,6 +8,7 @@ and can be called from server components or from client components.
 "use server";
 import { revalidatePath } from "next/cache";
 import { cookies } from "next/headers";
+import { signOut } from "next-auth/react";
```
dleard, line 11: Remove this unused import & the frontend-docker-build step will pass

dleard (changes requested):
Thanks Shon! I am cautiously optimistic that this will fix the issue. This PR needs pre-commit run over it to make it happy & one unused import removed.

dleard (approved):

Issue #656, opened 2024-01-22
operations/{operation_id} PUT endpoint assumes a point_of_contact_id


dleard:
operations/{operation_id} PUT endpoint assumes a point_of_contact_id where one will never exist to start. This is blocking all operation submissions for a BORO ID. We haven't caught this up until now because our fixtures all have a point_of_contact_id populated by default, however this will always be null until lthe user gets to the application_lead page.
The code starting [here](https://github.com/bcgov/cas-registration/blob/develop/bc_obps/registration/api/operation.py#L224) is the problem, where we attemtpt to create or update a contact based on the point_of_contatct_id, but it errors out because nothing exists to create it from.

To do:
- [x] Remove all point_of_contact_id values from the operations fixtures where status is 'Not Started', these operations should not have this ID by default
- [x] If you attempt to go through the register operation workflow, you'll get an JSON error in the frontend & a more descriptive error in the backend
Backend error:
![Image](https://github.com/bcgov/cas-registration/assets/26367840/69bb14f0-6a35-4a71-95c8-1178cbeef462)
- [x] Fix the endpoint, so that it does not try to create a contact until it has contact data (user is on the application_lead page)

PR #717, opened 2024-01-25
Fix/operations point of contact/656

andrea-williams:
Card: #656 

Fixing one bug revealed a bunch of other bugs (all discovered bugs are addressed in this ticket). In addition to the bug described in the card, I've also fixed it so that the text in the Operation form "Would you like to designate another person to be a point of contact for this application?" actually matches the intended behaviour (i.e., when the button is clicked, a new Contact for the "external" or other person is created in the database. If the button isn't selected, the current user will be marked as the Contact for the Operation).

Added several new test cases for the /operations endpoint to mimic the behaviour that will happen on data import (i.e., sending a PUT request for an operation when it doesn't have a `point_of_contact` yet)

Note that I didn't remove the Contact's FK reference to `address`, although we've since found out that addresses won't be collected.


bc_obps/registration/models.py
```diff
@@ -311,6 +311,10 @@ class Meta:
             models.Index(fields=["business_role"], name="contact_role_idx"),
         ]
 
+    def __str__(self) -> str:
+        fields = [f"{field.name}={getattr(self, field.name)}" for field in self._meta.fields]
+        return ' - '.join(fields)
+
```
dleard, line 317: What is this function for? It looks like it's setting a set of fields to be in the format field-field-field? Just a question for my general knowledge :)

dleard (approved):
Awesome work! And good catch on the extra bugs. I've tested this locally with some ciip-swrs data & I can get past the point now where I first found the bug.
:100:

bc_obps/registration/models.py
andrea-williams: It is helpful for debugging purposes (until I get around to setting up proper Python debugging in VS Code). It's so that when a dev throws in a `print(operation)` somewhere for debugging purposes, what gets printed to the python server console is the string representation of the `operation` object, rather than a very not-helpful `Operation <object>` (I forget exactly how it looks in the console, but basically you can't interrogate any of the values of the object through print statements)

andrea-williams: example output for an operation would be something like:

`id - 1 name - Operation 1 <other fields...>`

PR #725, opened 2024-01-25
fix: incorrect logic to show missing operator information message

marcellmueller:


andrea-williams:
@marcellmueller branch is out-of-date with develop, and I also suspect the k6 file commit was accidental...

marcellmueller:
> @marcellmueller branch is out-of-date with develop, and I also suspect the k6 file commit was accidental...

Hah, thanks should be good now. There is a .gitignore added for that k6 folder in that PR, I will have to be more careful context switching.

andrea-williams (approved):
I think this is right (although I've got end-of-day-brain and don't entirely trust my judgment right now). But 3 LOC is a low-risk commit ;)

PR #727, opened 2024-01-25
Chore: reset data in dev / test on deploy

dleard:
This PR fixes (removes) the broken init container in the backend deployment and replaces it with a couple pre/post upgrade jobs. Deploys to dev/test should now run without any manual intervention needed in the database

In dev:
- reset-database will run as a pre-upgrade hook, which will reset the database (including all Django migrations) & allow the backend to rerun all migrations from the beginning
- then the load-dev-data job runs at the end to re-populate our dev data

In test:
- reset-database will run as a pre-upgrade hook just like in dev, but as part of the test-specific command, will preserve the user table in a separate schema that persists until after the deployment is complete
- restore-test-users will run as a post-upgrade hook, this job copies the preserved users back into the live table & drops the preserve schema that temporarily contained the user data during deploy


Sepehr-Sobhani (approved):
🏅

PR #736, opened 2024-01-26
591: Operations tile is hidden

BCerki:
https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=50184016

PR #768, opened 2024-01-29
462 unique constraints

BCerki:
https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=49836096

There's a bug that [prevents editing of these fields on the operator](https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=51509212), but you can do the manual check on adding an operator

The sonarcloud fail is on repetition in db comments, which I think is fine to ignore but feel free to disagree!


andrea-williams:
@BCerki  I don't have any concerns about the duplication of DB comments. You could sign in to SonarCloud and mark the issues it's complaining about as "Won't fix" if you want it to stop failing you. 

I would like to see some additional test cases for POSTing or PUTting to the `/operators` (or is it `/user-operators` ?) API endpoint - we want to make sure that the API returns an error code (and maybe also an error message?) when a user tries to create/update an operator with a legal name or corporate registry ID that already exists in the DB. I think we already have similar tests for handling duplication of CRA numbers in that file...

bc_obps/registration/tests/endpoints/test_useroperator_endpoints.py
```diff
@@ -776,3 +832,87 @@ def test_put_user_operator_operator_malformed_data(self):
...
+    def test_put_duplicates_not_allowed(self):
+        baker.make(
+            Operator, legal_name='Legal Name', bc_corporate_registry_number="aaa1234321", cra_business_number=55555
+        )
+
+        operator = baker.make(Operator, bc_corporate_registry_number="yyy1234321")
+
+        user_operator = baker.make(UserOperator, user=self.user, operator=operator, role=UserOperator.Roles.ADMIN)
+
+        # duplicate CRA business number
```
BCerki, line 845: The [bug](https://github.com/bcgov/cas-registration/pull/736) is that cra business number and bc corporate number are never added to the db, so these tests currently fail (I wrote them before I realized that, oops.) In the bug ticket, I wrote a note to either uncomment them out or delete them depending on how the bug is handled.

andrea-williams (approved):
Awesome, thanks for adding those extra unit tests Bri!

PR #771, opened 2024-01-30
chore: text changes

BCerki:
card: https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=51308566

### DIFF
--- a/client/app/components/routes/select-operator/form/AccessRequestReceived.tsx
+++ b/client/app/components/routes/select-operator/form/AccessRequestReceived.tsx
@@ -2,6 +2,8 @@ import { Operator } from "@/app/components/routes/select-operator/form/types";
 import AccessTimeIcon from "@mui/icons-material/AccessTime";
 import { getOperator } from "@/app/components/routes/select-operator/form/ConfirmSelectedOperator";
 import { actionHandler } from "@/app/utils/actions";
+import Link from "next/link";
+import Button from "@mui/material/Button";
 
 export async function getOperatorHasAdmin(id: number) {
   return actionHandler(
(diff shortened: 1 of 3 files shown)


client/app/components/routes/select-operator/form/AccessRequestReceived.tsx
```diff
@@ -44,20 +46,29 @@ export default async function AccessRequestReceived({
...
+      <Link href="/dashboard/select-operator">
+        <Button
+          className="my-10"
+          aria-label={"View Submitted Information"}
+          color="primary"
+          variant="contained"
+        >
+          View Submitted Information
+        </Button>
+      </Link>
```
marcellmueller, line 58: I don't feel super strongly about this and have likely done it myself before but wrapping a button in an `<a>` tag is not valid by HTML 5 specification:
https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element


I played around a bit and got it styled the same aside from the hover elements:
```
      <Link
        href="/dashboard/select-operator"
        aria-label={"View Submitted Information"}
        className="m-auto w-fit no-underline text-lg text-white py-[6px] px-[16px] rounded flex items-center bg-bc-primary-blue"
      >
        View Submitted Information
      </Link>{" "}
```

To make this cleaner and more reusable we could add a custom tailwind class using `@apply` in `globals.css` under the `.form-heading-label` class. Something like this: 

```
  .link-button-blue {
    @apply w-fit no-underline text-lg text-white py-[6px] px-[16px] rounded flex items-center bg-bc-primary-blue;
  }
```

The use on the link like this:

```
      <Link
        href="/dashboard/select-operator"
        aria-label={"View Submitted Information"}
        className="m-auto link-button-blue"
      >
        View Submitted Information
      </Link>{" "}
```

BCerki: Thanks, didn't know that about the <link><button /></link>. That pattern does show up a few times in the app (for example where I stole this code from haha), so I've made a tech debt ticket to fix it everywhere: https://github.com/bcgov/cas-registration/issues/784

marcellmueller (approved):
Awesome, thanks for making that tech debt ticket!

Issue #765, opened 2024-01-29
Show MUI Datagrid footer so dashboard pagination works again [6]


marcellmueller:
Describe the Bug:

There is currently no way to view more than 50 (I think) items on the Operator or Operations dashboard. I disabled the footer as it wasn't in the design and realized that it's necessary to be able to view more than 50 items. Looks like disabling pagination isn't allowed unless you use datagrid pro or premium: 

https://mui.com/x/react-data-grid/pagination/

I think there will be some CSS adjustments needed since we turned the SVG icons white, they are hidden with the white background if you remove the `hideFooter` prop from `DataGrid`

Probability (how likely the bug is to happen, scored from 1-5): 2
(For example, probability of 5 is something like "it happens to all users every time they log in."
Probability of 1 "only happens to certain users when a really specific and unlikely path is followed.")

Effect (how bad the bug is when it does happen, scored from 1-5): 4
(For example, effect of 5 is "the entire app crashes and makes it unusable for all users" or "the bug causes the wrong data to be saved, with critical information (e.g. payment) being affected."
Effect of 1 is "It makes some styling look a little bit weird.")

Steps to reproduce the behaviour:

Screenshots:

Additional information:

- Browser: [e.g. chrome, safari]
- Version: [e.g. 22]
- Additional context

marcellmueller:
Follow up tech debt issue to support server side pagination: #782

dleard:
a developer should review this (Not a PO concern)

PR #780, opened 2024-01-30
fix: re-enable pagination on mui datagrid

marcellmueller:
Fix for #765 

Follow up tech debt issue to support server side pagination: #782 

Show the footer again and enable pagination even if it's not in the design:
Screenshot 2024-01-30 at 11 17 30 AM

Re-enables pagination and sets page size to 20 as discussed at standup. If you want to actually test the pagination you will have to somehow get over 20 records (Operators or Operations) in the database. I just ran my load testing PR for a few mins which created hundreds.

Screenshot 2024-01-30 at 10 50 52 AM


BCerki (approved):
:100:

PR #785, opened 2024-01-30
fix: operators can only edit their own operations

BCerki:
https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=51179688

### DIFF
--- a/bc_obps/registration/tests/utils/bakers.py
+++ b/bc_obps/registration/tests/utils/bakers.py
@@ -58,7 +58,6 @@ def operation_baker(operator_id=None):
     if operator_id:
         return baker.make(
             Operation,
-            operator=operator_baker(),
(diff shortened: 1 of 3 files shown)


bc_obps/registration/tests/utils/bakers.py
```diff
@@ -54,7 +54,17 @@ def operator_baker():
     )
 
 
-def operation_baker():
+def operation_baker(operator_id=None):
+    if operator_id:
+        return baker.make(
+            Operation,
+            operator=operator_baker(),
```
Sepehr-Sobhani, line 61: This `operator` assignment is redundant since we are passing the `operator_id` on line 65.

Sepehr-Sobhani (approved):
🤩 Looks great! Just a tweak is required.

PR #789, opened 2024-01-31
387 add custom history user

Sepehr-Sobhani:
[ISSUE 387](https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=48609626)


andrea-williams (commented):
Do we also need to capture the date of the change along with the user ID, or is that automatically captured by the django history package (don't know the proper name for it) we're using?

Sepehr-Sobhani:
> Do we also need to capture the date of the change along with the user ID, or is that automatically captured by the django history package (don't know the proper name for it) we're using?
> 
> Also flagging for you that there's a merge conflict in one of the migration files. Probably because I took too long to review this. Sorry!

That's correct, django simple history package by default takes care of the `created_at` and `updated_at` data fields.
No worries, Thanks for reviewing this PR :)

andrea-williams (approved):
Code LGTM so I'm happy, although it looks like the action for backend tests failed???

Sepehr-Sobhani:
> Code LGTM so I'm happy, although it looks like the action for backend tests failed???

Should be fixed now, Thank you!

Issue #750, opened 2024-01-26
Refine Landing Page Layout & Font Size


nanyangpro:
As requested by @hannavovk and discussed with @marcellmueller - The landing page design has been refined with the layout and font updated.

This ticket is for implementing the refinements according to [the latest design](https://www.figma.com/file/LsDmLDCdnJqI8UahyqvOD8/OBPS-App-Designs?type=design&node-id=2688-6375&mode=design&t=M6fDYE2TCClt3dDZ-4) (more details in the comment threads in Figma).

PR #792, opened 2024-01-31
750 refine homepage

marcellmueller:
Implements #750 

This also removes the login buttons in the nav which I will miss for ease of access. Though I'm sure we'll get used to using the landing page logins instead.

### DIFF
--- a/client/app/components/navigation/Session.tsx
+++ b/client/app/components/navigation/Session.tsx
@@ -1,4 +1,3 @@
-import Login from "@/app/components/navigation/Login";
 import Profile from "@/app/components/navigation/Profile";
 import { useSession } from "next-auth/react";
 import { getUserFullName } from "@/app/utils/getUserFullName";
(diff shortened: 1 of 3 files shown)


client/app/components/navigation/Session.tsx
```diff
@@ -14,10 +13,7 @@ export default function Session() {
   if (status === "loading") {
     //👇️ Handle loading state (e.g., show a loading spinner).
     return <div>Loading...</div>;
-  } else if (!session) {
-    //👇️ Handle unauthenticated state (e.g., show a "Sign In" button).
-    return <Login />;
```
marcellmueller, line 19: I didn't delete `Login.tsx` which isn't being used anymore, wasn't sure if I should keep it around.

andrea-williams:
@marcellmueller  flagging that the branch is out of sync with `develop` :)

andrea-williams (approved):

Issue #706, opened 2024-01-25
When Operation status is Approved, edits to Operation data should not reset the Operation's status


andrea-williams:
#### Describe the task

Current behaviour: when an external user updates any data fields for an already-Approved Operation, on Submit to our API, the status of that Operation is reset to "Pending". However, since the Operation was previously approved, it will still have an active BORO ID associated with it. This is pretty confusing for everyone.

Desired behaviour: when an external user updates any data fields for an already-Approved Operation, on Submit to our API, the status of that Operation should remain as it is (i.e., "Approved").

#### Acceptance Criteria

- [x] modify logic in our API on the PUT "/operations/{operation_id}" endpoint so that if the operation's status is already "Approved", the status remains unchanged 
- [x] PUT requests sent to "/operations/{operation_id}" should still trigger changes on the `updated_at` and `updated_by_id` columns in the Operation table
- [x] add a backend unit test confirming this desired behaviour

#### Additional context

- in another ticket, we'll be setting up Metabase to trigger notifications to internal staff whenever data about an approved Operation is updated. See #704

Dianadec:
Needed for MVP.

PR #793, opened 2024-01-31
Task/edit operation wo status reset/706

andrea-williams:
Card: [706](https://github.com/bcgov/cas-registration/issues/706)

Original purpose of card was to ensure that operations that are already Approved don't have their status reset to "Pending" when an edit is made (PUT request sent for the operation). This PR addresses that and some other issues:

- deletes/comments out Operation fields that are no longer required for MVP (such as multiple_operators, reporting_activities, addresses, etc.)
- fixes logic on the PUT endpoint for operations. We weren't saving some of the fields being sent to the backend correctly. There's still room for improvement on this IMO - it gets ugly because part of the payload being sent via the PUT request is actually default values from the frontend, which is why I introduced the `form_section` param on the endpoint so that we're only saving the parts of data in the payload that will come from the form section the user is on. Confusing? Yup.

### DIFF
--- a/bc_obps/registration/api/operation.py
+++ b/bc_obps/registration/api/operation.py
@@ -250,6 +250,7 @@ def update_operation(request, operation_id: int, submit: str, save_contact: str,
             "bcghg_id",
             "opt_in",
             "operation_has_multiple_operators",
+            "status",
         }
     )
 
@@ -262,13 +263,10 @@ def update_operation(request, operation_id: int, submit: str, save_contact: str,
             - if operation.status was already "Approved", it should remain Approved and the submission date should not be altered
             - if operation.status was "Changes Requested", it should switch to Pending
             - if operation.status was "Declined", it should switch to Pending
+            - if operation.status was "Not Started", it should switch to Pending
             - if operation.status was "Pending", it should remain as Pending
         """
-        if operation.status in [
-            Operation.Statuses.CHANGES_REQUESTED,
-            Operation.Statuses.DECLINED,
-            Operation.Statuses.PENDING,
-        ]:
+        if operation.status is not Operation.Statuses.APPROVED:
             operation.status = Operation.Statuses.PENDING
             operation.submission_date = datetime.now(pytz.utc)
 
@@ -279,6 +277,9 @@ def update_operation(request, operation_id: int, submit: str, save_contact: str,
 
     operation.save()
     operation.set_create_or_update(modifier=user)
+
+    print("partway through - operation in db:")
+    print(operation.__dict__)
 
     if payload.operation_has_multiple_operators:
         create_or_update_multiple_operators(payload.multiple_operators_array, operation, user)
--- a/bc_obps/registration/api/operation.py
+++ b/bc_obps/registration/api/operation.py
@@ -255,10 +255,22 @@ def update_operation(request, operation_id: int, submit: str, save_contact: str,
 
     for attr, value in payload_dict.items():
         setattr(operation, attr, value)
-    # set the operation status to 'pending' on update
     if submit == "true":
-        operation.status = Operation.Statuses.PENDING
-        operation.submission_date = datetime.now(pytz.utc)
+        """
+        if the PUT request has submit == "true" (i.e., user has clicked Submit button in UI form), the desired behaviour depends on
+        the Operation's status:
+            - if operation.status was already "Approved", it should remain Approved and the submission date should not be altered
+            - if operation.status was "Changes Requested", it should switch to Pending
+            - if operation.status was "Declined", it should switch to Pending
+            - if operation.status was "Pending", it should remain as Pending
+        """
+        if operation.status in [
+            Operation.Statuses.CHANGES_REQUESTED,
+            Operation.Statuses.DECLINED,
+            Operation.Statuses.PENDING,
+        ]:
+            operation.status = Operation.Statuses.PENDING
+            operation.submission_date = datetime.now(pytz.utc)
 
     operation.regulated_products.set(payload.regulated_products)  # set replaces all existing products with the new ones
     operation.reporting_activities.set(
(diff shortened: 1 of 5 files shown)


bc_obps/registration/api/operation.py
```diff
@@ -207,7 +207,27 @@ def update_operation(request, operation_id: int, submit: str, save_contact: str,
...
+            'type',
+            'naics_code_id',
+            'swrs_facility_id',
+            'bcghg_id',
+            'opt_in',
+        ]
+        payload_dict: dict = {}
+        for attr in data_fields:
+            if hasattr(payload, attr):
+                payload_dict[attr] = getattr(payload, attr)
```
Sepehr-Sobhani, line 225: I think it's cleaner to replace this with something like:
```python
        payload_dict: dict= payload.dict(include={
            "name",
            "type",
            "naics_code_id",
            "swrs_facility_id",
            "bcghg_id",
            "opt_in",
        })
```

bc_obps/registration/schema/operation.py
```diff
@@ -49,20 +49,15 @@ class OperationUpdateIn(ModelSchema):
     position_title: Optional[str] = None
     email: Optional[str] = None
     phone_number: Optional[str] = None
+    regulated_products: Optional[List[int]] = []
+    reporting_activities: Optional[List[int]] = []
```
Sepehr-Sobhani, line 53: Do we need these two? I believe they are handled by the `ModelSchema`

bc_obps/registration/schema/operation.py
```diff
@@ -77,6 +72,8 @@ class Config:
             "id",  # need to exclude id since it's auto generated and we don't want to pass it in
             "documents",  # excluding documents because they are handled by individual form fields
             *AUDIT_FIELDS,
+            "status",
+            "submission_date",
```
Sepehr-Sobhani, line 76: Since there are lots of fields that we don't want to handle while updating an operation (to name some of them:           `swrs_facility_id`, `bcghg_id`, `operation_has_multiple_operators`, `verified_at`, `verified_by`, `bc_obps_regulated_operation`, `reporting_activities`, `point_of_contact_id`), I suggest replace `model_exclude` with `model_fields` and only specify the fields that are required for updating and we care about them.

bc_obps/registration/api/operation.py
```diff
@@ -239,42 +259,25 @@ def update_operation(request, operation_id: int, submit: str, save_contact: str,
...
+        """
+        if the PUT request has submit == "true" (i.e., user has clicked Submit button in UI form), the desired behaviour depends on
+        the Operation's status:
+            - if operation.status was already "Approved", it should remain Approved and the submission date should not be altered
+            - if operation.status was "Changes Requested", it should switch to Pending
+            - if operation.status was "Declined", it should switch to Pending
+            - if operation.status was "Not Started", it should switch to Pending
+            - if operation.status was "Pending", it should remain as Pending
+        """
+        if Operation.Statuses(operation.status) is not Operation.Statuses.APPROVED:
```
Sepehr-Sobhani, line 274: nit:
```python
operation.status != Operation.Statuses.APPROVED
```

Sepehr-Sobhani (commented):
Looks great 🤩 ! I plan to continue reviewing this pull request tomorrow, only tests remain.

bc_obps/registration/tests/endpoints/test_operations_endpoints.py
```diff
@@ -657,3 +661,213 @@ def test_put_operation_with_no_point_of_contact(self):
...
+            naics_code_id=operation.naics_code_id,
+            regulated_products=[],
+            # reporting_activities=[],
+            is_external_point_of_contact=True,
+            external_point_of_contact_first_name='Bart',
+            external_point_of_contact_last_name='Simpson',
+            external_point_of_contact_position_title='Scoundrel',
+            external_point_of_contact_email='bart@email.com',
+            external_point_of_contact_phone_number='+17787777777',
+        )
```
Sepehr-Sobhani, line 857: I believe in section 2, we only get fields related to the user and contact and not operation-related fields.

bc_obps/registration/tests/endpoints/test_operations_endpoints.py
```diff
@@ -657,3 +661,213 @@ def test_put_operation_with_no_point_of_contact(self):
...
+            content_type_json,
+            update_from_form_section_2.json(),
+            self.endpoint + '/' + str(operation.id) + "?submit=false&form_section=2",
+        )
+        assert put_response_2.status_code == 200
+        assert Operation.objects.count() == 1
+        retrieved_op = Operation.objects.first()
+        # should be 2 contacts - the original Contact created by baker for the operation, and the updated "Bart" contact
+        assert Contact.objects.count() == 2
+        contact = Contact.objects.last()
```
Sepehr-Sobhani, line 870: I think using `retrieved_op.point_of_contact` is more accurate.

Sepehr-Sobhani:
@andrea-williams FYI: `point_of_contact_id = payload.point_of_contact_id or None`

Sepehr-Sobhani:
@andrea-williams 
Screenshot 2024-02-01 at 1 38 49 PM

bc_obps/registration/tests/endpoints/test_operations_endpoints.py
Sepehr-Sobhani: Skip this!Sorry.

andrea-williams:
@Sepehr-Sobhani updates in my latest push: I rebased against develop so it's up to date, and I've made most of the changes you requested (marked the conversations as resolved). Worth noting that I had to keep `point_of_contact_id: Optional[int] = None` in `OperationUpdateIn` (for now) because in some situations the value of `payload.point_of_contact` is an int, not an object. I don't fully understand what's going on there, but since it's related to the pending refactor/tech debt ticket you just created, I'd rather leave that code in this PR as-is (since it's working) for now. I also had to include `regulated_products` in the list of `model_fields` for `OperationUpdateIn` for django/pydantic to not throw an error about missing a value for a required field.

Hope you're ok with this!

Sepehr-Sobhani (approved):
💯

PR #794, opened 2024-02-01
Revise User Operator form

BCerki:
https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=50189801

To test the workflows in this PR:
- [x] user requests access to an existing operator that has an admin
-- Login as bc-cas-secondary
-- Delete the bc-cas-secondary record in the user operator table: `delete from erc.user_operator where id>2;`  (or temporarily edit the fixtures)
-- Request access to Existing Operator 2
-- You should see the correct confirmation message and no form
- [x] user requests access to an existing operator that DOESN'T have admin
-- login with any bceid and clear the user_operator table (`truncate erc.user_operator;` or adjust the fixtures)
-- request access to any operator
-- You should see the correct confirmation message and no form
- [x] user adds new operation
 -- login with any bceid and clear the user_operator table (`truncate erc.user_operator;` or adjust the fixtures)
-- add an operator
-- fill out the first page of the form and submit (there should no longer be a second page)
-- You should see the correct confirmation message on submission
- [x] edit an operator
--reset the db and login with any bceid
--click the my operator tile and you should be directed to the operator form
-- there should be no second page
-- editing should work without errors and you should be redirected to the dashboard

Since I touched the MultiStepForm component, which is used in every form, I manually checked that still works and it does other than this pre-existing bug: https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=52418646

If you back into the operator tile after requesting access, you should see the same confirmation message you saw when requesting access the first time. For me, the screen flashes here, which seems to be a known next bug: https://github.com/vercel/next.js/issues/57257


client/app/components/routes/select-operator/form/UserOperator.tsx
```diff
@@ -107,11 +98,8 @@ export default async function UserOperator({
     ...userOperatorData,
   };
 
-  return params?.id === "request-access" ? (
-    // If the operator exists then show the form from the second page
-    <UserOperatorContactForm formData={formData} schema={userOperatorPage2} />
-  ) : (
-    <UserOperatorMultiStepForm
```
BCerki, line 114: If we need to re-add the senior officer form later, I think we should be able to conditionally change the `UserOperatorForm` schema rather conditionally using a different form

client/app/utils/jsonSchema/userOperator.ts
```diff
@@ -375,7 +375,6 @@ export const userOperatorSchema: RJSFSchema = {
   type: "object",
   properties: {
     userOperatorPage1,
-    userOperatorPage2,
```
BCerki, line 378: I removed the schema here but left it above in case we need it later

bc_obps/registration/api/user_operator.py
```diff
@@ -306,72 +308,6 @@ def create_operator_and_user_operator(request, payload: UserOperatorOperatorIn):
         return 400, {"message": str(e)}
 
 
-@router.post("/user-operator/contact", response={200: SelectOperatorIn, codes_4xx: Message})
```
BCerki, line 309: I removed this instead of leaving it in in case we need it later because I think the contact-adding logic should be in a put on the user-operator endpoint

bc_obps/registration/tests/endpoints/test_useroperator_endpoints.py
```diff
@@ -176,15 +165,14 @@ def test_unauthorized_users_cannot_put(self):
...
         user_operator = user_operator_baker()
         user_operator.user_id = self.user.user_guid
         user_operator.save(update_fields=['user_id'])
-        response = TestUtils.mock_get_with_auth_role(
-            self, 'industry_user', f"{base_endpoint}user-operator-status-from-user"
-        )
+        response = TestUtils.mock_get_with_auth_role(self, 'industry_user', f"{base_endpoint}user-operator-from-user")
         assert response.status_code == 200
         assert response.json()['status'] == user_operator.status
+        assert response.json().get('is_new') is not None
```
BCerki, line 175: I tested one sample field from the UserOperator ModelField, and is_new because it's from operator

andrea-williams:
Notes while reviewing the Create Operator form:
- the Create Operator form has a section with a radio button for "Does this operator have multiple parent companies?" that doesn't appear in the wireframes. On the one hand, the wireframe just immediately displays a Parent Company info section by default. But on the other hand, in our implementation, if an operator has ONE parent company, the user has to click Yes to MULTIPLE parent companies to be able to enter any info
- I left the address fields blank and am getting "Required field" error messages even on fields that I have populated
Screenshot 2024-02-07 at 2 50 31 PM
- implementation says that physical address field is required, but based on the wireframe, it isn't? Not sure, although I thought we weren't collecting any address info for MVP
- after populating the physical address field, I tried to submit and still got validation errors on required fields that I've filled in 
Screenshot 2024-02-07 at 2 53 45 PM
- after fiddling with those fields giving me errors, I was allowed to Submit the form, but then got a "Server Error" page in the UI. This was in my backend server log:
```
[{'loc': ('path', 'operator_id'), 'msg': 'value is not a valid integer', 'type': 'type_error.integer'}]
Unprocessable Entity: /api/registration/operators/undefined
[07/Feb/2024 22:55:11] "GET /api/registration/operators/undefined HTTP/1.1" 422 115
[{'loc': ('path', 'operator_id'), 'msg': 'value is not a valid integer', 'type': 'type_error.integer'}]
Unprocessable Entity: /api/registration/operator-has-admin/undefined
[07/Feb/2024 22:55:11] "GET /api/registration/operator-has-admin/undefined HTTP/1.1" 422 115
```

BCerki:
> * I left the address fields blank and am getting "Required field" error messages even on fields that I have populated

I wasn't able to reproduce this--do you remember what order you filled the fields out in?

BCerki:
> * implementation says that physical address field is required, but based on the wireframe, it isn't? Not sure, although I thought we weren't collecting any address info for MVP

Also not sure... though there's a validation error on a blank municipality. (Also, I just noticed now that none of the wireframe fields have a * beside them or anything else I can see to indicate required)

BCerki:
> * after populating the physical address field, I tried to submit and still got validation errors on required fields that I've filled in

Also had trouble reproducing

BCerki:
> * after fiddling with those fields giving me errors, I was allowed to Submit the form, but then got a "Server Error" page in the UI. This was in my backend server log:

Fixed!

BCerki:
> * I looked to see what's in my database, and the new operator I was creating shows in the Operator table with a status of "Draft", but on the frontend I'm shown that my request to add my new operator and becomes its representative has been received, so I'm not able to go back and edit it. (This is another argument for prioritizing the ticket for atomic transactions in our API)

I think it's the desired workflow that you can't edit an operator once you've submited it. @nanyangpro and I discussed workflow and she said users shouldn't be able to edit the operator form until their user_operator has been approved. To test the edit, you can either approve the user_operator you just created, or reset the db because in the fixtures the user_operators are already approved.

(Sidenote, the atomic transaction work went in yesterday.)

client/app/components/routes/select-operator/form/UserOperator.tsx
marcellmueller: Yeah we could do that though the design showed a single page form. I was actually going to suggest removing the [formSection] route since I noticed we are still passing the form page. Though I think we can keep it in for now if there is a change we will use the multistep form again.

BCerki: Yes, I don't want to do any drastic refactoring until this is more settled.

client/app/components/form/MultiStepButtons.tsx
```diff
@@ -28,6 +29,11 @@ const SubmitButton: React.FunctionComponent<SubmitButtonProps> = ({
 }) => {
   const isFinalStep = step === steps.length - 1;
   const isDisabled = disabled || isSubmitting;
+  const { data: session } = useSession();
+
+  const isCasInternal =
+    session?.user.app_role?.includes("cas") &&
+    !session?.user.app_role?.includes("pending");
```
marcellmueller, line 36: I suggest changing this to ` const isIndustryUser = session?.user.app_role?.includes("industry");`

Same with the refactored `UserOperatorForm`, I noticed that this was removed since I added it in a recent PR to enhance the logic to display the new operator message.

The issue with using `!isCasInternal` is that using the client session takes a second or two to load so it will flash for a second if logged in as an internal user and look buggy. I confirmed this by logging in as internal user, scrolling to the bottom of a user operator form and refreshing the page.

If we are specific to who this displays for then we don't have the components flashing for internal users.

client/app/components/form/UserOperatorForm.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+  const isFormStatusDisabled =
+    formData?.status === Status.PENDING || formData?.status === Status.APPROVED;
+
+  const isNewOperatorMessage = formData.is_new && formSection === 1;
+  const operatorRoute = isCasInternal ? "operators" : "select-operator";
+
+  return (
+    <>
+      {!isCasInternal && (
```
marcellmueller, line 90: This should be `isIndustryUser` as mentioned in the other comment.

marcellmueller (commented):
Great work on this, it's working really well and I think is a nice improvement on the workflow.

Unrelated to this PR but I think we should suggest to the designers put a `Return to dashboard` or similar button here:

Screenshot 2024-02-09 at 8 13 05 AM

client/app/components/form/MultiStepButtons.tsx
BCerki: So using !isCasInternal or !isIndustryUser is going to cause a flash, but isCasInternal or isIndustryUser works? Or using isIndustryUser/!isIndustryUser everywhere works?

marcellmueller: I think just being specific to who we display the components to will stop components flashing for the wrong users. So avoid using `!isCasInternal` or `!isIndustryUser`. It's something I just recently noticed so there are likely other places to update this, I just noticed it removed with the new `UserOperatorForm` refactor.

Unfortunately there will still be some component pop in as the session loads, though it doesn't look too buggy when it loads in and stays there for the user since we were specific about the users we display it with.

Shon and I recently experimented with passing the server session to the client so there would be no loading issues whatsoever though for now it seems to break CI and e2e tests for some reason so it's a nice to have.

BCerki: Huh, odd. At least we've got a solution even if it's not perfect

marcellmueller (approved):
Awesome work, thanks for updating that. I confirmed that the button doesn't flash anymore for internal users.

Also thanks for the review checklist, this was super helpful 🥇

Issue #550, opened 2024-01-11
Form required fields do not trigger alert "Required field" if null value [10]


shon-button:
Describe the Bug:
Users should be alerted to required field (even if the field has never had a value, e.g. Position Title)

![image](https://github.com/bcgov/cas-registration/assets/120038448/4f5f7394-4fe0-4129-b507-518c1433aa71)

Probability (how likely the bug is to happen, scored from 1-5): 5
(For example, probability of 5 is something like "it happens to all users every time they log in."
Probability of 1 "only happens to certain users when a really specific and unlikely path is followed.")

Effect (how bad the bug is when it does happen, scored from 1-5): 5
(For example, effect of 5 is "the entire app crashes and makes it unusable for all users" or "the bug causes the wrong data to be saved, with critical information (e.g. payment) being affected."
Effect of 1 is "It makes some styling look a little bit weird.")

Steps to reproduce the behaviour:

Screenshots:

Additional information:

- Browser: [e.g. chrome, safari]
- Version: [e.g. 22]
- Additional context

marcellmueller:
https://github.com/rjsf-team/react-jsonschema-form/issues/289

According to this an empty string is still a valid string in the JSON schema spec. What is happening here is in `User.tsx` the default `formData` passes in empty strings for fields if there is no data to pre-populate it with causing this to pass validation.

The easiest fix would be to alter `UserProfileFormData` types to accept `string | undefined` and change all the `""` empty strings to `undefined` in the `User.tsx` formData.

We could also look at seeing if we can alter the form `validator` to trigger an error if a field has empty strings as it is a bit unintuitive even if it is valid by the JSON schema spec.

shon-button:
CHANGES:
```
export interface UserProfileFormData {
  first_name: string | undefined;
  last_name: string | undefined;
  position_title: string | undefined;
  email: string | undefined;
  phone_number: string | undefined;
}
```

```
      formData = {
        first_name: names?.[0] ?? undefined,
        last_name: names?.[1] ?? undefined,
        email: session?.user?.email ?? undefined,
        phone_number: undefined,
        position_title: undefined,
      };
```
RETURNS:
`position_title` is still deemed as valid...

(Perhaps) The reason this data passes validation is because the undefined value is considered a valid value for a property in TypeScript when the property is marked as string | undefined. TypeScript allows undefined to be assigned to a property even if the property is marked as required in the schema.


Also tried change:
```
export interface UserProfileFormData {
[...]

PR #798, opened 2024-02-01
fix: user profile empty string values not triggering form validation

marcellmueller:
Fix for #550 

This was caused since an empty string still counts as a string in JSON Schema. I have created a `UserProfilePartialFormData` type to be used when we won't have all of the user profile form data and kept the original `UserProfileFormData` type for when it all needs to be there.


andrea-williams (approved):
👌

Issue #797, opened 2024-02-01
Session component show "loading" on home page [6]


shon-button:
**Describe the Bug:**
Currently, the Session component `client/app/components/navigation/Session.tsx`  flashes "loading" on the home page.

**Background:**
With the  removal of the `Login` component from the Session component, there should not be any loading state indicator on the home page because there is nothing loading (anymore).

**Possible Resolution**

- Add useSession hook to `client/app/components/layout/Header.tsx`
- In Header, display `client/app/components/navigation/Profile.tsx` conditionally based on session?
- Delete `client/app/components/navigation/Session.tsx`  (YAGNI principle)
- Delete `client/app/components/navigation/Login.tsx`   (YAGNI principle)
- Possibly redesign the layout components as follows:
  - Add Header; Footer; Main to `client/app/layout.tsx`
  - Delete `client/app/(onboarding)/layout.tsx`
  - Modify `client/app/(authenticated)/layout.tsx` to nest Breadcrumbs within `client/app/layout.tsx`

Probability (how likely the bug is to happen, scored from 1-5): 5
(For example, probability of 5 is something like "it happens to all users every time they log in."
Probability of 1 "only happens to certain users when a really specific and unlikely path is followed.")

Effect (how bad the bug is when it does happen, scored from 1-5): 1
(For example, effect of 5 is "the entire app crashes and makes it unusable for all users" or "the bug causes the wrong data to be saved, with critical information (e.g. payment) being affected."
Effect of 1 is "It makes some styling look a little bit weird.")

Steps to reproduce the behaviour:

Screenshots:

Additional information:

- Browser: [e.g. chrome, safari]
- Version: [e.g. 22]
- Additional context

dleard:
developer to review (not a PO concern)

PR #803, opened 2024-02-01
797 session loading

marcellmueller:
Fix for #797 

I tried to pass the server session to the header through a wrapper component but it seems to break CI. Might be worth looking into down the road as it's a nice improvement on the front end and stops components from visibly loading in.

Also deleted `Session.tsx` and `Login.tsx`

 Implemented part of the layout suggestions in the issue. I kept the `(onboarding)` layout and just moved the `Header` and `Footer` components out of `(authenticated)/layout.tsx` and `(onboarding)/layout.tsx` to the top level `layout.tsx`.

Needed to keep `<Main>` in the `(onboarding)/layout.tsx` as it needs to have a custom top margin since it doesn't have the breadcrumbs component. This is due to the Nav components needing to use `position: absolute` so we can extend them past the max page width:

Screenshot 2024-02-01 at 3 40 40 PM


andrea-williams (approved):
Deleting more code than is added - yay! *chef's kiss*

Issue #806, opened 2024-02-02
Remove "Request Changes" feature for Operators


andrea-williams:
#### Describe the task

We accidentally introduced the "Request Changes" feature on both Operations and Operators, when it should have been for Operations only. 

Requesting Changes on an Operator isn't straightforward because the status of an Operator isn't displayed anywhere in the UI. To simplify things for MVP, we want to remove the option for any user to Request Changes on an Operator. 

#### Acceptance Criteria

- [ ] delete the "Request Changes" button from the UI on the Operator form (is only visible to internal users right now)
- [ ] ensure the "Request Changes" button remains in the UI on the Operation form (should only be visible to internal users)

#### Additional context

- Add any other context about the task here.
- Or here

PR #808, opened 2024-02-02
Remove "Request Changes" feature for Operators

marcellmueller:
Implements #806 

Added a couple comments below for your consideration thanks!

### DIFF
--- a/client/app/components/button/Review.tsx
+++ b/client/app/components/button/Review.tsx
@@ -18,8 +18,8 @@ interface Props {
   showRequestChanges?: boolean;
   onApprove: () => Promise<any>;
   onReject: () => Promise<any>;
-  onRequestChange: () => Promise<any>;
-  onUndoRequestChange: () => Promise<any>;
+  onRequestChange?: () => Promise<any>;
+  onUndoRequestChange?: () => Promise<any>;
 }
 
 interface CloseProps {
@@ -107,12 +107,12 @@ const Review = ({
   };
 
   const handleChangeRequestConfirm = () => {
-    onRequestChange();
+    onRequestChange?.();
     setShowRequestChangesUndo(true);
   };
 
   const handleUndoRequestChanges = async () => {
-    onUndoRequestChange();
+    onUndoRequestChange?.();
     setShowChangeConfirmation(false);
     setShowRequestChangesUndo(false);
   };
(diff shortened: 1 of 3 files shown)


client/app/components/button/Review.tsx
```diff
@@ -107,12 +107,12 @@ const Review = ({
...
 
   const handleChangeRequestConfirm = () => {
-    onRequestChange();
+    onRequestChange?.();
     setShowRequestChangesUndo(true);
   };
 
   const handleUndoRequestChanges = async () => {
-    onUndoRequestChange();
+    onUndoRequestChange?.();
```
marcellmueller, line 115: I haven't used optional chaining to invoke a function that may not exist but it works. Not sure how I feel about it though I used it because our eslint settings wouldn't let me write it how I normally would ie:

`onRequestChange && onRequestChange();`

client/app/components/form/UserOperatorMultiStepForm.tsx
```diff
@@ -120,19 +119,7 @@ export default function UserOperatorMultiStepForm({
...
-              userOperator={formData as UserOperatorFormData}
-              userOperatorId={Number(userOperatorId)}
-              isOperatorNew={formData?.is_new}
-              operatorId={formData?.operator_id}
-              showRequestChanges
-            />
-          </>
-        )}
-        {isReviewWithoutRequestChanges && (
+        {isCasInternal && (
```
andrea-williams, line 122: we still want to keep the `<Note>` component when an operator is new.

client/app/components/button/Review.tsx
andrea-williams: I'm also not v familiar with the `?` usage on a function, but I had a quick look at the docs and you seem to have used it correctly, so I'm happy :)

andrea-williams (changes requested):
Just requesting a change to not delete the "this operator is new" message that appears for internal users when an operator's `is_new` flag is `true`. Otherwise, looking good!

marcellmueller:
> Just requesting a change to not delete the "this operator is new" message that appears for internal users when an operator's `is_new` flag is `true`. Otherwise, looking good!

Ah good catch thanks! I've added it back with a new conditional.

andrea-williams (approved):
Re-ran the code locally, have confirmed that the Note to internal users about a new operator is back in, and the Request Changes button only shows on Operations, not Operators. 

Thanks @marcellmueller !

PR #814, opened 2024-02-05
Optimize endpoints - GET operations

Sepehr-Sobhani:
[ISSUE 815](https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=52270431)


bc_obps/registration/api/operation.py
```diff
@@ -192,7 +202,7 @@ def update_operation(request, operation_id: int, submit: str, form_section: int,
         raise HttpError(401, UNAUTHORIZED_MESSAGE)
     operator = Operator.objects.get(id=user_operator.operator_id)
 
-    approved_users = get_an_operators_approved_users(operator)
+    approved_users = get_an_operators_approved_users(operator.pk)
```
andrea-williams, line 205: Is `operator.pk` synonymous with `operator.id`?

andrea-williams (approved):
You are an absolute champion @Sepehr-Sobhani !

bc_obps/registration/api/operation.py
Sepehr-Sobhani: Unless we define a custom ID and primary key for our data model, Django will create an auto-increment integer field as the primary key by default, which is the same as ID.

PR #871, opened 2024-02-09
474 rejection

BCerki:
https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=48609943

To manually test this:
- login with your IDIR and go to the operators tile
- decline New Operator 3 Legal Name
- check in the db that New Operator 3 Legal Name and its user_operator have both been declined

PS: sonarcloud fails are all in test files so I'll dismiss them unless the reviewer has concerns
PPS: the approve operator workflow isn't working well atm, there's already a ticket: https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=52823560

### DIFF
--- a/bc_obps/registration/tests/utils/bakers.py
+++ b/bc_obps/registration/tests/utils/bakers.py
@@ -44,14 +44,19 @@ def contact_baker():
     return baker.make(Contact)
 
 
-def operator_baker():
-    return baker.make(
-        Operator,
-        bc_corporate_registry_number=generate_random_bc_corporate_registry_number(),
-        business_structure=BusinessStructure.objects.first(),
-        physical_address=address_baker(),
-        website='https://www.example-operator.com',
-    )
+def operator_baker(custom_properties=None):
+    properties = {
+        'bc_corporate_registry_number': generate_random_bc_corporate_registry_number(),
+        'business_structure': BusinessStructure.objects.first(),
+        'physical_address': address_baker(),
+        'website': 'https://www.example-operator.com',
+    }
+
+    # Update properties with custom_properties if provided
+    if custom_properties:
+        properties.update(custom_properties)
+
+    return baker.make(Operator, **properties)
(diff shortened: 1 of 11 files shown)


bc_obps/registration/tests/utils/bakers.py
```diff
@@ -44,14 +44,19 @@ def contact_baker():
...
 
-def operator_baker():
-    return baker.make(
-        Operator,
-        bc_corporate_registry_number=generate_random_bc_corporate_registry_number(),
-        business_structure=BusinessStructure.objects.first(),
-        physical_address=address_baker(),
-        website='https://www.example-operator.com',
-    )
+def operator_baker(custom_properties=None):
```
BCerki, line 47: My tests were getting repetitive to the point of sonarcloud fails, so tweaked these functions to allow for DRYer test code

marcellmueller:
This is working great! Though as mentioned in teams I wonder if we should filter out Prime Admin requests that are tied to an Operator that is declined? Currently there is no way to undo declining an Operator so we could end up with a bunch of stuck Prime Admin requests.

Also not part of the acceptance criteria but post MVP we could enhance this by hiding the Prime Admin review component if a request has been denied. Though as is it is fine since you can't approve a request with a denied Operator.

client/app/components/routes/select-operator/form/UserOperator.tsx
```diff
@@ -23,6 +23,7 @@ async function getBusinessStructures() {
 export async function getUserOperatorFormData(id: number | string) {
   if (!id || isNaN(Number(id))) return {};
   return actionHandler(
+    // brianna
```
marcellmueller, line 26: Is this intentional?

BCerki: Definitely, I'm signing my work like an artiste

PR #878, opened 2024-02-12
chore: reduce user cache timeout to 5 mins

dleard:
Reducing the timout on the cache to 5 minutes. This way we're still not sending a request to the database on every request (there will be a request every 5 minutes for the user) but we also don't have to wait 1 hour for manual changes to apply when assigning internal permissions. We can revisit the timeout and increase it after we have implemented the user management workflow for internal users.

### DIFF
--- a/bc_obps/registration/middleware/current_user_middleware.py
+++ b/bc_obps/registration/middleware/current_user_middleware.py
@@ -23,7 +23,7 @@ class CurrentUserMiddleware:
                     # If user is not in cache, fetch from database
                     user = get_object_or_404(User, user_guid=user_guid)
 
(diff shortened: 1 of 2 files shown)


bc_obps/registration/middleware/current_user_middleware.py
```diff
@@ -24,7 +24,7 @@ def __call__(self, request):
                     user = get_object_or_404(User, user_guid=user_guid)
 
                     # Cache the user for 1 hour
```
andrea-williams, line 26:
```suggestion
                    # Cache the user for 5 minutes
```

PR #899, opened 2024-02-13
Fix: 890 duplicate user operator records

dleard:
It was possible to create rows in the user_operator table with duplicate user-operator pairs. This PR adds a unique constraint so that is no longer possible.

To test in the app:
- A subsequent user should not be able to request access to the same operator twice

To test the constraint:
- Create a user_operator access request & then attempt to insert a new row in that table with the same user-operator pair. It should throw a unique constraint error


bc_obps/registration/tests/test_models.py
```diff
@@ -594,6 +603,21 @@ def setUpTestData(cls):
             ("verified_at", "verified at", None, None),
         ]
 
+    def test_unique_cra_business_number_constraint(self):
```
Sepehr-Sobhani, line 606: nit: copy/paste function name issue

Sepehr-Sobhani (approved):
Just needs a function renaming tweak otherwise good to go 🤩

Issue #866, opened 2024-02-09
The Access requests view seen by internal CAS users should only show admin requests for operators with no approved admin


dleard:
The list of access requests that is seen by an internal user should be filtered to exclude access requests to an operator that already has an approved admin. If an operator has an approved admin, then the user management should be offloaded to the admin(s) of that operator. CAS should only be responsible for approving access for the first admin.

It looks like the /user-operators endpoint is only used by this view. My suggestion is to change the name of this endpoint to be clearer about what it is actually being fetched, something like /user-operator-initial-requests. This endpoint should only fetch the user-operators for operators that do not already have an approved admin.

AC:
- [ ] The /user-operators endpoint is renamed to something that is more explicit about the data being fetched
- [ ] The endpoint filters out any rows where the related operator already has an approved admin
- [ ] The endpoint has been checked for performance with the django-silk profiler
- [ ] Tests have been written/updated

PR #911, opened 2024-02-14
866 internal access requests view

marcellmueller:
#866

I've also updated the fixtures so that we have multiple prime admin requests tied to different operators so the dashboard won't be empty once one is approved.

To test this you can sign in as `cas_admin` and approve the prime admin request of an operator. Once you return to the dashboard you should see no prime admin requests that are tied to that same operator.


shon-button:
@marcellmueller you might want to review the validity of the fixtures as some do not align correctly with the "Approval" flow:
image

PR #912, opened 2024-02-14
Refactor backend tests

Sepehr-Sobhani:


andrea-williams (approved):
It would probably be wise to update our developer guide to explain how to use `custom_reverse_lazy` at some point, but we can leave that for a future refactor post-launch.

Looks good!

Issue #914, opened 2024-02-14
URL-encode the DB password pulled from env vars [10]


andrea-williams:
Describe the Bug:
In `bc_obps/settings.py`, we build a postgres connection string using values pulled from environment variables. If any one of these values contains special characters (hint - the password that postgres generates), the parsing of the connection string explodes and affects our ability to deploy to Openshift.

In the `default_db_url`, we need to modify `{os.environ.get('DB_PASSWORD')}` to convert the value of DB_PASSWORD to a URL-encoded string. Manually changing the value of the secret in Openshift to be URL-encoded proved that this fix will work.

Probability (how likely the bug is to happen, scored from 1-5):
(For example, probability of 5 is something like "it happens to all users every time they log in."
Probability of 1 "only happens to certain users when a really specific and unlikely path is followed.")

Effect (how bad the bug is when it does happen, scored from 1-5):
(For example, effect of 5 is "the entire app crashes and makes it unusable for all users" or "the bug causes the wrong data to be saved, with critical information (e.g. payment) being affected."
Effect of 1 is "It makes some styling look a little bit weird.")

Steps to reproduce the behaviour:

Screenshots:

Additional information:

- Browser: [e.g. chrome, safari]
- Version: [e.g. 22]
- Additional context

PR #915, opened 2024-02-14
Fix: URL encode DB_PASSWORD in database conn string

andrea-williams:
Issue: #914 

**Manual testing**
I updated my local `bc_obps/.env` file with: `DB_PASSWORD=D)b7?m{hkukhklh*nk,8798}` and inserted a temporary print statement in `bc_obps/settings.py` to print out the parsed value of `default_db_url`, which output `postgres://awilliam:D%29b7%3Fm%7Bhkukhklh%2Ank%2C8798%7D@localhost:5432/registration`

### DIFF
--- a/bc_obps/bc_obps/settings.py
+++ b/bc_obps/bc_obps/settings.py
@@ -99,7 +99,7 @@ WSGI_APPLICATION = "bc_obps.wsgi.application"
 
 # Database
 # https://docs.djangoproject.com/en/4.2/ref/settings/#databases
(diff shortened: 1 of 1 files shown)


dleard (approved):

andrea-williams:
@dleard  FYI the GHA check failed at the check-migrations step. Turns out we need to explicitly cast the env var to a str (to handle digits) before we URL-encode it. I've pushed up the change and it's passed now. Just pointing it out in case you were curious.

PR #917, opened 2024-02-14
840 fix parent operators

BCerki:
card: https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=52562357

Please pay special attention to the python--I'm fairly confident that it works, but less confident it's good python and sonarcloud agrees

I'll dimiss the sonarcloud fails in the testing files if the reviewer is okay with it. There's an open PR with a slight refactor to the baker files to make for less repetition, but not in yet: https://github.com/bcgov/cas-registration/pull/871/files#diff-f0d287d6538eb3a422d5728bcffb9fa1cd9bc62c78eb9326a2394f2f715c1a9d

### DIFF
--- a/bc_obps/registration/api/user_operator.py
+++ b/bc_obps/registration/api/user_operator.py
@@ -1,5 +1,5 @@
 from django.db import transaction
-from django.db import  transaction
+from django.db import transaction
 from registration.api.utils.handle_parent_operators import handle_parent_operators
 from registration.utils import (
     generate_useful_error,
--- a/bc_obps/registration/api/user_operator.py
+++ b/bc_obps/registration/api/user_operator.py
@@ -86,18 +86,40 @@ def save_operator(payload: UserOperatorOperatorIn, operator_instance: Operator, 
         # create parent operator records
         operator_has_parent_operators: bool = payload.operator_has_parent_operators
         if operator_has_parent_operators:
-            po_operator_fields_mapping = {
-                "po_legal_name": "legal_name",
-                "po_trade_name": "trade_name",
-                "po_cra_business_number": "cra_business_number",
-                "po_bc_corporate_registry_number": "bc_corporate_registry_number",
-                "po_website": "website",
-            }
             for idx, po_operator in enumerate(payload.parent_operators_array):
-                new_po_operator_instance: ParentOperator = ParentOperator(
-                    child_operator=created_or_updated_operator_instance,
-                    operator_index=idx + 1,
+
+                existing_parent_operator_indices = list(
+                    ParentOperator.objects.filter(child_operator=created_or_updated_operator_instance).values_list(
+                        'operator_index', flat=True
+                    )
                 )
+
+                # archive any parent operators that have been removed
+                if existing_parent_operator_indices:
+
+                    def get_indices(po):
+                        return po.operator_index
+
+                    updated_parent_operator_indices = [
+                        get_indices(po) for po in payload.parent_operators_array if get_indices(po) is not None
+                    ]
+
+                    indices_to_delete = list(
+                        set(existing_parent_operator_indices) - set(updated_parent_operator_indices)
+                    )
+
+                    if indices_to_delete:
+                        for op_index in indices_to_delete:
+                            ParentOperator.objects.get(operator_index=op_index).set_archive(user.pk)
+
+                # assign an operator_index to new parent operators
+                if not po_operator.operator_index:
+                    highest_existing_index = (
+                        max(existing_parent_operator_indices) if len(existing_parent_operator_indices) > 0 else 0
+                    )
+
+                    po_operator.operator_index = highest_existing_index + 1
+
                 # handle addresses--if there's no mailing address given, it's the same as the physical address
                 po_physical_address = Address.objects.create(
                     street_address=po_operator.po_physical_street_address,
@@ -105,10 +127,9 @@ def save_operator(payload: UserOperatorOperatorIn, operator_instance: Operator, 
                     province=po_operator.po_physical_province,
                     postal_code=po_operator.po_physical_postal_code,
                 )
-                new_po_operator_instance.physical_address = po_physical_address
 
                 if po_operator.po_mailing_address_same_as_physical:
-                    new_po_operator_instance.mailing_address = po_physical_address
+                    po_mailing_address = po_physical_address
                 else:
                     po_mailing_address = Address.objects.create(
                         street_address=po_operator.po_mailing_street_address,
@@ -116,14 +137,23 @@ def save_operator(payload: UserOperatorOperatorIn, operator_instance: Operator, 
                         province=po_operator.po_mailing_province,
                         postal_code=po_operator.po_mailing_postal_code,
                     )
-                    new_po_operator_instance.mailing_address = po_mailing_address
 
-                new_po_operator_instance.business_structure = po_operator.po_business_structure
-                new_po_operator_instance = update_model_instance(
-                    new_po_operator_instance, po_operator_fields_mapping, po_operator.dict()
+                po_operator_instance, _ = ParentOperator.objects.update_or_create(
+                    child_operator=created_or_updated_operator_instance,
+                    operator_index=po_operator.operator_index,
+                    defaults={
+                        "legal_name": po_operator.po_legal_name,
+                        "trade_name": po_operator.po_trade_name,
+                        "cra_business_number": po_operator.po_cra_business_number,
+                        "bc_corporate_registry_number": po_operator.po_bc_corporate_registry_number,
+                        "business_structure": po_operator.po_business_structure,
+                        "website": po_operator.po_website,
+                        "physical_address": po_physical_address,
+                        "mailing_address": po_mailing_address,
+                    },
                 )
-                new_po_operator_instance.save()
-                new_po_operator_instance.set_create_or_update(user.pk)
+
+                po_operator_instance.set_create_or_update(user.pk)
 
         # get an existing user_operator instance or create a new one with the default role
         user_operator, created = UserOperator.objects.get_or_create(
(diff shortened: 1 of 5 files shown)


bc_obps/registration/tests/endpoints/test_useroperator_endpoints.py
```diff
@@ -687,7 +688,77 @@ def test_create_operator_and_user_operator_with_parent_operators(self):
...
+            'industry_user',
+            content_type_json,
+            mock_payload,
+            f"{self.user_operator_endpoint}/operator/{user_operator.id}",
+        )
+        assert response.status_code == 200
+
+        parent_operators: List[ParentOperator] = child_operator.parent_operators.all()
+
+        assert len(parent_operators) == 1  # archived records are not pulled
```
BCerki, line 745: @Sepehr-Sobhani , is it expected that archived records don't show up when doing a selection? I didn't add any code to do this; I just used the existing `set_archive` function

Sepehr-Sobhani: Yes, we're using the `TimeStampedModelManager` and that custom manager automatically filters out archived records for every data model inheriting `TimeStampedModel` and we only see the active ones. If we need to access the archived records for any particular data model, we'll need to override this custom manager specifically for that purpose.

bc_obps/registration/api/user_operator.py
```diff
@@ -87,44 +87,74 @@ def save_operator(payload: UserOperatorOperatorIn, operator_instance: Operator,
         # create parent operator records
         operator_has_parent_operators: bool = payload.operator_has_parent_operators
         if operator_has_parent_operators:
```
Sepehr-Sobhani, line 89: Should we archive all parent operators associated with an operator if the user removes all of them?

bc_obps/registration/api/user_operator.py
```diff
@@ -87,44 +87,74 @@ def save_operator(payload: UserOperatorOperatorIn, operator_instance: Operator,
...
-            po_operator_fields_mapping = {
-                "po_legal_name": "legal_name",
-                "po_trade_name": "trade_name",
-                "po_cra_business_number": "cra_business_number",
-                "po_bc_corporate_registry_number": "bc_corporate_registry_number",
-                "po_website": "website",
-            }
             for idx, po_operator in enumerate(payload.parent_operators_array):
```
Sepehr-Sobhani, line 90: Since we don't use `idx` any longer, we don't need to use `enumerate` to get the index.

bc_obps/registration/api/user_operator.py
```diff
@@ -87,44 +87,74 @@ def save_operator(payload: UserOperatorOperatorIn, operator_instance: Operator,
...
-                new_po_operator_instance: ParentOperator = ParentOperator(
-                    child_operator=created_or_updated_operator_instance,
-                    operator_index=idx + 1,
+
+                existing_parent_operator_indices = list(
+                    ParentOperator.objects.filter(child_operator=created_or_updated_operator_instance).values_list(
+                        'operator_index', flat=True
+                    )
                 )
```
Sepehr-Sobhani, line 96: I suggest modifying this code to enhance its readability and efficiency. Here is the updated code:
```python
existing_parent_operator_indices = list(created_or_updated_operator_instance.parent_operators.values_list('operator_index', flat=True))
```
We can move this code out of the loop and place it above `if operator_has_parent_operators:` statement. This way, we can avoid running the same query multiple times during the loop execution. Moreover, we can use `existing_parent_operator_indices` to check for my earlier comment. For instance, if the user removes all parent operators, `operator_has_parent_operators` would evaluate to false. However, if `existing_parent_operator_indices` is true, we must archive all existing parent operators related to this operator.

bc_obps/registration/api/user_operator.py
```diff
@@ -87,44 +87,74 @@ def save_operator(payload: UserOperatorOperatorIn, operator_instance: Operator,
...
+
+                # archive any parent operators that have been removed
+                if existing_parent_operator_indices:
+
+                    def get_indices(po):
+                        return po.operator_index
+
+                    updated_parent_operator_indices = [
+                        get_indices(po) for po in payload.parent_operators_array if get_indices(po) is not None
+                    ]
```
Sepehr-Sobhani, line 106: We can change this to:
```python
updated_parent_operator_indices = [po.operator_index for po in payload.parent_operators_array if po.operator_index]
```
This way we don't need to define `get_indices` function on every loop execution.

bc_obps/registration/api/user_operator.py
```diff
@@ -87,44 +87,74 @@ def save_operator(payload: UserOperatorOperatorIn, operator_instance: Operator,
...
+
+                    indices_to_delete = list(
+                        set(existing_parent_operator_indices) - set(updated_parent_operator_indices)
+                    )
+
+                    if indices_to_delete:
+                        for op_index in indices_to_delete:
+                            ParentOperator.objects.get(operator_index=op_index).set_archive(user.pk)
```
Sepehr-Sobhani, line 114: This function will archive all parent operators that are related to other operators but have the same operator index. We can replace it with something like:
```python
if indices_to_delete:
    parent_operators_to_delete: List[ParentOperator] = created_or_updated_operator_instance.parent_operators.filter(operator_index__in=indices_to_delete)
    for po in parent_operators_to_delete:
        po.set_archive(user.pk)
```

bc_obps/registration/api/user_operator.py
```diff
@@ -87,44 +87,74 @@ def save_operator(payload: UserOperatorOperatorIn, operator_instance: Operator,
...
+
+                # assign an operator_index to new parent operators
+                if not po_operator.operator_index:
+                    highest_existing_index = (
+                        max(existing_parent_operator_indices) if len(existing_parent_operator_indices) > 0 else 0
```
Sepehr-Sobhani, line 119: We don't need `len(existing_parent_operator_indices) > 0`, in Python, an empty list is considered falsy.
```python
highest_existing_index = max(existing_parent_operator_indices) if existing_parent_operator_indices else 0
```

bc_obps/registration/api/user_operator.py
```diff
@@ -87,44 +87,74 @@ def save_operator(payload: UserOperatorOperatorIn, operator_instance: Operator,
...
+
+                    po_operator.operator_index = highest_existing_index + 1
+
                 # handle addresses--if there's no mailing address given, it's the same as the physical address
                 po_physical_address = Address.objects.create(
                     street_address=po_operator.po_physical_street_address,
                     municipality=po_operator.po_physical_municipality,
                     province=po_operator.po_physical_province,
                     postal_code=po_operator.po_physical_postal_code,
                 )
```
Sepehr-Sobhani, line 130: I suggest updating a parent operator's Address(physical/mailing) with new data instead of creating new Addresses (of course when we are doing the update). If we keep creating new Addresses, we may end up with many orphaned addresses that are not associated with any parent operator.

bc_obps/registration/api/user_operator.py
Sepehr-Sobhani, line 1: My suggestion is to split the parent operator part of the `save_operator` function into two separate functions, such as `create_parent_operator` and `update_parent_operator`. This way, we can have more manageable functions that are easier to test and debug.

bc_obps/registration/tests/endpoints/test_useroperator_endpoints.py
```diff
@@ -740,7 +742,77 @@ def test_create_operator_and_user_operator_with_parent_operators(self):
...
+
+        # parent_operator_1 has been archived
+        parent_operator_1.refresh_from_db()
+        assert parent_operator_1.operator_index == 1
+        assert parent_operator_1.archived_by is not None
+        assert parent_operator_1.archived_at is not None
+
+        # parent_operator_2 has been edited
+        assert parent_operators[0].legal_name == "test po legal name-EDITED"
+        parent_operators[0].operator_index == 2
```
Sepehr-Sobhani, line 809: Missing `assert`?

bc_obps/registration/tests/endpoints/test_useroperator_endpoints.py
```diff
@@ -740,7 +742,77 @@ def test_create_operator_and_user_operator_with_parent_operators(self):
         assert parent_operators[0].operator_index == 1
         assert parent_operators[1].operator_index == 2
 
+    def test_edit_and_archive_parent_operators(self):
```
Sepehr-Sobhani, line 745: Based on the comments above, We might want to change tests a little bit.

Sepehr-Sobhani (changes requested):
I really appreciate that you are addressing this issue. 🙏 It looks really good. I have requested a couple of changes to fix some minor issues and make the code more manageable. Additionally, I have shared my comments about this pull request with @andrea-williams . Since I will be away tomorrow, you can follow up on the changes with her and avoid blocking other issues.

bc_obps/registration/api/user_operator.py
BCerki: Thanks for catching this!! 😱

bc_obps/registration/api/user_operator.py
BCerki: Yes, I agree. There's a different ticket for that: https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=51034863

bc_obps/registration/api/user_operator.py
BCerki: Thanks for the refactor! This has to stay in the loop or else highest_index won't change and we just keep overwriting the same id.

Sepehr-Sobhani (approved):
This looks great! 🏅  Two minor changes are required but I'm approving this to not block you once those are covered.

Issue #887, opened 2024-02-12
In the operator form, website says it is optional but RJSF validation blocks continuing with an empty website value


dleard:
In the operator form, when updating an operator's information, the website field says it is optional but when you try to submit the form with a blank website the RJSF validation fires & you cannot submit unless you add a website.

To reproduce:
- as an industry user with admin access to an operator
- go to the operator form & try to submit it with an empty website
- the form submission is blocked

Expected behaviour is that website is not a required field

PR #921, opened 2024-02-14
fix: empty string causing url widget validation to trigger

marcellmueller:
Fix for #887 

Steps to recreate this bug:
- Log in as `bc-cas-dev`
- Go to operator form
- Edit the `website` field and remove the URL all together and save
- Click edit again and try to save
- The URL validation triggers because django is returning empty string

To solve this I have added a useEffect which will convert an empty value to undefined so validation won't erroneously trigger.

### DIFF
--- a/client/app/components/form/widgets/URLWidget.tsx
+++ b/client/app/components/form/widgets/URLWidget.tsx
@@ -1,9 +1,20 @@
 "use client";
 
+import { useEffect } from "react";
 import TextWidget from "./TextWidget";
 import { WidgetProps } from "@rjsf/utils/lib/types";
 
 const URLWidget: React.FC<WidgetProps> = (props) => {
+  const { onChange, value } = props;
+
+  useEffect(() => {
+    // Set value to undefined if it's an empty string as this was causing validation to trigger
+    // If a user had previously saved a URL, then deleted it the field would have an empty string
+    if (!value) {
+      onChange(undefined);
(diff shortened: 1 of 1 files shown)


BCerki (changes requested):
Thanks for tackling this one; it's been such a pain!

The main operator form works well now, but something's gone wrong with the parent operator form. I have to double-click the button to show the form and then once I start typing, the radio box goes back to "no".

Another way to fix this bug might be to set the rjsf default value to "undefined" (haven't tested it, just a thought)

marcellmueller:
> Thanks for tackling this one; it's been such a pain!
> 
> The main operator form works well now, but something's gone wrong with the parent operator form. I have to double-click the button to show the form and then once I start typing, the radio box goes back to "no".
> 
> Another way to fix this bug might be to set the rjsf default value to "undefined" (haven't tested it, just a thought)

Wow what a catch haha. Seems like the useEffect gets triggered no matter what since RJSF re-renders the entire form as you use it. I had considered adding a default value but it seems like a bit of a band aid solution and someone might run into the same problem in the future.

I tried removing the useEffect dependencies so it would only run once and it was still causing issues. So I've removed the useEffect all together and just run the onChange without it and this seems to be working well and no funny business with the array fields. Let me know what you think.

BCerki (approved):
:cartwheeling:

PR #929, opened 2024-02-15
Fix: 885 declined request side effects

dleard:
We made an incorrect assumption in many of our endpoints that users will have a 1:1 relationship with the user_operator table. A user can be declined (for example if they requested access to the incorrect operator) and then create a new request that is then approved. 

In this PR, many of our endpoints now ignore DECLINED statuses when fetching information and on the frontend a couple screens have been added to show that they have been declined when a user attempts to select an operator where their status is declined.

This PR has touch a lot of endpoints and will need thorough walkthroughs of all the workflows during review. 
Before pushing I tested the following workflows:
- External can request admin access
- Internal can decline admin access
- Internal sees the DECLINED request row
- External can still select operators when they have a row with DECLINED status in the user_operator table
- Externalwhen selecting the operator they were declined for from the select-operator flow sees a message that they were declined by CAS
- External can request admin access to a different operator
- Internal can approve the admin access to a different operator
- External can see the admin tiles for the APPROVED operator
- External can edit the operator data for the APPROVED operator
- External can complete a BORO ID request for an existing operation under the APPROVED operator
- External can create new operations and complete a BORO ID request under the APPROVED operator
- External can view their access in the User Management page
- A subsequent user can request access to the operator
- External can decline the subsequent user's access
- Subsequent user can still select operators with a row with DECLINED status in the user_operator table
- Subsequent user, when selecting the operator they were declined for sees a message that they were declined by the Operator's Admin
- Subsequent user can request access to a different operator
- Subsequent user can be given access to a different operator

### DIFF
--- a/bc_obps/registration/api/user_operator.py
+++ b/bc_obps/registration/api/user_operator.py
@@ -107,6 +107,18 @@ def get_user_operator_from_user(request):
     user_operator = get_object_or_404(UserOperator, user_id=request.current_user.user_guid)
     operator = get_object_or_404(Operator, id=user_operator.operator_id)
     return 200, {**user_operator.__dict__, "is_new": operator.is_new, "operator_status": operator.status}
+def get_user_operator_status(request):
+    try:
+        user_operator = (
+          UserOperator.objects.only("id", "status", "operator__id", "operator__is_new")
+          .exclude(status=UserOperator.Statuses.DECLINED)
+          .select_related("operator")
+          .get(user_id=request.current_user.user_guid)
+
+        )
+    except UserOperator.DoesNotExist:
+        return 404, {"message": "User is not associated with any operator"}
+    return 200, PendingUserOperatorOut.from_orm(user_operator)
 
 
 @router.get(
@@ -134,8 +146,10 @@ def get_user_operator_operator(request):
     try:
         user_operator = (
             UserOperator.objects.only("operator__status", "operator__id")
+            .exclude(status=UserOperator.Statuses.DECLINED)
             .select_related("operator")
             .get(user=user.user_guid)
+
         )
     except UserOperator.DoesNotExist:
         return 404, {"message": "User is not associated with any operator"}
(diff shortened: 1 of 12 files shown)


dleard:
Admin request declined screen:
MicrosoftTeams-image (2)

Subsequent request declined by admin screen:
MicrosoftTeams-image (3)

bc_obps/registration/tests/endpoints/test_operations_endpoints.py
```diff
@@ -214,6 +214,39 @@ def test_operations_endpoint_get_method_with_mock_data(self):
         response_data = response.json().get('data')
         assert len(response_data) == PAGE_SIZE
 
+    def test_operations_endpoint_get_ignores_declined_user_operators(self):
+        # IRC users can get all operations except ones with a not Started status
```
Sepehr-Sobhani, line 218: Leftover!

bc_obps/registration/api/user_operator.py
```diff
@@ -389,9 +416,7 @@ def update_operator_and_user_operator(request, payload: UserOperatorOperatorIn,
 @authorize(AppRole.get_all_authorized_app_roles(), ["admin"])
 def update_user_operator_status(request, payload: UserOperatorStatusUpdate):
     current_user: User = request.current_user  # irc user or industry user admin
-    if payload.user_guid:  # to update the status of a user_operator by user_guid
-        user_operator = get_object_or_404(UserOperator, user_id=payload.user_guid)
```
Sepehr-Sobhani, line 393: As we no longer need the `user_guid`, I believe we can remove it from the UserOperatorStatusUpdate schema and a couple of places in front-end.

Sepehr-Sobhani (approved):
Great stuff, Thank you! 🤩 I have pointed out a few nits, but overall, everything looks good to me. Additionally, I observed that when we change the status of an admin request, the page refreshes twice and the status does not update until we perform a hard refresh. Is this problem related to the current pull request?

bc_obps/registration/tests/endpoints/test_useroperator_endpoints.py
```diff
@@ -759,7 +800,7 @@ def test_create_operator_and_user_operator_with_parent_operators(self):
         }
 
         parent_operators: List[ParentOperator] = operator.parent_operators.all()
-        assert len(parent_operators) == 2
```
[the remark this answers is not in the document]
dleard: Unintentional, I'll put it back in

Issue #883, opened 2024-02-12
Internal users should not be able to add / remove parent operators


dleard:
If fixing this requires updating the `save_operator` function, there will be conflicts with #840. I suspect this is mainly a frontend issue though (auth protections should stop internal users from actually being able to edit anything) so conflict should be minimal

It is currently possible as an internal user to add or remove parent operators when viewing the operator admin request where they should not be able to mutate any data.

To reproduce:
- create an admin request as an industry user
- log in as an internal user & view the admin request
- click the Add parent operator button
- you are able to add data to the form (the internal user should not be able to do this)

PR #930, opened 2024-02-15
fix: hide parent operator button if form is disabled

marcellmueller:
Fix for #883

### DIFF
--- a/client/app/styles/rjsf/ArrayFieldTemplate.tsx
+++ b/client/app/styles/rjsf/ArrayFieldTemplate.tsx
@@ -22,6 +22,7 @@ const MinusSVG = () => (
 
 const ArrayFieldTemplate = ({
   canAdd,
+  disabled,
   items,
   onAddClick,
   uiSchema,
@@ -67,8 +68,9 @@ const ArrayFieldTemplate = ({
           </div>
         );
       })}
-      {canAdd && (
+      {canAdd && !disabled && (
         <Button
+          disabled={disabled}
           variant="contained"
           className="w-fit my-8 normal-case"
           onClick={onAddClick}
--- a/bc_obps/registration/fixtures/mock/parent_operator.json
+++ b/bc_obps/registration/fixtures/mock/parent_operator.json
@@ -1,0 +1,17 @@
+[
+  {
+    "model": "registration.parentoperator",
+    "pk": 1,
+    "fields": {
+      "child_operator": 3,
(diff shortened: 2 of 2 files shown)


client/app/styles/rjsf/ArrayFieldTemplate.tsx
```diff
@@ -67,8 +68,9 @@ const ArrayFieldTemplate = ({
           </div>
         );
       })}
-      {canAdd && (
+      {canAdd && !disabled && (
         <Button
+          disabled={disabled}
```
marcellmueller, line 73: This line is probably redundant but I added it just incase the button showed up somehow. Initially I had just disabled it using this though I though we should just hide it altogether if the form is disabled.

client/app/styles/rjsf/ArrayFieldTemplate.tsx
```diff
@@ -67,8 +68,9 @@ const ArrayFieldTemplate = ({
           </div>
         );
       })}
-      {canAdd && (
+      {canAdd && !disabled && (
```
marcellmueller, line 71: Should we only hide this when disabled for internal users or is it fine to hide for everyone.

BCerki: Unknown, @nanyangpro ?

client/app/styles/rjsf/ArrayFieldTemplate.tsx
BCerki: makes sense to me

BCerki (commented):
Nice work! I looked at the form as an IDIR user and couldn't see the button, and the button was still available as an internal user while editing. I don't know the answer to your question; I tagged Nan. I also committed a fixture to add a parent operator because that made manual testing easier; feel free to edit it if you want

client/app/styles/rjsf/ArrayFieldTemplate.tsx
BCerki: Talked with Nan, will hide (rather than disable) button for everyone

BCerki (approved):

nanyangpro:
Thanks for confirming, @BCerki!
- As discussed, in read-only views for the form, either disabling / graying out OR hiding the button is fine. Ideally we hide it.
- For future enhancement, we should consider having a generic look and feel and behaviour for forms when read-only.

PR #932, opened 2024-02-16
fix-internal-users-can-approve-existing-operators

BCerki:
card: https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=53448161


andrea-williams:
I don't think this PR is needed any more - it's a duplicate of #946

dleard:
@BCerki Yep! This is a duplicate. Closing.

PR #933, opened 2024-02-16
923 save and continue

BCerki:
card: https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=53303290
To test:
- to see the save and continue, go into an approved or not started operation
- to see the next, go into a pending operation

### DIFF
--- a/client/app/components/form/MultiStepButtons.tsx
+++ b/client/app/components/form/MultiStepButtons.tsx
@@ -32,7 +32,6 @@ const SubmitButton: React.FunctionComponent<SubmitButtonProps> = ({
   const { data: session } = useSession();
 
   const isIndustryUser = session?.user.app_role?.includes("industry");
-
   return (
     <div className={`flex w-full mt-8 justify-between ${classNames}`}>
       {cancelUrl && (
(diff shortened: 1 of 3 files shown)


client/app/components/form/MultiStepButtons.tsx
```diff
@@ -67,7 +66,7 @@ const SubmitButton: React.FunctionComponent<SubmitButtonProps> = ({
               type="button"
               disabled={isFinalStep || isSubmitting}
             >
-              Next
+              {isSubmitting ? "Save and Continue" : "Next"}
```
BCerki, line 69: necessary or else the button flashes between "Save And continue" and "Next" while submitting

marcellmueller (approved):
Works perfectly. I was a little confused about the fixtures updates but I assume you just wanted an approved Operation for `bc-cas-dev`?

BCerki:
> Works perfectly. I was a little confused about the fixtures updates but I assume you just wanted an approved Operation for `bc-cas-dev`?

Yes, I wanted to check that it looked right for operations with different statuses

Issue #668, opened 2024-01-23
Add missing URLs to landing page and update key dates


marcellmueller:
#### Describe the task

This is a follow up ticket to #163 since the URLs weren't ready on time.

Notes about the URLs are in the [Wireframe](https://www.figma.com/file/LsDmLDCdnJqI8UahyqvOD8/OBPS-App-Designs?type=design&node-id=2688-6375&mode=design&t=y7smsRJhRtzkRitv-4).

The Key Dates table should also be updated according to the latest design above since we postpone the prod release date.

#### Acceptance Criteria

- [x] BC OBPS Program Website URL
- [ ] Carbon Tax Exemption URL (please see comments below)
- [x] Detailed Guidance URL
- [x] Update Key Dates in the table

Dianadec:
Once we have the info, we can bring into the "Isolated Simple Tasks"

nanyangpro:
The Key Dates table is updated as advised by @Maralsotoudehnia. I've also included it in the AC.

We're still waiting for the URLs. Once we get them, I'll include them in the design and inform the developers for implementation.

nanyangpro:
Maral advised that "FIN's carbon tax exemption link won't be ready until late March, so I suggest removing it from the page for now."
Please remove the paragraph with the carbon tax exemption link for the MVP prod release.

nanyangpro:
We got the links, this ticket is unblocked now! :)

Issue #163, opened 2023-10-03
Update Landing Page Content


Dianadec:
#### Describe the task

This card is meant to contain the approved content from the business area of the text/info to go onto the landing page of the app.

Please refer to the [design](https://www.figma.com/file/LsDmLDCdnJqI8UahyqvOD8/OBPS-App-Designs?type=design&node-id=2688-6375&mode=design&t=y7smsRJhRtzkRitv-4) in Figma.

#### Acceptance Criteria

- [ ] Meta tag for description has text = TBD
- [ ] Content is responsive for mobile and desktop
- [ ] Jest unit test (with snapshot?)
- [ ] e2e tests

#### Additional context

- Blocked by #112 
- Blocked by #260

suhafa:
Is this ticket blocked by #260 as well where we will get clear description of the content from business area?

suhafa:
For now, we will have one landing page for both external and internal users. Later we will have a separate link

Dianadec:
@suhafa  do we have updated info that we could add to this card to refine it?

suhafa:
@Dianadec currently this is blocked by progress on ticket #260

hannavovk:
Dec 4 update: We have a reasonably close draft page, but it doesn't look finalized enough. Working with BA to finalize.

hannavovk:
@nanyangpro, would you mind updating this card with the current info. I think it's ready to go for devs.

nanyangpro:
Sure thing, thanks @hannavovk!

Done. Feel free to remove the "Backlog Refinement" label unless the developers want to add in more AC, thanks @Dianadec!

hannavovk:
Nice to see the super old placeholder landing page gone!
I see the Program Admin and Industrial Operator log in buttons are still in the banner - will they be removed later or should it be part of this ticket? 

![Image](https://github.com/bcgov/cas-registration/assets/87211104/2b458d54-95f2-425a-a3e1-c8a2b49d2a64)

Also, would it be terribly annoying if I asked for some modifications to improve the styling of the page? That would be both design and app updates. Just thinking of making a good first impression :). I don't mind being told otherwise though. 

cc @marcellmueller @nanyangpro @Dianadec

marcellmueller:
Sorry I just redid the landing page content and didn't touch the nav. Though it is super easy to remove the buttons. 

[...]

PR #934, opened 2024-02-16
Add missing URLs to landing page and update key dates

marcellmueller:
#668 - accidentally numbered this branch #163 which this is a follow up issue for.

There is a comment in #668 that says the exemption URL won't be ready for MVP so I've commented out that section. Also reduced the font size and padding in the `Key dates` section as that has been changed in Figma.


BCerki (approved):
:boat:

PR #937, opened 2024-02-17
Fix: 879 bcghgid opt ins

dleard:
Some decisions around the operations form were made while investigating this bug:

- Remove the "Did you submit a GHG Report for 2022"  section entirely. It served no purpose and was the source of the bug
- Add a readonly bcghgid field to always display the bcghgid (if they have it, otherwise shows a message that it will be assigned by CAS)
- Add additional context around opt-ins and a link to the site with more information and opt-in application forms

### DIFF
--- a/client/app/components/form/titles/operationsTitles.tsx
+++ b/client/app/components/form/titles/operationsTitles.tsx
@@ -8,15 +8,18 @@ export const PointOfContactTitle = (
 export const OptInOperationTitle = (
   <div className="max-w-2xl">
     Operations within regulated sectors that emit less than 10,000 tonnes CO2e
-    per year may voluntarily opt-in to the B.C. OBPS.<br/><br/>
-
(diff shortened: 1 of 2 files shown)


andrea-williams (commented):
Nit: "...by the Climate Ac" - remainder gets cut off and there's no way to scroll in the box to see the rest of the sentence.

Screenshot 2024-02-20 at 12 02 57 PM

dleard:
> Nit: "...by the Climate Ac" - remainder gets cut off and there's no way to scroll in the box to see the rest of the sentence.
> 
> Screenshot 2024-02-20 at 12 02 57 PM

I was aware of this (though I think I was able to scroll the disabled text). I thought we had a card somewhere to create a readonly template where for readonly fields/forms we just show plain text, but I can't find it right now. If we don't have a card for that we should make one.

andrea-williams (approved):

PR #939, opened 2024-02-20
936 change id to UUID

Sepehr-Sobhani:
[ISSUE 936](https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=53565073)
AND
[ISSUE 822](https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=52300972)
AND
An issue with the Admin Requests Table which was not showing up because of removing `id` attribute in recent changes.

NOTE:
- Had to remove `orjson` package as it doesn't support 128-bit integers so we can't get UUIDs to work.(https://github.com/ijl/orjson/issues/116#issuecomment-661397959)


client/app/(authenticated)/bceidbusiness/industry_user_admin/dashboard/users/page.tsx
```diff
@@ -66,6 +66,7 @@ export default async function Page() {
   ];
 
   const statusRows: GridRowsProp = userOperatorStatuses.map((uOS) => ({
+    id: uOS.id, // This unique ID is needed for DataGrid to work properly
```
dleard, line 69: :+1:
We should also remove lines 14-20 above, showing the guid in the table is not helpful

dleard (approved):
Nice! Looks good to me. Just one suggestion that can be tackled later.
I did some basic testing locally on the main workflow & didn't see any issues:
- Request operator access
- Approve operator access
- Update operator
- Request BORO ID

PR #940, opened 2024-02-20
fix: fix authorize definition causing incorrect declined message

dleard:
Follow up fix for the work done in #929. A last minute change in that PR to the @authorize check for the new /operator-access-declined endpoint caused an issue where nobody could access it, which was in turn causing the app to always direct a user to the access declined message.


andrea-williams (approved):
Tested locally - am able to request access to operator now.

PR #941, opened 2024-02-20
661 parent operator addresses

BCerki:
card: https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=51034863

This card involved creating a `handle_operator_addresses` helper function that's used for both operators and parent operators. Sepehr and I had trouble with circular references in the tests, which is why so many of the import statements changed.

For manual testing:
part 1:
- create a new operator and add both business and mailing addresses, and add a parent operator with business and mailing addresses. Check db--should have four address records
- edit all addresses of this operator. In the db, we should have the 4 edited records, not any extra ones

part 2: 
- create a new operator with only a business address. There should be one address in the `Address` table and the operator physical and mailing address ids should be the same
- edit the operator and add a mailing address. now there should be two address, and the operator address ids should be different


bc_obps/registration/api/utils/operator_utils.py
```diff
@@ -0,0 +1,106 @@
...
+        id=physical_address_id,
+        defaults={
+            "street_address": address_data.get(f'{prefix}physical_street_address'),
+            "municipality": address_data.get(f'{prefix}physical_municipality'),
+            "province": address_data.get(f'{prefix}physical_province'),
+            "postal_code": address_data.get(f'{prefix}physical_postal_code'),
+        },
+    )
+    if address_data.get(f'{prefix}mailing_address_same_as_physical'):
+        mailing_address = physical_address
```
dleard, line 44: There is still a possibility of orphaned addresses where they used to have both a physical and mailing address, but then change mailing_address_same_as_physical to True.

In this case, the id of the mailing address will be set to the id of the physical address, but the original mailing address record will remain orphaned in the database.

Should we add a condition here opposite to the condirion below where if the current mailing address id != physical address id then we prune the old mailing address? Should we archive or hard delete this mailing address?

dleard (approved):
Really nice work with the new util functions!
I tested this manually updating, adding, removing and re-adding address and it all works as it should.
I left one comment around a scenario where we can still have orphaned addresses, but we can decided what to do about that in a follow up ticket as this PR covers the AC of the original ticket.

bc_obps/registration/api/utils/operator_utils.py
BCerki: https://github.com/bcgov/cas-registration/issues/970

Issue #470, opened 2023-12-21
As an admin IO user, I want to assign a role when approving my users


BCerki:
#### Description:

Admin IO users should be able to choose what role to assign their company's users.

Wireframe: https://www.figma.com/file/LsDmLDCdnJqI8UahyqvOD8/OBPS-App-Designs?type=design&node-id=2719-15135&mode=design&t=RTQrkxl7Upj6pRkn-0

#### Acceptance Criteria:

Given I am an admin IO user
When I go the the users tile and see the list of my company's users
Then I can approve users as either admins (same permissions as me) or reporters (can't see the user tile)

#### Development Checklist:

- [x] Add a dropdown to the users table so admins can select the `reporter` or `admin` role
- [x] When the IO user clicks approve:
- - [ ]  <strike>the approved user should get the appropriate app_role in the user table (admin=`'industry_user_admin`, reporter=`industry_user`)</strike> -
- - [x] the user_operator record for that user should have status=Statuses.APPROVED
- [ ] ~~IF we aren't using user_operator.role anywhere, we can delete this column (to avoid role and permission confusion). When determining which user role to show in the table, industry_user=reporter and industry_user_admin=admin~~
- [ ] Meets the DOD

**Definition of Ready** (Note: If any of these points are not applicable, mark N/A)

- [ ] User story is included
- [ ] User role and type are identified
- [ ] Acceptance criteria are included
- [ ] Wireframes are included (if required)
- [ ] Design / Solution is accepted by Product Owner
- [ ] Dependencies are identified (technical, business, regulatory/policy)
- [ ] Story has been estimated (under 13 pts)

·**Definition of Done** (Note: If any of these points are not applicable, mark N/A)

- [ ] Acceptance criteria are tested by the CI pipeline
- [ ] UI meets accessibility requirements
- [ ] Configuration changes are documented, documentation and designs are updated
- [ ] Passes code peer-review
- [ ] Passes QA of Acceptance Criteria with verification in Dev and Test
- [ ] Ticket is ready to be merged to main branch
- [ ] Can be demoed in Sprint Review
- [ ] Bugs or future work cards are identified and created
- [ ] Reviewed and approved by Product Owner

#### Notes:

-

Dianadec:
[...]

PR #958, opened 2024-02-22
Fix: 954 update draft operator flow

dleard:
This PR fixes some access control in the API & fixes the redirects under certain conditions:

- API: Only external admins can access the PUT endpoint for updating an operator
- Client: If operator.status = DRAFT && operator.is_new = false then the user is redirected to the access received page instead of the edit operator form

### DIFF
--- a/bc_obps/registration/api/user_operator.py
+++ b/bc_obps/registration/api/user_operator.py
@@ -335,7 +335,7 @@ def create_operator_and_user_operator(request, payload: UserOperatorOperatorIn):
     response={200: RequestAccessOut, codes_4xx: Message},
     url_name="update_operator_and_user_operator",
 )
-@authorize(["industry_user"], UserOperator.get_all_industry_user_operator_roles())
+@authorize(["industry_user"], ["admin"])
 def update_operator_and_user_operator(request, payload: UserOperatorOperatorIn, user_operator_id: UUID):
     user: User = request.current_user
     try:
--- a/bc_obps/registration/tests/endpoints/test_useroperator_endpoints.py
+++ b/bc_obps/registration/tests/endpoints/test_useroperator_endpoints.py
@@ -1100,27 +1100,22 @@ class TestUserOperatorEndpoint(CommonTestSetup):
         operator = operator_baker()
         operator.created_by = self.user
         operator.save(update_fields=["created_by"])
+        mock_payload = {
+            "legal_name": "Unauthorized",
+            "trade_name": "Unauthorized",
+            "cra_business_number": 678123654,
+            "bc_corporate_registry_number": "jkl1234321",
+            "business_structure": BusinessStructure.objects.first().pk,
+            "physical_street_address": "add",
+            "physical_municipality": "add",
+            "physical_province": "BC",
+            "physical_postal_code": "H0H0H0",
+            "mailing_address_same_as_physical": True,
+            "operator_has_parent_operators": False,
+        }
         user_operator = baker.make(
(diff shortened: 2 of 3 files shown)


bc_obps/registration/api/user_operator.py
```diff
@@ -390,7 +390,7 @@ def create_operator_and_user_operator(request, payload: UserOperatorOperatorIn):
     response={200: RequestAccessOut, codes_4xx: Message},
     url_name="update_operator_and_user_operator",
 )
-@authorize(["industry_user"], UserOperator.get_all_industry_user_operator_roles())
+@authorize(["industry_user"], ["admin"])
```
BCerki, line 393: As far as I remember, the first user who requests access gets an `admin` role, and everyone else who follows gets `pending`. We didn't build role assignment into MVP, so I think this change might block everyone except the first requester from accessing this endpoint. Approving a user_operator sets their status to `Approved` but doesn't change their role.

bc_obps/registration/tests/endpoints/test_useroperator_endpoints.py
```diff
@@ -1096,6 +1098,36 @@ def test_put_user_operator_operator(self):
...
+            "business_structure": BusinessStructure.objects.first().pk,
+            "physical_street_address": "add",
+            "physical_municipality": "add",
+            "physical_province": "BC",
+            "physical_postal_code": "H0H0H0",
+            "mailing_address_same_as_physical": True,
+            "operator_has_parent_operators": False,
+        }
+        user_operator = baker.make(
+            UserOperator, user=self.user, operator=operator, role=UserOperator.Roles.REPORTER, created_by=self.user
```
BCerki, line 1119: A `pending` role might be a better test case? I don't think we'll ever have anyone with a reporter role for MVP.

BCerki (commented):
Looking good! I tested that I could/couldn't GET/PUT the form with different Draft/is_new values, and I checked that an internal user can still see and approve user_operators.

I've got a couple inline comments and a couple general ones:
- This AC: "An external user can update data for a 'Draft' operator where 'is_new'=true (creating new operator)"--how would anyone end up in this state? As soon as they create the operator, the status changes to `Pending` (if I go mess with the database though, I can confirm this AC)
- If I bypass the tiles and go directly to user_operator URL (e.g. http://localhost:3000/dashboard/select-operator/user-operator/616e3872-85c9-4101-99dc-32c2ed066d2f/1), I can GET the form when the operator is Draft and not new: 
(can't PUT though)

bc_obps/registration/api/user_operator.py
dleard: This will be handled by #470 . I've moved up its priority in our backlog

bc_obps/registration/tests/endpoints/test_useroperator_endpoints.py
dleard: Good call, I'll test for both.

dleard:
Thanks for your review!
For your general comments:
1. That looks like it's not doing anything (an unreachable path) I'll remove the useless condition.
2. Good catch & I think I'll try hardening the auth parameters for that endpoint. At first glance, it doesn't look like restricting that GET to admin only will block any other necessary workflows

BCerki (approved):
Looks great! :cartwheeling: While testing, I found what I think is an unrelated bug because I see it on `develop` too: https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=54531016

PR #960, opened 2024-02-22
764 update cra business number and bc corporate registry number when updating operators

Sepehr-Sobhani:
[ISSUE 764](https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=51509212)
AND an issue with not showing API errors on the operation form (No card)

### DIFF
--- a/bc_obps/registration/api/user_operator.py
+++ b/bc_obps/registration/api/user_operator.py
@@ -69,7 +69,9 @@ def get_user_operator_from_user(request):
 )
 @authorize(["industry_user"], UserOperator.get_all_industry_user_operator_roles())
 def is_approved_admin_user_operator(request, user_guid: str):
-    approved_user_operator: bool = UserOperator.objects.filter(user_id=user_guid, role=UserOperator.Roles.ADMIN, status=UserOperator.Statuses.APPROVED).exists()
+    approved_user_operator: bool = UserOperator.objects.filter(
+        user_id=user_guid, role=UserOperator.Roles.ADMIN, status=UserOperator.Statuses.APPROVED
+    ).exists()
 
     return 200, {"approved": approved_user_operator}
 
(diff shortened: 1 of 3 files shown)


bc_obps/registration/tests/endpoints/test_useroperator_endpoints.py
```diff
@@ -1095,6 +1095,62 @@ def test_put_user_operator_operator(self):
...
+            "mailing_postal_code": operator.mailing_address.postal_code,
+            "operator_has_parent_operators": operator.parent_operators.exists(),
+            "parent_operators_array": list(operator.parent_operators.all()),
+        }
+
+    def test_put_user_operator_operator_with_an_existing_cra_business_number(self):
+        existing_operator = operator_baker()
+        new_operator = operator_baker()
+        new_operator.created_by = self.user
+        new_operator.save(update_fields=["created_by"])
```
BCerki, line 1121: The `operator_baker` function now takes a `custom_properties` arg, so instead of lines 1120-1121, you could use that, e.g. something like new_operator = operator_baker({created_by: self.user})

bc_obps/registration/api/user_operator.py
```diff
@@ -78,11 +78,13 @@ def save_operator(payload: UserOperatorOperatorIn, operator_instance: Operator,
             "mailing_address_id",
             "website",
             "business_structure",
+            "cra_business_number",
+            "bc_corporate_registry_number",
```
BCerki, line 82: If we updated all fields, we wouldn't have to remember to update this list every time we add a field (I can see this bug happening again). Was `operator_related_fields` for optimization?

BCerki (approved):
Works great! I made a couple suggestions but nothing blocking. If you'd like another look after rebasing, just let me know.

bc_obps/registration/api/user_operator.py
Sepehr-Sobhani: Yes, I want to pass this list to 'update_fields' to optimize the save method. However, I also want to prevent unintended field updates in the 'save_operator' function if we pass different data using the payload.

PR #965, opened 2024-02-22
Fix: changed mock child_operator from id to uuid

andrea-williams:


BCerki (approved):
thank you for that branch name

PR #967, opened 2024-02-22
chore: set cra # to 9 digits in db

BCerki:
card: https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=54181386

To manually test:
- front-end: try to enter a number that has less than 9 digits (field cuts you off at 9 so can't enter more) in the cra fields for both operator and parent operator--you should get an rjsf error
- back-end/db: comment out the max/min lines in the rjsf schema and subject the form with incorrect cra #s. should get a 422, or try to insert a bad value directly into db

Sonarcloud fail is a random number generator for testing, so I'll ignore that unless the reviewer feels otherwise


bc_obps/registration/schema/user_operator.py
```diff
@@ -123,6 +124,11 @@ class UserOperatorOperatorIn(Schema):
     def validate_business_structure(cls, value: str) -> BusinessStructure:
         return validate_business_structure(value)
 
+    @validator("cra_business_number")
+    @classmethod
+    def validate_cra_business_number(cls, value: str):
```
Sepehr-Sobhani, line 129: Should the value type be `int`?

bc_obps/registration/schema/validators.py
```diff
@@ -0,0 +1,7 @@
+def validate_cra_business_number(value: str):
```
Sepehr-Sobhani, line 1: Should the value type be `int`?

bc_obps/registration/schema/parent_operator.py
```diff
@@ -36,6 +37,11 @@ class ParentOperatorIn(Schema):
     def validate_po_business_structure(cls, value: str) -> BusinessStructure:
         return validate_business_structure(value)
 
+    @validator("po_cra_business_number")
+    @classmethod
+    def validate_cra_business_number(cls, value: str):
```
Sepehr-Sobhani, line 42: Should the value type be `int`?

Sepehr-Sobhani (approved):
Just commented a nit otherwise looks great! 🤩 @dleard Should we verify if all CRA numbers in prod data conform to this format?

PR #979, opened 2024-02-26
901 operation status

BCerki:
card: https://github.com/orgs/bcgov/projects/122/views/2?filterQuery=not+started&pane=issue&itemId=53169328

Check out the comments in the issue for more details/wireframe--this wasn't a bug so much as something we never implemented


client/app/components/datagrid/OperationDataGrid.tsx
shon-button, line 1: Possibly use `Status` enum for the switch case values.

bc_obps/registration/fixtures/mock/operation.json
```diff
@@ -67,17 +67,17 @@
   {
     "model": "registration.operation",
     "fields": {
-      "operator": "685d581b-5698-411f-ae00-de1d97334a71",
+      "operator": "4242ea9d-b917-4129-93c2-db00b7451051",
```
BCerki, line 70: NEW since last review, added a Draft one to make this easier to test

PR #981, opened 2024-02-27
900 user profile name

BCerki:
https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=53168930

to manually test:
- login as a bceid user
- change your first and last name in the profile page
- your name should update in the header
- go to the users table and you should the full updated name in the name column (previously it was getting cut off at last initial)


client/app/(authenticated)/bceidbusiness/industry_user_admin/dashboard/users/page.tsx
```diff
@@ -67,7 +67,7 @@ export default async function Page() {
 
   const statusRows: GridRowsProp = userOperatorStatuses.map((uOS) => ({
     id: uOS.id, // This unique ID is needed for DataGrid to work properly
-    name: `${uOS.user.first_name} ${uOS.user.last_name.slice(0, 1)}`,
+    name: `${uOS.user.first_name} ${uOS.user.last_name}`,
```
BCerki, line 70: confirming with Nan in figma that we want to show the user's full name

client/app/components/navigation/Profile.tsx
```diff
@@ -1,32 +1,16 @@
...
-  const name = getUserFullName(session);
-  // 👇️ run function whenever the session object changes e.g. session.error changes
-  useEffect(() => {
-    if (session?.error === "ErrorAccessToken") {
-      // Keycloak has expired; so, call next-auth Signout
-      signOut();
-    }
-  }, [session]);
-
+export default function Profile({ userFullName }: { userFullName: string }) {
```
BCerki, line 5: This is a client component, so we pass the name from a server component higher in the tree (rather than using fetch in useEffect)

BCerki: This ticket is about using the name someone entered on the user profile page instead of using the name associated with their IDIR/BCEID. So I don't think we can use the session provider if we want to use the name from the db?

bc_obps/registration/middleware/current_user_middleware.py
```diff
@@ -13,7 +14,7 @@ def __call__(self, request):
         auth_header = request.headers.get('Authorization')
         if auth_header:
             try:
-                user_guid = json.loads(auth_header).get('user_guid')
+                user_guid = UUID(json.loads(auth_header).get('user_guid'))
```
BCerki, line 17: New! From Sepehr: We are receiving user_guid from the front-end as a UUID without hyphens. We store it as the cache key. However, when we update the user, we attempt to invalidate their cache using the user_guid with hyphens, causing a mismatch and preventing successful invalidation.

client/app/components/navigation/Profile.tsx
shon-button: Yes, I understand...the suggested changes would allow for "using the name someone entered on the user profile page instead of using the name associated with their IDIR/BCEID" with the added benefit of encapsulation of session data within the auth session provider.
Shall we discuss?

shon-button (approved):

PR #982, opened 2024-02-27
976 authorize decorator

BCerki:
card: https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=54526586

This PR prevents unapproved industry_users from accessing endpoints by default. If an endpoint should be accessible to unapproved users, we can add an argument to the @authorize decorator. Therefore, manual testing for this PR is walking through all the industry user workflows to make sure people can still access everything they need to. I went through the workflows before posting the PR, but the reviewer should probably take a look too.

Sonarcloud fail is on duplicated code in tests

### DIFF
--- a/bc_obps/registration/api/operator.py
+++ b/bc_obps/registration/api/operator.py
@@ -46,6 +46,7 @@ def get_operators_by_cra_number_or_legal_name(
     return 404, {"message": "No matching operator found. Retry or add operator."}
 
 
+# We have to let unapproved users to reach this endpoint otherwise they can't see operator info when they select it
 @router.get(
     "/operators/{operator_id}", response={200: ConfirmSelectedOperatorOut, codes_4xx: Message}, url_name="get_operator"
 )
--- a/bc_obps/registration/api/operator.py
+++ b/bc_obps/registration/api/operator.py
@@ -22,7 +22,7 @@ from django.core.exceptions import ValidationError
     response={200: Union[List[OperatorSearchOut], OperatorSearchOut], codes_4xx: Message, codes_5xx: Message},
     url_name="get_operators_by_cra_number_or_legal_name",
 )
-@authorize(AppRole.get_all_authorized_app_roles(), UserOperator.get_all_industry_user_operator_roles())
+@authorize(AppRole.get_all_authorized_app_roles(), UserOperator.get_all_industry_user_operator_roles(), False)
 def get_operators_by_cra_number_or_legal_name(
     request, cra_business_number: Optional[int] = None, legal_name: Optional[str] = ""
 ):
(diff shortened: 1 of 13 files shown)


bc_obps/registration/api/user_operator.py
```diff
@@ -48,7 +48,8 @@
     response={200: PendingUserOperatorOut, codes_4xx: Message},
     url_name="get_user_operator_from_user",
 )
-@authorize(["industry_user"], UserOperator.get_all_industry_user_operator_roles())
+# brianna I think it's fine for them to see this unapproved
```
BCerki, line 51: Thoughts?

bc_obps/registration/api/user_operator.py
```diff
@@ -97,7 +98,8 @@ def get_user_operator_operator(request):
 
 
 @router.get("/user-operator-id", response={200: UserOperatorIdOut, codes_4xx: Message}, url_name="get_user_operator_id")
-@authorize(["industry_user"], UserOperator.get_all_industry_user_operator_roles())
+# brianna I think it's fine for them to see this
```
BCerki, line 101: thoughts

bc_obps/registration/api/user_operator.py
```diff
@@ -293,6 +296,7 @@ def request_access(request, payload: SelectOperatorIn):
             if status != 200:
                 return status, message
 
+            # brianna how could one exist if this is an access request?
             # Making a draft UserOperator instance if one doesn't exist
```
BCerki, line 300: Not really related to the PR, but this endpoint is for requesting access. Users can't have have more than one request, so how could a previously existing UserOperator instance come about?

bc_obps/registration/api/operator.py
```diff
@@ -46,6 +46,7 @@ def get_operator_by_legal_name_or_cra(
     response={200: List[OperatorOut], codes_4xx: Message, codes_5xx: Message},
     url_name="get_operators_by_legal_name",
 )
+@authorize(AppRole.get_all_app_roles(), UserOperator.get_all_industry_user_operator_roles(), False)
```
BCerki, line 49: Sepehr's got a PR that narrows down the return type to only the info a non-approved user can see

bc_obps/registration/api/user_operator.py
dleard: It is possible for a user to have multiple records in the user_operator table in the scenario where:
They are declined access to an operator & then they request access to a different operator

bc_obps/registration/api/operator.py
Sepehr-Sobhani: I think we have to let this endpoint open to unapproved users otherwise they can't search for operators to request access!

bc_obps/registration/api/user_operator.py
Sepehr-Sobhani: I think it's fine too!

dleard: Agreed. This is their own data & contains nothing they couldn't see while creating a request

bc_obps/registration/api/user_operator.py
dleard: also agree

Sepehr-Sobhani: Where do we use this endpoint? :D

Sepehr-Sobhani (commented):
looks great! 🤩  I made `def get_operator(request, operator_id: UUID):` endpoint available to unapproved users because they need to see the operation info when they select one. Additionally, I removed a couple of comments and made a quick refactor.

bc_obps/registration/api/user_operator.py
BCerki: It used to be in the `select-operator` page but it must have been cut at some point

Sepehr-Sobhani (approved):
🤩

PR #989, opened 2024-02-28
985 Fix BCGHG ID validation error

Sepehr-Sobhani:
[ISSUE 985](https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=54763048)

### DIFF
--- a/bc_obps/registration/schema/operation.py
+++ b/bc_obps/registration/schema/operation.py
@@ -86,6 +86,7 @@ class OperationOut(ModelSchema):
     # multiple_operators_array: Optional["List[MultipleOperatorOut]"] = Field(None, alias="multiple_operator")
     statutory_declaration: Optional[str] = None
     bc_obps_regulated_operation: Optional[str] = Field(None, alias="bc_obps_regulated_operation.id")
+    bcghg_id: Optional[str] = ""
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


bc_obps/registration/schema/operation.py
```diff
@@ -94,6 +95,10 @@ def resolve_statutory_declaration(obj: Operation):
             return file_to_data_url(statutory_declaration)
         return None
 
+    @staticmethod
+    def resolve_bcghg_id(obj: Operation):
+        return obj.bcghg_id or ""
```
BCerki, line 100: Do we need both this and line 89? I thought 89 on its own would do it

bc_obps/registration/schema/operation.py
```diff
@@ -86,6 +86,7 @@ class OperationOut(ModelSchema):
     # multiple_operators_array: Optional["List[MultipleOperatorOut]"] = Field(None, alias="multiple_operator")
     statutory_declaration: Optional[str] = None
     bc_obps_regulated_operation: Optional[str] = Field(None, alias="bc_obps_regulated_operation.id")
+    bcghg_id: Optional[str] = ""
```
BCerki, line 89: Why an empty string instead of None?

BCerki (approved):
:cartwheeling:

PR #991, opened 2024-02-29
880 utc

BCerki:
card: https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=53041800

This is tricky one to test because the problem only shows up in `dev`. On local, at least for me in my timezone, everything already shows up in PST, but in `dev` it's UTC. I ended up testing with CI since it also seems to use UTC like `dev`. I temporarily put a date on the homepage and looked at the playright report before making changes:
image
After the fix, the playwright report shows the time in PST, so I think/hope `dev` will react the same way.
image

### DIFF
--- a/client/app/components/routes/operations/Operations.tsx
+++ b/client/app/components/routes/operations/Operations.tsx
@@ -20,14 +20,14 @@ async function getOperations() {
     throw error;
   }
 }
-const formatTimestamp = (timestamp: string) => {
+export const formatTimestamp = (timestamp: any) => {
   if (!timestamp) return undefined;
 
   const date = new Date(timestamp).toLocaleString("en-CA", {
     month: "short",
     day: "numeric",
     year: "numeric",
-    timeZone: "America/Vancouver",
+    // timeZone: "America/Vancouver",
   });
 
   const timeWithTimeZone = new Date(timestamp).toLocaleString("en-CA", {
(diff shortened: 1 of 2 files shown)


client/app/components/routes/operations/Operations.tsx
```diff
@@ -20,21 +20,22 @@ async function getOperations() {
...
 }
-
 const formatTimestamp = (timestamp: string) => {
   if (!timestamp) return undefined;
 
   const date = new Date(timestamp).toLocaleString("en-CA", {
     month: "short",
     day: "numeric",
     year: "numeric",
+    // timeZone: "America/Vancouver",
```
marcellmueller, line 30: Hey is this a mistake? The only changes are some added commented code.

BCerki: Sorry, something went funky with the revert/rebase. Fixed now

marcellmueller (approved):
Nice work, I checked this locally and it is still displaying the PST.

I think I remember some related timezone issues on CCBC. Our local computers will have whatever our local timezone is but Kubernetes defaults to UTC so specifying here is safer and should work correctly.

BCerki:
> Nice work, I checked this locally and it is still displaying the PST.
> 
> I think I remember some related timezone issues on CCBC. Our local computers will have whatever our local timezone is but Kubernetes defaults to UTC so specifying here is safer and should work correctly.

We had issues in CIF too. I tried to change my local timezone (my computer timezone, my browser timezone, my postgres.conf timezone) and couldn't get it out of PST even before I made the change. So hoping this works :crossed_fingers:

PR #999, opened 2024-02-29
Limit user profile put endpoint schema

Sepehr-Sobhani:


dleard (approved):
Nice work. I tested it and when I update my user and introspect the payload, only the necessary values are included :+1:

Sepehr-Sobhani:
> Nice work. I tested it and when I update my user and introspect the payload, only the necessary values are included 👍

@dleard lol Thanks for approving this before even posting it 🙏

PR #1009, opened 2024-03-01
fix: store the state to prevent loosing it when switching tabs

Sepehr-Sobhani:
[ISSUE 986](https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=54765288)

### DIFF
--- a/client/app/components/form/widgets/FileWidget.tsx
+++ b/client/app/components/form/widgets/FileWidget.tsx
@@ -150,8 +150,12 @@ const FileWidget = ({
   options,
   registry,
 }: WidgetProps) => {
+  // We need to store the value in state to prevent loosing the value when user switches between tabs
+  const [localValue, setLocalValue] = useState(value);
   const [filesInfo, setFilesInfo] = useState<FileInfoType[]>(
-    Array.isArray(value) ? extractFileInfo(value) : extractFileInfo([value]),
+    Array.isArray(localValue)
+      ? extractFileInfo(localValue)
+      : extractFileInfo([localValue]),
   );
   const { data: session } = useSession();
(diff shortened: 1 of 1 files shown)


shon-button:
Test steps:
1. "Create Operation"
2.  Upload an attachment
3. Click the (Download [template]) 
4. **Confirmed** attachment field retains text of the atachment
5. Click 'Submit`
6. Validation error on attachment field
image

shon-button:
@Sepehr-Sobhani 

Perhaps a harder task to accomplish but, wanted to note that the attachment field text is cleared on `Back` Navigation

Sepehr-Sobhani:
> @Sepehr-Sobhani
> 
> Perhaps a harder task to accomplish but, wanted to note that the attachment field text is cleared on `Back` Navigation

@shon-button Thanks for reviewing this pull request! I think the back navigation is not working intentionally because we handle the attachment upload to our buckets in the backend once the user hits the submit button. Could you please correct me if I'm wrong, @BCerki?

BCerki:
> > @Sepehr-Sobhani
> > Perhaps a harder task to accomplish but, wanted to note that the attachment field text is cleared on `Back` Navigation
> 
> @shon-button Thanks for reviewing this pull request! I think the back navigation is not working intentionally because we handle the attachment upload to our buckets in the backend once the user hits the submit button. Could you please correct me if I'm wrong, @BCerki?

Yes, the attachment only saves once the user clicks submit. The rest of the form works that way too. If you fill out page 2 and click "Back", the form will wipe. It only saves if you click "Save and Continue" before clicking "Back"

shon-button (approved):
Nice ✨

PR #1011, opened 2024-03-01
chore: add sentry dsn config for backend

dleard:
Backend was missing DSN information necessary to report to sentry


Sepehr-Sobhani (approved):
🙏

PR #1021, opened 2024-03-04
470 assign role

marcellmueller:
Implements #470 

Thanks for taking a look, take a good look at the api changes for me. They are fairly simple but I could use another look 🙇


client/app/components/datagrid/UserOperatorDataGrid.tsx
```diff
@@ -0,0 +1,93 @@
+"use client";
```
marcellmueller, line 1: Moving this to it's own file was unnecessary but happened during my debugging that issue where the User Operator Datagrid would refresh every time you did anything with it. I kept this anyways since it will be necessary to have a client component to pass in actions once we do the server side pagination ticket for this table.

client/app/components/datagrid/cells/ChangeUserOperatorStatusColumnCell.tsx
```diff
@@ -50,12 +39,13 @@ const handleUpdateStatus = async (
   }
 };
 
-export async function ChangeUserOperatorStatusColumnCell(
-  params: Readonly<ButtonRenderCellParams>,
-) {
```
marcellmueller, line 55: It's been a few days but if you were curious I believe that it was this component being wrapped in an `async` function that was causing the User Operator Datagrid to rerender every time we did anything. It took me a long time to figure out 😅

andrea-williams:
@marcellmueller  when you run `make pythontests` on your local from the `bc_obps` dir, what results do you get? I'm getting one test failure on registration/tests/endpoints/test_user_endpoints.py::TestUserEndpoint::test_create_user_profile_idir with output

```
>       assert 'app_role' in content and content["app_role"]["role_name"] == "cas_pending"
E       AssertionError: assert ('app_role' in {'app_role': {'role_name': 'cas_admin'}, 'bceid_business_name': 'bcgov test', 'email': 'idir.user@email.com', 'first_name': 'Idir', ...} and 'cas_admin' == 'cas_pending'
E         - cas_pending
E         + cas_admin)
```

but I'm v confused why it isn't failing in the GH action

marcellmueller:
> @marcellmueller when you run `make pythontests` on your local from the `bc_obps` dir, what results do you get? I'm getting one test failure on registration/tests/endpoints/test_user_endpoints.py::TestUserEndpoint::test_create_user_profile_idir with output
> 
> ```
> >       assert 'app_role' in content and content["app_role"]["role_name"] == "cas_pending"
> E       AssertionError: assert ('app_role' in {'app_role': {'role_name': 'cas_admin'}, 'bceid_business_name': 'bcgov test', 'email': 'idir.user@email.com', 'first_name': 'Idir', ...} and 'cas_admin' == 'cas_pending'
> E         - cas_pending
> E         + cas_admin)
> ```
> 
> but I'm v confused why it isn't failing in the GH action

If you disable `BYPASS_ROLE_ASSIGNMENT=True` in your backend .env it should pass :)

bc_obps/registration/api/user_operator.py
```diff
@@ -269,7 +269,7 @@ def request_admin_access(request, payload: SelectOperatorIn):
                 return status, message
         # Making a pending UserOperator instance if one doesn't exist
         user_operator, created = UserOperator.objects.get_or_create(
-            user=user, operator=operator, role=UserOperator.Roles.ADMIN, status=UserOperator.Statuses.PENDING
+            user=user, operator=operator, role=UserOperator.Roles.PENDING, status=UserOperator.Statuses.PENDING
```
andrea-williams, line 272: why did you change this one to `PENDING`? I would have expected it to remain as `ADMIN` since the endpoint is only used for external users requesting prime admin access to their operator

marcellmueller: I changed it to be safe just incase there was some mistake and someone only did a check on a user operators role. By default all roles should set to pending just like status and then set to `admin` or `reporter` on approval.

client/app/components/routes/access-requests/form/UserOperatorReview.tsx
```diff
@@ -50,7 +55,11 @@ export default function UserOperatorReview({
         "PUT",
         "",
         {
-          body: JSON.stringify({ status, user_operator_id: userOperatorId }),
+          body: JSON.stringify({
+            role: UserOperatorRoles.ADMIN,
```
marcellmueller, line 59: @andrea-williams here we set the role from the default `pending` to `admin` when they approve the prime admin request.

andrea-williams:
I spent a very long time wrapping my head around this. At the moment I have some reservations because it feels like the `status` and `role` columns are starting to basically be the same thing, but I'm approving because all the AC are met, and when I was running it locally I aggressively tried to poke holes in the logic and couldn't.

Nicely done Marcel!

andrea-williams (approved):

PR #1042, opened 2024-03-07
chore: add reporting app and model in django

BallardRobinett:
Issue: https://github.com/bcgov/cas-reporting/issues/124

I added a new reporting Django app with ```python manage.py startapp reporting``` and added a basic Report class to models.py. 

I added reporting to the list of INSTALLED_APPS in settings.py because I got an error complaining about that when I launched the Django shell.

I modified the reporting apps.py file to match registration because the card AC said the report table should be accessible from the ```erc``` schema, however I'm not sure if ```erc``` will be a shared schema between registration and reporting going forward or if some other structure will be used.

Tagging moose devs to sanity check if this is a sensible way to make changes in django. Thanks everyone!


bc_obps/reporting/migrations/0001_initial.py
Sepehr-Sobhani, line 1: I don't see this data model in the `models.py` file. probably a leftover?

bc_obps/reporting/migrations/0003_profile.py
Sepehr-Sobhani, line 1: I don't see this data model in the `models.py` file. probably a leftover?

bc_obps/reporting/views.py
Sepehr-Sobhani, line 1: We are not using Django views since we are utilizing a REST API. This means we can remove this file.

bc_obps/reporting/models.py
```diff
@@ -0,0 +1,12 @@
+from django.db import models
+
+class Report(models.Model):
+    title = models.CharField(max_length=100)
```
Sepehr-Sobhani, line 4: It is recommended to include a `db_comment` for each field in order to provide a clear description of its main purpose. For example: `db_comment="The physical address of an operator (where the operator is physically located)"`.

Additionally, metadata needs to be added to each data model. This could be something like:
```python
class Meta:
        db_table_comment = "Naics codes"
        db_table = 'erc"."naics_code'
```
The `db_table` specifies the schema to which the table is associated.

Sepehr-Sobhani (commented):
Awesome job! Left some comments.

bc_obps/reporting/migrations/0003_profile.py
BallardRobinett: I'm not sure I quite understand the migrations files. They get generated when I use the ```python manage.py makemigrations reporting``` command based on the changes I've made in the reporting directory, but is it correct to remove old ones if the changes are overwritten in a future migration? I left them all in because I was unsure.

BallardRobinett: I did originally add a class in models.py called profile, but I changed in to report because of the AC of the card.

Sepehr-Sobhani: That's correct. As your migration files have not been applied to the PROD database yet, you can delete all migration files in the reporting app and create new ones once you've completed your table. This will help avoid generating redundant migration files, and it's always better to maintain less code :)

bc_obps/reporting/models.py
```diff
@@ -0,0 +1,14 @@
+from django.db import models
+
+class Report(models.Model):
+    title = models.CharField(max_length=100, db_comment="The title of the report")
+    description = models.TextField(db_comment="The description of the report")
+    created_at = models.DateTimeField(auto_now_add=True, db_comment="The timestamp when the report was created")
+
+    class Meta:
+        db_table_comment = "A table to store reports"
+        db_table = "erc.report"
```
Sepehr-Sobhani, line 10: This line should be revised to `'erc"."report'` in order to correctly define the schema for this table. To test if it was added to the schema correctly, you can use `psql registration` and run `\d erc.report`.

PR #1057, opened 2024-03-08
feat: add Nx project reporting app and shared_components placeholders

mikevespi:
Add an Nx project to the repo. It will initially hold the reporting app and shared libraries. The placeholders are barebones and just intended to give us a starting point with working examples.

When creating an Nx project, you enter what would have been the parent directory. I called it bciers, and for now have kept it as it created a nice sandbox for us rather than poluting the root dir. We can consider removing it once we have all of our configuration worked out.

To run the "reporting app" cd into the `bciers` directory and run `yarn` to install the packages, then `nx dev reporting`.


.gitignore
```diff
@@ -16,7 +16,6 @@ downloads/
 eggs/
 .eggs/
 k6_results
-lib/
```
mikevespi, line 19: Since `lib` is one of the default folder names created when generating packages/components with Nx I've removed it from the gitignore. This line seems to have come from a template .gitignore, and removing it didn't cause any new files to be discovered by git.

joshgamache (approved):
Tooling all works and all the base infra makes sense! Great to see this get started 👍

pbastia (changes requested):

pbastia (approved):

PR #1066, opened 2024-03-11
Chore: mailhog

dleard:
Adds mailhog to the dev & test environments to ensure test emails are not actually delivered outside of our environment.

This implementation was pulled mostly from what we're doing in CIIP. It may need tweaking once we have CHES set up if there are any differences in that setup from what we have in CIIP.


marcellmueller (approved):
🥇

Issue #971, opened 2024-02-23
Forced logout after a period of inactivity


shon-button:
**Background:**
Keycloak authentication server provides a flow for issuing an access token, the expiration time of the access token, and a refresh token to be used to request a new access token once that token has expired,`access_token`; `expires_at`,`refresh_token`, which can be used to determine access to the protected resources needed by our application users.

Previously we had implemented a [refresh token strategy](https://next-auth.js.org/v3/tutorials/refresh-token-rotation) from Keycloak but, decision to remove it as an implementation bug was causing a forced logout after 30 minutes and a resolution of the bug would essential just keep refreshing the token without reusing the token's refreshed information.
Currently, we use the token from a successful client authentication with the Keycloak authentication server in `client/middlewares/withAuthorization.tsx` to access user information `app_role` for front end protection and  `user_guid` and 
 for API protection but, **we do not use** `access_token`; `expires_at`,`refresh_token`.

As such, we need to discuss using or not using of the Keycloak `access_token`; `expires_at`,`refresh_token` for user access to the protected resources:

- If we do not use the `access_token` for protecting resources do we need to implement a refresh token strategy?
- Should we implement a forced session logout after some x interval?

**Forced Session Logout**

**Background:**
A session is initiated when a user logs in or starts interacting with the application and is typically terminated when the user logs out or after a period of inactivity. 

These are the main components of a session:

- Session Identifier (Session ID): A unique string that distinguishes one session from another. It is sent by the server to the client after successful login and is usually stored in a cookie on the client side.
- Session Store: A storage mechanism, often on the server side, where session data, such as user details and preferences, are stored. This could be in-memory storage, a database, a file system, or a distributed cache, depending on the application's requirements.
- Session Data: Information stored in the session, often including user preferences, user identification and authentication data, and temporary application state. This data is used to personalize the user experience and to maintain the state of the application between different requests from the client.
- Session Cookie: A type of HTTP cookie sent from the server to the client’s browser to store the session ID.
- **Session Timeout:** A mechanism to terminate sessions after a predefined period of inactivity to minimize the risk of unauthorized access. Once a session is timed out, the user needs to re-authenticate to continue interacting with the application.

Suggesting v1 is implementation of **session timeout** - forced logout after a **period of inactivity**

[...]

PR #1070, opened 2024-03-12
971 session expired logout

shon-button:
[Card](https://github.com/bcgov/cas-registration/issues/971)


marcellmueller (approved):
Nice work! I signed in and left It while I had my lunch and when I returned I was signed out 🥇 

It's beyond the scope of this ticket but it would be nice if we could redirect users to the homepage after their logout expires in the future.

shon-button:
> Nice work! I signed in and left It while I had my lunch and when I returned I was signed out 🥇
> 
> It's beyond the scope of this ticket but it would be nice if we could redirect users to the homepage after their logout expires in the future.

Thank-you for the review! :)

Agree on the `it would be nice` but, we need the NextAuth SignOut redirect to go to the Siteminder then Keycloak page as a mechanism to ensure complete logout from NextAuth and Keycloak so that users can re-login with different IDs from the same browser.

PR #1082, opened 2024-03-13
480 bceid in table

BCerki:
card: https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=49865585

### DIFF
--- a/bc_obps/registration/fixtures/mock/user.json
+++ b/bc_obps/registration/fixtures/mock/user.json
@@ -5,7 +5,7 @@
     "fields": {
       "user_guid": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
       "business_guid": "efb76d57-88b7-4eb6-9f26-ec12b49c14c1",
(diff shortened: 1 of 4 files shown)


bc_obps/registration/fixtures/mock/user.json
```diff
@@ -5,7 +5,7 @@
     "fields": {
       "user_guid": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
       "business_guid": "efb76d57-88b7-4eb6-9f26-ec12b49c14c1",
-      "bceid_business_name": "Example Business",
+      "bceid_business_name": "New Operator 3 Legal Name",
```
BCerki, line 8: I only changed a couple of these because I think it's useful to see that the bceid business name can be different from the legal name (suspicious suspicious)

marcellmueller (approved):
Beautiful, thanks for updating the action column 🙇

Issue #1037, opened 2024-03-07
Unique identifier seen by internal users in the "Request ID" field in Operator table


hannavovk:
As an internal CAS user, I want to see a meaningful user friendly unique identifier of the user access requests in the Operator access table instead of the current "Request ID". See screenshot below.

Could this be a user-friendly 'User ID' of some sort? This could also connect to a 'user ID' in the User Management table of external users.

Initial request from BA: "As an internal user, in the Operators dashboard, the request ID appears as a long key (see screenshot). Business area is requesting sequential numbers starting at 1."

![Image](https://github.com/bcgov/cas-registration/assets/87211104/2ca4f4a0-e87f-475c-b255-e99a2307cd1d)

Acceptance Criteria:
- [ ] keep UUID as the Primary Key on the `UserOperator` table, add a secondary column with auto-integer type (calling it "Friendly ID" for the sake of this ticket)
- [ ] populate the `RequestID` values in the UI table with the "Friendly ID" values instead of the UUID
- [ ] if necessary, write a script that back-populates existing prod data with values for "Friendly ID"

andrea-williams:
We want to keep the record identifier in the database as is (UUID), but we can choose to add another column with a more user-friendly unique identifier (?)

nanyangpro:
Re: Refinement - @andrea-williams  will refine the ticket description to include more technical details. Thanks!

PR #1090, opened 2024-03-13
Unique identifier seen by internal users in the "Request ID" field in Operator table

marcellmueller:
Implements #1037 

Creating a sequence in Django was challenging since Django only allows one `AutoField` since it has to have `primary_key=true` set to work. Thanks to help from Sepehr we have come up with our own solution.

To test that this will work with User Operator rows that are already in prod, you can do the following steps:
```
git checkout develop
cd /bc_obps && make reset_db
git checkout 1037-user-operator-user-friendly-unique-id
make migrate
psql -d registration
select user_friendly_id from erc.user_operator;
```
You should see the fixtures we set up in the develop branch now have a `user_friendly_id` column with the user friendly ids added:

Screenshot 2024-03-14 at 2 29 09 PM

Screenshot 2024-03-14 at 3 01 35 PM


bc_obps/registration/fixtures/mock/userOperator.json
```diff
@@ -7,7 +7,8 @@
       "role": "admin",
       "status": "Approved",
       "verified_at": "2024-02-26 06:24:57.293242-08",
-      "verified_by": "ba2ba62a121842e0942aab9e92ce8822"
+      "verified_by": "ba2ba62a121842e0942aab9e92ce8822",
+      "user_friendly_id": 1
```
marcellmueller, line 11: Needed to add this to the fixtures as the fixtures are loaded after the migration so the function that adds this to existing user operators won't run on these.

bc_obps/registration/models.py
```diff
@@ -19,6 +19,7 @@
 from django.core.exceptions import ValidationError
 from django.core.files.storage import default_storage
 from django.core.cache import cache
+from django.utils.translation import gettext as _
```
Sepehr-Sobhani, line 22: Leftover!

Sepehr-Sobhani (approved):
Just a redundant import otherwise looks great, Thanks! 🙏

bc_obps/registration/models.py
marcellmueller: Nice catch, thanks 🙇

Issue #1060, opened 2024-03-08
Remove  'User ID' column from external user management table


hannavovk:
#### Description:

As an external Admin user, I want to see a meaningful user friendly unique user identifier in the "User ID" column  instead of the uuid.

![Image](https://github.com/bcgov/cas-registration/assets/87211104/b39eab22-2848-4366-8ba7-25ce735ebf5e)

#### Acceptance Criteria:

Given I am an authenticated external (operator) admin user
When I view the User Management page
Then I don't see the User IDs (UUIDs) in the table because they're useless to me

### Additional Context
- just delete the column from the table (on the frontend only), no further action necessary

#### Dependencies

- Blocked by
- Blocking

nanyangpro:
Re: Refinement
@dleard suggested removing the column from external view since it brings no value to industry users.
@hannavovk suggested allowing searching by name to help with locating the proper record.

nanyangpro:
@andrea-williams will refine this.

hannavovk:
Awesome, thanks @marcellmueller

PR #1093, opened 2024-03-13
chore: remove user id from user management table

marcellmueller:
Implements #1060

Screenshot 2024-03-13 at 3 22 36 PM


Sepehr-Sobhani (approved):
🤩

PR #1098, opened 2024-03-14
Fix sentry env variable for django

Sepehr-Sobhani:


Sepehr-Sobhani:
@andrea-williams Here is the PR I was talking about.

andrea-williams (approved):
I can confirm that ENVIRONMENT is now spelled correctly.

I have no idea how Sentry works though 🙃

Sepehr-Sobhani:
> I can confirm that ENVIRONMENT is now spelled correctly.
> 
> I have no idea how Sentry works though 🙃

Thanks for the quick review. If you're interested, I can share what I know about Sentry.

Issue #1074, opened 2024-03-12
Flaky Playwright Test: Test Profile\Test Update\Role None [8]


shon-button:
Describe the Bug:

The 'client/e2e/poms/profile.ts\updateSuccess()`  designed to assert the success of a profile update by verifying the presence of a success message after submitting the updated profile information. However, for role `none` the test occasionally fails due to possible timing issues, where the success message is not reliably detected before navigation to the dashboard.

PR #1103, opened 2024-03-14
1074 fix e2e flaky tests

shon-button:
[Card](https://github.com/bcgov/cas-registration/issues/1074)

## 🚀 Impact:
   - Fixes flaky ` Test Profile Page › Test Role none › Test Update `
   - Redesign looping to improve test describe details

##  🔬 Testing:

**Option 1:** CI e2e test
Validate [e2e test](https://github.com/bcgov/cas-registration/actions/runs/8300523821) status is  `pass` 
**OR**
Download artefact files from `GitHub Actions\Test Registration App` to your `client/playwright-report` folder and run command `cd client && yarn playwright show-report`
[CI artefact](https://github.com/bcgov/cas-registration/actions/runs/8300523821#artifacts)
[CI artefact download](https://github.com/bcgov/cas-registration/actions/runs/8300523821/artifacts/1330478519)

image

**Option 2:** Local e2e testing:

**Pre-reqs** 

1. From file `client/e2e/.env.local.example` create `client/e2e/.env.local` with values reflecting instructions in file `client/e2e/.env.local.example`

2.  From terminal command, start the api server:
 ```
cd bc_obps
make reset_db
make run
```  
3.  From terminal command, start the app development server:
 ```
cd client && yarn build && yarn start
```  
 
**e2e tests**

1. From terminal command, run Playwright e2e tests:
```
cd client && yarn e2e
```   
2. Test Results:
image

### DIFF
--- a/client/e2e/pages/authenticated/dashboard/page.spec.ts
+++ b/client/e2e/pages/authenticated/dashboard/page.spec.ts
@@ -41,7 +41,7 @@ test.beforeAll(async () => {
     await upsertUserOperatorRecord(
       process.env.E2E_INDUSTRY_USER_ADMIN_GUID as string,
       AppRole.ADMIN,
-      UserOperatorStatus.APPROVED
+      UserOperatorStatus.APPROVED,
     );
     // Scenario FrontEndRoles.INDUSTRY_USER where userOperatorStatus !== UserOperatorStatus.APPROVED
     // Shows "Select Operator\...1 pending action(s) required" bceidSelectOperatorTile
@@ -49,7 +49,7 @@ test.beforeAll(async () => {
     // Upsert a User record: bc-cas-dev-secondary
     await upsertUserRecord(UserRole.INDUSTRY_USER);
     await deleteUserOperatorRecord(
-      process.env.E2E_INDUSTRY_USER_GUID as string
+      process.env.E2E_INDUSTRY_USER_GUID as string,
     );
   } catch (error) {
     // eslint-disable-next-line no-console
--- a/client/e2e/pages/authenticated/dashboard/page.spec.ts
+++ b/client/e2e/pages/authenticated/dashboard/page.spec.ts
@@ -41,7 +41,7 @@ test.beforeAll(async () => {
     await upsertUserOperatorRecord(
       process.env.E2E_INDUSTRY_USER_ADMIN_GUID as string,
       AppRole.ADMIN,
-      UserOperatorStatus.APPROVED
+      UserOperatorStatus.APPROVED,
     );
     // Scenario FrontEndRoles.INDUSTRY_USER where userOperatorStatus !== UserOperatorStatus.APPROVED
     // Shows "Select Operator\...1 pending action(s) required" bceidSelectOperatorTile
@@ -49,7 +49,7 @@ test.beforeAll(async () => {
     // Upsert a User record: bc-cas-dev-secondary
     await upsertUserRecord(UserRole.INDUSTRY_USER);
     await deleteUserOperatorRecord(
-      process.env.E2E_INDUSTRY_USER_GUID as string
+      process.env.E2E_INDUSTRY_USER_GUID as string,
     );
   } catch (error) {
     // eslint-disable-next-line no-console
@@ -76,7 +76,7 @@ test.describe("Test Dashboard Page", () => {
     test.describe(`Test Role ${value}`, () => {
       // 👤 run test as this role
       test.use({ storageState: storageState });
-      test("Test Selfie", async ({ page }, testInfo) => {
+      test("Test Selfie", async ({ page }) => {
         // 🛸 Navigate to dashboard page
         const dashboardPage = new DashboardPOM(page);
 
--- a/client/e2e/pages/authenticated/dashboard/page.spec.ts
+++ b/client/e2e/pages/authenticated/dashboard/page.spec.ts
@@ -41,7 +41,7 @@ test.beforeAll(async () => {
     await upsertUserOperatorRecord(
       process.env.E2E_INDUSTRY_USER_ADMIN_GUID as string,
       AppRole.ADMIN,
-      UserOperatorStatus.APPROVED,
+      UserOperatorStatus.APPROVED
     );
     // Scenario FrontEndRoles.INDUSTRY_USER where userOperatorStatus !== UserOperatorStatus.APPROVED
     // Shows "Select Operator\...1 pending action(s) required" bceidSelectOperatorTile
@@ -49,7 +49,7 @@ test.beforeAll(async () => {
     // Upsert a User record: bc-cas-dev-secondary
     await upsertUserRecord(UserRole.INDUSTRY_USER);
     await deleteUserOperatorRecord(
-      process.env.E2E_INDUSTRY_USER_GUID as string,
+      process.env.E2E_INDUSTRY_USER_GUID as string
     );
   } catch (error) {
     // eslint-disable-next-line no-console
--- a/client/e2e/pages/authenticated/dashboard/page.spec.ts
+++ b/client/e2e/pages/authenticated/dashboard/page.spec.ts
@@ -41,7 +41,7 @@ test.beforeAll(async () => {
     await upsertUserOperatorRecord(
       process.env.E2E_INDUSTRY_USER_ADMIN_GUID as string,
       AppRole.ADMIN,
-      UserOperatorStatus.APPROVED,
+      UserOperatorStatus.APPROVED
     );
     // Scenario FrontEndRoles.INDUSTRY_USER where userOperatorStatus !== UserOperatorStatus.APPROVED
     // Shows "Select Operator\...1 pending action(s) required" bceidSelectOperatorTile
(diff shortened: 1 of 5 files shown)


client/e2e/pages/authenticated/dashboard/profile/page.spec.ts
```diff
@@ -6,39 +6,62 @@ import { DashboardPOM } from "@/e2e/poms/dashboard";
...
+// 📚 Declare a beforeAll hook that is executed once per worker process before all tests.
+// 🥞 Set DB for e2e login roles
+/*
+For "new user":
+-  delete record in the db so that on "new user" login the ID will have no app_role
+*/
+test.beforeAll(async () => {
+  try {
+    // 👤 delete new user: bc-cas-dev-three
+    await deleteUserRecord(process.env.E2E_NEW_USER_GUID as string);
```
BCerki, line 24: Further to our conversation about seeding/resetting, are there any tests where we need the new user to already be in the db? If not, maybe we can remove them from global so we don't have to add/delete them, and they can just start out not existing

client/e2e/pages/authenticated/routes/page.spec.ts
```diff
@@ -50,119 +58,121 @@ function buildAccessLists(currentRole: UserRole): AppRoute[] {
...
+            case AppRoute.USERS:
+              pomPage = new UsersPOM(page);
+              break;
+          }
+          if (pomPage) {
+            // 🚨 Check if route is in the role's allow list
+            const isAllowedRoute = accessLists.includes(route);
+            // 🛸 Navigate to route
+            await pomPage.route();
+            // 📌 Some roles have exceptions to the expected results
```
BCerki, line 107: I found this test hard to follow--lots of exceptions to the defaults and different cases. If the test changes to only cover not-allowed routes though, I think that'll naturally improve readability.

client/e2e/pages/authenticated/routes/page.spec.ts
```diff
@@ -50,119 +58,121 @@ function buildAccessLists(currentRole: UserRole): AppRoute[] {
...
-for (let [role, value] of Object.entries(UserRole)) {
-  role = "E2E_" + role;
-  const storageState = process.env[role + "_STORAGE"] as string;
-  test.describe(`Test Route Access for ${value}`, () => {
-    // 👤 Run test as this role
-    test.use({ storageState: storageState });
-    test("Test Navigate to Routes", async ({ page }) => {
-      // 🚨 Build routes allowed list for this user role
-      const accessLists = buildAccessLists(value);
+test.describe("Test Route Access", () => {
```
BCerki, line 61: If I understand this test correctly, it checks that every role can/can't get to the appropriate routes. If that's correct, then I'm not sure this test needs to check accessible routes because I think the role-based workflow tests will end up covering that. For example, the industry_user_admin has to go to dashboard > operations > operation in order to test the approve/deny boro id workflow

client/e2e/poms/profile.ts
```diff
@@ -46,26 +46,20 @@ export class ProfilePOM {
     await fieldsUpdate(this.page);
     // Click the Submit button
     await this.buttonSubmit.click();
-    // Wait for the success message to appear
-    await this.page.waitForSelector('div:has-text("✅ Success")', {
-      timeout: 15000,
+    // Response from submit either shows errors or triggeres handleSubmit which handles state changes on submit button etc.
+    // 🔍 Assert that the error selector is not available
+    await this.page.waitForSelector(DataTestID.ERROR_PROFILE, {
```
BCerki, line 51: `waitForSelector` is discouraged: https://playwright.dev/docs/api/class-page. Playwright's supposed to have good async/waiting support--would something as simple as page.getByText("✅ Success") work? Or is this one more complicated because it's state-related?

client/e2e/pages/authenticated/dashboard/page.spec.ts
```diff
@@ -68,35 +68,38 @@ test.afterEach(async () => {
...
-  role = "E2E_" + role;
-  const storageState = process.env[role + "_STORAGE"] as string;
-  test.describe(`Test Dashboard for ${value}`, () => {
-    // 👤 run test as this role
-    test.use({ storageState: storageState });
-    test("Test Selfie", async ({ page }) => {
-      // 🛸 Navigate to dashboard page
-      const dashboardPage = new DashboardPOM(page);
+test.describe("Test Dashboard Page", () => {
+  // ➰ Loop through the entries of UserRole enum
```
BCerki, line 72: This looks like major diffs in the github UI, but I think the only real change is to wrap everything in a `test.describe("Test Dashboard Page")`? (If there are more changes I'll look more closely)

client/e2e/pages/authenticated/dashboard/profile/page.spec.ts
```diff
@@ -6,39 +6,62 @@ import { DashboardPOM } from "@/e2e/poms/dashboard";
...
-      await profilePage.route();
-      // 🔍 Assert that the current URL
-      await profilePage.urlIsCorrect();
-      // 🔍 Assert profile update validates required fields
-      await profilePage.updateFail();
-      // 🔍 Assert profile update success
-      await profilePage.updateSuccess();
-      //🔍 Assert profile name reflects the updated user profile full name
-      await profilePage.userFullNameIsCorrect("e2e first name* e2e last name*");
+test.describe("Test Profile Page", () => {
```
BCerki, line 34: Same as above, many diffs in the UI but I don't think there are real changes below?

client/e2e/pages/authenticated/routes/page.spec.ts
```diff
@@ -50,119 +58,121 @@ function buildAccessLists(currentRole: UserRole): AppRoute[] {
...
+  // ➰ Loop through the entries of UserRole enum
```
BCerki, line 62: Are these changes related to fixing the flakiness, or separate thing?

BCerki (commented):
Thanks for working on this on, flakiness is the worst!

client/e2e/pages/authenticated/routes/page.spec.ts
shon-button: @BCerki  If you are suggesting to test only `not-allowed routes` I would suggest that in our implementation of the app route access it is easier to **not** have a route a role should have over having a route a role should not have. So, allow routes are important to test IMO.

Note: this test, with its exceptions,  already exists in develop, I have only re-arranged the looping so to get better details from the test descriptions

client/e2e/pages/authenticated/routes/page.spec.ts
shon-button: @BCerki there was 1 flaky test in this file for so I reconfigured the loop to get better details on the role and the route in hopes to help troubleshoot.

client/e2e/poms/profile.ts
shon-button: @BCerki  I believe the issue is with a race condition for new user having a successful submit and getting re-directed to dashboard from profile.

client/e2e/pages/authenticated/dashboard/page.spec.ts
shon-button: @BCerki correct, attempt to `Redesign looping to improve test describe details`

client/e2e/pages/authenticated/dashboard/profile/page.spec.ts
shon-button: @BCerki correct, attempt to `Redesign looping to improve test describe details`

client/e2e/pages/authenticated/routes/page.spec.ts
shon-button: @BCerki agree there is overlap but, I think there is value in having a separate test specific to access rights for all routes for all roles.

client/e2e/pages/authenticated/dashboard/profile/page.spec.ts
shon-button: @BCerki there is no need for `new user to already be in the db`.  The profile update test will create the record for new user which is negated by the berforeAll hook.  If we have agreed to tread lightly in terms of implementing seeding/resetting with a watchful eye on resource hogging then I might leave the pages\ directory setup as is until further review of the seeding process.

client/e2e/pages/authenticated/routes/page.spec.ts
BCerki: I agree that allow routes are important to test--I was only suggesting removing them here because they're already going to be covered in the other tests, and testing the same thing twice is going to slow down our e2e runs. I'm fine to make a tech debt card to look into removing duplicate tests later though; if the sharding etc. stuff you're working on speeds things up a lot then double testing is less of a big deal.

client/e2e/poms/profile.ts
BCerki: Sorry, I think I put the wrong link, if we have to use a selector that involves waiting, the recommendation is `locator.waitFor()` instead of `waitForSelector`
image

shon-button: @BCerki  for your review:
https://github.com/bcgov/cas-registration/pull/1103/commits/8ccde178e4d05186b94f5c06f4092dda898f50a0

BCerki (approved):

Issue #816, opened 2024-02-05
Phone number field lock-out error [6]


Dianadec:
Describe the Bug:
As an external user, I logged in and entered my Profile info, submitted it, then clicked back. 
Once back on the user info form I changed the phone number to an invalid one and tried to submit. 
This threw the error state, but I was then locked out of the field and unable to update it with a valid phone number.
(found by Mike)

![Image](https://github.com/bcgov/cas-registration/assets/35785153/75646e61-6f32-452a-9eb4-2e9a73956020)

Probability (how likely the bug is to happen, scored from 1-5): [2]
(For example, probability of 5 is something like "it happens to all users every time they log in."
Probability of 1 "only happens to certain users when a really specific and unlikely path is followed.")

Effect (how bad the bug is when it does happen, scored from 1-5): [4]
(For example, effect of 5 is "the entire app crashes and makes it unusable for all users" or "the bug causes the wrong data to be saved, with critical information (e.g. payment) being affected."
Effect of 1 is "It makes some styling look a little bit weird.")

Steps to reproduce the behaviour:

Screenshots:

Additional information:

- Browser: [e.g. chrome, safari]
- Version: [e.g. 22]
- Additional context

marcellmueller:
Posting my comments from PR here:

I had difficulty recreating this at first though it turns out it's an error with US phone numbers. Since both Canada and US use the same phone format a US number can pass our validation.

To recreate it log in as either `bc-cas-dev-three` or through IDIR on a fresh database so it brings you to the profile page. For the phone number use an American phone number eg `(234)-555-1234`, fill out the position field and submit the form.

Return to the profile page either by clicking back or clicking the user on the top right and the phone field will be a US number. If you try to edit it the widget will be locked out and show an error:
<img width="862" alt="Screenshot 2024-03-14 at 3 42 03 PM" src="https://github.com/bcgov/cas-registration/assets/14259474/e8ae19b0-234a-4947-9ae0-431a7ead4b3b">

There are a couple options to fix this and I think we should go with adding US to the `onlyCountries` prop. In the edge case that someone uses a US number this will allow them to edit it with no problems.

The other option is to check if it's a Canadian phone number which the library supports. Though I hesitate to do this as there are potential edge cases to block users who have valid phone numbers, either because they have a US number or else the validation is outdated due to new area codes. 

```
import { matchIsValidTel } from 'mui-tel-input'

 const isValid = matchIsValidTel(val, {
       defaultCountry: "CA",
[...]

PR #1121, opened 2024-03-14
fix: phone widget lockout error

marcellmueller:
Fix for #816 

I had difficulty recreating this at first though it turns out it's an error with US phone numbers. Since both Canada and US use the same phone format a US number can pass our validation.

To recreate it log in as either `bc-cas-dev-three` or through IDIR on a fresh database so it brings you to the profile page. For the phone number use an American phone number eg `(234)-555-1234`, fill out the position field and submit the form.

Return to the profile page either by clicking back or clicking the user on the top right and the phone field will be a US number. If you try to edit it the widget will be locked out and show an error:
Screenshot 2024-03-14 at 3 42 03 PM

There are a couple options to fix this and I think we should go with adding US to the `onlyCountries` prop. In the edge case that someone uses a US number this will allow them to edit it with no problems.

The other option is to check if it's a Canadian phone number which the library supports. Though I hesitate to do this as there are potential edge cases to block users who have valid phone numbers, either because they have a US number or else the validation is outdated due to new area codes. 

```
import { matchIsValidTel } from 'mui-tel-input'

 const isValid = matchIsValidTel(val, {
       defaultCountry: "CA",
 }),
```

### DIFF
--- a/client/app/components/form/widgets/PhoneWidget.tsx
+++ b/client/app/components/form/widgets/PhoneWidget.tsx
@@ -34,7 +34,7 @@ const PhoneWidget: React.FC<WidgetProps> = ({
       onChange={handleChange}
       forceCallingCode
       defaultCountry="CA"
-      onlyCountries={["CA"]}
+      onlyCountries={["CA", "US"]}
       sx={styles}
     />
   );


Sepehr-Sobhani (approved):
🥇

PR #1127, opened 2024-03-15
1062 api refactor example

BCerki:
moved to new PR #1259


bc_obps/registration/api/user_operator/is_approved_admin_user_operator.py
```diff
@@ -0,0 +1,25 @@
...
+)
+from registration.api.api_base import router
+
+from registration.models import (
+    UserOperator,
+)
+from ninja.responses import codes_4xx
+
+
+# BRIANNA CHANGE THESE TO BE PROPERLY UNDER USER-OPERATOR/OPERATOR/WHATEVER MAKES SENSE
```
BCerki, line 16: whoopsie

bc_obps/registration/decorators.py
```diff
@@ -34,3 +36,22 @@ def wrapper(request: HttpRequest, *args: Any, **kwargs: Any) -> Any:
         return wrapper
 
     return decorator
+
+
+def try_except_wrapper():
```
pbastia, line 41: This should probably reflect what it's doing in the name! Maybe something like `with_http_errors()`?

bc_obps/registration/service/handle_exception.py
```diff
@@ -0,0 +1,16 @@
...
+    generate_useful_error,
+)
+from django.core.exceptions import ValidationError
+
+from django.http import Http404
+
+
+def handle_exception(error):
+    if isinstance(error, ValidationError):
+        return 400, {"message": generate_useful_error(error)}
```
pbastia, line 11: We could use the [`422 unprocessable entity`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422) code here to discriminate from the generic `400 bad request`

BCerki: @Sepehr-Sobhani , do you know if django ninja expects a 400 for validation errors? When I changed it to 422 I got this error: `ninja.errors.ConfigError: Schema for status 422 is not set in response dict_keys([200, 416, 418, 451, 425, 429, 400, 401, 402, 403, 404, 405, 406, 407, 408, 409, 410, 411, 412])`. I'm using the `from ninja.responses import codes_4xx` so I'm wondering why that wouldn't include 422 if we were meant to be using it

Sepehr-Sobhani: You are correct. HTTP status code 422 is reserved and triggers errors when the pydantic schema fails to process the data. So, If there's an issue with the schema we don't even get to this part of the code to raise an error and django-ninja automatically returns the 422 error for us.
A related topic(somehow :) ): https://github.com/vitalik/django-ninja/issues/491#issuecomment-1170990866

BCerki: Wow, really? I was totally just guessing haha

Sepehr-Sobhani: Told you, you nailed Python!

Sepehr-Sobhani: Still, I think we should add `422` manually since it's the most suitable code for informative errors.

bc_obps/registration/service/select_operator/user_operator/UpdateStatus.py
```diff
@@ -0,0 +1,71 @@
+from django.db import transaction
+from registration.utils import (
+    generate_useful_error,
+)
+from django.core.exceptions import ValidationError
+import pytz
```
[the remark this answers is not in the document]
Sepehr-Sobhani: In case you want to resolve this issue too:
Screenshot 2024-03-22 at 10 57 40 AM

Otherwise, it could be a tech debt.

Sepehr-Sobhani (changes requested):
Thank you for tackling this issue.
I have reviewed the code and provided some minor comments and suggestions. Overall, the changes look great. I strongly recommend renaming all service files to follow the Python file naming convention (Snake case). Additionally, as you mentioned in the pull request description, we could consider removing the `select-operator` component and relocating all related files to the `user_operator` folder as an enhancement.

bc_obps/registration/service/handle_exception.py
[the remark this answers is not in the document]
pbastia: Agreed! The service layer should probably live outside of a registration or app specific place

pbastia (changes requested):
Nice work! This is great to see that we can break things down and be a little closer to the Single Responsibility Principle. 

About the service layer, I think there's a way of making the different services have a bot of a cleaner API. We don't really need one service per function, although the implementation should still separate the functions as independently as possible.
I was imagining an API like:
`DataAccessService.create_operator(...):` 
`DataAccessService.list_user_operators(userguid: abc, exclude: Statuse.DECLINED):`
- These are very much examples and will need more thought on how to make the data access layer flexible in its usage. 
- Maybe if we want to namespace a little better like `DataAccessService.Operators.create(...):`), this would be easy with python's folder/module structure

Another comment is about being careful with the signature of the service methods. We should only include in the parameters the data needed to perform the operation, instead of API data structures (like http requests). Otherwise, this means:
- the service layer needs to know about API http request data structures, which means they're now tightly coupled
- other consumers can't use the service without crafting an http request object, which means they also need to know about the API service...we can already visualize the spaghetti ;)
- If the service needs complex data to perform some operations, then the service needs to be responsible for exposing the data structures and interfaces to be used. This way the consumer only needs to know about the service they're trying to use, and not the other way around.


Happy to hear your thoughts! We can discuss all this tomorrow during our check-in, or async as you prefer.

PR #1160, opened 2024-03-16
Giraffe/feat/login page

pbastia:
Adding a login page for the reporting app, using the registration components


BallardRobinett (commented):
This looks good to me, but I'm not sure how to actually navigate to the login page in my local dev environment

BallardRobinett (approved):
working on my local, looking good!

joshgamache (approved):
Running through the login and getting to the authenticated page is working well! Code looks good too.

I think we'll have to establish some norms with how we want to leverage Nx. Once we get some of the core in place, I think it would be good to pull in the components we want to use through a Nx `lib` vs importing them from `/client`. 

I ran into a hiccup when setting it up due to not having the right packages installed. I was running on a fresh clone of the repo, so I hadn't run `yarn` in `/client` yet. Looking through the code I noticed you're pulling from that, so all it took was that.

pbastia:
@joshgamache yes! I 100% agree that importing from outside the project scope is a little wonky. Once we merge giraffe-develop and develop, we'll be able to pull apart the registration code and extract modules and libraries. Until then I'm quite worried about minimizing the changes we make to the side moose is responsible for, so that the merge goes smothly

Issue #1007, opened 2024-03-01
Dashboard/operators/user-operator displays User Information\Approve\Decline button when Operator is denied [9]


shon-button:
Describe the Bug:
Dashboard/operators/user-operator displays User Information\Approve\Decline button when Operator is denied
![image](https://github.com/bcgov/cas-registration/assets/120038448/01ee9c4c-4b63-4cd2-ba1b-f42ba7395a0e)

Probability (how likely the bug is to happen, scored from 1-5): 4
(For example, probability of 5 is something like "it happens to all users every time they log in."
Probability of 1 "only happens to certain users when a really specific and unlikely path is followed.")

Effect (how bad the bug is when it does happen, scored from 1-5): 5
(For example, effect of 5 is "the entire app crashes and makes it unusable for all users" or "the bug causes the wrong data to be saved, with critical information (e.g. payment) being affected."
Effect of 1 is "It makes some styling look a little bit weird.")

Steps to reproduce the behaviour:

Screenshots:

Additional information:

- Browser: [e.g. chrome, safari]
- Version: [e.g. 22]
- Additional context

PR #1162, opened 2024-03-18
fix: hide prime admin review when operator is declined

marcellmueller:
Fix for #1007 

Also fixed the User Operator fixtures that were tied to a new Operator. Many of them were tied to the same new Operator which can't happen in the app since you can't create a new operator with the same CRA number or BC corporate registry number. To fix this I created a bunch of new Operators to tie to each of the offending User Operators.


BCerki (approved):
:cartwheeling:

PR #1163, opened 2024-03-18
Happo merge reports

marcellmueller:
Implements issue #1168

Merging Happo reports after the recent PR which runs parallel using shards.

Happo docs for this here: https://docs.happo.io/docs/playwright#parallel-builds

One thing I could use some feedback on is the addition of yet another e2e job, `happo-finalize`. I had considered putting it in the `e2e report` job to keep the job list cleaner but left it here for now.

Screenshot 2024-03-18 at 2 40 34 PM

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -281,9 +281,9 @@ jobs:
           E2E_NEW_USER_GUID: ${{ secrets.E2E_NEW_USER_GUID }}
           E2E_NEW_USER_PASSWORD: ${{ secrets.E2E_NEW_USER_PASSWORD }}
           E2E_NEW_USER_STORAGE: ${{ secrets.E2E_NEW_USER_STORAGE}}
-          HAPPO_ENABLED: true
           HAPPO_API_KEY: ${{ secrets.HAPPO_API_KEY }}
           HAPPO_API_SECRET: ${{ secrets.HAPPO_API_SECRET }}
+          HAPPO_NONCE: ${{ github.event.pull_request.head.sha }}
         working-directory: ./client
       - name: 💾 save ${{ matrix.project }} report artifact
         # prefer to upload the report only in case of test failure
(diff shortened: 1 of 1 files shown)


shon-button (approved):
@marcellmueller super duper!

One "for your consideration", would it makes sense to only create happo report if e2e-test fails...as is being done for the Playwright HTML report (only created on e2e-tests failure)
Happy with whatever you decide :)

shon-button:
> Implements issue #1168
> 
> Merging Happo reports after the recent PR which runs parallel using shards.
> 
> Happo docs for this here: https://docs.happo.io/docs/playwright#parallel-builds
> 
> One thing I could use some feedback on is the addition of yet another e2e job, `happo-finalize`. I had considered putting it in the `e2e report` job to keep the job list cleaner but left it here for now.
> 
> Screenshot 2024-03-18 at 2 40 34 PM

I think modular `happo-finalize` job is fine...it's not on you that the file is getting long! ;)

marcellmueller:
> @marcellmueller super duper!
> 
> One "for your consideration", would it makes sense to only create happo report if e2e-test fails...as is being done for the Playwright HTML report (only created on e2e-tests failure) Happy with whatever you decide :)

I thought about this and I think that Happo is uploading screenshots as the tests run. I'm not sure it would make any difference as a Happo report is not created if the e2e tests fail with or without this job eg this run from another PR with a failing e2e run: 

Screenshot 2024-03-20 at 11 52 29 AM

Though one way we could cut the time this job takes is to separate out the `yarn cache` step from the `dev-env-setup` action in it's own tiny composite action as it is the only part of that needed in this job. I think it could be useful for other jobs going forward as well. What are your thoughts on that? 

I am going to merge this as is as I want to monitor our Happo diffs today for any issues. Though let me know if you think that is worth our time and how we might structure that if it's doable.

PR #1166, opened 2024-03-18
chore: add new make target and logic for shipit.yml

BallardRobinett:
task: https://github.com/orgs/bcgov/projects/123/views/2?pane=issue&itemId=54817515

Created a new make target for deploying from the ```giraffe-develop``` branch to the ```CIF-tools``` namespace. I added some logic to the ```shipit.yml``` file to determine which branch it is being run from and choose the appropriate make target. 

The new make target is hard-coded to pull values from values-giraffe.yaml, which has a frontend route of cas-reg-giraffe-dev.apps.silver.devops.gov.bc.ca, which is subject to change.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,12 +70,12 @@ install_giraffe: check_environment
 install_giraffe:
 install_giraffe: GIT_SHA1=$(shell git rev-parse HEAD)
 install_giraffe: IMAGE_TAG=$(GIT_SHA1)
-install_giraffe: NAMESPACE=$(OBPS_NAMESPACE_PREFIX)-$(ENVIRONMENT)
+install_giraffe: NAMESPACE=$(CIF_NAMESPACE_PREFIX)-tools
 install_giraffe: CHART_DIR=./helm/cas-registration
 install_giraffe: CHART_INSTANCE=cas-registration
 install_giraffe: HELM_OPTS=--atomic --wait-for-jobs --timeout 2400s --namespace $(NAMESPACE) \
 										--set defaultImageTag=$(IMAGE_TAG) \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -64,6 +64,30 @@ install:
 		helm upgrade $(HELM_OPTS) $(CHART_INSTANCE) $(CHART_DIR); \
 	fi;
 
+.PHONY: install_giraffe
+install_giraffe: ## Installs the helm chart on the OpenShift cluster
+install_giraffe: check_environment
+install_giraffe:
+install_giraffe: GIT_SHA1=$(shell git rev-parse HEAD)
+install_giraffe: IMAGE_TAG=$(GIT_SHA1)
+install_giraffe: NAMESPACE=$(OBPS_NAMESPACE_PREFIX)-$(ENVIRONMENT)
```
mikevespi, line 73: Since we're deploying to the CIF tools namespace, this line can change to 
`install_giraffe: NAMESPACE=$(CIF_NAMESPACE_PREFIX)-tools`

mikevespi (changes requested):
This looks good to me! Just he one change request, and once you've made that you're safe to temporarily change this to run on  your branch to test it, then change it back. If that goes smoothly, we're good to go!

Makefile
```diff
@@ -64,6 +64,30 @@ install:
...
+install_giraffe: CHART_DIR=./helm/cas-registration
+install_giraffe: CHART_INSTANCE=cas-registration
```
joshgamache, line 75: Would we want to change the `CHART_INSTANCE` to something else, just to avoid colissions if we ever chose to deploy to the same namespace as Moose? I've been using `cas-bciers` in my testing, not sure if that's best though.

Issue #1113, opened 2024-03-14
 e2e CI cache for Playwright binaries


shon-button:
#### 🎯 Target

The goal is to reduce run time of CI pipeline's e2e Playwright tests over the current run time by implementing Playwright binary caching strategy.

**Current Strategy:**
- Using  GitHub Actions `jobs.<job_id>.strategy` configuration to implement a matrix strategy to generate multiple job runs based on combinations of matrix variables. [jobs matrix](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs) 
- Implement Playwright [test sharding](https://playwright.dev/docs/test-sharding) as the job strategy matrix

**Possibilities for Next Strategy:**

1. Configure Github Actions YML file includes a section that creates a Cache and stores the playwright binaries so Github Actions doesn't have to re-download every single time the action runs...
[with cache](https://github.com/microsoft/playwright/issues/20603#top)
**Issue:** some shard runs find the cache but then test fails due to missing deps (seems to be webkit)

2. Configure .yaml to use Playwright image
[with container setup](https://radekmie.dev/blog/on-playwright-in-github-actions/)
**Issue:** No Happo in image used; so, no happy

3. Change matrix variables based on browsers over shards
[with matrix of three browser engines](https://dev.to/drakulavich/fast-and-reliable-end-to-end-tests-with-playwright-on-github-actions-2mkh)
**Issue:** No show stoppers, yet?...

**Dev Tasks**
- [x] Modify `.github/workflows/test.yaml` to use install Playwright browser deps if not found in cache
- [x] Modify `.github/workflows/test.yaml` to  **change matrix variables** from shard index to browser based

#### Acceptance Criteria

- [ ] GitHub e2e tests execution time is reduced
- [ ] Playwright report displays all FAILED test shards
- [ ] Playwright e2e CI status returns pass

PR #1177, opened 2024-03-19
1113 e2e ci matrix cache

shon-button:
[Card](https://github.com/bcgov/cas-registration/issues/1113)

## 🚀 Impact:
   - Reduces CI pipeline's e2e Playwright tests run time
   - Restricts CI pipeline's e2e Playwright tests report artifacts to failures
   - Cancels current job when pushing new commit into the PR

## ✏️ Notes:
- Implementing playwright binary cache using `shard` matrix variables results in test(s) finding the cache but then failing due to missing browsers installs 🤷 
- Implementing playwright binary cache using `browser` matrix results in 1 of 3 tests, the `e2e test webkit`,  NOT finding cache; so, that test takes extra 30 seconds to install Playwright deps 🤷 
**Possible Explanation** [This is expected...](https://github.com/microsoft/playwright/issues/22146#issuecomment-1493902369)
- Overall, we have reduced e2e test run by ~40 seconds 🎯

### AFTER CACHING IMPLEMENTATION
image

### BEFORE CACHING IMPLEMENTATION
image

##  🔬 Testing:

Validate `1113-e2e-ci-matrix-cache` [e2e test](https://github.com/bcgov/cas-registration/actions/) status is  `pass`

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -202,33 +202,57 @@ jobs:
           fail_action: false
 
   e2e-tests:
-    needs:
-      ["backend-docker-build", "frontend-docker-build", "install-dev-tools"]
-    timeout-minutes: 60
+    name: 🧪 e2e tests ${{ matrix.project }}
     runs-on: ubuntu-latest
+    needs:
+      - backend-docker-build
+      - frontend-docker-build
+      - install-dev-tools
+    timeout-minutes: 60
     strategy:
       fail-fast: false
       matrix:
-        shardIndex: [1, 2, 3, 4]
-        shardTotal: [4]
+        include:
+          - project: chromium
+            os: ubuntu-latest
+            cache_dir: ~/.cache/ms-playwright
+          - project: firefox
+            os: ubuntu-latest
+            cache_dir: ~/.cache/ms-playwright
+          - project: webkit
+            os: macos-12
+            cache_dir: ~/Library/Caches/ms-playwright
     steps:
       - uses: actions/checkout@v3
-      - name: dev env setup
+
+      - name: 🎁 setup dev env
         uses: ./.github/actions/dev-env-setup
-      - name: run app locally
+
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yaml
```diff
@@ -202,33 +202,57 @@ jobs:
...
       fail-fast: false
       matrix:
-        shardIndex: [1, 2, 3, 4]
-        shardTotal: [4]
+        include:
+          - project: chromium
+            os: ubuntu-latest
+            cache_dir: ~/.cache/ms-playwright
+          - project: firefox
+            os: ubuntu-latest
```
marcellmueller, line 220: I wonder if we should test in Edge given that this is a BC Gov project and many people are likely using Edge.

marcellmueller (approved):
Another amazing PR 😍 I live for optimization work like this haha

I believe it's unrelated to this PR specifically but since the auth handling was changed I noticed we are getting screenshot diffs with your name being used for the IDIR users instead of IDIR USER. Do you know why that is happening?

.github/workflows/test.yaml
shon-button: Good idea...but, `Edge`  😬

shon-button:
> Another amazing PR 😍 I live for optimization work like this haha
> 
> I believe it's unrelated to this PR specifically but since the auth handling was changed I noticed we are getting screenshot diffs with your name being used for the IDIR users instead of IDIR USER. Do you know why that is happening?

@marcellmueller thank-you very much for the review!

Suggest we discuss the expected logged in user name in consequence with the change in the storageStates functionality  further...again, great insight! ✨

.github/workflows/test.yaml
marcellmueller: Haha I know, I only brought it up because of that recent post that some users were getting errors logging into Edge 🙃

Issue #1168, opened 2024-03-18
Merge Happo reports after parallel runs in CI


marcellmueller:
With the recent PR that runs Playwright tests in parallel we need to set a `HAPPO_NONCE` env variable and finalize the Happo report to merge the screenshots from these jobs together.

https://docs.happo.io/docs/playwright#parallel-builds

PR #1197, opened 2024-03-20
happo e2e optimize

marcellmueller:
Follow up PR to #1168 

- Fixes a bug from using `github.event.pull_request.head.sha` as the `HAPPO_NONCE` variable. Since the pull request head sha doesn't exist when this is ran from develop it fails. I am using `github.sha` instead which is the sha of the commit that started the job. 
- Move the `yarn cache` step from `dev-env-action` into it's own composite action. This greatly reduces the time of the `happo-finalize` step and hopefully will be useful in the future when we need to run a command from a dependency that is in our `node_modules` folder.
- Added a timeout since we keep getting a Firefox screenshot diff which gives the text 500ms to settle and render correctly. If you use the zoom feature in Happo, set it to 400% and zoom into one of the affected characters using the swipe tab you can see that text characters are slightly moving once they appear. Though I know that using timeouts is considered bad practice so I'd love any other suggestions to stop this issue. 

Example of the failing diffs:
https://happo.io/a/336/p/1755/compare/8d06beaf31ebf85d9cca9d6d918d7ee7464e2859/e2ffbfa1350ede06bdccd06e519ba250037b3b6c

Failing happo-finalize job from main:
https://github.com/bcgov/cas-registration/actions/runs/8366960287/job/22908575263

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -283,7 +283,7 @@ jobs:
           E2E_NEW_USER_STORAGE: ${{ secrets.E2E_NEW_USER_STORAGE}}
           HAPPO_API_KEY: ${{ secrets.HAPPO_API_KEY }}
           HAPPO_API_SECRET: ${{ secrets.HAPPO_API_SECRET }}
-          HAPPO_NONCE: ${{ github.event.pull_request.head.sha }}
+          HAPPO_NONCE: ${{ github.sha }}
         working-directory: ./client
       - name: 💾 save ${{ matrix.project }} report artifact
         # prefer to upload the report only in case of test failure
@@ -342,7 +342,7 @@ jobs:
         env:
           HAPPO_API_KEY: ${{ secrets.HAPPO_API_KEY }}
           HAPPO_API_SECRET: ${{ secrets.HAPPO_API_SECRET }}
-          HAPPO_NONCE: ${{ github.event.pull_request.head.sha }}
+          HAPPO_NONCE: ${{ github.sha }}
         run: npx happo-e2e finalize
         working-directory: ./client
 
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -336,8 +336,8 @@ jobs:
     needs: [e2e-tests]
     steps:
       - uses: actions/checkout@v4
-      - name: dev env setup
-        uses: ./.github/actions/dev-env-setup
+      - name: yarn cache
+        uses: ./.github/actions/yarn-cache
       - name: finalize happo e2e tests
         env:
           HAPPO_API_KEY: ${{ secrets.HAPPO_API_KEY }}
--- a/client/e2e/workflows/bceidbusiness/industry_user.spec.ts
+++ b/client/e2e/workflows/bceidbusiness/industry_user.spec.ts
@@ -167,7 +167,9 @@ test.describe("Test Workflow industry_user", () => {
     );
 
     // Add short timeout to mitigate the Firefox text rendering issue causing spurious screenshot failures
-    await page.waitForTimeout(500);
+    if (test.info().project.name === "firefox") {
+      await page.waitForTimeout(500);
+    }
 
     pageContent = page.locator("html");
     await happoPlaywright.screenshot(page, pageContent, {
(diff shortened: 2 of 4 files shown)


.github/workflows/test.yaml
```diff
@@ -283,7 +283,7 @@ jobs:
           E2E_NEW_USER_STORAGE: ${{ secrets.E2E_NEW_USER_STORAGE}}
           HAPPO_API_KEY: ${{ secrets.HAPPO_API_KEY }}
           HAPPO_API_SECRET: ${{ secrets.HAPPO_API_SECRET }}
-          HAPPO_NONCE: ${{ github.event.pull_request.head.sha }}
+          HAPPO_NONCE: ${{ github.sha }}
```
marcellmueller, line 286: In retrospect it makes sense this wouldn't work after merging into main since there is no pull request sha. So I'm using `github.sha`:

Screenshot 2024-03-20 at 2 26 37 PM

shon-button: @marcellmueller and now you are an expert in yaml as well- nice!

shon-button:
@marcellmueller 🤩 **Fantastic!** `Move the yarn cache step from dev-env-action into it's own composite action`

shon-button:
@marcellmueller it seems a shame to spend time with `waitForTimeout` after you gained so much time in the yaml composite change...

perhaps, at a minimum, you could wait only for FireFox...
```
 // Add short timeout to mitigate the Firefox text rendering issue causing spurious screenshot failures
    if (test.info().project.name === "firefox") {
      await page.waitForTimeout(500);
    }
```

and potentially we could solve with happo config tweaks?

Some thoughts...that you may have already thought about
- are the `https://happo.io/a/336/projects/1755/thresholds` configurations a best guess, or could we use strategy to determine the best  threshold value:
```
To help find the right value to use, you can make dry-run comparisons. Find one or a few comparisons (via https://happo.io/dashboard) and run happo compare <sha1> <sha2> --dry-run on the SHAs and look at what's being logged to figure out what threshold value you want to use.
```
- can we leverage [permanently ignore the diff](https://docs.happo.io/docs/ignoring-diffs)
- should we make a tech debt ticket for possible future possible resolutions to spurious diffs

shon-button (approved):

client/e2e/workflows/bceidbusiness/industry_user.spec.ts
shon-button, line 1: Perhaps wait just for firefox: 
 // Add short timeout to mitigate the Firefox text rendering issue causing spurious screenshot failures
    if (test.info().project.name === "firefox") {
      await page.waitForTimeout(500);
    }

marcellmueller:
> * are the [threshold configurations](https://happo.io/a/336/projects/1755/thresholds) a best guess, or could we use strategy to determine the threshold value:
> 
> ```
> To help find the right value to use, you can make dry-run comparisons. Find one or a few comparisons (via https://happo.io/dashboard) and run happo compare <sha1> <sha2> --dry-run on the SHAs and look at what's being logged to figure out what threshold value you want to use.
> ```

Sepehr suggested increasing the diff threshold but some of those Firefox diffs have 10+ characters shifting so we would likely have to increase the threshold to 100+ pixels to accommodate.  I think that's a great strategy you suggested that we can use in the future if we are getting smaller diffs.

> * can we leverage [permanently ignore the diff](https://docs.happo.io/docs/ignoring-diffs)
> * should we make a tech debt ticket for future possible resolutions to spurious diffs

I also played with permanently ignoring the diff though do you think that has a risk of another issues getting through? I'm not sure how 'smart' the feature is so if we had something broken with the validation then would that Firefox diff fail then?

I do think a tech debt ticket might be helpful. Screenshot testing is tough and always has small issues like this so maybe some research, documentation and a team agreement would be helpful. Personally I have found that using timeouts is unavoidable when screenshot testing. There is likely another solution for this specific issue but especially in the case of animated components like our accordions that expand, a timeout is sometimes necessary to let the animation complete so we take the screenshots at the right time in CI.

> **Side note:** I was surprised that the `client/e2e/workflows/bceidbusiness/industry_user.spec.ts` had test "Test Redirect to Dashboard" seems superfluous since we test dashboards in `client/e2e/pages/authenticated/dashboard/page.spec.ts` - any thoughts on this?

Oh I had left that one in my `industry_user_admin` PR as well, though in retrospect I realize it was a placeholder setup test. I checked Happo and we have two identical screenshots with slightly different names, I will remove this test.

PR #1198, opened 2024-03-20
Reduce sentry trace rate

Sepehr-Sobhani:
[ISSUE 1189](https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=57110752)
### Notes:

We have made some changes to the sentry `traceSampleRate`. It will now only send 10% of the transactions to the sentry server. Additionally, we have added `traceSampleRate` as a deployment variable for better control. We can now manually change its value in Openshift and shuffle pods instead of making a release every time we want to change its value.

Furthermore, we have put the sentry exception handler in a try/catch block to track the number of errors we receive with this implementation. However, if we receive a lot of emails, we might want to remove the handler from the `try` block.

### DIFF
--- a/bc_obps/.env.example
+++ b/bc_obps/.env.example
@@ -16,6 +16,8 @@ ENVIRONMENT=develop
 BYPASS_ROLE_ASSIGNMENT=<True/False>
 # Sentry
 SENTRY_DSN=your_sentry_dsn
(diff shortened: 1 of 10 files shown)


pbastia (approved):
Looks good! One small comment about the dev setup

pbastia:
...and needs rebase :)

PR #1248, opened 2024-03-27
Setup Vitest and update broken tests

marcellmueller:
Implements #1176

Experimenting with [vitest](https://vitest.dev/) to replace Jest. It's a faster, modern testing framework that supports ES6 modules and hot reloading. This came as a result of problems fixing our Jest tests which are broken in part due to es6 module transpilation issues.

To test this first run `yarn` to install vitest packages and run `yarn test`. The hot reloading works great, it will run whichever tests that are in the file you saved.

When it came to fixing the broken tests there were 3 tests which were all server components. This required various mocks which I tried to make reusable. This was more challenging than I thought due to the way Jest/Vitest hoists mocks on runtime. I ended up adding some global mocks in `/tests/setup/globals.ts`. To change the return value of these mocks you can import and use them in tests like this: 

```
import { useRouter } from `@/tests/setup/mocks`;

useRouter.mockReturnValue({
    query: {
      formSection: "3",
      operation: "test-operation-id",
    },
    replace: vi.fn(),
});
```

Another thing to note is previously we were using `jest-fetch-mock` to mock our fetch requests. I had originally switched it to [vitest-fetch-mock](https://www.npmjs.com/package/vitest-fetch-mock) though while these libraries work great we were still getting warnings in our console from `actionHandler`.

Now to mock our fetch requests import `actionHandler` from `@/tests/setup/mocks` and mock a return value. There is a bit more context to this in the comments and updated documentation. 

Thankfully the Jest docs didn't need to be updated much as Vitest is so similar. I checked and vitest supports everything in our docks ie: `DEBUG_PRINT_LIMIT`, updating snapshots with `-u` flag etc. 

As discussed in our dev team agreement session we would leave our broken Jest setup for now incase we decide to switch back. I have created a tech debt ticket here as a reminder: 
https://github.com/bcgov/cas-registration/issues/1260


client/app/components/datagrid/DataGrid.tsx
```diff
@@ -146,10 +146,10 @@ const DataGrid: React.FC<Props> = ({
           "& .MuiDataGrid-columnHeaders": {
             backgroundColor: BC_GOV_BACKGROUND_COLOR_BLUE,
           },
-          "& .MuiDataGrid-columnHeader:first-child": {
+          "& .MuiDataGrid-columnHeader:first-of-type": {
             borderLeft: "none",
           },
-          "& .MuiDataGrid-columnHeader:last-child": {
+          "& .MuiDataGrid-columnHeader:last-of-type": {
```
marcellmueller, line 152: Changed these to clean up warning when running tests: 

`The pseudo class ":first-child" is potentially unsafe when doing server-side rendering. Try changing it to ":first-of-type".`

client/app/components/form/widgets/SelectWidget.tsx
```diff
@@ -51,12 +51,16 @@ const SelectWidget: React.FC<WidgetProps> = (props) => {
...
         value={value || ""}
-        id={id}
         disabled={disabled || readonly}
         name={id}
         onChange={handleChange}
         sx={styles}
+        inputProps={{
+          id: id,
+          "aria-label": id,
+        }}
```
marcellmueller, line 63: Fixed an error that revealed itself using `getByLabel`: 
`Found a label with the text of: /Operation Type+/i, however the element associated with this label (<div />) is non-labellable`

client/e2e/poms/operation.ts
```diff
@@ -136,7 +136,7 @@ export class OperationPOM {
     // Fill out the operation information form
     await this.page.getByLabel("Operation Name*").fill("Test Operation Name");
 
-    await this.page.locator("#root_type").click();
+    await this.page.locator("#root_type_select").click();
```
marcellmueller, line 139: Had to update this due to fixing the incorrect label id in `SelectWidget`.

client/tests/setup/mocks.ts
```diff
@@ -0,0 +1,21 @@
+// To return a different value for each test, you can import this file and use mockReturnValue
+// to set the return value for the function you want to mock using mockReturnValue
+// or mockReturnValueOnce. Example:
```
BCerki, line 3: I don't understand this part--the example doesn't show how to do different values for each test/resets

client/tests/setup/SessionProviderMock.tsx
```diff
@@ -0,0 +1,7 @@
+// Needed to put this in a tsx file since .ts files are not allowed to have jsx
```
BCerki, line 1: If you haven't already tried this, can we put all the mocks in a tsx file to keep them together?

client/tests/setup/mocks.ts
```diff
@@ -0,0 +1,21 @@
+// To return a different value for each test, you can import this file and use mockReturnValue
+// to set the return value for the function you want to mock using mockReturnValue
```
BCerki, line 2: I find this comment a bit unclear--the example shows how to use mockReturnValue, but it doesn't seem to explain how to  change the value for each tests. I see a reset in the `beforeEach`; maybe mention that?

client/tests/routes/operations/Operations.test.tsx
```diff
@@ -1,64 +1,54 @@
...
+      row_count: 2,
+    });
     render(await Operations());
+
     // Check if the grid of mock data is present
-    await expect(screen.getByText(/Operation 1/i)).toBeVisible();
-    await expect(screen.getByText(/Operation 2/i)).toBeVisible();
+    expect(screen.getByText(/Operation 1/i)).toBeVisible();
+    expect(screen.getByText(/Operation 2/i)).toBeVisible();
     // temporarily commented out because render only renders half the grid
```
BCerki, line 48: Do we still need this comment? I think the whole grid must be rendering now if we can assert the start registration link is visible

client/tests/components/form/OperationsForm.test.tsx
```diff
@@ -1,184 +1,173 @@
...
+      replace: vi.fn(),
+    } as Router);
+    useParams.mockReturnValue({
+      formSection: "1",
+      operation: "create",
+    } as QueryParams);
+
+    render(<OperationsForm schema={testOperationSchema} formData={{}} />);
+
+    // Test for the Operation Information form heaer
```
BCerki, line 88: typo

client/tests/components/form/OperationsForm.test.tsx
```diff
@@ -1,184 +1,173 @@
...
+    expect(screen.getByTestId("field-template-label")).toBeVisible();
     // Operation Name
-    expect(screen.getByLabelText(/Operation Name+/i)).toBeInTheDocument();
+    expect(screen.getByLabelText(/Operation Name+/i)).toHaveValue("");
```
BCerki, line 91: Why are we expecting "" as a value? Wouldn't that mean it isn't showing up correctly?

client/tests/setup/mocks.ts
```diff
@@ -0,0 +1,21 @@
...
+//
+// useRouter.mockReturnValue({
+//  query: {
+//      formSection: "3",
+//      operation: "test-operation-id",
+//    },
+//    replace: vi.fn(),
+//  });
+
+const actionHandler = vi.fn();
```
BCerki, line 15: These are going to be super convenient for mocking; thanks for setting these up!

BCerki (commented):
This looks great! Thanks for doing all the setup and the helpers. And loving that hot reload.

client/tests/components/form/OperationsForm.test.tsx
marcellmueller: This is the `renders the empty OperationsForm when no formData is passed` test. Using `toBeNull()` returns and error, same with `toHaveValue(undefined)`. Though the undefined value works on plain text fields, I think it's likely something with how some of the MUI widgets handle empty fields.

client/tests/setup/mocks.ts
marcellmueller: Thanks. This setup looks fairly simple but it took me a lot of time and trial and error to figure out how to handle this and work my way through mocking errors 🙈

client/tests/setup/mocks.ts
marcellmueller: Thanks, I think that comment was slightly outdated since I initially had some default values being returned from the functions in this file. I decided that it would cause more trouble than help so I just made them plain `vi.fn()`.

Anyways I've updated this comment and example a bit, hopefully it's more clear if not let me know what would be helpful.

BCerki (approved):
:cartwheeling:

PR #1249, opened 2024-03-27
Create deploy script

BallardRobinett:
Modifying the deploy process to run a bash script instead of having the commands coded into the shipit.yml file. I was unable to create another stack in shipit to test the branch on, so this is the test, on giraffe-develop.


joshgamache (approved):
Looks good to me, let's see how ShipIt handles it!

Issue #843, opened 2024-02-07
e2e test\workflow\bceidbusiness\industry_user_admin


shon-button:
#### Description:

Tests of `\bceidbusiness\industry_user_admin` [workflow](https://miro.com/app/board/uXjVN6uBj9Y=/)

**Page by Roles Tests**
- [x] Cannot navigated to restricted app routes (**NOTE:** tested in `client/e2e/pages/authenticated/routes/page.spec.ts`)
- [x] Login redirected to Dashboard page (**NOTE:** tested in `client/e2e/pages/onboarding/home/page.spec.ts`)
- [x] Profile page submit required data (**NOTE:** tested in `client/e2e/pages/authenticated/dashboard/profile/page.spec.ts`)
- [x] Dashboard design: "My Operator", "Operations", and "Report a Problem"  (**NOTE:** tested in `client/e2e/pages/authenticated/dashboard/page.spec.ts`)

**WIP**

**Tests using storageState for `industry_user_admin`:**

- [x] clicking on "Operations" tile should redirect them to table of all existing operations associated with their Operator only, including operations with a status of "Draft" or "Not Started"
- [x]  is able to click "View Details" button for any operation in the table
- [x]  is able to click "Add New Operation" from the Operations page and create a new Operation
- [x]  is able to click on "My Operator" tile, can see readonly data about the Operator
- [x]  from "My Operator" tile, is able to click the "Edit information" button and edit data fields (at the moment, can edit any field) about the Operator and submit the updated data successfully

andrea-williams:
Closing this - we don't have an `industry_user_admin` role anymore

PR #1255, opened 2024-03-28
test: add extra assertion to fix flakey breadcrumb screenshot tests

marcellmueller:
Follow up to #843 as I introduced some flakey screenshots in that PR.

- Add assertions to make sure the `Operation name` breadcrumb is visible before taking a screenshot as it takes a second to appear.
- Add `created_at` values to ensure the Operations table is always in the same order
- Removed the wait added to fix the Firefox spurious diff for the `Add a new Operator - required errors` spurious diff. It made it happen less but it still popped up sometimes and was the only diff that showed when I pushed the fixtures updates. I am going to propose we add it to the `Permentantly ignored diffs` as suggested in the [Happo docs here](https://docs.happo.io/docs/ignoring-diffs). We test in 4 different browsers so losing this screenshot in Firefox should be safe. If you look in the sidebar to the left of the Happo repo report you can see some similar diffs that were auto ignored. 
- Looks like there is one spurious diff. I think this one can get added to the ignored diffs, not sure why it wasn't automatically added as it's very similar to the two that Happo auto added.

### DIFF
--- a/client/e2e/workflows/bceidbusiness/industry_user.spec.ts
+++ b/client/e2e/workflows/bceidbusiness/industry_user.spec.ts
@@ -153,9 +153,6 @@ test.describe("Test Workflow industry_user", () => {
       selectOperatorPage.buttonSubmit,
     );
 
-    // Add short timeout to mitigate the Firefox text rendering issue causing spurious screenshot failures
-    await page.waitForTimeout(500);
-
     pageContent = page.locator("html");
     await happoPlaywright.screenshot(page, pageContent, {
       component: "Add a new operator",
--- a/bc_obps/registration/fixtures/mock/operation.json
+++ b/bc_obps/registration/fixtures/mock/operation.json
@@ -13,7 +13,8 @@
       "swrs_facility_id": 1001,
       "bcghg_id": "23219990005",
       "regulated_products": [],
-      "status": "Not Started"
+      "status": "Not Started",
+      "created_at": "2024-2-01T15:27:00.000Z"
     }
   },
   {
@@ -27,7 +28,8 @@
       "opt_in": false,
       "bcghg_id": "23219990004",
       "regulated_products": [],
-      "status": "Not Started"
+      "status": "Not Started",
+      "created_at": "2024-2-02T15:27:00.000Z"
     }
   },
   {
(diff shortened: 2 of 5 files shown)


client/e2e/workflows/bceidbusiness/industry_user.spec.ts
```diff
@@ -153,9 +153,6 @@ test.describe("Test Workflow industry_user", () => {
       selectOperatorPage.buttonSubmit,
     );
 
-    // Add short timeout to mitigate the Firefox text rendering issue causing spurious screenshot failures
```
BCerki, line 156: Did you already adding an assertion? Could just check the url or something to slow it down

BCerki (commented):
Re the happo spurious diffs, have we already got the threshold set to where we want it? If yes, then I think ignoring in one browser is fine. If not, maybe we could try adjusting threshold a bit first. When I was having trouble with diffs in the Button component library, the maintainer showed me how to change the threshold and that helped a lot. It's safer than ignoring because if there's ever a bit change, it'll still show up as a diff

marcellmueller:
> Re the happo spurious diffs, have we already got the threshold set to where we want it? If yes, then I think ignoring in one browser is fine. If not, maybe we could try adjusting threshold a bit first. When I was having trouble with diffs in the Button component library, the maintainer showed me how to change the threshold and that helped a lot. It's safer than ignoring because if there's ever a bit change, it'll still show up as a diff

I think the issue with increasing the diff threshold for this is that it would likely have to be increased to hundreds of pixels looking at an example at the top of [this run]:(https://happo.io/a/336/p/1755/compare/23fb66cf3cfba6ab69a2fce30639c1b2ad320ff1/30981fcdea6df3f9e73af00091eea569e029e645)

Screenshot 2024-04-03 at 8 05 10 AM

client/e2e/workflows/bceidbusiness/industry_user.spec.ts
marcellmueller: Yeah I tried a few things and did a timeout as a last resort. Shon suggested adding it to ignored diffs though I was trying to avoid that, and the timeout works _most_ of the time, though I think if the CI machine is running really slow it still pops up sometimes.  
  
I suspect another reason that our form screenshots seem to be so vulnerable to spurious diffs is also because RJSF rerenders on every update.

BCerki (approved):
Just needs a rebase

Issue #1062, opened 2024-03-11
Do #952 and #1061 for one endpoint to validate new API structure


dleard:
This point of this card is to test our new API / service structure in #952 & #1061 by doing the work of both cards for one endpoint. This way we can validate that the new structure is going to work for us before applying it to all of our endpoints (minimizing work if we run into problems and want to re-think our strategy).

Suggest using user_operator since it is one of the larger endpoints & will cover the most cases.

- [ ] user_operator has been split as described in #952
- [ ] user_operator has a service layer as described in #1061
- [ ] assess validity & efficiency of the new structure

PR #1259, opened 2024-04-02
1062 2 api refactor example

BCerki:
card: https://github.com/bcgov/cas-registration/issues/1062

Notes for the reviewer:
- docs explaining the separation decisions in the developer guidelines
- endpoints (files in /api) now organized by endpoint (e.g. user_operator/operator contains all methods that use this route)
- actual logic is in service files and there are currently three
1. database access  (multiple files, to ask how to do the init thing)
2. request access service
3. user-operator-y one that needs a better name
- all endpoints should have the handle_http decorator _underneath_ the authorize. (Otherwise, it will catch auth errors.) This means that I had to update some of the authorization endpoint test setup, because if setup is only partial, the test will fail on the http error instead of the auth one.
- all db services that change data (so anything except get) should have the transaction.atomic decorator
- removed `select-operator` from `/select-operator/user-operator/update-status`
- removed `select-operator` from `/select-operator/user-operator/{uuid:user_operator_id}`
- added some new endpoint tests when I noticed we were missing coverage
- restrict the update-status endpoint
- some new tech debt cards I plan to make:
1. What to do about getting the current user. To properly separate things, in the endpoints we pass the user_guid to the services, and then if we need the full user in the service, we look it up using the guid. This feels wasteful because we already have the full user object in the endpoints (from the middleware). This double lookup may cause performance issues.https://github.com/orgs/bcgov/projects/122/views/2?filterQuery=user&pane=issue&itemId=59071266
2. Rename some of the more mysterious endpoints: https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=56218560
3. Can we use url name instead of url in the actionHandler (not sure if we can do this from the front end, but it would save us having to update files when the routes change) https://github.com/orgs/bcgov/projects/122/views/2?filterQuery=&pane=issue&itemId=59071560
4. endpoints related to user_operator should start with `user_operator` (e.g. @router.get(
    "/is-approved-admin-user-operator/{user_guid}" becomes @router.get(
    "user_operator/is-approved-admin-user-operator/{user_guid}": https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=56218560
5. operator-related endpoints should be under operator instead (operator-access-declined, operator_has_admin are possible candidates. `user-operator/operator` creates/edits both the user_operator and operator so not sure where that one should go)--addressed in this PR
6. refactor endpoint tests into separate files in: https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=54068291
7. use zoneinfo instead of pytz: https://github.com/orgs/bcgov/projects/122/views/2?filterQuery=&pane=issue&itemId=59070949
8. move the parent operators helpers into a service https://github.com/orgs/bcgov/projects/122/views/2?filterQuery=split+the+logic&pane=issue&itemId=59073618
9. use @classmethod and `cls` in the service layer classes (@Sepehr-Sobhani this made a couple tests fail so decided to leave it out of this PR)


bc_obps/service/data_access_service/operator_service.py
```diff
@@ -0,0 +1,23 @@
+from registration.models import Operator
```
BCerki, line 1: Bri figure out how to do the model thing Pierre was talking about

bc_obps/service/user_operator_service.py
```diff
@@ -0,0 +1,248 @@
+from uuid import UUID
```
BCerki, line 1: b better name

bc_obps/service/data_access_service/user_operator_service.py
```diff
@@ -0,0 +1,62 @@
...
+
+        pending_user_operators = UserOperator.objects.filter(
+            operator=operator, role=UserOperator.Roles.ADMIN, status=UserOperator.Statuses.PENDING
+        )
+        pending_users = User.objects.filter(user_guid__in=pending_user_operators.values('user_id'))
+
+        return pending_users
+
+    def get_an_operators_user_operators_by_user_guid(user_guid: UUID):
+        user = UserDataAccessService.get_user_by_guid(user_guid)
```
BCerki, line 40: To properly separate things we just pass the user_guid to the service layer and then if we need the full user, we look it using the guid. This feels wasteful because we already have the full user object from the middleware and therefore in the endpoint. So what happens is in the endpoint we pull the guid off the user, pass that into a service, and then in the service we look up the user object again

bc_obps/handle_exception.py
```diff
@@ -0,0 +1,20 @@
+"""
+Module: handle_exception.py
+Description: This module handles http exceptions.
+"""
+from django.http import Http404
+from django.core.exceptions import ValidationError, ObjectDoesNotExist
+from registration.utils import generate_useful_error
```
BCerki, line 7: This fails pre-commit in CI but is okay locally, thoughts?

Sepehr-Sobhani: I believe adding a dot before `registration.utils` will fix the issue. Once we tackle ticket 114 or 176 we no longer face these kinds of issues.

docs/developer-guide.md
Sepehr-Sobhani, line 1: Thanks for adding this documentation 🙏

bc_obps/registration/api/utils/current_user_utils.py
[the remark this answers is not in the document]
BCerki: No strong feelings, maybe one of the other reviewers will have some feelings, if not, I'll change it

bc_obps/service/data_access_service/operator_service.py
```diff
@@ -0,0 +1,23 @@
...
+from uuid import UUID
+
+
+class OperatorDataAccessService:
+    def get_operator_by_id(operator_id: UUID):
+        return Operator.objects.get(id=operator_id)
```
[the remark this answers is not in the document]
BCerki: I did this for maximum separation of concerns, and also in case we need to optimize this endpoint and it gets more complicated

bc_obps/service/data_access_service/operator_service.py
```diff
@@ -0,0 +1,23 @@
...
+
+    def get_operator_by_user_operator_id(user_operator_id: UUID):
+        # brianna why circular import??
```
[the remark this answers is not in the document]
BCerki: A note-to-self to remember to ask the reviewers if circular imports are a problem (I would have thought separating concerns would have helped with circular imports)

bc_obps/service/data_access_service/user_operator_service.py
```diff
@@ -0,0 +1,62 @@
...
+    Operator,
+    UserOperator,
+)
+
+
+class UserOperatorDataAccessService:
+    def get_user_operator_by_id(user_operator_id: UUID):
+        # brianna this is the optimized one?
+        return UserOperator.objects.select_related('operator').get(id=user_operator_id)
+        # return UserOperator.objects.get(id=user_operator_id)
```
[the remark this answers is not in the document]
BCerki: Is there a reason not to always use the optimized `UserOperator` instance? (Brianna doesn't know about optimization)

bc_obps/service/data_access_service/user_operator_service.py
BCerki: I've passed the `user_guid` instead of the `user` instance for maximum separation of concerns (I'm avoiding doing anything even remotely db-related in the endpoints). I agree that this is an extra lookup though, so I'm going to make a tech debt ticket to explore alternatives (and block it on the endpoint optimization--I don't know how much effect the extra lookup is having at the moment).

bc_obps/service/application_access_service.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def request_access(operator_id: UUID, user_guid: UUID):
+        if ApplicationAccessService.is_user_eligible_to_request_access(operator_id, user_guid):
+            # Making a draft UserOperator instance if one doesn't exist
+            user_operator = UserOperatorDataAccessService.get_or_create_user_operator(user_guid, operator_id)
+
+        return {"user_operator_id": user_operator.id, "operator_id": user_operator.operator.id}
+
+    def request_admin_access(operator_id: UUID, user_guid: UUID):
+        if ApplicationAccessService.is_user_eligible_to_request_admin_access(user_guid, operator_id):
```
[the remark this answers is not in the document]
BCerki: As discussed, this wasn't straightforward, so it's added to the make-tech-debt cards list in the PR description

bc_obps/registration/tests/test_utils.py
```diff
@@ -1,16 +1,15 @@
+from service.application_access_service import ApplicationAccessService
```
pbastia, line 1: Looks like we can split a bunch of these tests into the individual test file for the services

bc_obps/service/data_access_service/user_operator_service.py
```diff
@@ -0,0 +1,52 @@
...
+    def get_an_operators_user_operators_by_user_guid(user_guid: UUID):
+        user = UserDataAccessService.get_user_by_guid(user_guid)
+        operator = (
+            UserOperator.objects.select_related("operator")
+            .exclude(status=UserOperator.Statuses.DECLINED)
+            .get(user=user.user_guid)
+            .operator
+        )
+        user_operator_list = UserOperator.objects.select_related("user").filter(
+            operator_id=operator, user__business_guid=user.business_guid
```
pbastia, line 38: `operator_id=operator` will always be false since they're of different types

bc_obps/service/data_access_service/user_operator_service.py
pbastia: I think the confusion here is that we don't need the whole `User` object. The only missing piece is the business_guid that we could fetch individually? 

Either that, or we can ask for an entire `User` object as a parameter (since that is an object the Data Access Layer controls, this is not necessarily a bad thing. It just puts the burden on the caller to figure out how to get one and we conveniently have it in a middleware)

bc_obps/service/handle_addresses_service.py
```diff
@@ -0,0 +1,38 @@
+from registration.api.utils.operator_utils import AddressesData
+from registration.models import Address
+
+
+class HandleAddressesService:
```
pbastia, line 5: I think we could use a more descriptive name than "Handle".
Maybe an `AddressesService` class with a `upsert_addresses_from_data(...)` method?

I would also add a bit of documentation, the phsical-, mailing- and prefix parameters are pretty confusing

pbastia (changes requested):
A few small comments. This is amazing!! 💯 💯

bc_obps/registration/api/user_operator/is_approved_admin_user_operator.py
```diff
@@ -0,0 +1,25 @@
...
+
+from registration.models import (
+    UserOperator,
+)
+from registration.api.custom_codes_4xx import custom_codes_4xx
+from registration.api.utils.current_user_utils import get_current_user_guid
+
+
+@router.get(
+    "/is-approved-admin-user-operator/{user_guid}",
```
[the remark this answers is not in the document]
BCerki: Yes! There is a separate ticket for renaming endpoints: https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=56218560 so I'll leave it as-is here unless you think it's important to do now

bc_obps/registration/tests/test_utils.py
BCerki: Also have a separate tech debt ticket for this: https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=54068291 (there's a list of tech debt tickets in the PR description; this PR is already pretty big)

bc_obps/service/data_access_service/user_operator_service.py
BCerki: :scream:

bc_obps/service/data_access_service/operator_service.py
BCerki: @pbastia , I didn't understand what you meant by this comment in the other PR: `Maybe if we want to namespace a little better like DataAccessService.Operators.create(...):), this would be easy with python's folder/module structure`. Have I got it set up like you were imagining?

bc_obps/service/data_access_service/user_operator_service.py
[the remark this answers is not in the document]
Sepehr-Sobhani: Yes, using `select_related` is not always useful: https://www.webtechnika.pl/en/news/7/django-select_related-to-use-or-not-to-use-that-is/#:~:text=How%20not%20to%20use%C2%A0select_related

bc_obps/service/data_access_service/user_operator_service.py
Sepehr-Sobhani: Yes, most of the time, we don't need the entire `User` object. However, we need to be cautious about when to pass the user object and when to pass only the `user_guid`. This is because adding an extra query to fetch the user could cause performance issues, given the number of functions that rely on the user and their roles. This was the main reason we cached the user instance in the middleware to avoid additional database calls.

bc_obps/service/data_access_service/user_operator_service.py
BCerki: This is interesting because this wasn't a refactor whoopsie; it's always been like this.

bc_obps/service/addresses_service.py
```diff
@@ -0,0 +1,51 @@
...
+        address_data: This is the address information including street, postal code, etc.
+        physical_address_id: This is the id of the physical address record if an address record already exists.
+        mailing_address_id: This is the id of the mailing address record if an address record already exists. If physical and mailing address are the same, this id is the same as the physical_address_id
+        prefix: Sometimes address data contains a prefix. For example, parent operator address keys contain `po` (e.g. po_mailing_street_address). This function strips out the prefix if included.
+
+    Returns the physical and mailing address information.
+
+    """
+
+    def upsert_addresses_from_data(address_data: AddressesData, physical_address_id, mailing_address_id, prefix=""):
```
pbastia, line 19: Eventually the `AddressesData` structure should probably live with this service (it's part of the API  of this service) but that can be a later refactor

pbastia (approved):
Looks great!! Awesome work @BCerki

bc_obps/service/addresses_service.py
BCerki: https://github.com/orgs/bcgov/projects/122/views/2?filterQuery=split+the+logic&pane=issue&itemId=59073618

PR #1268, opened 2024-04-03
Giraffe-develop merge

pbastia:
Making this a long-standing draft PR to follow the divergence between those 2 branches.
Conflicts, rebases and CI issues will be highlighted by this process


.github/workflows/test.yaml
```diff
@@ -172,6 +172,71 @@ jobs:
...
+        run: |
+          npx nx container reporting --skip-nx-cache
+        # Temp fix
+        # https://github.com/docker/build-push-action/issues/252
+        # https://github.com/moby/buildkit/issues/1896
+      - name: Move cache
+        run: |
+          rm -rf /tmp/.buildx-cache
+          mv /tmp/.buildx-cache-new /tmp/.buildx-cache
+
```
joshgamache, line 239: This code will need to be addressed, as it is flagged to only run on merge to `giraffe-develop`. The makefiles/helm templates will be what actually pulls the image for deploy, so it should be fine.

Makefile
```diff
@@ -64,6 +64,30 @@ install:
 		helm upgrade $(HELM_OPTS) $(CHART_INSTANCE) $(CHART_DIR); \
 	fi;
 
+.PHONY: install_giraffe
+install_giraffe: ## Installs the helm chart on the OpenShift cluster
+install_giraffe: check_environment
+install_giraffe:
+install_giraffe: GIT_SHA1=$(shell git rev-parse HEAD)
+install_giraffe: IMAGE_TAG=$(GIT_SHA1)
+install_giraffe: NAMESPACE=$(CIF_NAMESPACE_PREFIX)-tools
```
dleard, line 73: Are we planning to continue deploying this to -tools for now?

shipit.yml
```diff
@@ -6,8 +6,9 @@ review:
 
 deploy:
   override:
-    - make install:
-        timeout: 5000
+    - |
+      bash deploy.sh
```
dleard, line 10: Is this deploy.sh still necessary? The shell script checks if branch=giraffe-develop, which won't be used anymore & therefore won't deploy at all. I think maybe we should change our environments in our stacks in shipit to be registration_<environment> /  reporting_<environment> and make specific shipit.<app>_<environment>.yaml files.

Makefile
pbastia: I'd be tempted to? Unless we're fine having everything in the same place, and we have the resources to do that

shipit.yml
pbastia: We can probably revert this yes

helm/cas-registration/templates/reporting/deployment.yaml
```diff
@@ -0,0 +1,38 @@
+{{- if (.Values.reportingFrontend.enabled) }}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "cas-registration.fullname" . }}-reporting-frontend
```
dleard, line 5: This is going to read as `cas-registration-reporting-frontend`.Perhaps we should change the chart name to something more generic?

dleard: ah I see there is an override in the values.yaml for this :)

shipit.yml
dleard: We might be able to do something with the shipit stack environments here to coordinate which FE app we want to deploy

helm/cas-registration/templates/NetworkPolicy/backendPostgresIngress.yaml
```diff
@@ -11,9 +11,9 @@ spec:
   - from:
     - podSelector:
         matchLabels:
-          release: {{ .Release.Name }}
+          release: {{ include "cas-registration.fullname" . }}
```
dleard, line 14: I think this probably needs to stay as .Release.Name since the matcher is "release", otherwise it may not match properly

dleard (approved):

PR #1270, opened 2024-04-03
811 myself pill

BCerki:
card: https://github.com/orgs/bcgov/projects/122/views/2?filterQuery=&pane=issue&itemId=52081252

### DIFF
--- a/client/app/components/datagrid/UserOperatorDataGrid.tsx
+++ b/client/app/components/datagrid/UserOperatorDataGrid.tsx
@@ -5,7 +5,7 @@ import DataGrid from "@/app/components/datagrid/DataGrid";
 import { OperatorStatus, UserOperatorRoles } from "@/app/utils/enums";
 import AccessTypeColumnCell from "@/app/components/datagrid/cells/AccessTypeColumnCell";
 import ChangeUserOperatorStatusColumnCell from "@/app/components/datagrid/cells/ChangeUserOperatorStatusColumnCell";
-import { statusStyle } from "@/app/components/datagrid/helpers";
+import { StatusStyleColumnCell } from "./cells/StatusStyleColumnCell";
 
 const columns: GridColDef[] = [
   {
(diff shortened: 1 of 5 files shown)


client/app/utils/users/adminUserOperators.ts
```diff
@@ -49,10 +50,9 @@ export async function processExternalDashboardUsersTileData() {
   const selfIndex = transformedTileData.findIndex((userOperator) => {
     return userOperator.user.user_guid.replace(/-/g, "") === uid;
   });
-
   // Ensure selfIndex is within the valid range before modifying the array
   if (selfIndex !== -1 && selfIndex < transformedTileData.length) {
-    transformedTileData[selfIndex].status = Status.MYSELF;
+    transformedTileData[selfIndex].isUserCurrentUser = true;
```
BCerki, line 55: In client/app/components/datagrid/cells/ChangeUserOperatorStatusColumnCell.tsx, we have a `buttonsToShow` function that shows the approve/reject/undo buttons based on the statuses. This means that as soon as I changed "Myself" to "Approved", the buttons showed up. 

This file is the only place we have access to the user_guid (this function strips out the user_guid and anything else that shouldn't show in the table). We needed some way to keep track of who the current user is besides the "Myself" status, so I've added an `isUserCurrentUser` key. This key is passed to the grid but is hidden in the table. I'm not super happy with this solution, but wasn't able to come up with anything better. Thoughts very welcome!

client/app/utils/enums.ts
```diff
@@ -46,7 +46,6 @@ export enum UserOperatorStatus {
 
 // user statuses
 export enum Status {
-  MYSELF = "Myself",
```
BCerki, line 49: I would have liked to change this to MYSELF = "Approved", but we use the status text to decide whether or not to show approve/reject buttons, more details below

client/app/utils/users/adminUserOperators.ts
marcellmueller: Wouldn't it be far simpler to just change the name in the status pill itself? In `datagrid/helpers.tsx`:

```
  const status = (params.value = Status.MYSELF
    ? Status.APPROVED
    : (params.value as string));
```

As an aside, I think the `statusStyle` cell in `helpers` should be moved to its own file in `/cells` to be easier to find 😅

marcellmueller (approved):
🥇

PR #1095, opened 2024-03-14
505 industry user e2e

BCerki:
card: https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=49836206

Note: sonarcloud fail is in a test endpoint, so I'll ignore unless the reviewer feels otherwise

PR #1005, opened 2024-03-01
test: create Operations POM and helpers

BCerki:
card: https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=52567202

Still to dos:
- something is going wrong with the operations table in CI: 
image I couldn't confirm if it works locally for IDIR users, but it does work locally for the `industry_user`: 
image
So I think it's CI-specific
- I couldn't get past the operations table fail, so not sure if any of the test beyond that works

Issue #845, opened 2024-02-07
e2e test\workflow\idir\cas_analyst


shon-button:
#### Description:

Tests of `\idir\cas_analyst` [workflow](https://miro.com/app/board/uXjVN6uBj9Y=/)

**Page by Roles Tests**
- [x] Cannot navigated to restricted app routes (**NOTE:** tested in `client/e2e/pages/authenticated/routes/page.spec.ts`)
- [x] Login redirected to Dashboard page (**NOTE:** tested in `client/e2e/pages/onboarding/home/page.spec.ts`)
- [x] Profile page submit required data (**NOTE:** tested in `client/e2e/pages/authenticated/dashboard/profile/page.spec.ts`)
- [x] Dashboard design:  " "Operators", "Operations", "Report a Problem" tiles (**NOTE:** tested in `client/e2e/pages/authenticated/dashboard/page.spec.ts`)

**Tests using storageState for `cas_analyst`:**
- [x] Operator tile workflow
    - [ ]  Dashboard tile click
    - [ ] Operator grid ux
        - [ ]  grid view
        - [ ]  grid table
        - [ ]  grid table columns
        - [ ]  grid column values
    - [ ] View Details click
    - [ ] Details form ux
    - [ ] Workflow New Operator, Pending status, Deny Action
    - [ ] Workflow New Operator, Pending status, Approve Action
    - [ ] Workflow Existing Operator, Pending status, Deny Action
    - [ ] Workflow Existing Operator, Pending status, Approve Action
- [x] Operation tile workflow
    - [ ]  Dashboard tile click
    - [ ] Operation grid ux
        - [ ]  grid view
        - [ ]  grid table
        - [ ]  grid table columns
        - [ ]  grid column values
    - [ ] View Details click
    - [ ] Details form ux
    - [ ] Workflow Pending status,Request Changes\Unrequest Changes
    - [ ] Workflow Pending status,Approve Action
    - [ ] Workflow Pending status, Decline Action
    - [ ] Workflow  Approved Preview the Statutory Declaration PDF

andrea-williams:
De-prioritizing this - for MVP, all internal users will have `cas_admin` roles.

shon-button:
## 🚧 
**Issue**
The following Playwright tests work locally but, fail in our CI environment.

1. `client/e2e/poms/operations.ts\await expect(this.messageOperationApproved).toBeVisible();`

**Local:**
[...]

PR #1275, opened 2024-04-04
refactor: e2e poms, specs, utils

shon-button:
## 📌 

This PR is an iteration on the valuable working  contributions currently existing within our E2E pipeline. 👏 
- [e2e cas admin PR](https://github.com/bcgov/cas-registration/pull/1005) 
- [e2e industry_user PR](https://github.com/bcgov/cas-registration/pull/1095) 
- [e2e industry_user admin PR](https://github.com/bcgov/cas-registration/pull/1180) 
- 
## 🚧 

The changes in this branch aim to enhance adhering to Playwright Test specs and POMs best practices and principles of clean code.

1. If we look to Playwright documentation as the source of best practice guidance then we should try to abstract and encapsulate test logic in the higher-level Page Object Model API.  This abstraction will make test scripts more readable by reducing code logic in the specs and will allow for code sharing across test spec.  Principles applied, the POM should identify and encapsulate the following:
- Identify properties
- Identify locators
- Identify actions
- Identify assertions

2. For decisions on locator patterns, Playwright recommends using built in **locators**  that **prioritize user-facing attributes** and semantic attributes (such as role, label, or title).
- Semantic methods encourage best practices. As you’ll see in the passages that follow, the most durable selector patterns have helper methods.
- Semantic methods are typed, improving your IDE experience and alerting you to errors. Plain selector strings can only be debugged at runtime, resulting in more bugs.
- Semantic methods are chainable. This dramatically simplifies the queries themselves. It also results in better error messages when queries fail.

**These are the recommended built in locators.**
[page.getByRole()](https://playwright.dev/docs/locators#locate-by-role) to locate by explicit and implicit accessibility attributes.
[page.getByText()](https://playwright.dev/docs/locators#locate-by-text) to locate by text content.
[page.getByLabel()](https://playwright.dev/docs/locators#locate-by-label) to locate a form control by associated label's text.
[page.getByPlaceholder()](https://playwright.dev/docs/locators#locate-by-placeholder) to locate an input by placeholder.
[page.getByAltText()](https://playwright.dev/docs/locators#locate-by-alt-text) to locate an element, usually image, by its text alternative.
[page.getByTitle()](https://playwright.dev/docs/locators#locate-by-title) to locate an element by its title attribute.
[page.getByTestId()](https://playwright.dev/docs/locators#locate-by-test-id) to locate an element based on its data-testid attribute (other attributes can be configured).

[Locators are the central piece of Playwright's auto-waiting and retry-ability...](https://playwright.dev/docs/locators#introduction)

**Locate by role**
Playwright recommends prioritizing role locators to locate elements, as it is the closest way to how users and assistive technology perceive the page.

**Locate by test id**
Testing by test ids is the most resilient way of testing as even if your text or role of the attribute changes the test will still pass. QA's and developers should define explicit test ids and query them with [page.getByTestId()](https://playwright.dev/docs/api/class-page#page-get-by-test-id). However testing by test ids is not user facing. If the role or text value is important to you then consider using user facing locators such as [role](https://playwright.dev/docs/locators#locate-by-role) and [text locators](https://playwright.dev/docs/locators#locate-by-text).

**Examples:**
```
// Target Test IDs when you have control of the HTML
// E.G. `<section data-testid="delete-modal" />`
const modal = page.getByTestId('delete-modal');
// Unlike every other HTML attribute, there is never any reason for data-testid to change, making it the least brittle of all possible selectors.

// Target explicit or implied element roles
// "button" matches `<button>`, `<input type="button">`, or `<div aria-role="button">`
const button = page.getByRole('button', {hasText: 'Buy'});

// Target text HTML attributes that are unlikely to change
const input = page.getByLabel('Email');
const search = page.getByPlaceholder(/^search/i);
const image = page.getByAltText('Profile Picture');
const icon = page.getByTitle('Info', {exact: false});

// Or target elements by `innerText`, **when you can be sure it's stable**
const dialog = page.getByText(/^confirm/i);
```
**Key Points to Remember:**

Prioritize user-facing attributes: When possible, use locators that reflect how users interact with the page (roles, labels, text).
Strictness: Locators are strict and expect a unique element. Use page.getByRole().first(), etc. only if necessary, but focus on creating robust locators.
Filtering locators: Refine your locators using filter() to target specific elements within a list based on text content, presence of child elements, etc.
Iterating over lists: Use all() to get all elements matching a locator and then iterate using loops or evaluateAll().

3. Reduce code duplication (DRY principle):
- Repeated code blocks have been consolidated into single reusable blocks.
- Test redundancy has been trimmed as **Playwright performs a range of actionability checks on the elements before making actions to ensure these actions behave as expected**. It auto-waits for all the relevant checks to pass and only then performs the requested action. 
- Imported enums are preferred over static text to enhance code clarity and maintainability.

4. Optimize test execution speed:
- Parallel assertions using Playwright's Promise.all() function have been utilized to perform visibility checks simultaneously, improving test execution time.

## 📝 Notes

- Hopefully this PR adds value to our E2E pipeline emphasizing abstracting test logic to the poms, static text to the enums, and trying to ensure DRY code. Of course, the architecture not perfect, and further discussions for best practice and optimization might include locator methodology(ies), enums architecture, naming standards, etc. 

- Note: `Test Registration App / 📊 e2e report artifact` was not merging the blob files so to review e2e errors with the run condition of ` if: ${{ needs.e2e-tests.result == 'failure' }}`; so, I changed to `    if: ${{ always() }}` so to ensure a report on errors. This change produces a "failing" status when there are no blobs to merge i.e. no e2e errors. [Need to implement correct if condition ticket](https://github.com/bcgov/cas-registration/issues/1283)
- There are 2 FIXME assertions that need to be fix that I will handle in: [e2e cas_analyst](https://github.com/bcgov/cas-registration/issues/845)


client/e2e/workflows/bceidbusiness/industry_user.spec.ts
```diff
@@ -138,60 +131,54 @@ test.describe("Test Workflow industry_user", () => {
...
-    await expect(page.getByText(/Operator Information/i)).toBeVisible();
-
+    // 👉 Action create a new operator
+    await selectOperatorPage.clickAddOperator();
+    // 📷 Cheese!
     pageContent = page.locator("html");
     await happoPlaywright.screenshot(page, pageContent, {
       component: "Add a new operator",
       variant: "default",
     });
```
marcellmueller, line 143: I was looking through the Happo diffs and it looks like the majority of them are intended diffs due to various changes in the test. At the very top of the run is this diff which is problematic though, I think we need an assertion that the form is visible before taking a pic!

Screenshot 2024-04-05 at 11 54 53 AM

client/e2e/utils/enums.ts
```diff
@@ -60,18 +222,42 @@ export enum OperatorUUID {
   DEFAULT = "4242ea9d-b917-4129-93c2-db00b7451051",
 }
 
+// 📝 Status LOV Operation
+export enum OperationStatus {
+  PENDING = "Pending",
+  APPROVED = "Approved",
+  DECLINED = "Declined",
```
marcellmueller, line 229: nit: I don't think we need to write these out again, we have them in `@/app/utils/enums.ts`

bc_obps/registration/fixtures/mock/operator.json
```diff
@@ -255,8 +255,8 @@
       "contacts": [1],
       "status": "Draft",
       "is_new": true,
-      "verified_at": "2024-01-03 08:09:51.133556-08",
-      "verified_by": "00000000-0000-0000-0000-000000000002"
+      "verified_at": null,
+      "verified_by": null
```
marcellmueller, line 259: Out of curiosity, why set this to `null` rather than remove it?

client/e2e/pages/authenticated/dashboard/page.spec.ts
```diff
@@ -71,34 +76,46 @@ test.describe.configure({ mode: "serial" });
...
             // 🔍 Assert that the current URL ends with "/profile"
-            const profilePage = new ProfilePOM(page);
-            await profilePage.urlIsCorrect();
+            await new ProfilePOM(page).urlIsCorrect();
             break;
           default:
             // 🔍 Assert that the current URL ends with "/dashboard"
             await dashboardPage.urlIsCorrect();
-
+            // 📷 Cheese!
```
marcellmueller, line 98: Haha I love this

client/e2e/poms/dashboard.ts
```diff
@@ -36,68 +42,88 @@ export class DashboardPOM {
...
-    this.operatorsTileIndustry = page.locator("#My-Operator-link");
-    this.operationsTileIndustry = page.locator("#My-Operations-link");
     this.reportProblemLink = page.getByRole("link", {
-      name: "Report problems to GHGRegulator@gov.bc.ca",
+      name: TileTextDashboard.TILE_REPORT_PROBLEM,
     });
-    this.userAccessManagementTileIndustry = page.locator(
-      "#User-Access-Management-link",
+    this.selectOperatorTile = page.getByText(
+      TileTextDashboard.TILE_OPERATOR_SELECT,
```
marcellmueller, line 62: I don't feel too strongly about this but I wonder if we could somehow make a data-testid for each tile type that used this format and would be user independent eg: `tile-operator`, `tile-operation`, `tile-user` etc. That way we wouldn't have to worry about users or label changes.

client/e2e/workflows/idir/cas_admin.spec.ts
```diff
@@ -1,24 +1,113 @@
...
+});
+
 test.describe("Test Workflow cas_admin", () => {
   // 👤 run test using the storageState for this role
   const storageState = JSON.parse(process.env.E2E_CAS_ADMIN_STORAGE as string);
   // Note: specify storageState for each test file or test group, instead of setting it in the config. https://playwright.dev/docs/next/auth#reuse-signed-in-state
   test.use({ storageState: storageState }); // this will error if no such file or directory
-  test("Test Redirect to Dashboard", async ({ page }) => {
+
+  test("Operators Tile workflow", async ({ page }) => {
```
marcellmueller, line 38: This test is looking so nice and so clean 😍

marcellmueller (commented):
Incredible work here! 😍  This will make our e2e tests easier to write and maintain going forward and I think will provide a great example for the reporting app once they get there as well.

I just had a couple of very minor nits.

.github/workflows/test.yaml
```diff
@@ -299,7 +299,7 @@ jobs:
     name: 📊 e2e report artifact
     runs-on: ubuntu-latest
     needs: [e2e-tests]
-    if: ${{ needs.e2e-tests.result == 'failure' }}
+    if: ${{ always() }}
```
marcellmueller, line 302: A comment here might be useful since an `if: true` statement is redundant

bc_obps/registration/fixtures/mock/operator.json
shon-button: @marcellmueller capturing that we will defer to @Sepehr-Sobhani or @BCerki for answers to the api changes

client/e2e/poms/dashboard.ts
shon-button: @marcellmueller great discussion point for the team to weigh-in on: if the role or text value is important then consider using user facing locators OR if test resiliency is prioritized the consider using test ids.

bc_obps/registration/fixtures/mock/operator.json
BCerki: If it was me who put it in, I just didn't think it through, no reason

marcellmueller (approved):
Amazing work 😍

PR #1291, opened 2024-04-05
1211 database setup for email notification

Sepehr-Sobhani:
[ISSUE 1211](https://github.com/orgs/bcgov/projects/122/views/2?filterQuery=&pane=issue&itemId=57274383)

**NOTES**

- In order to run CHES emails locally, we need to obtain CHES credentials from the updated `OBPS backend ENV` file in 1Password.

- It is advisable to avoid sending actual emails while testing features to prevent spamming people. For instance, when testing the `send_admin_access_request_email` function, we may replace our email with `user_operator.user.email` to receive the real email and confirm its contents.

- To add more email templates and continue working on the Email Notification feature, we need to create a new migration file. This file should include the `subject`, `name`, and `body` of the new email notification. This process is similar to the migration file `0002_...` in the common app. After creating the new template, we can use it to generate new functions to create the email context and send the email, just like the `send_admin_access_request_email` function.

- To test submission confirmation state:
- Add CHES credentials to your `bc_obps/.env` file
- In `def request_admin_access` function replace your personal email with `user_operator.user.email` in `send_admin_access_request_email` function
- Log in as `bc-cas-dev-three` and request admin access to an operator that doesn't have an administrator already
- Look at your email :) You should receive something like below:
Screenshot 2024-04-16 at 4 13 12 PM

- To test Approve and Declined state:
- Add CHES credentials to your `bc_obps/.env` file
- in `def update_user_operator_status` function replace your personal email with `user_operator.user.email` in `send_admin_access_request_email` function
- Log in as `bc-cas-dev` and go to the user access management page and Decline/Approve admin requests
- Look at your email :) You should receive something like below:
Screenshot 2024-04-16 at 4 17 57 PM


Sepehr-Sobhani:
Replaced by: https://github.com/bcgov/cas-registration/pull/1413

PR #1292, opened 2024-04-05
Create deploy script

BallardRobinett:
card: https://github.com/bcgov/cas-reporting/issues/164

Creating a deploy process for the giraffe team to use on the ```giraffe-develop``` branch. 

A bash script ```deploy.sh``` is run from the ```shipit.yaml``` file, which checks if the current branch is ```giraffe-develop``` and runs a different make target to deploy to the CIF-tools namespace (c53ff1-tools). 

Also modified the ```nx-docker-build``` step in our github actions so that aan image is correctly built and pushed to ghcr in our CI and fixed a secret name being generated for the backend.


client/app/utils/getUserFullName.ts
pbastia, line 1: I don't think we should modify the registration code in place to fix typing issues

The build error comes from the fact that the next-auth `Session` interface has been redefined in `client/app/types/next-auth.d.ts`, and we can fix the build error by including this file in our `tsconfig.json` in the `include` section. I would also be as specific as possible, as this will need to be reverted when we integrate all the apps under the Nx structure.

bceirs/apps/reporting/tsconfig.json
```
...
    "include": [
      ... ,
      "../../../client/app/types/next-auth.d.ts"
    ]
```

helm/cas-registration/templates/backend/secret.yaml
```diff
@@ -9,7 +9,7 @@
 apiVersion: v1
 kind: Secret
 metadata:
-  name: {{ include "cas-registration.fullname" . }}-backend
+  name: {{ .Release.Name }}-backend
```
pbastia, line 12: we need to change the `deployment.yaml` file instead to use the helm named template

pbastia (changes requested):
Nice work! just a couple of things to avoid making changes to the registration code

bciers/apps/reporting/src/app/api/auth/[...nextauth]/route.ts
```diff
@@ -1,7 +1,5 @@
 import NextAuth, { NextAuthOptions } from 'next-auth';
-import { authOptions as regAuthOptions } from 'registration/api/auth/[...nextauth]/route';
-
-export const authOptions = regAuthOptions as NextAuthOptions;
```
Comment (pbastia) [inline bciers/apps/reporting/src/app/api/auth/[...nextauth]/route.ts:4]: Nx enforces the Next conventions and verifies that we're not exporting extra stuff from our routes files

bciers/yarn.lock
```diff
@@ -2887,19 +2887,10 @@
...
-  dependencies:
-    "@types/prop-types" "*"
-    "@types/scheduler" "*"
-    csstype "^3.0.2"
-
-"@types/react@18.2.33":
-  version "18.2.33"
-  resolved "https://registry.yarnpkg.com/@types/react/-/react-18.2.33.tgz#055356243dc4350a9ee6c6a2c07c5cae12e38877"
-  integrity sha512-v+I7S+hu3PIBoVkKGpSYYpiBT1ijqEzWpzQD62/jm4K74hPpSP7FF9BnKG6+fg2+62weJYkkBWDJlZt5JO/9hg==
+"@types/react@*", "@types/react@^18.2.33":
```
pbastia, line 2890: This is to match the registration app down to the minor version, to avoid type incompatibilities

mikevespi (approved):
Nice work!

bc_obps/reporting/migrations/0001_initial.py
```diff
@@ -17,7 +16,10 @@ class Migration(migrations.Migration):
                 ('id', models.BigAutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
                 ('title', models.CharField(db_comment='The title of the report', max_length=100)),
                 ('description', models.TextField(db_comment='The description of the report')),
-                ('created_at', models.DateTimeField(auto_now_add=True, db_comment='The timestamp when the report was created')),
+                (
```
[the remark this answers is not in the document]
pbastia: Ooooh good call. I'll add a rule in prettierignore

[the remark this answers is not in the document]
pbastia: on second thought - now that prettier runs on those files, the issue is the files being modified after a release. Prettier _should_ flag this so far

Issue #1165, opened 2024-03-18
Set up Growthbook


andrea-williams:
#### Describe the task

As we implement user stories for Reg Part II, we'll need feature flagging set up so that we can enable new features/user stories into the Registration module only when they're entirely ready to be launched into production.

#### Acceptance Criteria

- [x] add [growthbook](https://www.growthbook.io/) to repo
- [x] test by creating a mock feature (e.g., "Hello World" print statement on a page, hide it)
- [x] add some quick documentation to `docs/developer-guide.md` explaining how to use feature flagging

#### Additional context

- we've set up growthbook in other repos of ours - for reference if needed
- reuse our existing Growthbook account - add tags for feature flags to tag by project

PR #1293, opened 2024-04-05
chore: growthbook - 1165

dleard:
Adds two packages:
growthbook & next-runtime-env (This one allows us to get env vars in client components at runtime)

Includes documentation in the /docs directory on how to use a feature flag in client/server NextJS components

#1165


BCerki (commented):
I wasn't able to get the feature flag working on either server or client components--everything still showed up. 
image
image

This is what I did for setup:
- `yarn`
- added NEXT_PUBLIC_GROWTHBOOK_CLIENT_KEY to env file
- added code from the docs to the profile (server) and operations grid (client)
I've added a commit with my setup in case it helps with debugging.

PR #1319, opened 2024-04-08
1241 accessibility testing

marcellmueller:


client/app/styles/rjsf/ArrayFieldTemplate.tsx
```diff
@@ -47,6 +47,7 @@ const ArrayFieldTemplate = ({
                 <button
                   onClick={item.onDropIndexClick(item.index)}
                   className="border-none bg-transparent p-0"
+                  aria-label="Remove item"
```
marcellmueller, line 50: Buttons need to have visible text, or at least an aria-label for screenreaders.

PR #1346, opened 2024-04-10
chore: release v1.2.3

dleard:


Sepehr-Sobhani (approved):
Approving to not block you!

PR #1368, opened 2024-04-10
chore: fix shipit timeout definition

dleard:
I think the shipit timeout was not being applied correctly in the yaml, the deploy was either timing out at 300 seconds or it would deploy and then throw an error that it doesn't know what "timeout" meant after a deploy was complete. I think this is the correct way to apply a timeout to a shell script. 

docs: https://github.com/Shopify/shipit-engine?tab=readme-ov-file#shell-commands-timeout

### DIFF
--- a/shipit.yml
+++ b/shipit.yml
@@ -6,9 +6,8 @@ review:
 
 deploy:
   override:
-    - |
-      bash deploy.sh
-      timeout: 5000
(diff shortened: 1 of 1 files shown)


mikevespi (approved):
👍 good catch!

dleard:
went with the simpler fix in #1379

PR #1379, opened 2024-04-10
chore: alternative fix to 1368

dleard:
Since the deploy.sh script won't work to deploy giraffe stuff anyway, this is an alternative fix to #1368. It just calls make install & we can figure out a combined deploy strategy later.


mikevespi (approved):
👍

PR #1296, opened 2024-04-08
817 rjsf read only theme

marcellmueller:
Implements #817 

Styling differences can be seen in the Happo diffs.

Some widgets that required finessing the values since they weren't plain strings:

`ReadOnlyComboBoxWidget` - this was just saving the province code ie: `BC` so I had to make it display the full name.

`ReadOnlyMultiSelectWidget` - added some values to `regulated_products` to the `Operation 14` fixture (it was the first Pending/read only one) if you are logged in as `bc-cas-dev`, go to the `Operations` datagrid and click view details for `Operation 14`. I am displaying each item on a new line:
Screenshot 2024-04-08 at 11 10 49 AM

`FileWidget`: - Display the filename with the preview (download) link. I don't think we have fixtures for any download so just log in as `bc-cas-dev` and either add a new or continue an Operation and attach a file. Once submitted the form can be viewed in read only state as it's Pending:
Screenshot 2024-04-08 at 11 12 01 AM

Issue #1241, opened 2024-03-26
Accessibility testing


BCerki:
###Description of the Tech Debt###

Accessibility is important! Let's make sure our app is accessible! Playwright has axe testing: https://playwright.dev/docs/accessibility-testing.
- [ ] install the axe package
- [ ] add an assertion to scan all screens in the app (look for where the happo screenshots are taken, that should catch most of them)
- [ ] if a manageable number of items come up in the scan, fix as part of this ticket. If there are a lot of problems, create more broken-down tickets to address them

### Additional Context ###
- if we only have a small number/minor accessibility issues, fix them as part of this ticket
- if we have a large number of issues or major accessibility issues, this ticket may have to "hang out" and been on pause while we fix other issues (in a separate ticket). One option would be to comment-out accessibility tests by file ON THE CONDITION that a new ticket is created capturing the work to fix the issues in that file

###Tech Debt Triage#
The purpose of our technical debt triage process is to analyze technical debt to determine risk level of the technical debt and the value in tackling that technical debt.

Risk Value Scoring:

| Level  | Value |
| ------ | ----- |
| High   | 3     |
| Medium | 2     |
| Low    | 1     |

| Technical Debt - Risk Types                                                                                                   | Level | Value |
| ----------------------------------------------------------------------------------------------------------------------------- | ----- | ----- |
| Business Area Risk - Risk of business area visibility / damage to user experience                                             |       |       |
| Developer Fault Risk - How likely will this tech debt cause a future error related to coding on top of it                     |       |       |
| System Fault Risk - Risk of system errors or application downtime                                                             |       |       |
| Time Scale Risk - Compound risk effect if left alone. How much more difficult to fix or dangerous will this become over time? |       |       |
| Time Sink Risk - How much will this tech debt slow the development process down                                               |       |       |
| TOTAL SCORE:                                                                                                                  |       |       |

PR #1383, opened 2024-04-11
1241 accessibility testing

marcellmueller:
Implements #1241

rebased this on #1411 to fix the yarn cache issue so this could pass CI

For the tests themselves the initial plan was to write totally separate tests for accessibility, though I soon realized we were repeating so many steps that this seemed like the wrong approach. With this in mind I think we will have to be mindful to not repeat tests unnecessarily. I don't think we need to test the same areas with different roles and focus on making sure we test each page and test the forms in different states ie read only, with errors etc.

I've placed them after most of our Happo screenshots and I think we will have to consider if that scenario is tested before adding more screenshots for other roles. Testing each form page *might* seem excessive though I thought it was worth it to make sure all of our widgets were compliant.

We had some accessibility issues though it wasn't too bad:

- Our pages needed an `h1` header. A lot of our pages don't have an obvious, unique `h1` in the design so I've wrapped the title in the nav in one. There is some contention about using a non-unique page title as an h1 though given the design of our pages this seemed like the best option.
- The disabled form field text colour not having enough contrast. Fixed by #1296 - RJSF read only theme
- The red minus button in our Parent Operator array field doesn't have text in it which failed, so I added an `aria-label`.
- There was an error with the DataGrid pagination component. This one eluded me though I found an option for the DataGrid `experimentalFeatures={{ ariaV7: true }}` which fixed it. While looking into this I noticed that `mui-x/DataGrid` version 7 just came out and tried installing that but it had further accessibility errors that seemed unrelated to any customizations we added (I disabled a bunch of our custom cells and pagination). I will likely open a tech debt ticket before merging this to keep an eye on v7 and update it in the future.


BCerki:
> Implements #1241
> 
> rebased this on #1411 to fix the yarn cache issue so this could pass CI
> 
> For the tests themselves the initial plan was to write totally separate tests for accessibility, though I soon realized we were repeating so many steps that this seemed like the wrong approach. With this in mind I think we will have to be mindful to not repeat tests unnecessarily. I don't think we need to test the same areas with different roles and focus on making sure we test each page and test the forms in different states ie read only, with errors etc.
> 
> I've placed them after most of our Happo screenshots and I think we will have to consider if that scenario is tested before adding more screenshots for other roles. Testing each form page _might_ seem excessive though I thought it was worth it to make sure all of our widgets were compliant.
> 
> We had some accessibility issues though it wasn't too bad:
> 
> * Our pages needed an `h1` header. A lot of our pages don't have an obvious, unique `h1` in the design so I've wrapped the title in the nav in one. There is some contention about using a non-unique page title as an h1 though given the design of our pages this seemed like the best option.
> * The disabled form field text colour not having enough contrast. Fixed by [817 rjsf read only theme #1296](https://github.com/bcgov/cas-registration/pull/1296) - RJSF read only theme
> * The red minus button in our Parent Operator array field doesn't have text in it which failed, so I added an `aria-label`.
> * There was an error with the DataGrid pagination component. This one eluded me though I found an option for the DataGrid `experimentalFeatures={{ ariaV7: true }}` which fixed it. While looking into this I noticed that `mui-x/DataGrid` version 7 just came out and tried installing that but it had further accessibility errors that seemed unrelated to any customizations we added (I disabled a bunch of our custom cells and pagination). I will likely open a tech debt ticket before merging this to keep an eye on v7 and update it in the future.

This all makes sense to me. I think testing each form page is a good idea; certain widgets could be non-compliant.

client/e2e/workflows/bceidbusiness/industry_user.spec.ts
```diff
@@ -50,12 +54,16 @@ test.describe("Test Workflow industry_user", () => {
...
+    // 🔍 Assert the form is visible - needed to prevent analyzeAccessibility from failing
+    await selectOperatorPage.formIsVisible();
     // 📷 Cheese!
     pageContent = page.locator("html");
     await happoPlaywright.screenshot(page, pageContent, {
       component: "Select operator page",
       variant: "default",
     });
+    // ♿️ Analyze accessibility
+    await analyzeAccessibility(page);
```
BCerki, line 66: It looks like every happo shot is accompanied by the accessibility test, so could consider making a helper to do both.

BCerki (approved):
Thanks for your work on this, yay accessibility testing! And awesome that you were able to fix the fails in the same PR.

Everything in here looks good (pending CI fixes), so I'm going to approve, but this PR only covers accessibility for the industry user roles. Cas role tests are getting happo tests in @shon-button's https://github.com/bcgov/cas-registration/pull/1337, so we can block the existing card on that, or make a new one for cas role accessibility testing.

client/e2e/workflows/bceidbusiness/industry_user.spec.ts
marcellmueller: Thanks, I have made a separate PR to simplify this and make sure I don't break our Happo screenshots here (still WIP): https://github.com/bcgov/cas-registration/pull/1426

PR #1384, opened 2024-04-11
chore: deploy reporting with specific shipit environment

dleard:
- Re-adds using the deploy.sh script in shipit.yml
- Updates deploy.sh to parse a reporting-specific environment from shipit into an environment that can be used by helm to deploy to openshift

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -4,8 +4,9 @@
 echo "Branch: $BRANCH"
 
 # Check if the current branch is 'giraffe-develop'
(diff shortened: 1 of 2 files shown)


deploy.sh
```diff
@@ -4,8 +4,9 @@
 echo "Branch: $BRANCH"
 
 # Check if the current branch is 'giraffe-develop'
```
pbastia, line 6: comment is now obsolete

pbastia (approved):
just an old comment that can go!

Issue #1282, opened 2024-04-05
Adding Search Filters to Operations table


hannavovk:
#### Description:

As a CAS internal user, I want to be able to search operation BORO ID applications so I don't have to scroll through a long list to find the application I am looking for.
 
The wireframes of Registration pt. 1 in Figma included search boxes, that would work as filters, under column titles in the dashboards (screenshot 1). 
 
Could we please get these added in the dashboard (screenshot 2)? They would be very useful as applications are piling up.

![Image](https://github.com/bcgov/cas-registration/assets/87211104/81a10b25-c5fb-4654-bafb-328e2cfbd3ec)

![Image](https://github.com/bcgov/cas-registration/assets/87211104/54335373-4869-4489-858b-22f55ef9ce6e)

 
 
#### Acceptance Criteria:

Given
When
Then

#### Development Checklist:

- [ ] Checklist item
- [ ] Checklist item
- [ ] Checklist item
- [ ] Meets the DOD

**Definition of Ready** (Note: If any of these points are not applicable, mark N/A)

- [ ] User story is included
- [ ] User role and type are identified
- [ ] Acceptance criteria are included
- [ ] Wireframes are included (if required)
- [ ] Design / Solution is accepted by Product Owner
- [ ] Dependencies are identified (technical, business, regulatory/policy)
- [ ] Story has been estimated (under 13 pts)

·**Definition of Done** (Note: If any of these points are not applicable, mark N/A)

- [ ] Acceptance criteria are tested by the CI pipeline
- [ ] UI meets accessibility requirements
- [ ] Configuration changes are documented, documentation and designs are updated
- [ ] Passes code peer-review
- [ ] Passes QA of Acceptance Criteria with verification in Dev and Test
- [ ] Ticket is ready to be merged to main branch
- [ ] Can be demoed in Sprint Review
- [ ] Bugs or future work cards are identified and created
- [ ] Reviewed and approved by Product Owner

#### Dependencies

- Blocked by
- Blocking

hannavovk:
@marcellmueller investigating how to make a quick fix and will create a separate ticket for the search box design after

shon-button:
@marcellmueller 

[Next.js 14: The Power of the URL...](https://blog.gogrow.dev/next-js-14-the-power-of-url-in-state-management-7d38a4e34ca8)

```
One of the most compelling aspects of using the URL as a state manager is the seamless ability to share curated experiences effortlessly...the experience becomes dynamic and easily shareable.

[...]

PR #1385, opened 2024-04-11
1282 search filters

marcellmueller:
Implements #1282 

Make sure to run `yarn` before trying this as the search fields won't show up on the older version of `MUI DataGrid`.

This includes an update to all of our DataGrids  - `Operators`, `Operations` and `Users`. Now when you search or sort that gets added to the URL `searchParams`. You can now refresh the page or save a bookmark and load your previous sort order 🎉 

- Since I've refactored all the DataGrids it would be good to make sure the `Operators` DataGrid for `cas` users is working still as well as the `Users` DataGrid for `industry_user_admin`.
- As discussed at standup, the `submission_date` field doesn't have a search filter due to it being saved in a timestamp format, it wouldn't return any data if the user typed in a month ie `jan`. We may add this in the future if it's decided to be important.
- No search for the `Action` column either as it's not a database value
- Refactored all of the base `page.tsx` routes and children for our DataGrid components. `page.tsx` files receive `searchParams` props so we can pass those to the server components and fetch the correct page/sort/filter data on the serverside first render
- I initially installed the new `DataGrid v7` which supports column groups which I used for the search. However there are accessibility errors with it that I found in the accessibility PR so I upgraded to the latest v6 and used the experimental `columnGrouping` feature. I created a [tech debt ticket to upgrade in the future ](https://github.com/bcgov/cas-registration/issues/1410). The upgrade is pretty straightforward aside from some style tweaks and testing selectors. Hopefully they fix the aria issues soon!


bc_obps/registration/api/operations.py
```diff
@@ -122,6 +134,16 @@ def list_operations(request, page: int = 1, sort_field: str = "created_at", sort
...
+            .filter(
+                Q(bcghg_id__icontains=bcghg_id) if bcghg_id else Q(),
+                Q(bc_obps_regulated_operation__id__icontains=bc_obps_regulated_operation)
+                if bc_obps_regulated_operation
+                else Q(),
+                Q(name__icontains=name) if name else Q(),
+                Q(operator__legal_name__icontains=operator) if operator else Q(),
+                Q(status__icontains=status) if status else Q(),
+                Q(submission_date__icontains=submission_date) if submission_date else Q(),
+            )
```
marcellmueller, line 146: @Sepehr-Sobhani 

This is working for the multi-field search. What do you think of this method? Thanks :)

Sepehr-Sobhani: I think your approach is good, but I highly recommend following the suggestions provided by django-ninja: https://django-ninja.dev/guides/input/filtering/.

client/app/(authenticated)/bceidbusiness/industry_user/dashboard/operations/page.tsx
```diff
@@ -1,6 +1,11 @@
...
+import { OperationsSearchParams } from "@/app/components/routes/operations/types";
 
-export default async function Page() {
-  return <OperationsPage />;
+export default async function Page({
+  searchParams,
+}: {
+  searchParams: OperationsSearchParams;
+}) {
+  return <OperationsPage searchParams={searchParams} />;
```
marcellmueller, line 10: These page files are all refactored because for reasons I don't understand, there is no hook or easy way to get `search params` in server components in NextJS besides this way. Search params are pass to all `page.tsx` props so I am passing them through so we can use them to fetch data in the server components.

client/e2e/utils/helpers.ts
```diff
@@ -303,3 +309,90 @@ export async function setupTestEnvironment(
...
+    .locator(`[aria-label="${columnLabel}"]`)
+    .getAttribute("aria-sort");
+
+  expect(ariaSort).toBe(expectedSortDirection);
+
+  // wait for response to complete
+  await page.waitForResponse((response) => response.status() === 200);
+
+  // TODO: Find a better way to wait for the table to be sorted
+  await page.waitForTimeout(1000);
```
marcellmueller, line 337: I spent a lot of time trying alternatives to this timeout (same with the filter helper) and came up with nothing 😞  Would love a suggestion on how to get rid of it, though for now I am stuck and left this TODO here.

Without it the assertion at the bottom of the function fails as it returns the text that exists in the first cell before the sort finishes.

client/app/components/datagrid/styles.ts
```diff
@@ -0,0 +1,92 @@
...
+  },
+  "& .MuiDataGrid-columnHeader": {
+    border: "1px white solid",
+    borderBottom: "none",
+    borderTop: "none",
+    color: "white",
+    backgroundColor: BC_GOV_BACKGROUND_COLOR_BLUE,
+    fontWeight: "bold",
+  },
+  "& .MuiDataGrid-columnHeadersInner": {
```
marcellmueller, line 21: This `sx` prop in `DataGrid.tsx` was getting rather large so I've moved it to it's own file and spread it in the `sx` style prop. Dynamic styles go underneath the spread.

Sepehr-Sobhani:
Go get some rest man 😜

BCerki (commented):
First of all, apologies in advance for this suggestion because I know you've put a lot of time and effort into developing this feature.

I've just learned that the free version of MUI Datagrid has a search feature called [Quick Filter](https://mui.com/x/react-data-grid/filtering/quick-filter/). It doesn't show up above every column like the design, but the designers did the grid based on CIF because they thought that would be easiest for us, so the exact placement of the search may not be that important.

Quick search is interesting because it searches every column at once and actually worked on dates. I threw it into a test branch (I didn't want to suggest overhauling this PR without checking if it would actually work) and you can see how it looks here: https://www.loom.com/share/0398dfc3a99746cbb1db0ae462f0d0e9. (branch [here](https://github.com/bcgov/cas-registration/pull/1438/files). It's a lot less code and therefore more maintainable, so even though this PR is working, I think we should at least consider switching to quick search/column filters (column filters are also free and again not exactly like the design, but they're so much less code I think it's worth bringing up with the designers).

marcellmueller:
> First of all, apologies in advance for this suggestion because I know you've put a lot of time and effort into developing this feature.
> 
> I've just learned that the free version of MUI Datagrid has a search feature called [Quick Filter](https://mui.com/x/react-data-grid/filtering/quick-filter/). It doesn't show up above every column like the design, but the designers did the grid based on CIF because they thought that would be easiest for us, so the exact placement of the search may not be that important.
> 
> Quick search is interesting because it searches every column at once and actually worked on dates. I threw it into a test branch (I didn't want to suggest overhauling this PR without checking if it would actually work) and you can see how it looks here: https://www.loom.com/share/0398dfc3a99746cbb1db0ae462f0d0e9. (branch [here](https://github.com/bcgov/cas-registration/pull/1438/files). It's a lot less code and therefore more maintainable, so even though this PR is working, I think we should at least consider switching to quick search/column filters (column filters are also free and again not exactly like the design, but they're so much less code I think it's worth bringing up with the designers).

I saw the quick filter as well when I was researching how to implement the design. The reason it's seemingly working so well out of the box is that it's just doing a search for the single page data that we have fed into the page (including nice readable dates). It's not doing a database search so the dates wouldn't work once we did a proper implementation of this.

As for the column filters as discussed we would have implemented that in this ticket if I didn't find a way to implement the design. Most of the changes and refactoring in this ticket would stay the same aside from the `columnGroups`, the `HeaderSearchCell` and some minor `DataGrid` styling updates.

BCerki:
> > First of all, apologies in advance for this suggestion because I know you've put a lot of time and effort into developing this feature.
> > I've just learned that the free version of MUI Datagrid has a search feature called [Quick Filter](https://mui.com/x/react-data-grid/filtering/quick-filter/). It doesn't show up above every column like the design, but the designers did the grid based on CIF because they thought that would be easiest for us, so the exact placement of the search may not be that important.
> > Quick search is interesting because it searches every column at once and actually worked on dates. I threw it into a test branch (I didn't want to suggest overhauling this PR without checking if it would actually work) and you can see how it looks here: https://www.loom.com/share/0398dfc3a99746cbb1db0ae462f0d0e9. (branch [here](https://github.com/bcgov/cas-registration/pull/1438/files). It's a lot less code and therefore more maintainable, so even though this PR is working, I think we should at least consider switching to quick search/column filters (column filters are also free and again not exactly like the design, but they're so much less code I think it's worth bringing up with the designers).
> 
> I saw the quick filter as well when I was researching how to implement the design. The reason it's seemingly working so well out of the box is that it's just doing a search for the single page data that we have fed into the page (including nice readable dates). It's not doing a database search so the dates wouldn't work once we did a proper implementation of this.
> 
> As for the column filters as discussed we would have implemented that in this ticket if I didn't find a way to implement the design. Most of the changes and refactoring in this ticket would stay the same aside from the `columnGroups`, the `HeaderSearchCell` and some minor `DataGrid` styling updates.

OK, thanks! So all of the data grid complication is because of pagination, and we have to paginate so we don't have load problems.

marcellmueller:
> OK, thanks! So all of the data grid complication is because of pagination, and we have to paginate so we don't have load problems.

Yeah the reason there is so many changes in this PR is because of refactoring all DataGrids due to changes with handling the state in the query params. This is a big quality of life upgrade and and is seen as best practice by many - Shon emailed me [this article](https://blog.gogrow.dev/next-js-14-the-power-of-url-in-state-management-7d38a4e34ca8).

So even if we went to a single field search most of the changes would remain. As an aside, I definitely considered pushing to use the quick search and have a single search for all fields, though I don't feel strongly about which is better UX so I implemented the design as is.

BCerki (commented):
Partway through the review, just "finishing" now so I don't lose my comments

BCerki (commented):
Great work on this! I didn't completely follow everything that's going on in the grid refactor, so if you'd like some more feedback, maybe we can do a walkthrough.

Some overall notes:
- I looked at all the grids in all the roles and was able to paginate and click into the details pages with no problems
- I was got some glitches with the search focus both as industry user and cas: https://www.loom.com/share/84b053ef244d4d60a661066847df3a7a
- If I try and search on the second page of the grid, I get an error: https://www.loom.com/share/fa861240986e4bbc9435682f0af9a894
- I think we could use fewer e2e tests and more python and vitest tests to be more in line with the testing trophy

client/e2e/workflows/idir/cas_admin.spec.ts
```diff
@@ -110,4 +117,50 @@ test.describe("Test Workflow cas_admin", () => {
...
+    await tableRowCount(operationsPage.page, 1);
+
+    // Check junk search
+    await filterTableByFieldId(
+      operationsPage.page,
+      OperationTableDataField.OPERATION,
+      OperationTableHeaders.OPERATION,
+      "this search will return no results",
+      true,
+    );
```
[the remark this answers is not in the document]
marcellmueller: I added a test for Operator sorting which was missing.

client/app/utils/buildQueryParams.ts
```diff
@@ -0,0 +1,15 @@
+// Function to build a query string from an object of key-value pairs
+// eg: buildQueryParams({ page: "1", search: "search" }) => "?page=1&search=search"
+
+const buildQueryParams = (params: {
```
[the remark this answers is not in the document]
marcellmueller: I agree though this was brought up at standup and I was asked to just do e2e tests for this ticket. That said I added a small unit test for this anyways so I didn't add more tech debt 🙈

marcellmueller (commented):
Thanks for the good review!

- I updated the focus handling so it should work better now. This part of the UX might have been the most difficult part of the whole ticket.
- Thanks for catching the page 2 bug 💀 Luckily it was a relatively easy fix and should be working well now.

Let me know if you want to walk through this PR or discuss the e2e stuff haha 😅

BCerki (approved):
I manually re-tested the focus and pagination and everything's working well. I found one styling bug: file:///home/briannacerkiewicz/Pictures/Screenshots/Screenshot%20from%202024-04-24%2007-30-40.png that could be tech debt or part of a larger ticket for mobile display (I'm not on mobile but I use the accessibility settings on my computer to make everything bigger and it kind of has the same effect). I also had a mini row to the right of actions, but part of it being blue is new from this PR.

marcellmueller:
> I manually re-tested the focus and pagination and everything's working well. I found one styling bug: Screenshot from 2024-04-24 07-30-40 that could be tech debt or part of a larger ticket for mobile display (I'm not on mobile but I use the accessibility settings on my computer to make everything bigger and it kind of has the same effect). I also had a mini row to the right of actions, but part of it being blue is new from this PR.

Thanks, I fixed this!

PR #1403, opened 2024-04-15
Fix: is approved admin user operator endpoint

Sepehr-Sobhani:


BCerki (approved):
Thanks for catching this!!
I also wonder why our e2e didn't catch this--I see the assertion for the tile right here: 
image and there's a whole test for that tile: `User Access Management Tile workflow - Approve and Decline users`
To investigate once e2e tests are working again, I guess

BCerki:
> Thanks for catching this!! I also wonder why our e2e didn't catch this--I see the assertion for the tile right here: image and there's a whole test for that tile: `User Access Management Tile workflow - Approve and Decline users` To investigate once e2e tests are working again, I guess

Also, how did you get this through CI? Nothing else is passing e2e

Sepehr-Sobhani:
> > Thanks for catching this!! I also wonder why our e2e didn't catch this--I see the assertion for the tile right here: image and there's a whole test for that tile: `User Access Management Tile workflow - Approve and Decline users` To investigate once e2e tests are working again, I guess
> 
> Also, how did you get this through CI? Nothing else is passing e2e

I didn't do anything :) I only fixed those endpoints. Probably @shon-button did the magic.

shon-button:
> Thanks for catching this!! I also wonder why our e2e didn't catch this--I see the assertion for the tile right here: image and there's a whole test for that tile: `User Access Management Tile workflow - Approve and Decline users` To investigate once e2e tests are working again, I guess

@Sepehr-Sobhani @BCerki 
Since we have **moved** from generating a user's storageState using "live" login **to** using a "static" storageState within a GitHub secret, the storageState user session information for `industry_user_admin` tests has a valid app_role.  Therefore, it is expected the test should pass if the GutHub secret `E2E_INDUSTRY_USER_ADMIN_STORAGE` is valid.

Issue #1402, opened 2024-04-15
Add PostGIS extension to database


andrea-williams:
#### Describe the task

Based on anticipated future work for Registration Part 2 regarding mapping/GIS stuff, it would be useful for us to have PostGIS installed on our databases to enable geospatial queries and overall better handling of geospatial data.

Installation instructions: https://postgis.net/documentation/getting_started/

#### Acceptance Criteria

- [ ] install PostGIS extension on our PROD database
- [ ] install PostGIS extension on our database in TEST
- [ ] install PostGIS extension on our database in DEV
- [ ] add the PostGIS extension installation command to the `create_db` Make command for our local development

#### Additional context

- geographic coordinates are included in the `Facility` model (#1320 ) so ideally this ticket will be done soon
- example usage: we can use PostGIS to compare a facility's user-inputted geographic coordinates against an imported shapefile of BC to ensure that all facilities are within BC.

Sepehr-Sobhani:
Useful doc: https://docs.djangoproject.com/en/5.0/ref/contrib/gis/install/postgis/

hannavovk:
Better requirements needed. If we pull this in first need a card to gather requirements.

PR #1406, opened 2024-04-16
1402 postgis

BCerki:
https://github.com/bcgov/cas-registration/issues/1402


BCerki:
Probably (?) moving away from postgis, closing

PR #1413, opened 2024-04-16
1211 database setup for email notification latest

Sepehr-Sobhani:
[ISSUE 1211](https://github.com/orgs/bcgov/projects/122/views/2?filterQuery=&pane=issue&itemId=57274383)

**NOTES**

- In order to run CHES emails locally, we need to obtain CHES credentials from the updated `OBPS backend ENV` file in 1Password.

- It is advisable to avoid sending actual emails while testing features to prevent spamming people. For instance, when testing the `send_admin_access_request_email` function, we may replace our email with `user_operator.user.email` to receive the real email and confirm its contents.

- To add more email templates and continue working on the Email Notification feature, we need to create a new migration file. This file should include the `subject`, `name`, and `body` of the new email notification. This process is similar to the migration file `0002_...` in the common app. After creating the new template, we can use it to generate new functions to create the email context and send the email, just like the `send_admin_access_request_email` function.

- To test submission confirmation state:
- Add CHES credentials to your `bc_obps/.env` file
- In `def request_admin_access` function replace your personal email with `user_operator.user.email` in `send_admin_access_request_email` function
- Log in as `bc-cas-dev-three` and request admin access to an operator that doesn't have an administrator already
- Look at your email :) You should receive something like below:
Screenshot 2024-04-16 at 4 13 12 PM

- To test Approve and Declined state:
- Add CHES credentials to your `bc_obps/.env` file
- in `def update_user_operator_status` function replace your personal email with `user_operator.user.email` in `send_admin_access_request_email` function
- Log in as `bc-cas-dev` and go to the user access management page and Decline/Approve admin requests
- Look at your email :) You should receive something like below:
Screenshot 2024-04-16 at 4 17 57 PM


andrea-williams:
Review notes:
- pulled code to local, received emails when requesting admin access as external user, and when admin access was approved.

.github/workflows/test.yaml
```diff
@@ -13,10 +13,6 @@ env:
   KEYCLOAK_CLIENT_ID: ${{ secrets.KEYCLOAK_CLIENT_ID }}
   KEYCLOAK_CLIENT_SECRET: ${{ secrets.KEYCLOAK_CLIENT_SECRET }}
   NEXTAUTH_SECRET: ${{ secrets.NEXTAUTH_SECRET }}
-  CHES_CLIENT_ID: ${{ secrets.CHES_CLIENT_ID }}
```
andrea-williams, line 16: thanks for having my back Sepehr! I forgot to remove these in my PR _even after you reminded me about it_! 🤦‍♀️

bc_obps/common/tests/utils/helpers.py
```diff
@@ -0,0 +1,42 @@
+from django.db import models
```
andrea-williams, line 1: this is awesome - thanks for setting up these helpers!!

docs/developer-guide.md
```diff
@@ -662,3 +662,10 @@ Sometimes, an endpoint needs to do something more complicated than simply call a
 
 - if multiple database services are called, ensure they're atomic using the @transaction.atomic() decorator
 - check if users should be allowed to do things (in the regisration app, role-based authentication is handled in the endpoints, but anything more specific (e.g., if an operator already has an admin, subsequent users can't request admin access) is handled in an intermediary service)
+
+## Email Notifications
+
+Email notifications are an essential part of the application's communication strategy. They provide users with important information, updates, and alerts. We use CHES (Common Hosted Email Service) to send email notifications to users.
+To send real emails, you need to set up CHES credentials in your environment variables. You can find the required CHES credentials in the 1Password document `OBPS backend ENV`.
+
+**Note**: Make sure to not send real emails in the development environment by commenting out the CHES credentials in the `bc_obps/.env` file.
```
andrea-williams, line 671: 💙

bc_obps/common/service/email/email_service.py
```diff
@@ -1,27 +1,37 @@
...
 
-    def __init__(self):
-        self.api_url: str = settings.CHES_API_URL
-        self.client_id: str = settings.CHES_CLIENT_ID
-        self.client_secret: str = settings.CHES_CLIENT_SECRET
-        self.token_endpoint: str = settings.CHES_TOKEN_ENDPOINT
-        self.token: Optional[str] = None
-        self.token_expiry: datetime = datetime.now()
-        logger.info(f'Initializing EmailService for clientID {self.client_id} to connect to {self.api_url}')
+    _instance = None
```
andrea-williams, line 19: this is extremely nitpicky (but I'm in a nitty headspace because I spent all afternoon in interviews) - I'd suggest adding an extra line of documentation/comment here explaining how to use the EmailService class in other files. I think that would be helpful to future devs.

I think something as simple as this would suffice:
"Use `email_service = EmailService()` to access the EmailService singleton object in other .py files"

andrea-williams (approved):
It's a beauty! 🌟

.github/workflows/test.yaml
Sepehr-Sobhani: Anytime 🥷

bc_obps/common/service/email/email_service.py
Sepehr-Sobhani: This is a very valid point, thank you 🙏

Issue #1362, opened 2024-04-10
Admin\Facilities - External User has data imported from SWRS


zoeyli-46:
Blocked by #2749 , #2750 

As an External User, given that my operation is not a new entrant, I want any pre-existing data about my facility to be imported from SWRS

Acceptance Criteria:
- [ ] import `if_a`, `if_b`, and `lc` facilities from SWRS
- [ ] import each facility's name, BCGHG ID, SWRS Facility ID (wherever data is available). This data should be inserted into the `facility` table
- [ ] where possible, associate each imported facility with its operation (adding a new record in `FacilityDesignatedOperationTimeline`). For some facilities in SWRS this will be possible by referencing the facility's "Parent Facility ID" (the LFO). The LFO also has a SWRS Facility ID (and may have a BCGHG ID), which can be used to FK reference to our existing prod data about operations
- [ ] in cases where it isn't possible to connect an imported facility to an existing operation in our prod data --- **what should we do? TBD** --> decide after completing the practice run; assess based on how many orphaned facilities there are. May need manual reconciliation.

dleard:
@BCerki @hannavovk The facility model will need bcghgid and swrs_facility_id columns added to it if we decide that we are going to import facility data from swrs. I think we will end up doing an import, but since it's a question right now the columns were not added in the initial setup of the facility model.

patriciarussellCAS:
@andrea-williams

patrickisaac:
Refinement Notes:
1. Doing a practice run
2. Once Reg1 is shutdown, then we can do a complete run

patrickisaac:
Work has been complete. Closing ticket

PR #1417, opened 2024-04-17
1320 2 facility model

BCerki:
card: https://github.com/orgs/bcgov/projects/122/views/2?filterQuery=facil&pane=issue&itemId=59088186

Note: 
- address is currently not mandatory because we haven't done any front-end work to collect operation addresses
- happo looks like flake to me; hoping it'll sort itself out when I inevitably rebase
- I ~think I've correctly set well authorization number to optional but would appreciate another set of eyes


bc_obps/registration/migrations/0012_facility_alter_operation_table_comment_and_more.py
```diff
@@ -0,0 +1,307 @@
...
+        migrations.AlterModelTableComment(
+            name='operation',
+            table_comment='Operations',
+        ),
+        migrations.AlterField(
+            model_name='operation',
+            name='address',
+            field=models.ForeignKey(
+                blank=True,
+                db_comment='The address of the operation or facility',
```
dleard, line 91: ditto here, if this comment is on a field on the operation table should it just say 'operation'?
```suggestion
                db_comment='The address of the operation',
```

bc_obps/registration/migrations/0012_facility_alter_operation_table_comment_and_more.py
```diff
@@ -0,0 +1,307 @@
...
+                        null=True,
+                        on_delete=django.db.models.deletion.PROTECT,
+                        related_name='%(class)s_updated',
+                        to='registration.user',
+                    ),
+                ),
+            ],
+            options={
+                'db_table': 'erc"."facility',
+                'db_table_comment': 'Facilities ',
```
dleard, line 79: This table comment needs more information
```suggestion
                'db_table_comment': 'Contains data on facilities that emit carbon emissions and must report them to Clean Growth. A linear facility operation is made up of several different facilities whereas a single facility operation has only one facility. In the case of a single facility operation, much of the data in this table will overlap with the parent record in the operation table',
```

bc_obps/registration/migrations/0012_facility_alter_operation_table_comment_and_more.py
```diff
@@ -0,0 +1,307 @@
+# Generated by Django 4.2.8 on 2024-04-17 14:19
```
dleard, line 1: Can this migration be squashed into 0011_... or is there a dependency?

dleard (changes requested):

bc_obps/registration/tests/test_models.py
```diff
@@ -802,6 +813,57 @@ def test_unique_swrs_facility_id_constraint(self):
...
+        cls.test_object = Facility.objects.first()
+        cls.field_data = [
+            *timestamp_common_fields,
+            ("id", "ID", None, None),
+            ("name", "name", 1000, None),
+            ("type", "type", 1000, None),
+            ("operation", "operation", None, None),
+            ("address", "address", None, None),
+            ("new_entrant", "new entrant", None, None),
+            ("well_authorization_numbers", "well authorization number", None, None),
```
BCerki, line 839: @Sepehr-Sobhani , I've got these tests working, but I don't think I completely understand how they work. `well_authorization_numbers` is my related name in the Facility table:
```
facility = models.ForeignKey(Facility, on_delete=models.DO_NOTHING, related_name="well_authorization_numbers")
```
But what is `well authorization number`?

bc_obps/registration/models.py
```diff
@@ -684,37 +684,21 @@ def save(self, *args, **kwargs):
 class OperationAndFacilityCommonInfo(TimeStampedModel):
     """Operation and facility common information abstract base class"""
 
-    name = models.CharField(max_length=1000, db_comment="An operation or facility's name")
-    type = models.CharField(max_length=1000, db_comment="An operation or facility's type")
-    naics_code = models.ForeignKey(
-        NaicsCode,
+    name = models.CharField(max_length=1000, db_comment="The name of a placeholder")
```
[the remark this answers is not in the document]
Sepehr-Sobhani: I gave a few different approaches a shot, but it looks like this might not be doable.

bc_obps/registration/tests/test_models.py
Sepehr-Sobhani: That is the label name of that field. It's being mainly used for Django forms, but we thought it would be handy to test it out from the very beginning in case we find a use for it later on.

bc_obps/registration/migrations/0011_facility_alter_operation_table_comment_and_more.py
```diff
@@ -0,0 +1,418 @@
...
+from django.db import migrations, models
+import django.db.models.deletion
+import simple_history.models
+import uuid
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('registration', '0010_V1_3_0'),
```
dleard, line 12: this will need updating now that our 1.4.0 release has been added

dleard (approved):
Looks good to me. I've added a comment to the import facility data card (#1362) to remind us to add the bcghgid & swrs_facility_id columns to the facility table if we decide that we do want to import facility data from swrs. I'm pretty sure we will end up doing an import, but since we're not 100% sure, we can leave those columns out of the model for now.

PR #1425, opened 2024-04-17
1305 2 refactor user endpoints

BCerki:
https://github.com/bcgov/cas-registration/issues/1305


bc_obps/registration/api/user/user_app_role.py
```diff
@@ -1,15 +1,13 @@
...
     "user/user-app-role/{user_guid}", response={200: UserAppRoleOut, codes_4xx: Message}, url_name="get_user_role"
 )
+@handle_http_errors()
 def get_user_role(request, user_guid: str):
-    try:
-        user = User.objects.only('app_role').select_related('app_role').get(user_guid=user_guid)
-    except User.DoesNotExist:
-        return 404, {"message": "No matching user found"}
-    return 200, user.app_role
+    return 200, UserDataAccessService.get_app_role(user_guid)
```
BCerki, line 13: @Sepehr-Sobhani , I think returning a 200 is optional because django ninja will do it on success. Would you recommend including it anyway? Sometimes I did and sometimes I didn't in the refactor tickets, so I can standardize to whatever you think is best

client/app/components/routes/profile/User.tsx
```diff
@@ -24,7 +24,7 @@ export default async function User() {
   let formData: UserProfilePartialFormData | { error: string } =
     await getUserFormData();
   if ("error" in formData) {
-    if (formData.error.includes("404")) {
+    if (formData.error.includes("Not Found")) {
```
BCerki, line 27: @Sepehr-Sobhani , this surprised me a bit--the http error decorator does still return a 404, but when I log `formData` here, I get `{error: Not Found}`. So it seems like the decorator isn't returning exactly the same response as the individual endpoints. This is the first time it's come up--do you think it's a problem?

bc_obps/registration/api/user/user_app_role.py
Sepehr-Sobhani: Correct, Django Ninja does default to a 200 status code on successful requests. However, it's still beneficial to explicitly set it to 200 for consistency and clarity, especially when different status codes are used within the same function. Standardizing this approach can help prevent confusion among developers who are new to the codebase.

client/app/components/routes/profile/User.tsx
Sepehr-Sobhani: Sorry, I'm not sure if I'm following the issue here. Is there a way I can test this locally?

BCerki: If you pull the branch and leave it as `if (formData.error.includes("404"))`, the e2e profile none test will fail because after the refactor, `formData.error` no longer includes "404". However, as far as I can tell, I am returning a 404 in the decorator. Thanks for your help with this!

shon-button:
`client/app/api/auth/[...nextauth]/route.ts` fails to assign `app_role` due to missing `role_name` from 
```
const responseRole = await actionHandler(
            `registration/user/user-app-role/${token.user_guid}`,
            "GET"
          );
```

client/app/components/routes/profile/User.tsx
Sepehr-Sobhani: We previously used `get_object_or_404` to retrieve the user instance, which raised an `Http404` error without any message. Now, we've changed our approach and we still return a `404` error, but it includes a descriptive message. This helps clarify the error's context, making it a more effective solution.

shon-button (approved):

Issue #1013, opened 2024-03-01
Automated Email for Scenario 1.ii "External user logs in for the first time - Operator data exists, and there’s an Admin"


nanyangpro:
When external user logs in for the first time,
and If Operator data exists,
and there’s an Admin,
the user sees the screen below after submitting the access request
![Image](https://github.com/bcgov/cas-registration/assets/58751681/fe824e43-bf8f-438f-b6d5-6d61f488a509)

- [x] Submission confirmation
- [x] Approved
- [x] Declined

emails are required. Please see the [email templates copy](https://bcgov.sharepoint.com/teams/00608-CPRInternal/_layouts/15/Doc.aspx?sourcedoc={0d7ce55e-da9d-42d2-94b6-05171abd4452}&action=edit&wd=target%28Registration.one%7Ca745a153-ea3f-4b6e-8f16-9163bfe64932%2FScenario%201%28ii%5C%29%20Existing%20operator%5C%2F%20Existing%20admin%7C6acfc579-9175-4e6d-9ca2-a5207fbf820c%2F%29&wdorigin=703) here.

Blocked by #141, #266

nanyangpro:
Hi @hannavovk - Details are added to this ticket. Feel free to review/refine!

Once Jairo finished breaking down the email templates copy into separated pages, I'll add in the url, and move the ticket to the "Next up" column.

PR #1427, opened 2024-04-18
1013 automated email existing operator and admin

Sepehr-Sobhani:
[ISSUE 1013](https://github.com/bcgov/cas-registration/issues/1013)


dleard (approved):
:100: Manually tested email notifications when a user requests access to an operator that has an admin, when the admin user approves & when the admin user rejects. All 3 scenarios send the proper email to the correct user.

Issue #1433, opened 2024-04-19
Move DataGrid data transformation to Django Ninja schema


marcellmueller:
###Description of the Tech Debt###

There is a good description of the benefits of doing this in #755 which covers other pages which we do data transformation on the front end. 

- [ ] Industry User Admin DataGrid - `processExternalDashboardUsersTileData` in `@/app/utils/users/adminUserOperators.ts`. Also could likely remove this file altogether and move `getExternalDashboardUsersTileData` to `industry_user_admin/dashboard/users/page.tsx`
- [ ] Operations DataGrid - `formatOperationRows` in `routes/operations/Operations.tsx`
- [ ] Operations DataGrid - `formatUserOperatorRows` in `routes/access-requests/AccessRequests` - It looks like this function may be completely unnecessary - may just need to remove it with no extra data transormation needed.

###Tech Debt Triage#
The purpose of our technical debt triage process is to analyze technical debt to determine risk level of the technical debt and the value in tackling that technical debt.

Risk Value Scoring:

| Level  | Value |
| ------ | ----- |
| High   | 3     |
| Medium | 2     |
| Low    | 1     |

| Technical Debt - Risk Types                                                                                                   | Level | Value |
| ----------------------------------------------------------------------------------------------------------------------------- | ----- | ----- |
| Business Area Risk - Risk of business area visibility / damage to user experience                                             |   1    |       |
| Developer Fault Risk - How likely will this tech debt cause a future error related to coding on top of it                     |    3   |       |
| System Fault Risk - Risk of system errors or application downtime                                                             |    2   |       |
| Time Scale Risk - Compound risk effect if left alone. How much more difficult to fix or dangerous will this become over time? |    2   |       |
| Time Sink Risk - How much will this tech debt slow the development process down                                               |     2  |       |
| TOTAL SCORE:                                                                                                                  |    10   |       |

BCerki:
This may be addressed by the work on the part 2 dashboard

marcellmueller:
@BCerki is this still relevant? I'm putting a deletion candidate tag on it.

PR #1434, opened 2024-04-19
1234 util tests

marcellmueller:
Implements #1234 

Added tests most of the functions in the `/utils` folder

Re-added the `vitest-fetch-mock` library and added it to our mocks where it can be imported as `fetch`.

Please let me know if you can think of useful scenarios I haven't tested, especially for the action handler which is probably the most important util tested here 🙇 

What is not tested:
- the `jsonSchema` folder. These json schema files should probably live elsewhere and we aren't likely to test them on their own. They will get tested when we get to each form and are being tested in our e2e tests.
- the `/users/adminUserOperators.ts` file. It will likely get completely removed in #1433 and some of it is completely redundant and left over from early in the project (the `transformedTileData` map doesn't do anything anymore).

### DIFF
--- a/client/app/utils/getUUIDFromEndpoint.ts
+++ b/client/app/utils/getUUIDFromEndpoint.ts
@@ -1,0 +1,22 @@
+// 🛠️ Function to get the last non-empty segment as a UUID from an endpoint URL
+function getUUIDFromEndpoint(endpoint: string): string | null {
+  // Split the endpoint URL by '/'
+  const segments = endpoint.split("/");
+
+  // Filter out empty segments
+  const nonEmptySegments = segments.filter((segment) => segment.trim() !== "");
+
+  // Find the last non-empty segment
+  const lastSegment: string | null =
+    nonEmptySegments.length > 0
+      ? nonEmptySegments[nonEmptySegments.length - 1]
+      : null;
+
+  // Validate if the last segment is a UUID (a more permissive pattern)
+  const isUUID = /^[0-9a-fA-F]{32}$/.test(lastSegment as string);
+
+  // Return the last non-empty segment as a UUID or null if not a UUID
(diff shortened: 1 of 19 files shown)


client/app/utils/getUUIDFromEndpoint.ts
shon-button, line 1: @marcellmueller For your consideration, since the `function getUUIDFromEndpoint` is a solution to obtain a user's Keycloak `user_guid` BEFORE the nextauth authenticated JWT is created, for use in the `export async function actionHandler` that sends the `user_guid` in the Authorization header for API security; perhaps, we should create an allow list in  `function getUUIDFromEndpoint` of the endpoint urls that can get `user_guid` value from the url and not from the JWT.  These endpoint urls are just the endpoints in file `client/app/api/auth/[...nextauth]/route.ts`.  Note, if we do this change it will effect your test file `client/tests/utils/getUUIDFromEndpoint.test.ts`

marcellmueller: That sounds like a great idea! I think one of the silver linings in testing as tech debt is it allows us to take a fresh look at our code.

Looking at this I think I can implement this allow list similar to how we do it for the `getEnvValue` function, I will add this and write some tests and get back to you.

shon-button, line 1: @marcellmueller `registration/user/user-profile` should also be in the list, and perhaps take the opportunity to change the call to `registration/user/user-profile` in `client/app/api/auth/[...nextauth]/route.ts: 74` to actually include the user_guid:
```
        const response = await actionHandler(
          `registration/user/user-profile/${token.user_guid}`,
          "GET"
        );
```

shon-button (approved):

PR #1436, opened 2024-04-19
1014 automated email new operator

Sepehr-Sobhani:
[ISSUE 1014](https://github.com/bcgov/cas-registration/issues/1014)

## Notes:

- In order to run CHES emails locally, we need to obtain CHES credentials from the updated OBPS backend ENV file in 1Password.

- To see the actual email replace your email with `external_user_email_address` in `create_email_notification_record` function to receive the real email and confirm its contents.

- Refactored `save_operator` function to have a single responsibility (this was needed to be able to implement the feature)
- To test different scenarios (Email template is different based on different scenarios) of this feature:

1. Log in as `bc-cas-dev-three` -> Create new operator -> `bc-cas-dev-three` should receive confirmation email 
2. Log in as `bc-cas-dev-three` -> Create new operator -> Log in as IRC user -> Accept the newly created operator -> Accept the user operator -> `bc-cas-dev-three` should receive an approval email
3. Log in as `bc-cas-dev-three` -> Create new operator -> Log in as IRC user -> Decline the newly created operator -> `bc-cas-dev-three` should receive a rejection email.
4. Log in as `bc-cas-dev-three` -> Create new operator -> Log in as IRC user -> Accept the newly created operator -> Decline the user operator -> `bc-cas-dev-three` should receive a rejection email.


bc_obps/service/operator_service.py
```diff
@@ -127,17 +125,43 @@ def handle_parent_operators(
...
+            Operator: The updated operator object.
+        """
+        operator: Operator = OperatorDataAccessService.get_operator_by_id(operator_id)
         with transaction.atomic():
-
             operator.status = updated_data.status
             if operator.is_new and operator.status == Operator.Statuses.DECLINED:
                 # If the operator is new and declined, we need to decline all user operators tied to this operator
-                UserOperator.objects.filter(operator_id=operator_id).update(
+                user_operators_to_decline: QuerySet[UserOperator] = UserOperator.objects.filter(operator_id=operator_id)
```
BCerki, line 144: teachable moment for Brianna--what's the advantage of using `QuerySet`?

bc_obps/service/operator_service.py
```diff
@@ -127,17 +125,43 @@ def handle_parent_operators(
...
+                user_operators_to_decline.update(
                     status=UserOperator.Statuses.DECLINED,
                     verified_at=datetime.now(pytz.utc),
                     verified_by=user_guid,
                 )
+                for user_operator in user_operators_to_decline:
+                    user_operator.refresh_from_db()
```
BCerki, line 151: when/why do we need to refresh?

bc_obps/service/user_operator_service.py
```diff
@@ -78,14 +77,7 @@ def save_operator(updated_data: UserOperatorOperatorIn, operator_instance, user_
...
-        user_operator, created = UserOperator.objects.get_or_create(
-            user=user, operator=created_or_updated_operator_instance
-        )
-        if created:
-            user_operator.set_create_or_update(user.pk)
-        return {"user_operator_id": user_operator.id, 'operator_id': user_operator.operator.id}
+        return created_or_updated_operator_instance
 
     # brianna not sure how to fit this into which service
     # Used to show cas_admin the list of user_operators to approve/deny
```
BCerki, line 83:
```suggestion

```
Sorry about this one!

bc_obps/service/user_operator_service.py
```diff
@@ -164,8 +156,22 @@ def list_user_operators(page: int = 1, sort_field: str = "created_at", sort_orde
...
 
     @transaction.atomic()
-    def create_operator_and_user_operator(updated_data: UserOperatorOperatorIn, user_guid: UUID):
-        "Function to create a user_operator and an operator (new operator that doesn't exist yet)"
+    def create_operator_and_user_operator(updated_data: UserOperatorOperatorIn, user_guid: UUID) -> Dict:
+        """
+        Function to create a user_operator and an operator (new operator that doesn't exist yet).
+
+        Parameters:
+            updated_data: Updated data for the user operator.
```
BCerki, line 164: Since this function is about creation, maybe the arg should be `new_data` instead of `updated_data`? Updated implies editing

bc_obps/registration/tests/endpoints/test_operator_endpoints.py
```diff
@@ -227,11 +232,27 @@ def test_put_decline_new_operator(self):
...
+                mocker.call(
+                    AccessRequestStates.DECLINED,
+                    AccessRequestTypes.NEW_OPERATOR_AND_ADMIN
+                    if operator.created_by == user_operator.user
+                    else AccessRequestTypes.ADMIN,
+                    operator.legal_name,
+                    user_operator.user.get_full_name(),
+                    user_operator.user.email,
+                )
+            )
```
BCerki, line 251: Another teachable moment--what's happening here? Above I see that we're checking that the mock function was called three times after we hit the endpoint. What is this check after the API call for?

bc_obps/registration/tests/endpoints/test_useroperator_endpoints.py
```diff
@@ -487,6 +487,37 @@ def test_cas_admin_can_update_status_of_a_user_operator(self, mocker):
             user_operator.user.email,
         )
 
+    def test_cas_admin_can_update_status_of_a_user_operator_with_new_operator(self, mocker):
+        operator = operator_baker({'status': Operator.Statuses.APPROVED, 'is_new': False, 'created_by': self.user})
```
BCerki, line 491: If it's a new operator shouldn't `is_new` be true? Status is also probably pending?

BCerki (commented):
This is neat stuff, thanks for the intro to emails! My only overall note is about test coverage--I see a couple new tests, but they don't cover everything listed in the AC. Is that because we already had tests (e.g. for the confirmation email) from previous PRs?

I manually tested all the scenarios listed in the PR description and everything worked as expected.

bc_obps/service/operator_service.py
Sepehr-Sobhani: Using QuerySet as the type hint is more accurate than using List in this case. What we get from calling `.filter` is basically a QuerySet and by type hinting that we can get better code suggestions in vs code.

bc_obps/service/operator_service.py
Sepehr-Sobhani: I couldn't find a good doc to share with you about this case but in django `.update` performs a SQL UPDATE in the database and doesn't update instances using the ORM so we often don't get the updated value in Django after updating. TBH, I don't think we had any issue WO `refresh_from_db` but I only added that to make sure we always get the updated value in case we change the code later on.
This doc is somehow related though: https://docs.djangoproject.com/en/5.0/ref/models/instances/#:~:text=For%20example%2C%20to%20test%20that%20an%20update()%20call%20resulted%20in%20the%20expected%20update%2C%20you%20could%20write%20a%20test%20similar%20to%20this%3A

bc_obps/service/user_operator_service.py
Sepehr-Sobhani: Np at all.

bc_obps/registration/tests/endpoints/test_operator_endpoints.py
Sepehr-Sobhani: Here I'm making sure that we are calling those mocked calls with the correct args. (different user operators, and different access request types)

bc_obps/registration/tests/endpoints/test_useroperator_endpoints.py
Sepehr-Sobhani: Sorry, copy/pasting 🤦‍♂️

Sepehr-Sobhani: actually, this is correct 😆  because here we are testing the user operator status change and not the operator change so we have to mark the operator as `is_new=False` and `status=APPROVED` so we can bypass the below part and can get to the email sending part:

```python
# We can't update the status of a user_operator if the operator has been declined or is awaiting review, or if the operator is new
        if operator.status == Operator.Statuses.DECLINED or operator.is_new:
            raise Exception("Operator must be approved before approving or declining users.")
```

bc_obps/service/user_operator_service.py
Sepehr-Sobhani: Valid point 👍

Sepehr-Sobhani:
> This is neat stuff, thanks for the intro to emails! My only overall note is about test coverage--I see a couple new tests, but they don't cover everything listed in the AC. Is that because we already had tests (e.g. for the confirmation email) from previous PRs?
> 
> I manually tested all the scenarios listed in the PR description and everything worked as expected.

I believe I've addressed all scenarios for `Confirming`, `Declining`, and `Approving`. However, I fully trust your judgment. Could you please point out any test cases that I might have overlooked?

bc_obps/service/operator_service.py
BCerki: Good to know, thanks! I'll try to remember to refresh whenever I use update

bc_obps/registration/tests/endpoints/test_useroperator_endpoints.py
BCerki: Right! In that case, maybe we don't need two tests then?

Sepehr-Sobhani:
> test_create_operator_and_user_operator_with_parent_operators

Thanks for noticing this 🙏

BCerki (approved):
Nice work! Just needs a rebase and happo rerun (this one actually has different data but hoping it's just flake: https://happo.io/a/336/p/1755/compare/5eb468f65dc41f23f4533adb544564de66b886d3/404b1e2ff888e4a87d5a13ffceb8aaa7bc6cabe0?t=diffs&target=firefox#Operators%5C%20Details%5C%20Page%5C%20cas_analyst-pending)

PR #1443, opened 2024-04-22
fix owasp-zap scan ci issue

marcellmueller:
Looks like their DockerHub images went down, so switching to GHCR images since I saw they are hosted there as well:

https://www.zaproxy.org/blog/2023-06-13-ghcr-docker-images/

Fixed run here: 
https://github.com/bcgov/cas-registration/actions/runs/8790774852/job/24123733697?pr=1443


shon-button (approved):
Brilliant!
👌💡

Issue #957, opened 2024-02-22
Refactor API user operator endpoint test files[11]


Sepehr-Sobhani:
###Description of the Tech Debt###

Blocked by #1061 
This will have some conflicts with the operator refactor because some of the tests were in the user operator test file.

The API endpoint tests are hard to read and navigate due to their size. It would be beneficial to divide the tests into smaller classes that handle a specific HTTP method or endpoint. This will make the tests easier to manage.

This should not be done until #952 and #1061 are complete to avoid re-work.

NEW: A test refactor AC has been added to some other cards, so this card is only about refactoring the user_operator endpoint tests.

###Tech Debt Triage#
The purpose of our technical debt triage process is to analyze technical debt to determine risk level of the technical debt and the value in tackling that technical debt.

Risk Value Scoring:

| Level  | Value |
| ------ | ----- |
| High   | 3     |
| Medium | 2     |
| Low    | 1     |

| Technical Debt - Risk Types                                                                                                   | Level | Value |
| ----------------------------------------------------------------------------------------------------------------------------- | ----- | ----- |
| Business Area Risk - Risk of business area visibility / damage to user experience                                             |       |  1     |
| Developer Fault Risk - How likely will this tech debt cause a future error related to coding on top of it                     |       |   3    |
| System Fault Risk - Risk of system errors or application downtime                                                             |       |   1    |
| Time Scale Risk - Compound risk effect if left alone. How much more difficult to fix or dangerous will this become over time? |       |   3    |
| Time Sink Risk - How much will this tech debt slow the development process down                                               |       |   3    |
| TOTAL SCORE:                                                                                                                  |       |    11   |

nanyangpro:
Re: Refinement - Another ticket for naming conventions would be created. Thanks @andrea-williams!

BCerki:
Update: I split this card so it's only about user operator testing

PR #1444, opened 2024-04-22
refactor: separate user operator api endpoint tests

marcellmueller:
Implements #957 

- Tests mirror the files in `/api/user_operator` and `/api/select_operator` except each test filename starts with `test_`
- The only file that doesn't mirror those files is `test_authorization` which has the `test_user_operator_unauthorized_users_cannot_get`, `test_user_operator_unauthorized_users_cannot_post`, `test_user_operator_unauthorized_users_cannot_put` files. I wasn't sure where to put them and thought it made sense to keep them separate.


marcellmueller:
@Sepehr-Sobhani @BCerki I am a bit out of the loop with the API endpoint refactoring and there isn't an example of a test refactor yet. 

Is this what you had in mind for the tests? Also I put the endpoint auth tests in their own file. Thanks for taking a quick look and any suggestions are appreciated :)

Sepehr-Sobhani:
> @Sepehr-Sobhani @BCerki I am a bit out of the loop with the API endpoint refactoring and there isn't an example of a test refactor yet.
> 
> Is this what you had in mind for the tests? Also I put the endpoint auth tests in their own file. Thanks for taking a quick look and any suggestions are appreciated :)

Thanks for the refactor, @marcellmueller  —it looks great! I suggest mirroring the file structure in `api/user_operator` to simplify finding tests related to specific endpoints. I hope this doesn't add too much extra work to this PR.

Sepehr-Sobhani (approved):
Super 🤩

Issue #1183, opened 2024-03-20
Front end cleanup


BCerki:
###Description of the Tech Debt###

Will have lots of conflicts with #1385 and @joshgamache 's file-moving PR

The Next App router creates makes files routable when they're called `route.tsx` or `page.tsx` (https://nextjs.org/docs/app/building-your-application/routing/pages-and-layouts).

Our app has a folder called `routes` and files called `Page.tsx` (note the capital) which are not routable, but because the names are so similar to what Next uses, there can be confusion. (Also, older versions of Next used a folder called `pages` to create routes, so there's a lot of opportunity for naming confusion.)

Suggestions to improve front-end clarity:

For the industry_user roles (files in `app/(authenticated)/bceidbusiness`):
- [x] Routing is self-documenting (you can look at the file structure to see how to write a URL, things in () are skipped, things in [] are dynamic). We have `withAuthorization` middleware that cuts off the `(authenticated/bceidbudiness/industry_user)` bit, so add a section in the `developer-guide.md` about routing that explains why the frontend URLs look different from the file structure
- [ ] most of the time the `[operation]` parameter is the operation's UUID, but when a operation is first being created and doesn't have an UUID yet, the url shows the word "create". I find this a little confusing (especially in the frontend files if we use `operationId` to refer to this parameter), but that would be a huge refactor, so maybe the solution is just to mention this in the routing section of the `developer_guide.md`
- [x] in the components/routes folder, rename anything called `Page.tsx` to the name of the React component within it (if component names could be improved, feel free to edit)
- [x] in `components/routes`, move any form-related components to `components/form` folder. Note: this change will affect import paths throughout the app, but not any routes, since components were never used for routing
- [ ] rename the `routes` folder to something else that doesn't imply routing (NOT `pages`, will interfere with Next's routing/cause confusion). Maybe `screens` or `body` or something? 
- [ ] move `user-operator/[id]/[formSection]` (this is the user-operator form) outside of `select-operator` into the `dashboard` folder. (Once the user reaches the form, they've already selected their operator; they're now adding or editing it.) Note: this will require updating routes within the app, and a possible way to handle this could be to make a `pageRoutes` folder like CIF has: https://github.com/bcgov/cas-cif/blob/develop/app/routes/pageRoutes.ts. This change may also require a tweak to the breadcrumbs (could potentially rename `user-operator` to `my-operator`)

For the cas roles (files in `app/(authenticated)/idir`):
[...]

PR #1487, opened 2024-04-24
Front end cleanup

BCerki:
https://github.com/bcgov/cas-registration/issues/1183

This is PR 1/2 for this card. This PR only moves files and updates imports

In this refactor:
- cleaned up components/folders that weren't being used
- All files in /routes called Page.tsx
- Made component folders for files that were previously in /routes but that aren't actually routes or pages (e.g. the buttons to accept/decline an operation are just a component)
- I'm still not convinced that the /routes folder is the way to go. We'll have to update it anytime we change our routes, and in some cases we have the same route (users) for different user roles and the different roles need to see different pages at the same route.

PR #2 will:
- update some file and component names
- consolidate some of the components (so we don't have three operations components that don't do anything except import each other.)
- possibly do something about the routes folder


client/app/components/routes/operations/Page.tsx
```diff
@@ -2,12 +2,12 @@
...
 import { Button } from "@mui/material";
 import { Suspense } from "react";
-import Operations from "@/app/components/routes/operations/Operations";
 import Loading from "@/app/components/loading/SkeletonGrid";
 import { getServerSession } from "next-auth";
 import { authOptions } from "@/app/api/auth/[...nextauth]/route";
 import Note, { registrationRequestNote } from "../../datagrid/Note";
-import { OperationsSearchParams } from "@/app/components/routes/operations/types";
+import { OperationsSearchParams } from "../../operations/types";
+import Operations from "../../operations/Operations";
```
marcellmueller, line 10: I think we should avoid using relative imports. Using it when it's in the same directory ie `./Operations` isn't too bad though I think going up multiple directories is confusing.  

Suggest to change all relative imports to aliased imports which are easier to read and more informative ie `@/app/components/operations/Operations`.

marcellmueller (commented):
Looking good, nice work! Just one other suggestion besides the aliased imports.

While doing the `user operator` api test refactor I noticed our api now separates user operator stuff and `select operator`. Should we have a `selectOperator` directory and keep that stuff separate?

client/app/components/routes/operations/Page.tsx
BCerki: I've fixed all the multiple directories ones but left ./

BCerki:
> Looking good, nice work! Just one other suggestion besides the aliased imports.
> 
> While doing the `user operator` api test refactor I noticed our api now separates user operator stuff and `select operator`. Should we have a `selectOperator` directory and keep that stuff separate?

I'm going to look into this in PR #2. Right now in the app folder, we've got the user operator pages within select-operator, e.g. client/app/(authenticated)/bceidbusiness/industry_user/dashboard/select-operator/(request-access)/user-operator/[id]/[formSection], so everything's mixed together. I think it probably does make sense to separate user operator and select operator in both in the app folder and in the components (there's an AC in the card to look into it).

client/app/components/routes/operations/Page.tsx
marcellmueller: I can still see about 6 imports in the changee files here that use `../../componentname`, other than that this is good to go!

docs/developer-guide.md
```diff
@@ -11,7 +11,7 @@ Architectural design using Next.js 14 includes employing a hierarchical structur
 Next.js 14 foundation is the App Router which boasts a range of features, including:
 
 [Routing](https://nextjs.org/docs/app/building-your-application/routing)
-Creating routes involves creating folders within the app directory and nesting a page.tsx file within the appropriate folder to define your route.
+Creating routes involves creating folders within the app directory and nesting a page.tsx file within the appropriate folder to define your route. Note: our `withAuthorization` middleware removes the first three folders (e.g. authenticated/bceidbudiness/industry_user) from the route, so all the URLs within the app will start with `dashboard` instead of `authenticated/.../dashboard`.
```
marcellmueller, line 14: Thanks for this!

marcellmueller (approved):
Nice job on this refactor, feels good! 🥇

PR #1495, opened 2024-04-24
Automate make release target

Sepehr-Sobhani:


dleard (changes requested):
Tested it & it looks good! Just one change request for the reporting django app

dleard (approved):

PR #1505, opened 2024-04-25
chore: add facilities page, add BCeID login

BallardRobinett:
card: https://github.com/bcgov/cas-reporting/issues/123

Adding a placeholder page to fetch and display operations and facilities for the user viewing the page. If a user without operators is signed in, the space after ```operators:``` will be blank.

To test:

1. navigate to ```/bc_obps/``` in the terminal and follow the instructions in ```/docs/developer-environment-setup.md/``` to get it running
2. in another terminal run ```yarn dev``` from ```/client/```, open on localhost:3000 and sign in using BCeID. Go through first time setup with one of the accounts in the 1password (bc-cas-dev etc)
3.  in another terminal access the django shell by navigating to ```/bc_obps/``` and running ```python manage.py shell```
4. in the django shell, modify the ```app_role``` value of your user from before. The commands I used to do this are:

from registration.models import User
from registration.models import AppRole
my_user = User.objects.get(first_name=’cas-dev-three’)
cas_admin_role = ppRole.objects.get(role_name=’cas-admin’)
my_user.app_role = cas_admin_role
my_user.save()

replacing 'cas-dev-three' with whatever your user's first name is. You can check this with the command

[print(”name: “ + user.first_name + “ role: “ + str(user.app_role)) for user in User.objects.all()]

5. in another terminal, navigate to ```/bciers/``` and run ```yarn dev```, and open on localhost:3000 (or 3001 if 3000 is in use by /client/). Log in with the same BCeID account. You should be redirected to a welcome page at localhost:3000/dashboard. Click the link to the facilities page. This should display a wall of text with the operator data for this user.


bciers/apps/reporting/src/app/(authenticated)/authenticated/facilities/page.tsx
```diff
@@ -0,0 +1,37 @@
+"use client"
+
+import React, { useState, useEffect } from 'react';
+import { actionHandler } from "@/app/utils/actions";
+
+const Page: React.FC = () => {
+  const [message, setMessage] = useState<string>('');
```
pbastia, line 7: At the risk of having blinking pages initially before we have all the `<Suspense> ... </ >` straightened out, we shouldn't create dynamic states and unnecessary effects when we don't need them.
I'd suggest following the pattern used in the registration app for now, like [here](https://github.com/bcgov/cas-registration/blob/23b68db8fe1c86a450619f62e1010b8dceb9deb8/client/app/components/operations/Operation.tsx#L172)

Benefit is also to not have a bunch of async methods firing everywhere, but to have the page data loading happening in a more controlled way (and rendered server-side as much as possible).

bciers/apps/reporting/src/app/(authenticated)/authenticated/facilities/page.tsx
```diff
@@ -0,0 +1,37 @@
...
+    };
+  
+    fetchData();
+  }, []);
+
+  return (
+    <div>
+      <h1>Facilities and operators page</h1>
+      <h2>Operators:</h2>
+      <p>{message}</p>
```
pbastia, line 32: probably we can call this `operator_list` or something similar

bciers/apps/reporting/src/app/(authenticated)/authenticated/facilities/page.tsx
```diff
@@ -0,0 +1,45 @@
...
+  return (
+    <div>
+      <h1>Facilities and operators page</h1>
+      <h2>Operators:</h2>
+      <OperatorsDisplay fetchOperators={fetchOperators} />
+    </div>
+  );
+};
+
+const OperatorsDisplay: React.FC<{ fetchOperators: () => Promise<any> }> = ({ fetchOperators }) => {
```
pbastia, line 29: - no reason to have the function as a parameter here, it's not required to be in the scope of the component
- make the component an async function

Something like:
```typescript
const OperatorsComponent = async () => { ... }
```

bciers/apps/reporting/src/app/(authenticated)/authenticated/facilities/page.tsx
```diff
@@ -0,0 +1,45 @@
...
+  const [operators, setOperators] = React.useState<any>(null);
+
+  React.useEffect(() => {
+    fetchOperators().then((operatorsData) => {
+      setOperators(operatorsData);
+    });
+  }, [fetchOperators]);
```
pbastia, line 36:
```suggestion
const operators = await fetchOperators();
```

pbastia (changes requested):
More comments! Hopefully it's helpful

pbastia (approved):
Looks great!! Once the Operators component has meaningful stuff in it we can probably look at putting it somewhere else, but for now this is great.

The PR will need a rebase

mikevespi (commented):
🥳

PR #1506, opened 2024-04-26
Vitesting form components

BCerki:
https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=57614651

Sonarcloud warnings are on tests so will ignore if reviewer agrees


client/tests/components/form/MultiStepFormBase.test.tsx
```diff
@@ -0,0 +1,275 @@
...
+    //     resolve = _resolve;
+    //   })
+    // );
+    render(<MultiStepFormBase {...defaultProps} disabled={false} />);
+    const saveAndContinueButton = screen.getByRole("button", {
+      name: /Save and Continue/i,
+    });
+    await fireEvent.click(saveAndContinueButton);
+    await waitFor(() => {
+      expect(saveAndContinueButton).toBeDisabled();
```
BCerki, line 241: I wasn't able to get this working. This was the best example I found so far but it didn't work: https://github.com/orgs/react-hook-form/discussions/8866

client/app/components/form/MultiStepButtons.tsx
```diff
@@ -4,19 +4,19 @@ import { Button } from "@mui/material";
...
 
-interface SubmitButtonProps {
+interface MultiStepButtonsProps {
   baseUrl: string;
   cancelUrl: string;
   classNames?: string;
   disabled?: boolean;
   isSubmitting: boolean;
-  step: number;
+  step: number; // 0-indexed
```
BCerki, line 13: My tests kept failing because I didn't realize this, so added some comments as a reminder

client/tests/components/form/MultiStepButtons.test.tsx
```diff
@@ -0,0 +1,197 @@
...
+  it("Save and Continue has correct link", () => {
+    render(<MultiStepButtons {...defaultProps} step={1} />);
+
+    expect(screen.getByRole("link", { name: "Next" })).toHaveAttribute(
+      "href",
+      "anything.com/3",
+    );
+  });
+  it("Submit button navigates correctly", () => {
+    // brianna can we actually test this without the form?
```
BCerki, line 195: thoughts?

client/tests/components/form/MultiStepHeader.test.tsx
```diff
@@ -0,0 +1,63 @@
...
+  it("shows the line between steps", () => {
+    render(<MultiStepHeader {...defaultProps} />);
+    expect(screen.getAllByRole("separator")).toHaveLength(2);
+  });
+
+  it("does something to do with the last step", () => {
+    render(<MultiStepHeader {...defaultProps} />);
+    expect(screen.getAllByRole("brianna fix this test")).toHaveLength(2);
+
+    // B what is this doing, how should I be testing it?
```
BCerki, line 55: thoughts?

client/tests/components/form/MultiStepButtons.test.tsx
BCerki: It's tested in the form component test with a mock test schema, yes. I don't think we need to test it in this component because it's not doing anything here (unlike something like the cancel button, which I can test has the correct link) but wanted a second opinion

client/tests/components/form/MultiStepFormBase.test.tsx
BCerki: I've used `tobeDisabled` as well; I think the issue here is timing. Have you come across good alternatives to `waitFor`?

client/tests/components/form/MultiStepHeader.test.tsx
BCerki: Gotcha, thanks

client/tests/components/form/MultiStepButtons.test.tsx
marcellmueller: Oh sorry that makes sense, I forgot we handle that stuff in say `OperationsForm`. Seems out of scope for this test 👍

client/app/components/form/MultiStepButtons.tsx
BCerki: https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=61566260

docs/developer-guide.md
```diff
@@ -159,6 +159,14 @@ actionHandler.mockReturnValueOnce({
...
 ```
 
+To maintain test isolation, we should reset (note that this is different from clear) mocks before every test:
+
+````
+beforeEach(() => {
+    vi.resetAllMocks();
+
+  });```
+
```
marcellmueller, line 169: Thanks for this, I think this is a great idea and have refactored some of my tests. However I'm not sure if we need to add it before every test or just tests that we use mocks for? 

I don't feel strongly either way, do you think it's useful to do before even simple tests for example some of the util tests?

marcellmueller: I think it would only possibly affect tests with functions/components that use those imports.

Another file I looked at refactoring is this [actions.test.ts](https://github.com/bcgov/cas-registration/blob/develop/client/tests/utils/actions.test.ts) test. The global `consoleMock` and `sentryMock` that we call throughout the tests will break if run `resetAllMocks()` and we can't put them in the `beforeEach` since they are called through the test. 

Anyways I think this is a great best practice that we might have to deviate from in certain tests like the actions test unless you have any refactoring ideas.

client/app/components/form/MultiStepButtons.tsx
marcellmueller: Thanks, I think the `MultiStepFormBase` was refactored to make this easier to follow and we didn't touch this 😬

client/tests/components/form/FormBase.test.tsx
```diff
@@ -0,0 +1,98 @@
...
+      type: "string",
+    },
+  },
+};
+
+const testUiSchema = {
+  type: {
+    "ui:widget": "TextWidget",
+  },
+};
```
marcellmueller, line 21: nit: you don't need to declare `TextWidget` as it is the default widget for string and number fields. Though it does make sense to make sure the `uiSchema` prop is working, it could be worth adding an item to `uiSchema` that will actually change something. I removed this and the `uiSchema` props and the test still passed.

client/tests/components/form/MultiStepButtons.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+import { Session } from "@/tests/setup/types";
+
+const defaultProps = {
+  allowBackNavigation: true,
+  baseUrl: "anything.com",
+  cancelUrl: "https://www.test.com/",
+  classNames: "garbage",
+  disabled: true,
+  isSubmitting: false,
+  step: 0, // steps are 0-indexed in the props but 1-indexed in the app
```
marcellmueller, line 15: 🙈

docs/developer-guide.md
BCerki: I'm going to update these docs to talk about reset vs clear because I think there are cases, like actions, where one is preferred over the other. I found this in the actions file:

```
// disable the global mock since we are testing actions here
vi.unmock("@/app/utils/actions");
``` so it looks like tests could be using mocks without explicit imports (?), so I'll mention that too.

File: client/tests/components/form/FormBase.test.tsx (hunk 6 above)
Comment (BCerki) [reply client/tests/components/form/FormBase.test.tsx:21]: I added a test to check if the phone number widget's +1 was visible

File: client/tests/components/form/MultiStepFormBase.test.tsx [hunk 8]
```diff
@@ -0,0 +1,304 @@
...
+    );
+    expect(screen.getByRole("button", { name: /Cancel/i })).not.toBeDisabled();
+    expect(screen.getByRole("link", { name: /Cancel/i })).toHaveAttribute(
+      "href",
+      "cancelurl.com",
+    );
+    expect(
+      screen.getByRole("button", { name: /test submit button text/i }),
+    ).not.toBeDisabled();
+    await fireEvent.click(submitButton);
```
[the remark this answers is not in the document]
BCerki: Where/how are you seeing that warning?

client/tests/components/form/MultiStepHeader.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+  steps: ["step one", "step two", "step three"],
+};
+
+describe("The MultiStepHeader component", () => {
+  beforeEach(() => {
+    vi.resetAllMocks();
+  });
+  it("reduces the width of the title if there are more than 2 steps", () => {
+    render(<MultiStepHeader {...defaultProps} />);
+    expect(screen.getAllByTestId("multistep-header-title")[0]).toHaveClass(
```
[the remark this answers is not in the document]
BCerki: I didn't test for grow-0; just that the header width is reduced

client/tests/components/form/MultiStepFormBase.test.tsx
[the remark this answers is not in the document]
marcellmueller: It's just a regular typescript/eslint warning, though it's more visible in vim than VSCode:

Await has a small underline:
Screenshot 2024-05-03 at 8 52 25 AM
When I hover over the 3 dot underline:
Screenshot 2024-05-03 at 8 52 36 AM

marcellmueller (approved):
Nice work on this! 🍦

PR #1500, opened 2024-04-24
Feat/email boro id non opt/1015

andrea-williams:
[ISSUE 1058](https://github.com/bcgov/cas-registration/issues/1058)
[ISSUE 1015](https://github.com/bcgov/cas-registration/issues/1015)

PR #1509, opened 2024-04-26
1058 boro id opted in operations

Sepehr-Sobhani:
[ISSUE 1058](https://github.com/bcgov/cas-registration/issues/1058)


Sepehr-Sobhani:
Addressed in: #1500

Issue #1300, opened 2024-04-08
Use zoneinfo instead of pytz


BCerki:
###Description of the Tech Debt###

We should replace pytz with "zoneinfo" since we're using Python 3.9 and later
See sonarcloud for details: 
![Image](https://github.com/bcgov/cas-registration/assets/77306764/e2061d89-63a6-44b8-8905-598698c2251e)

###Tech Debt Triage#
The purpose of our technical debt triage process is to analyze technical debt to determine risk level of the technical debt and the value in tackling that technical debt.

Risk Value Scoring:

| Level  | Value |
| ------ | ----- |
| High   | 3     |
| Medium | 2     |
| Low    | 1     |

| Technical Debt - Risk Types                                                                                                   | Level | Value |
| ----------------------------------------------------------------------------------------------------------------------------- | ----- | ----- |
| Business Area Risk - Risk of business area visibility / damage to user experience                                             |       |       |
| Developer Fault Risk - How likely will this tech debt cause a future error related to coding on top of it                     |       |       |
| System Fault Risk - Risk of system errors or application downtime                                                             |       |       |
| Time Scale Risk - Compound risk effect if left alone. How much more difficult to fix or dangerous will this become over time? |       |       |
| Time Sink Risk - How much will this tech debt slow the development process down                                               |       |       |
| TOTAL SCORE:                                                                                                                  |       |       |

Sepehr-Sobhani:
Sonarcloud details:


> In Python 3.9 and later, the zoneinfo module is the recommended tool for handling timezones, replacing the pytz library. This recommendation is based on several key advantages.
> 
> First, zoneinfo is part of Python’s standard library, making it readily available without needing additional installation, unlike pytz.
> 
> Second, zoneinfo integrates seamlessly with Python’s datetime module. You can directly use zoneinfo timezone objects when creating datetime objects, making it more intuitive and less error-prone than pytz, which requires a separate localize method for this purpose.
> 
> Third, zoneinfo handles historical timezone changes more accurately than pytz. When a pytz timezone object is used, it defaults to the earliest known offset, which can lead to unexpected results. zoneinfo does not have this issue.
> 
> Lastly, zoneinfo uses the system’s IANA time zone database when available, ensuring it works with the most up-to-date timezone data. In contrast, pytz includes its own copy of the IANA database, which may not be as current.
> 
[...]

PR #1514, opened 2024-04-26
chore: replace pytz with zoneinfo

Sepehr-Sobhani:
[ISSUE 1300](https://github.com/bcgov/cas-registration/issues/1300)

### DIFF
--- a/bc_obps/poetry.lock
+++ b/bc_obps/poetry.lock
@@ -1955,4 +1955,4 @@ brotli = ["Brotli"]
 [metadata]
 lock-version = "2.0"
 python-versions = "^3.9"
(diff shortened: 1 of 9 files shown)


marcellmueller (commented):
Nice PR, happy to get rid of these warnings about `pytz`!

I was surprised to see that it is saving in a slightly different date time format:
`2024-04-30 08:08:19.537118-07` instead of the previous timestamp`2024-1-13T15:27:00.000Z`. 

It seems to be fine, even on our frontend conversion so I think this is safe:
Screenshot 2024-04-30 at 8 21 26 AM

Do you think we should save timestamps in the same format for consistency sake or is this okay?

marcellmueller (approved):
Oh sorry, I switch to another branch to continue my work and checked the time zones as I updated some things and it is consistent. It's just the fixtures that use the unix style timestamp so I think this is good to go :)

Issue #114, opened 2023-09-15
Use ruff linter


Sepehr-Sobhani:
###Description of the Tech Debt###
We have added `pylint` as the Python linter, but it may be more efficient to use `ruff` instead, as it is faster and can accelerate our CI process.

Link to `ruff` pre-commit hook: https://beta.ruff.rs/docs/usage/

andrea-williams:
@Sepehr-Sobhani do we still need this ticket?

Sepehr-Sobhani:
@andrea-williams Yes, ruff is much faster and can speed up our local development and CI.
![Screenshot 2024-03-26 at 8 41 46 AM](https://github.com/bcgov/cas-registration/assets/52381975/94e25997-ca25-4868-82c8-bdef24b756b7)

PR #1540, opened 2024-04-29
114 add ruff

Sepehr-Sobhani:
[ISSUE 114](https://github.com/bcgov/cas-registration/issues/114)

ruff rules if interested: https://docs.astral.sh/ruff/rules/

**Notes:**

To update your packages, please run `poetry install`. To test this, you can either run `poetry run ruff` from within the `bc_obps` directory or `pre-commit run -a` from the root directory. Please note that we do not use the ruff formatter as it is being handled by `black`.


andrea-williams (commented):
Holy heck, ruff is FAST! 

Are we going to be using ruff strictly for linting, or for formatting too? I ran `ruff check` locally on this branch and it was happy, but when I ran `ruff format` there were >100 files changed

andrea-williams (approved):

PR #1546, opened 2024-04-29
Fix django sentry when not passing `DEBUG` env variable

Sepehr-Sobhani:
**Note:**

Sentry was not able to capture backend errors since we don't pass `DEBUG` env variable in our namespaces. This fix will run the sentry on the backend again.

### DIFF
--- a/bc_obps/bc_obps/settings.py
+++ b/bc_obps/bc_obps/settings.py
@@ -172,7 +172,7 @@ DATA_UPLOAD_MAX_MEMORY_SIZE = 20000000
 # Only enable sentry in production
 SENTRY_ENVIRONMENT = os.environ.get('SENTRY_ENVIRONMENT')
 SENTRY_TRACE_SAMPLE_RATE = os.environ.get('SENTRY_TRACE_SAMPLE_RATE')
-if SENTRY_ENVIRONMENT == 'prod' and DEBUG == 'False':
+if SENTRY_ENVIRONMENT == 'prod':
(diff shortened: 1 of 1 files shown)


bc_obps/bc_obps/settings.py
```diff
@@ -172,7 +172,7 @@
 # Only enable sentry in production
 SENTRY_ENVIRONMENT = os.environ.get('SENTRY_ENVIRONMENT')
 SENTRY_TRACE_SAMPLE_RATE = os.environ.get('SENTRY_TRACE_SAMPLE_RATE')
-if SENTRY_ENVIRONMENT == 'prod' and DEBUG == 'False':
+if SENTRY_ENVIRONMENT == 'prod':
```
BCerki, line 175: I don't think I understand the problem/fix here. If we were never passing a DEBUG variable then DEBUG would be None which is still falsey in python right?

Sepehr-Sobhani: We are strictly comparing the variable `DEBUG` with the string `"False"`, but since we never pass `DEBUG` in any of our environment variables, it will always be `None`. Therefore, it doesn't matter to keep this extra check here. We are already making sure that the `SENTRY_ENVIRONMENT` is set to `"prod"`, which should be sufficient.

BCerki: Gotcha, thanks!

BCerki (approved):

Issue #313, opened 2023-11-23
Update ERD as part of pre-commit hooks


Sepehr-Sobhani:
###Description of the Tech Debt###

We utilize the `django-diagram` package to produce ERD diagrams that are tailored to our Django models. It would be helpful to have a feature in our `.pre-commit-config.yaml` file that verifies whether our ERD aligns with our data models and automatically updates it if it doesn't.

PR #1552, opened 2024-04-30
313 update erd diagrams as a pre-commit hook

Sepehr-Sobhani:
[ISSUE 313](https://github.com/bcgov/cas-registration/issues/313)

**Notes:**

To test this, change a data model field, make the migration, and run `pre-commit run -a` from the root directory. You should see the updated erd.


BCerki (approved):
This is really cool, thanks for putting it together! I ran pre-commit locally and it updated the ERDs for me (it even worked if I just edited the model files and forgot to make migrations). Looks good in CI too. I made a commit and then reverted it to test CI; if you'd prefer to drop it, feel free (just didn't want to do any destructive rebasing on someone else's branch).

PR #1581, opened 2024-05-02
Chore/registration nx merge

dleard:
Moves the /client directory into /bciers/apps/registration to be housed in the nx monorepo alongside reporting.

#### Migration steps for in-progress PRs:
1. Stash or reset all your changes.
2. Pull the latest `develop` branch, post-Nx-merge.
3. In the root directory of the repository, run `asdf current`, install or update any local tools as needed.
4. Run `corepack enable`.
5. Run `yarn` to fetch the version of yarn needed and update root packages.
6. Navigate to `bciers`.
7. Run `yarn` to fetch monorepo packages.
8. Copy any `.env`, `.env.local` and any other ignored local files you require from the old `../client` directory to `/bciers` (the Monorepo root).
9. Run `nx run registration:dev`
10. Anything you would have done in `/client` you can now do in `/bciers/apps/registration`, with the exception of adding packages. See the updated developer documentation.

#### Additional work needed
Here are some things where I think we need some follow up work:
- shared lib for types (there are duplicate next-auth.d.ts types in the 2 projects)
- there's no local-app-run for reporting in CI
- update things in registration to model the “nx monorepo” patterns. The reporting structure is still quite different than what's inside the apps/registration directory. There is some additional refactoring here that could be done to make this more in line with the nx patterns.
- I see that the backend-tests CI step needs the frontend-build...does it actually need it?


Sepehr-Sobhani:
> * I see that the backend-tests CI step needs the frontend-build...does it actually need it?

I don't think we need it.

BCerki (approved):

PR #1618, opened 2024-05-07
Fix/ci issues

joshgamache:
CI had issues brought about by bcgov/cas-obps#257. Updated actions and workflows to get some passing again. 

## Changes 🚧

- Updated any yarn commands to use `--immutable`.
- Removed vestigial files from the client repo transfer.
- Removed unneded steps that referenced old commands or folders.
- Removed old build, now supplanted by its own Nx build.

## To test 🔬

- CI should pass?

## Notes 📝

Digging into this, we'll likely benefit from some further edits down the line to the CI related to Nx. Nothing critical, but switching to us to some of it's tooling for E2E, linting, testing, and it's caching should help with speed and flakiness. Some of this should be reflected in tickets spawned from bc-gov/cas-reporting#174.


.github/workflows/test.yaml
```diff
@@ -27,15 +27,22 @@ jobs:
...
+  nx-tests:
     needs: install-dev-tools
     runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ./bciers
     steps:
-      - uses: actions/checkout@v4
+      - name: Checkout
+        uses: actions/checkout@v3
```
BCerki, line 38: Do we have to go to v3? Marcel updated us to v4 awhile ago

BCerki (commented):
Pre-commit is failing but otherwise looking great! Thanks for sorting all this CI stuff out!

.github/workflows/test.yaml
shon-button, line 1: @joshgamache as mentioned on Teams, for `e2e-tests` [see example](https://github.com/bcgov/cas-registration/blob/debug/1223-vitest-form-components/.github/workflows/test.yaml)  to successfully upload e2e report artifact on e2e failures.

.github/workflows/test.yaml
joshgamache: Thanks for the catch. I was moving a step from one place to anther that just happens to use the same action at a different version!

.github/workflows/test.yaml
```diff
@@ -27,15 +27,22 @@ jobs:
...
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
       - name: dev env setup
         uses: ./.github/actions/dev-env-setup
-      - run: yarn test
-        working-directory: ./client
+      - name: Run Nx Tests
+        run: yarn nx run-many -t test --skipNxCache
```
mikevespi, line 44: Let's get a card to look at the caching, I think this is one of the places that Nx can really improve our CI run times

mikevespi (approved):
:shipit: Nice work! Let's confirm that the e2e fail is indeed being handled in another PR and everything else is looking good 🥳

PR #1620, opened 2024-05-07
chore: create placeholder page for API results, add BCeID login

BallardRobinett:
Recreated this PR off develop after nx migration

card: https://github.com/bcgov/cas-reporting/issues/123

Adding a placeholder page to fetch and display operations and facilities for the user viewing the page. If a user without operators is signed in, the space after operators: will be blank.

To test:

    navigate to /bc_obps/ in the terminal and follow the instructions in /docs/developer-environment-setup.md/ to get it running
    in another terminal run yarn dev from /client/, open on localhost:3000 and sign in using BCeID. Go through first time setup with one of the accounts in the 1password (bc-cas-dev etc)
    in another terminal access the django shell by navigating to /bc_obps/ and running python manage.py shell
    in the django shell, modify the app_role value of your user from before. The commands I used to do this are:

from registration.models import User
from registration.models import AppRole
my_user = User.objects.get(first_name=’cas-dev-three’)
cas_admin_role = ppRole.objects.get(role_name=’cas-admin’)
my_user.app_role = cas_admin_role
my_user.save()

replacing 'cas-dev-three' with whatever your user's first name is. You can check this with the command

[print(”name: “ + user.first_name + “ role: “ + str(user.app_role)) for user in User.objects.all()]

    in another terminal, navigate to /bciers/ and run yarn dev, and open on localhost:3000 (or 3001 if 3000 is in use by /client/). Log in with the same BCeID account. You should be redirected to a welcome page at localhost:3000/dashboard. Click the link to the facilities page. This should display a wall of text with the operator data for this user.

### DIFF
--- a/bciers/apps/reporting/src/app/(authenticated)/authenticated/facilities/Operators.tsx
+++ b/bciers/apps/reporting/src/app/(authenticated)/authenticated/facilities/Operators.tsx
@@ -1,0 +1,19 @@
+import { actionHandler } from "@/app/utils/actions";
+
+async function getOperators() {
+    try {
+      return await actionHandler(
+        "registration/operations",
+        "GET",
+      );
+    } catch (error) {
+      // Handle the error here or rethrow it to handle it at a higher level
+      throw error;
+    }
+  }
+
+export default async function Operators() {
+    const operators = await getOperators();
+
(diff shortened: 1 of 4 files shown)


pbastia (approved):
🎉

Issue #1628, opened 2024-05-09
Spike Beautiful API versioning


Sepehr-Sobhani:
#### Describe the task

we need a way to introduce new features and modifications without affecting existing functionalities. Implementing versioning will help manage different versions of the API and facilitate easier maintenance and upgrades.

**Research Django-Ninja Versioning:**
Review the [Django-Ninja documentation](https://django-ninja.dev/guides/versioning/) specifically focusing on the versioning capabilities.

**Proof of Concept:**

- [x] Create a basic versioned API on top of our existing APIs.
~- [ ] Implement a small change based on regulation Part 2 that affects our API, Schema and service layers. @dleard , do you have any suggestions for this change?~
- [x] Test to ensure that everything is working as expected for both Part 1 and Part 2.

**Documentation:**
Prepare guidelines for developers on how to add new versions or modify existing versions of the API.

PR #1647, opened 2024-05-13
1628 spike api versioning

Sepehr-Sobhani:
[ISSUE 1628](https://github.com/bcgov/cas-registration/issues/1628)

### Notes:

- Added `v1` and `v2` sub-folders for `api`, `schema` and `test` folders
- Every new API endpoint, schema and test should go to `v2` sub-folders
- Reg part 2 API endpoints MUST start with `/v2/...`
- Added tags to API endpoints for clarity and separation

Screenshot 2024-05-13 at 2 53 39 PM


docs/registration_part_2/backend-structure.md
```diff
@@ -0,0 +1,64 @@
...
+# user_service.py (v1)
+class UserDataAccessService:
+    def get_user_operator_by_user(user_guid: UUID):
+        # Existing implementation
+```
+
+If you need to add a new function with the same name, use the `_v2` suffix:
+
+```python
+# user_service.py (v2)
```
andrea-williams, line 58: does `(v2)` mean that we're looking at the `v2/user_service.py` file?

Sepehr-Sobhani: No, I didn't plan to create separate folders for different versions of our services, but if we make significant changes, we might consider organizing our services into 'v1' and 'v2' folders. Initially, I intended to use the same service class and add '_v2' at the end of the function name that is going to be updated.

andrea-williams (approved):
Can confirm - it's a beautiful API!

PR #1656, opened 2024-05-15
805 refactor update operation

BCerki:
card: https://github.com/orgs/bcgov/projects/122/views/2?filterQuery=805&pane=issue&itemId=51964186

The card was written back before we had a service layer, so I haven't implemented the AC exactly as written. This PR:
- breaks the update_operation service function into smaller service functions (instead of breaking up the endpoint as described in the card)


bc_obps/service/data_access_service/document_service.py
```diff
@@ -3,11 +3,11 @@
...
 class DocumentDataAccessService:
     @classmethod
-    def create_statutory_declaration(cls, user_guid, statutory_declaration, operation):
+    def create_statutory_declaration(cls, user_guid, statutory_declaration):
         document = Document.objects.create(
             file=statutory_declaration,
             type=DocumentType.objects.get(name="signed_statutory_declaration"),
             created_by_id=user_guid,
         )
-        operation.documents.set([document])
```
BCerki, line 12: I moved this to better separate the data access services

shon-button: For your consideration, perhaps this could be a generic parameterized function?  Something like:
```
   @classmethod
    def create_document_file(cls, user_guid, file_data, document_type_name):
        document_type = DocumentType.objects.get(name=document_type_name)
        document = Document.objects.create(
            file=file_data,
            type=document_type,
            created_by_id=user_guid,
        )
        return document
```

shon-button:
@BCerki 
👌 making a beautiful API more beautiful!

shon-button (approved):

PR #1661, opened 2024-05-16
Rename endpoints

BCerki:
card: https://github.com/orgs/bcgov/projects/122/views/2?filterQuery=&pane=issue&itemId=56218560

The renaming approach:
- add UUID parameters when we're requesting a specific resource (previously we were getting operator.id etc. via the payload)
- removes parameters that aren't UUIDs and puts them in the body instead (eg `identity_provider`)
- put parameters right after the thing they reference, e.g. /operation/{operation_id}/update-status (vs the things between slashes in any other order)
- remove the select-operator endpoints and put them under operator instead (so becomes operator/{operator_id}/request_access etc.)
- updates url_name to be more logical/consistent

This PR:
- renames endpoints
- moves endpoints so that the file structure is docs for the api structure
- renames and moves tests to match the api folder
- adds to docs


bc_obps/registration/api/v1/_user_operators/current.py
```diff
@@ -1,20 +1,25 @@
...
 )
 from service.error_service.custom_codes_4xx import custom_codes_4xx
-from registration.api.utils.current_user_utils import get_current_user_guid
+
+# We don't need a user_operator_id parameter (using "current" instead) because we can look up the user_operator via the user data we receive in the middleware
 
 
+# GET
+# This endpoint retrieves data about both the user-operator and the operator.
 @router.get(
```
BCerki, line 19: `/user-operator/user-operator-from-user` was similar to `user-operator/user-operator-operator`, so I've combined them into `/user-operators/current`. This means that `cas-registration/bciers/apps/registration/app/(authenticated)/dashboard/page.tsx` now receives some additional info from the API (legal name and isNew)

bciers/apps/registration/middlewares/withAuthorization.tsx
```diff
@@ -116,8 +116,9 @@ export const withAuthorization: MiddlewareFactory = (next: NextMiddleware) => {
                   }),
                 }),
               };
+              // brianna here too
               const response = await fetch(
-                `${process.env.API_URL}registration/user-operator/user-operator-operator`,
+                `${process.env.API_URL}registration/user-operators/current`,
```
BCerki, line 121: This endpoint gets more than it needs since I consolidated endpoints

bc_obps/registration/api/v1/_user_operators/user_operator_id.py
```diff
@@ -15,22 +15,34 @@
...
+from registration.schema.v1 import UserOperatorOut
+
+from registration.models import (
+    AppRole,
+)
+
+# GET
+@router.get(
+    # brianna method problems if it's user-operators
+    "user-operator/{uuid:user_operator_id}",
```
BCerki, line 29: This should be called `user-operators` not `user-operator` (singular), but if I change it the endpoint tests start failing with 405 errors. Thoughts? I'm still working on it simultaneously as this PR is being reviewed

bc_obps/registration/api/v1/_user_operators/user_operator_id.py
```diff
@@ -15,22 +15,34 @@
...
+    response={200: UserOperatorOut, custom_codes_4xx: Message},
+    url_name="get_user_operator",
     tags=USER_OPERATOR_TAGS,
 )
-@authorize(["industry_user"], UserOperator.get_all_industry_user_operator_roles(), False)
+@authorize(AppRole.get_all_authorized_app_roles(), UserOperator.get_all_industry_user_operator_roles())
```
BCerki, line 34: b check this out

bc_obps/service/user_profile_service.py
```diff
@@ -16,5 +17,5 @@ def create_user_profile(cls, user_guid: UUID, identity_provider: str, user_data:
             else AppRole.objects.get(role_name="cas_pending"),
             IdPs.BCEIDBUSINESS.value: AppRole.objects.get(role_name="industry_user"),
         }
-        role: AppRole = role_mapping.get(identity_provider)  # type: ignore[assignment] # we know this will not be None
+        role: AppRole = role_mapping.get(identity_provider, None)  # type: ignore[arg-type] # we know this will not be None
```
BCerki, line 20: I got mypy fails on this, okay to change the ignore?

pbastia: `None` is the default value anyways, this has no effect except triggering a different linter error.
A way of fixing this would be to make the signature of the method take an `IdPs` object instead of the string

pbastia (approved):
@BCerki 🥇  I've left some comments that you're welcome to address, but that are not necessary.

bc_obps/registration/api/v1/_user_operators/_current/access_requests.py
```diff
@@ -13,15 +13,15 @@
 
 
 @router.get(
-    "/user-operator/user-operator-list-from-user",
+    "/user-operators/current/access-requests",
```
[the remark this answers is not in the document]
BCerki: It is the current user-operator record. Users can only be associated with one operator. This endpoint is for admin industry users to view the list of people who have requested to their operator.

bc_obps/registration/api/v1/_operators/_operator_id/is_declined.py
```diff
@@ -16,7 +16,7 @@
 
 
 @router.get(
-    "/operator/operator-access-declined/{operator_id}",
+    "/operators/{operator_id}/is-declined",
```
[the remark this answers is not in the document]
BCerki: This endpoint isn't about requesting access. It checks if an operator is declined or not.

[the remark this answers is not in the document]
BCerki: I'm going to change this to a /user-operators endpoint because it's not about the operator

Sepehr-Sobhani (commented):
This is great, Thank you! 🙏
Just a few nits and a quick fix in the test function indentation. I also suggest moving user-related tests to its own folder so it mirrors the API structure.

bc_obps/registration/api/v1/_user_operators/_current/is_current_user_approved_admin.py
```diff
@@ -15,13 +14,13 @@
 
 
 @router.get(
-    "/user-operator/is-approved-admin-user-operator",
+    "/user-operators/current/is-current-user-approved-admin",
```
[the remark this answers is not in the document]
BCerki: I'm using `current` to mean the current user_operator, not the current user. I'll add a comment

bc_obps/registration/api/v1/_user_operators/current.py
```diff
@@ -1,29 +1,30 @@
...
 from service.error_service.custom_codes_4xx import custom_codes_4xx
-from registration.api.utils.current_user_utils import get_current_user_guid
+
+# We don't need a user_operator_id parameter (using "current" instead) because we can look up the user_operator via the user data we receive in the middleware
 
 # this endpoint retrieves data about both the user-operator and the operator
 @router.get(
-    "/user-operator/user-operator-operator",
+    "/user-operators/current",
     response={200: OperatorFromUserOperatorOut, custom_codes_4xx: Message},
```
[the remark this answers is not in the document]
BCerki: Hmm... I don't know why. Maybe just because we were in a hurry for MVP and didn't have time to think it through? I've cleaned it up a little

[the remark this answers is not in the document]
Sepehr-Sobhani: Thank you 🙏

bc_obps/service/user_profile_service.py
BCerki: The mypy issue seemed to be a mismatch between the type Optional[AppRole] and AppRole

PR #1674, opened 2024-05-22
Fix happo flake

BCerki:
card: https://github.com/orgs/bcgov/projects/122/views/2?filterQuery=&pane=issue&itemId=61334238

This PR:
- fixes the NAICS code diff
- fixes the table row diff
- attempts to fix the accordion diffs (adding a check for the dropdown arrows to slow down the test; check for stability). 
- renames "table" to "grid" in happo screenshot names (I left table in the code itself because it's everywhere)

I'll have to make a follow-up PR to check  if it truly is fixed because it's not visually obvious

### DIFF
--- a/bciers/apps/registration/e2e/utils/helpers.ts
+++ b/bciers/apps/registration/e2e/utils/helpers.ts
@@ -442,8 +442,8 @@ export async function clearTableFilter(page: Page, fieldId: string) {
   expect(await filter.inputValue()).toBe("");
 }
 
-export async function waitForElementToStabilize(page: Page, element: any) {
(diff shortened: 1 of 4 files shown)


bciers/apps/registration/e2e/poms/operations.ts
shon-button, line 1: For your consideration, the ` await this.page.locator(` code in the methods could be defined in the constructor...which is generally considered better practice in a POM

shon-button, line 1: ` '[data-testid="ArrowDropDownIcon"]',` could follow the use of ENUMS pattern

shon-button:
Nice strategies!

PR #1679, opened 2024-05-24
Chore: updated requests package manually because dependabot didnt work

andrea-williams:
Updating this package the difficult way after dependabot's PR failed.

Happo failures = flake.


BCerki (approved):
:mechanical_arm:

PR #1685, opened 2024-05-24
chore: debug happo

BCerki:
card: https://github.com/orgs/bcgov/projects/122/views/2?filterQuery=&pane=issue&itemId=61334238

This PR is part 2/? and:
- fixes the table row diff from hover
- fixes the pending operator detail page diff
- fixes accordions (?) I hope

Once the diffs settle, I'll refactor the tests so the assertions are in the POMs. It's easier to write experimental code directly in the test files so it's to find/edit/remove.

Note: Anytime someone merges a PR, since the diffs are currently flaky, our baseline accepted screenshots change. The screenshots I'm introducing in this PR should be the new more stable baseline accepted shots.

### DIFF
--- a/bciers/apps/registration/e2e/workflows/idir/cas_admin.spec.ts
+++ b/bciers/apps/registration/e2e/workflows/idir/cas_admin.spec.ts
@@ -201,7 +201,7 @@ test.describe("Test Workflow cas_admin", () => {
       const arrowDropDownElements = await page.locator(
         '[data-testid="ArrowDropDownIcon"]',
       );
-      expect(await arrowDropDownElements.count()).toBe(2);
+      expect(await arrowDropDownElements.count()).toBe(4);
       await waitForElementToStabilize(page, "section");
       pageContent = page.locator("html");
       await happoPlaywright.screenshot(operationsPage.page, pageContent, {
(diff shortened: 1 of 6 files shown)


bciers/apps/registration/e2e/workflows/idir/cas_admin.spec.ts
shon-button, line 1: I am curious why the choice of putting the assertions in the .spec files instead of the POM?  

If left in the POM you could reduce code duplication by having a reusable function, something like:

```
async  getElementsByTestIdAndAssertCount(dataTestId, expectedCount) {
  const elements = await this.page.locator(`[data-testid="${dataTestId}"]`);
  const actualCount = await elements.count();
 await  expect(actualCount).toBe(expectedCount);
}
```

BCerki: When I'm experimenting, I put code directly in the spec files because then if doesn't work, I can easily tweak it or take out the whole block. If I put code in the POMs, I have to wade through multiple files and functions to find things again and/or delete them. I'll refactor to use POMs in another PR once I get it working.

shon-button (approved):

Issue #1617, opened 2024-05-07
Move the logic currently done in the Tiles component to the backend & return via API


dleard:
In navigation/Tiles.tsx we are currently doing a bunch of logic that determines which tiles to display to a user based on their role. This can probably be offloaded to the backend & returned via the API. For registration part 2 implementation, lets create an endpoint to do this work on the backend.

**Possible Architecture** (SH)
Leveraging JSONField to simplify database schema while providing the flexibility needed for dynamic or evolving data structures and requirements.
JSONField natively supports nested structures, allowing storing and querying complex hierarchical data directly within a single field.

Advantages:
- Flexibility: JSONField offers a flexible way to store various data structures, including dictionaries, lists, and nested objects. This eliminates the need for multiple database tables to represent complex relationships between data points.
- Simplified Model Definition: Compared to defining multiple relational tables and foreign keys, JSONField keeps your model definition concise and easier to maintain.
- Efficient Storage: JSONField stores data in a compact format, potentially using less space than multiple relational tables.

**Steps:**

1. Create common API Router:
- [x] create `bc_obps/common/api/router.py` using `Router`
- [x] create `bc_obps/common/api/__init__.py` to allow import of API code from other parts of your project using dot notation
- [x] create  `bc_obps/common/api/v1/dashboard_data/tiles.py/example_endpoint`
- [x] create `bc_obps/common/api/v1/__init__.py` 
- [x] in  `bc_obps/common/api/v1/__init__.py`  import tile
- [x] in `bc_obps/bc_obps/api.py` import `common.api.router`
- [x] start API server `make run`
- [x] validate endpoint `example_endpoint`
2. Create common API DashboardData endpoint service
- [x] add `class DashboardData` to `bc_obps/common/models.py`
- [x] create `bc_obps/common/schema/v1/dashboard_data.py`
- [x] create `bc_obps/common/schema/v1/__init__.py` to `from .dashboard_data import *`
- [x] create `bc_obps/service/data_access_service/dashboard_service.py`
- [x] create  `bc_obps/common/api/v1/dashboard_data/tiles.py/dashboard-data` get endpoint
- [x] create `bc_obps/common/fixtures/bciers/external.json`
- [x] create `bc_obps/common/fixtures/bciers/internal.json`
- [x] load prod data using fixtures ^
- [x] run terminal command `make migrations` which will make model migrations 
- [x] run terminal command `make reset_db` which will run model migrations and load fixtures.

AC:
[...]

PR #1708, opened 2024-05-28
API: Dashboard Tiles

shon-button:
Addresses:
 [1617](https://github.com/bcgov/cas-registration/issues/1617)

## 🚀 Impact:

-  Leveraging Django's JSONField for storing and querying hierarchical dashboard data removing front-end (FE) display logic requirements. This approach also offers flexibility by enabling complex and/or conditional logic within the back-end (BE) API.
-  Using Django JSONField removes the need to define multiple relational tables keeping the model definition concise and easier to maintain.
- Provides dashboard data management using hierarchical fixture folder structure mimicking FE dashboard directories.

## ✏️ Notes:
-  Limited python tests to authorization assuming test of dashboard specific JSON will be tested in e2e tests (preliminary tests of API without @authorized decorator using ` http://localhost:8000/api/common/dashboard-data?dashboard=all` seem okily dokily)
- 🚀 http://localhost:8000/api/docs; 
- 🛸 http://localhost:8000/api/common/dashboard-data?dashboard=all

##  🔬 Testing:

1. From terminal command, reset the db and start the api server:
 ```
cd bc_obps
make reset_db
make run
```  
2. Run `/dashboard-data api tests` from terminal: 
```
cd bc_obps && make pythontests ARGS='common/tests/endpoints/v1/dashboard_data/test_tiles.py' 
```
3. Run all test from terminal: 
```
cd bc_obps && make pythontests
```

### CODE REVIEW

**Comment by Sepehr-Sobhani** [comment]
This looks great, Thank you Shon 🙏  Some suggestions:

- We need to include these fixtures as the initial data for our model using data migration, Otherwise, we won't have access to these values in other namespaces such as DEV, TEST, and PROD. This way we no longer need those fixtures in the `load_fixtures` command.

- I suggest adding `dashboard` and `role` query params within the endpoint function definition so we can add some type hints and get better and nicer validation and docs. Something like:

```python
def list_dashboard_data(request: HttpRequest, dashboard: Optional[Literal['all', 'bciers', 'registration', 'reporting', ...]])
```

This becomes:
Screenshot 2024-05-28 at 10 25 02 AM

- Alternatively, we can define a schema with `dashboard` and `role` fields (With type hints) and then use a POST request to define how we want to respond.

- And one final thought: We might be able to use the role associated with the user object (`current_user`) and don't have to pass it as a query param or payload.

bc_obps/common/tests/endpoints/v1/dashboard/test_dashboard_authorization.py
```diff
@@ -0,0 +1,18 @@
+from registration.tests.utils.helpers import CommonTestSetup
+from registration.tests.utils.helpers import TestUtils
+from common.constants import BASE_ENDPOINT
+
+# make pythontests ARGS='common/tests/endpoints/v1/dashboard/test_dashboard_authorization.py' 
+class TestDashboardDataEndpointAuthorization(CommonTestSetup):
```
Sepehr-Sobhani, line 6: I suggest adding another test to ensure we get the right data based on different roles and dashboard parameters.

shon-button: @Sepehr-Sobhani as mentioned in "Notes", my thinking is "Limited python tests to authorization assuming test of dashboard specific JSON will be tested in e2e tests" but, happy to add here if you think it advisable.

Sepehr-Sobhani: Sorry for not paying enough attention to your notes. Of course, we can ignore that. I always assume that I don't know what is happening on the front-end side, so I try to approach everything in a way that ensures the backend is independent. This way, we aren't reliant on tests from the other side.

Sepehr-Sobhani (approved):
This is great Shon, Thank you 🙏 
I have one more request. Could you please merge those three migration files into one? This will reduce the number of files and code we need to maintain. To do this, you can delete the three migration files, run the `make migrations` command, and then add the data fixtures part as a `migrations.RunPython` operation after operations related to creating the data model. 
I'm going to approve this so it doesn't block you.

shon-button:
> This is great Shon, Thank you 🙏 I have one more request. Could you please merge those three migration files into one? This will reduce the number of files and code we need to maintain. To do this, you can delete the three migration files, run the `make migrations` command, and then add the data fixtures part as a `migrations.RunPython` operation after operations related to creating the data model. I'm going to approve this so it doesn't block you.

Yes, of course!  Thank-you so much for your guidance! 💫

Issue #113, opened 2023-09-15
Add pre-commit-hooks-django


Sepehr-Sobhani:
###Description of the Tech Debt###
We couldn't set up this pre-commit hook initially because of the following reasons. However, this hook has some useful checks that would be beneficial when we have more migration files:
- It takes a lot of time to troubleshoot.
- Implementing it was a bit tricky due to issues with the poetry virtual environment.

Link to hook: https://github.com/ecugol/pre-commit-hooks-django

NOTE: Due to the issue with the virtual environment (Hook is not aware of poetry env), I've implemented this as a standalone command to check django migrations.

Sepehr-Sobhani:
This is not needed, we always run the migrate command when we run the backend image so no need to check this in the CI.

PR #1721, opened 2024-05-29
Add new django migration check

Sepehr-Sobhani:
[ISSUE 113](https://github.com/bcgov/cas-registration/issues/113)

NOTE: Due to the issue with the virtual environment (Hook is not aware of the poetry env), I have implemented this as a standalone command.


Sepehr-Sobhani:
This is not needed, we always run migrate command when we run the backend image so no need to check this in the CI.

PR #1730, opened 2024-06-03
Refactor CI to use Nx and best practices

BallardRobinett:
Card: https://github.com/bcgov/cas-reporting/issues/232

Adding nx-affected to limit tests to projects that have been affected by changes, added tags to workflows, updated action versions, and changed workflow names to be more descriptive of when they are triggered.

Nx replay (caching) will require nxCloud connections to be enabled in bciers/nx.json. 

Workflow names are inspired by Iceland's digital services CI: https://github.com/island-is/island.is/tree/main/.github/workflows


.github/workflows/test.yaml
```diff
@@ -40,8 +46,11 @@ jobs:
           fetch-depth: 0
       - name: dev env setup
         uses: ./.github/actions/dev-env-setup
-      - name: Run Nx Tests
-        run: yarn nx run-many -t test --skipNxCache
+      - name: Run Nx Affected Tests with Remote Caching
+        id: nx-tests
+        run: |
+          yarn nx affected:build --base=origin/develop --target=test --parallel
```
joshgamache, line 52:
```suggestion
          yarn nx affected --base=origin/develop --target=test --parallel
```

The `something:other` format is for `nx run project:target`. Nx's affected is it's own thing akin to run, so you can drop it here. Target is defined as a `--target` attribute like you already had.

bciers/package.json
```diff
@@ -17,7 +17,8 @@
     "reg1:e2e:ui": "nx run registration1:e2e:ui",
     "reg1:e2e:report": "nx run registration1:report",
     "reg1:coverage": "nx run registration1:test --coverage",
-    "reg1:test": "nx run registration1:test"
+    "reg1:test": "nx run registration1:test",
+    "affected:test": "nx affected:test --base=remotes/origin/main"
```
joshgamache, line 21:
```suggestion
    "affected:test": "nx affected --base=remotes/origin/main --target=test"
```

Same as other comment, the `target` is handled by the attribute.

joshgamache (approved):
Approved, conditional on the two changes I posted. Nice work on this!

I added tags into the project files in a quick commit, which was one of the intended steps but wasn't clear in the issue ticket (my bad!). Will be useful to have the workflow tags you added in.

Issue #1587, opened 2024-05-02
Registration part II components - Task list: Single-page form version


BCerki:
#### Describe the task

Check with giraffe before scheduling this--they have a ticket too: https://github.com/bcgov/cas-reporting/issues/237

Registration part II has some new components such as the side tabs and note.

#### Acceptance Criteria

- [ ] create the `Task list: Single-page form version` (not the multi page one) component using the instructions here: https://www.figma.com/design/LsDmLDCdnJqI8UahyqvOD8/OBPS-App-Designs?node-id=9812%3A1767&t=i99A8cQ2Z2W6IGzP-1
- [ ] create a mock page somewhere in registration part II to show the two components (this will be deleted later but is important for testing this PR)

#### Additional context

- Needs talking to Giraffe

patrickisaac:
Notes:
- Dev and Designer to pair on refining ticket

hannavovk:
@nanyangpro , can you work with a dev to refine this ticket?

nanyangpro:
Hi @marcellmueller - I see you've started development of this ticket. Just wondering if we can remove the "Backlog Refinement" label already or if there's anything that I can help clarify?

shon-button:
@marcellmueller 

For the `Task list: Single-page form version`; specifically, the state management of the "completed checkmark" per section...
![Image](https://github.com/bcgov/cas-registration/assets/120038448/af0973e2-793e-4a64-b721-7001aa06ca5f)



I believe you mentioned in your demo that you are currently using separate forms but, would prefer to use one form; so, I thought perhaps you could use `Jotai` to listen to field section updates and manage the checkmarks.

This example demonstrates a basic form of reactive state management using Jotai, where the state of one part of your application can automatically affect other parts.

### Steps:

1. **Install Jotai**:
   If you haven't already installed Jotai, you can do so using npm or yarn:
   ```sh
   npm install jotai
   # or
   yarn add jotai
   ```

2. **Create Atoms for Your Fields**:
   Create atoms for the fields you want to listen to.

3. **Create a Derived Atom**:
   Create a derived atom that checks if the fields are complete.

4. **Update the `div` Based on the Derived Atom**:
   Use the derived atom to conditionally render or update the `div`.

### Example:

Here’s a complete example demonstrating how to listen to input fields and update a `div` when both fields are complete:

```jsx
import React from 'react';
import { atom, useAtom } from 'jotai';

// Create atoms for the fields
const field1Atom = atom('');
[...]

PR #1745, opened 2024-06-05
1587 single step task list

marcellmueller:
#1587 

**Note**: For ease of review I've added a mock form setup to the registration 2 landing page since it's currently empty anyways and not going to prod anytime soon. I will remove this before merging.

This one took me longer than expected. It's really not that complex but it took awhile to figure out how to validate the form without triggering validation and making all fields go red on load. I learned a bunch about RJSF and React refs though so it wasn't too bad.

In RJSF 4 we could import a validator function to externally validate the forms. Though in RJSF 5 we have to pass in a validator in our `FormBase` and can't import it. Though we can access the validator through the onChange event as well as by adding a form `ref`. It doesn't seem to be in their documentation so it took me awhile to find.

- Added [documentation for externally validating forms here](https://github.com/bcgov/cas-registration/blob/a4e6ace82b5f7b1a2d0d614cea744f3b7047e7a4/docs/frontend/rjsf-forms.md).

- Added [a new RJSF doc for our form base components here](https://github.com/bcgov/cas-registration/blob/a4e6ace82b5f7b1a2d0d614cea744f3b7047e7a4/docs/frontend/rjsf-form-base-components.md)
- Updated RJSF library


bciers/apps/registration/app/page.tsx
shon-button, line 1: @marcellmueller 

Thank-you for the excellent RJSF documentation, very helpful in filling in the gaps of my RJSF understanding  🙌
This is a very impressive implementation of the task list form requirement, as always, you're a star 🤩

A few concerns\asks\suggestions for your consideration:

1. ` onSubmit={async (data) => console.log(data)}` prints a blank object `{}` on `Submit` - I was expecting formData
2. can the validation be decoupled from the form and applied specifically to the fields determined in `const errorData` obtained in `bciers/libs/shared/components/src/lib/form/SingleStepTaskListForm.tsx\handleFormChange`?  If so I think it would be more user friendly to have the section validation and the field(s) validation in sync.
3. if the task list (sections) could be sticky, I think form navigation/filling would be greatly improved

shon-button (approved):

bciers/apps/registration/app/page.tsx
marcellmueller: Hey sorry I meant to respond to this as well though just wanted to show you my somewhat failed attempt at single field validation this morning. 

Thanks for catching that `onSubmit` bug!! 🙇  Really appreciate it, I guess it's easy to miss when this isn't being used in anything yet.

As for the sticky, I don't think it's a bad idea at all. Personally I'm not a huge fan of sticky navs but user research shows that they are helpful and a lot of people request them. I think this is one to ask the designers for their say, I will try and remember to bring it up at dev-design sync :)

PR #1746, opened 2024-06-05
1686 2 external operations grid

BCerki:
card: https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=64781153

This one can't be fully completed until we have auth (can't confirm that only the allowed people can see stuff), so I'll do a follow up PR later.

This PR:
- new v2 operations endpoint to retrieve the p2 data
- new v2 schemas
- tests for ^
- new `page.tsx` file for p2
- new FE p2 files that handle operation grid columns. This resulted in new front-end components because I had to update the imports all the way up to the `page.tsx` file
- new version of the existing unit tests (coverage wasn't great, we have an epic to add more: https://github.com/bcgov/cas-registration/milestone/4)
- handles the internal grid as well because it ended up having almost the same requirements as the external one

Notes:
- there are some VS code errors related to not having auth but they don't seem to be breaking the build like TS errors, I'll address them in the follow up PR if they persist
- the operations page doesn't have the button/note because that's conditional on role (and related to auth). Again, if it still looks weirdly spaced after auth is in I'll address in the follow up
- test imports aren't working well atm so I put the tests in v1. Also to be addressed in follow up PR. Some work on this in commit  841d268894745f3195dcec82360f0fdc66ad677c, but will likely be solved by the auth pr)

Since we don't have auth yet, to test this PR you need to make a few temporary changes locally:
- in `bc_obps/registration/api/v2/operations.py`, comment out the `@authorize` decorator
- replace the return line with `return 200, OperationService.list_operations('ba2ba62a-1218-42e0-942a-ab9e92ce8822', filters)` (this is the uuid of bc-cas-dev in the fixtures)
- in `bciers/apps/registration1/app/utils/getUUIDFromEndpoint.ts`, add `"registration/v2/operations",` to the allow list

### DIFF
--- a/bciers/apps/registration/app/components/datagrid/models/operationColumns.ts
+++ b/bciers/apps/registration/app/components/datagrid/models/operationColumns.ts
@@ -7,11 +7,13 @@ const operationColumns = (
   isInternalUser: boolean,
   ActionCell: (params: GridRenderCellParams) => JSX.Element,
 ) => {
-  const columns = [
+  const columns: GridColDef[] = [
     {
       field: "name",
       headerName: "Operation Name",
       width: isInternalUser ? 720 : 320,
+      // Set flex to 1 to make the column take up all the remaining width if user zooms out
+      flex: 1,
     },
     { field: "bcghg_id", headerName: "BC GHG ID", width: 200 },
     { field: "type", headerName: "Operation Type", width: 200 },
@@ -21,10 +23,8 @@ const operationColumns = (
       renderCell: ActionCell,
       sortable: false,
       width: 120,
-      // Set flex to 1 to make the column take up all the remaining width if user zooms out
-      flex: 1,
     },
-  ] as GridColDef[];
+  ];
 
   if (isInternalUser) {
     columns.splice(OPERATOR_COLUMN_INDEX, 0, {
--- a/bciers/apps/registration/app/components/datagrid/models/operationColumns.ts
+++ b/bciers/apps/registration/app/components/datagrid/models/operationColumns.ts
@@ -1,0 +1,40 @@
+import { GridColDef, GridRenderCellParams } from "@mui/x-data-grid";
+import StatusStyleColumnCell from "@bciers/components/datagrid/cells/StatusStyleColumnCell";
+
+export const OPERATOR_COLUMN_INDEX = 1;
+
+const operationColumns = (
+  isInternalUser: boolean,
+  ActionCell: (params: GridRenderCellParams) => JSX.Element,
+) => {
+  const columns = [
+    {
+      field: "name",
+      headerName: "Operation Name",
+      width: isInternalUser ? 720 : 320,
+    },
+    { field: "bcghg_id", headerName: "BC GHG ID", width: 200 },
+    { field: "type", headerName: "Operation Type", width: 200 },
+    {
+      field: "action",
+      headerName: "Action",
+      renderCell: ActionCell,
+      sortable: false,
+      width: 120,
+      // Set flex to 1 to make the column take up all the remaining width if user zooms out
+      flex: 1,
+    },
+  ] as GridColDef[];
+
+  if (isInternalUser) {
+    columns.splice(OPERATOR_COLUMN_INDEX, 0, {
+      field: "operator",
+      headerName: "Operator Legal Name",
+      width: 400,
+    });
+  }
+
+  return columns;
+};
+
+export default operationColumns;
--- a/bciers/apps/registration/app/components/datagrid/models/operationGroupColumns.ts
+++ b/bciers/apps/registration/app/components/datagrid/models/operationGroupColumns.ts
@@ -9,7 +9,7 @@ const operationGroupColumns = (
   isInternalUser: boolean,
   SearchCell: (params: GridColumnGroupHeaderParams) => JSX.Element,
 ) => {
-  const columnGroupModel = [
+  const columnGroupModel: GridColumnGroupingModel = [
     {
       groupId: "bcghg_id",
       headerName: "BC GHG ID",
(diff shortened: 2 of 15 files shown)


bciers/apps/registration/app/components/datagrid/models/operationGroupColumns.ts
```diff
@@ -0,0 +1,52 @@
...
+      children: [{ field: "type" }],
+    },
+
+    {
+      groupId: "action",
+      headerName: "Action",
+      renderHeaderGroup: EmptyGroupCell,
+      children: [{ field: "action" }],
+    },
+  ] as GridColumnGroupingModel;
```
marcellmueller, line 38: Roasting my previous code here again, we should declare this type:
`const operationGroupColumns: GridColumnGroupingModel = (`

bciers/apps/registration/app/components/operations/OperationDataGrid.tsx
```diff
@@ -0,0 +1,78 @@
...
+}: {
+  lastFocusedField: string | null;
+  setLastFocusedField: (value: string | null) => void;
+}) => {
+  const RenderCell = (params: GridColumnGroupHeaderParams) => {
+    const { groupId, headerName } = params;
+    return (
+      <HeaderSearchCell
+        field={groupId as string}
+        fieldLabel={headerName as string}
```
marcellmueller, line 26: Hmm not sure what the best way fix these type assertions are, do you think something like `groupId ?? ""` is a good idea?

bciers/apps/registration/app/components/operations/Operations.tsx
```diff
@@ -0,0 +1,74 @@
...
+  try {
+    const queryParams = buildQueryParams(searchParams);
+    // fetch data from server
+    const pageData = await actionHandler(
+      `registration/v2/operations${queryParams}`,
+      "GET",
+      "",
+    );
+    return {
+      rows: formatOperationRows(pageData.data) as OperationRow[],
```
marcellmueller, line 37: `rows: formatOperationRows(pageData.data) ?? []` seems to make typescript happy without asserting the type.

bciers/apps/registration/app/components/operations/Operations.tsx
```diff
@@ -0,0 +1,74 @@
...
+  return rows.map(({ id, operator, name, bcghg_id, type }) => {
+    return {
+      id,
+      name,
+      bcghg_id,
+      operator: operator,
+      type,
+    };
+  });
+};
```
marcellmueller, line 22: Is this function even necessary? It doesn't look like it's doing anything useful.

marcellmueller (commented):
Nice work! This is working well. Most of my review comments are nits about code that I previously implemented, or other legacy code that I think could be cleaned up (formatOperationRows).

I suggest creating a `layout.tsx` in `/operations` alongside the page file so you can ensure this fits the design.

Layout.tsx:
```
import Main from "@bciers/components/layout/Main";

export default function Layout({
  children,
}: {
  children: React.ReactNode;
}) {
  return <Main>{children}</Main>;
}
```

The action column is too large:
Screenshot 2024-06-07 at 7 52 28 AM

Design:
Screenshot 2024-06-07 at 7 50 44 AM


Aside from that I think we could do some more unit testing. I'm not a coverage puritan and I know the DataGrid tests are tech debt but I'm not sure if we should pile on more tech debt. 

Not saying to do a in depth DataGrid test but maybe some more specific to this implementation. At the very least there should be tests that show the search fields are present. There are some DataGrid versions and options that can cause them to not appear at all.

bciers/apps/registration/app/components/operations/OperationDataGrid.tsx
BCerki: Sure

bciers/apps/registration/app/components/operations/Operations.tsx
BCerki: I didn't need to cast it after removing the function

bciers/apps/registration/app/components/datagrid/models/operationColumns.ts
```diff
@@ -0,0 +1,40 @@
...
+    },
+    { field: "bcghg_id", headerName: "BC GHG ID", width: 200 },
+    { field: "type", headerName: "Operation Type", width: 200 },
+    {
+      field: "action",
+      headerName: "Action",
+      renderCell: ActionCell,
+      sortable: false,
+      width: 120,
+      // Set flex to 1 to make the column take up all the remaining width if user zooms out
```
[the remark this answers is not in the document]
BCerki: It would look like this (I'd have to fix the border): 
image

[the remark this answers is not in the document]
marcellmueller: Can't you add flex 1 to Operation Name column since it is the big one in the design?

BCerki:
> Nice work! This is working well. Most of my review comments are nits about code that I previously implemented, or other legacy code that I think could be cleaned up (formatOperationRows).
> 
> I suggest creating a `layout.tsx` in `/operations` alongside the page file so you can ensure this fits the design.
> 
> Layout.tsx:
> 
> ```
> import Main from "@bciers/components/layout/Main";
> 
> export default function Layout({
>   children,
> }: {
>   children: React.ReactNode;
> }) {
>   return <Main>{children}</Main>;
> }
> ```
> 
> The action column is too large: Screenshot 2024-06-07 at 7 52 28 AM
> 
> Design: Screenshot 2024-06-07 at 7 50 44 AM
> 
> Aside from that I think we could do some more unit testing. I'm not a coverage puritan and I know the DataGrid tests are tech debt but I'm not sure if we should pile on more tech debt.
> 
> Not saying to do a in depth DataGrid test but maybe some more specific to this implementation. At the very least there should be tests that show the search fields are present. There are some DataGrid versions and options that can cause them to not appear at all.

I took a crack at a DataGrid test and ran into trouble with some of the mocks, so I'll add more testing in the follow up PR.

marcellmueller (approved):
Nice, this is looking great! 🥇

PR #1747, opened 2024-06-05
Chore: 254 wci config models

dleard:
adding the database models required for dynamic form configuration.

First part of the implementation that follows the minispike exploration done in #1644


bc_obps/reporting/models/base_schema.py
```diff
@@ -0,0 +1,20 @@
...
+class BaseSchema(BaseModel):
+    """Static base json schema for emissions reporting forms"""
+
+    # No history needed, these elements are immutable
+    slug = models.CharField(
+        max_length=1000,
+        db_comment="Name of the base schema. Should describe what form it is used to generate and when the base schema took effect. For example: general_stationary_combustion_2024",
+    )
+    schema = models.JSONField(
+        max_length=100000,
```
pbastia, line 14: `max_length` is not a valid option for `JSONField` https://docs.djangoproject.com/en/5.0/ref/models/fields/#jsonfield

bc_obps/reporting/models/configuration_element.py
```diff
@@ -0,0 +1,41 @@
...
+        db_table = 'erc"."configuration_element'
+        constraints = [
+            UniqueConstraint(
+                fields=[
+                    'reporting_activity',
+                    'source_type',
+                    'gas_type',
+                    'methodology',
+                    'valid_from',
+                    'valid_to',
```
pbastia, line 37: Same comment as before for the `valid_from`/`to`

bc_obps/reporting/models/reporting_field.py
```diff
@@ -0,0 +1,19 @@
...
+class ReportingField(BaseModel):
+    """Conditional Reporting Fields"""
+
+    # No history needed, these elements are immutable
+    field_name = models.CharField(
+        max_length=1000, db_comment="Name of field needed for the related configuration element."
+    )
+    field_type = models.CharField(max_length=1000, db_comment="Type definition for field.")
+
+    def serialize(self) -> "dict[str,str]":
```
pbastia, line 14: Apparently mypy doesn't mind quoted type hints 🤷 

```suggestion
    def serialize(self) -> dict[str, str]:
```

pbastia (changes requested):
Just a few nitpicky things! This looks awesome

bc_obps/reporting/models/activity_source_type_base_schema.py
```diff
@@ -0,0 +1,38 @@
...
+        db_table_comment = "Intersection table that assigns a base_schema as valid for a period of time given an activity-sourceType pair"
+        db_table = 'erc"."activity_source_type_base_schema'
+        constraints = [
+            UniqueConstraint(
+                fields=[
+                    'reporting_activity',
+                    'source_type',
+                    'base_schema',
+                    'valid_from',
+                    'valid_to',
```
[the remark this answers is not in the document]
dleard: Shouldn't the valid_from / valid_to overlap check go in the `configuration` table then? And this table does not need constraints?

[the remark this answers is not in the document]
dleard: I think we'll need to use the save() method in this table, the exclusion constraint won't work here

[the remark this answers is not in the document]
pbastia: There should be an exclusion check for the dates in the `configuration` table, but we also need to make sure that we don't let those `activity_source_type_base_schema` records overlap on configuration IDs, otherwise we'll have ambiguous data

Overriding the `save()` method might be more readable, and it's not going to be a performance penalty since it'll only run during data migration

bc_obps/reporting/models/reporting_field.py
dleard: My editor says this needs to be quoted
Screenshot from 2024-06-14 15-32-37

bc_obps/reporting/migrations/0004_baseschema_configuration_gastype_methodology_and_more.py
```diff
@@ -0,0 +1,318 @@
...
+            fields=[
+                ('id', models.BigAutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('name', models.CharField(db_comment='The name of a reporting methodology', max_length=1000)),
+            ],
+            options={
+                'db_table': 'erc"."methodology',
+                'db_table_comment': 'Table contains the set of reporting methodologies that can be applied to an emission as outlined in GGERR (Greenhous Gas Emission Reporting Regulation)',
+            },
+        ),
+        migrations.RunPython(init_methodology_data, reverse_init_methodology_data),
```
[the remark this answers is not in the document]
dleard: I think we should merge all our migrations into one 'initial' migration & then create a separate 'prod_data' migration (and then probably separate migrations for the larger data migrations like configuration_element & the m2m reporting field data)

bc_obps/reporting/models/configuration_element.py
```diff
@@ -0,0 +1,51 @@
...
+        for y in all_ranges:
+            if (
+                (self.valid_from.valid_from >= y.valid_from.valid_from)
+                and (self.valid_from.valid_from <= y.valid_to.valid_to)
+                or (self.valid_to.valid_to <= y.valid_to.valid_to)
+                and (self.valid_to.valid_to >= y.valid_from.valid_from)
+            ):
+                raise Exception(
+                    f'This record will result in duplicate configurations being returned for the date range {self.valid_from.valid_from} - {self.valid_to.valid_to} as it overlaps with a current record or records'
+                )
```
pbastia, line 50: There is code duplication with the act_source_type_base_schema table, maybe we can write some tech debt about that

bc_obps/reporting/models/reporting_field.py
pbastia: weird! I guess it's correct either way

pbastia (approved):
just a few nitpicky things that don't need to be addressed

bc_obps/reporting/models/configuration_element.py
dleard: Thanks! I'll tackle this and the other last few comments in my subsequent PR

PR #1754, opened 2024-06-06
single field validation 

marcellmueller:
Just threw up this old branch I had worked on as part of the `SingleStepTaskListForm` PR. Single field validation is not supported in RJSF for some reason. Though in my attempt at doing this feature I can see how it would get difficult for complex nested schemas. There is a bit more details in a comment.


bciers/libs/shared/components/src/lib/form/FormBase.tsx
```diff
@@ -0,0 +1,162 @@
...
+      (error: Error) => error.property === `.${field}`,
+    );
+
+    const fieldErrorSchema = _errorSchema[field];
+
+    $this.setState({
+      errors: [...new Set([...prevOtherFieldErrors, ...fieldErrors])],
+      errorSchema: { ...errorSchema, [field]: fieldErrorSchema },
+    });
+  };
```
marcellmueller, line 107: @shon-button this was my branch I saved a copy of and abandoned for single field validation. This is most of the relevant code, aside from the onBlur and onFocus functions, as well as updating our widgets to handle onBlur and onFocus.

The main issue with this is that it just handles flat schemas which is why it works with my previous multi-form approach.  

As soon as you start nesting objects this all falls apart. I could likely get it working with my single form approach but once we start adding complex forms and array field it will all fall apart. In the single form approach RJSF will append the section names to each field that it passes through the onBlur and onFocus. So a field called `first_name` will be `root_section1_first_name`. 

I do believe I could get this working with my mock schema you viewed in the actual PR, but as I said more complex schemas than that will be extremely hard to do single field validation on.

Issue #1726, opened 2024-05-31
Registration part II components - Notes component


marcellmueller:
#### Describe the task

Create the new notes component. 

#### Acceptance Criteria

- [ ] Create the new notes component
- [ ] Accept either a string or a JSX element to support rich text, such as bolded words etc.
- [ ] Add variants since there are a couple different colours

#### Additional context

- Add any other context about the task here.
- Or here

PR #1756, opened 2024-06-06
Registration part II components - Notes component

marcellmueller:
#1726

Edit: I removed the dynamic height using Javascript, it was needlessly complex for this specific component and I refactored to use all CSS. It was nice and easy, just changed the `absolute` to `relative` and now it will automatically change the height of the parent element :)

Since the `SingleStepTaskListForm` example got merged in I think it's okay to leave these notes in until we do the landing page, though happy to remove it as well.

Screenshot 2024-06-06 at 12 16 03 PM
Screenshot 2024-06-06 at 12 16 23 PM

### DIFF
--- a/bciers/apps/registration/app/page.tsx
+++ b/bciers/apps/registration/app/page.tsx
@@ -1,6 +1,7 @@
 "use client";
 
 import Main from "@bciers/components/layout/Main";
+import Note from "@bciers/components/layout/Note";
 
 // Will remove this before merging, just leaving it in for PR review
 import SingleStepTaskListForm from "@bciers/components/form/SingleStepTaskListForm";
@@ -120,6 +121,14 @@ export default function Page() {
         },
       }}
     >
+      <Note variant="important">
+        <b>Important:</b> It can be used to highlight important information or
+        provide context to the content on the page.
+      </Note>
+      <Note>
+        <b>Note:</b> It can be used to provide additional context to the content
+        on the page.
+      </Note>
       <h1>Registration Part II</h1>
       {/* Added a large bottom margin to test task list onClick smooth scroll */}
       <div className="mb-[50vh]">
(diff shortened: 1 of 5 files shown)


bciers/libs/shared/components/src/lib/layout/Note.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+    render(<Note>Test note component</Note>);
+
+    expect(addEventListener).toHaveBeenCalled();
+
+    expect(addEventListener).toHaveBeenCalledWith(
+      "resize",
+      expect.any(Function),
+    );
+  });
```
marcellmueller, line 51: I tried to test the height itself though I learned that JSDom just simulates the DOM and doesn't seem to simulate the height of elements.

bciers/apps/registration/app/page.tsx
```diff
@@ -120,6 +121,14 @@ export default function Page() {
         },
       }}
     >
+      <Note variant="important">
+        <b>Important:</b> It can be used to highlight important information or
```
BCerki: Do the important notes always have "Important" in front of them, and regular notes "Note"? If so, we could add this text to the component itself based on the variant

BCerki (commented):
I didn't get very far on the review before seeing your note, if the comment's not relevant to the new approach no worries

bciers/apps/registration/app/page.tsx
marcellmueller: I thought about that and am not opposed to it, but I thought a general component that would allow the user to add whatever they need to in it is more flexible as there are some `Info` variants in the design that have multiple bolded text areas. It's not difficult to add a `<b>Important:</b>` and maybe there would be a future case where they would want an icon to start or something else. 

The argument against that is consistency, and we could easily allow them to pass a React node for further bolded text or whatever else is needed after the `Important:` prefix. 

I'll leave this at your mercy haha.

marcellmueller: @BCerki

Issue #1734, opened 2024-06-03
Registration Part II - setup CI/CD


marcellmueller:
#### Describe the task

A clear and concise description of what the task is.

#### Acceptance Criteria

- [x] Add DockerFile
- [x] Setup gchr repo for separate image files from Part I
- [x] Setup CI to build image
- [ ] Deploy to OpenShift (Still need to discuss the best course of action for having Registration 1 & 2 accessible at the same)

#### Additional context

- Add any other context about the task here.
- Or here

marcellmueller:
@dleard here's that ticket if you had any more context to add

PR #1758, opened 2024-06-06
1734 registration2 ci cd

marcellmueller:
Implements #1734 

- Adds Dockerfile to `/bciers/registration`. I added some updates from #1856 and Sonarcloud is flagging the `chown` commands for `Make sure no write permissions are assigned to the executable.` which needs review
- Builds the dockerfile and pushes it to GHCR
- Added a helm chart though Josh is going to get this all working in OpenShift in another PR he is working on with the multizone auth stuff.

#### This can't be merged without updating the branch protections first. As you can see at the bottom of this PR, renaming the jobs broke required status protection and they are all stuck on `Expected — Waiting for status to be reported`.

This also includes a pretty hefty reorganization of our GitHub Actions pipeline. Most of our jobs including our build steps were in `test.yaml` and it was getting rather large and needed to be split up. This will just be part 1 of our pipeline reorganization and a good starting base to break up jobs and optimize CI going forward.

- Added a `main.yaml` which orchestrates all workflows
- Moved docker image builds to their own files and call them in `main.yaml` 
- Added `run-backend` action. Currently the backend test starts up the entire app including the front end which is unnecessary. 
- Created a reusable action called `nx-app-build` that builds and pushes a docker image for our nx projects with just a few env vars/inputs. The backend docker build was different enough and is unique so it stays the same.
- Trigger the Trivy and CodeQL workflows from `main.yaml` so we don't see separate actions for each push when clicking the `Actions` tab which was always confusing when trying to debug issues through prior workflows. Now they appear as a job in the main workflow.
- `test-code.yaml` is for tests that don't require containers to be built first and can run immediately. 
- Moved e2e tests to `test-e2e.yaml` which can be called when the backend image and `registration1` image is built. We can add more required jobs for `registration` and `reporting` as we set up e2e for those projects.


.github/actions/nx-app-build/action.yaml
```diff
@@ -0,0 +1,73 @@
...
+      run: |
+        corepack enable
+        yarn install --immutable
+        cd bciers && yarn install --immutable
+    - uses: actions/setup-node@v3
+    - name: Derive appropriate SHAs for base and head for `nx affected` commands
+      uses: nrwl/nx-set-shas@v4
+      with:
+        # This will need to be updated when giraffe-develop merges with develop
+        main-branch-name: "develop"
```
marcellmueller, line 46: I assume this should be set to `main`, just wanted to double check

.github/workflows/test-e2e.yaml
```diff
@@ -0,0 +1,154 @@
+# This file is used for tests that use Nx app containers.
+# For tests that don't require containers to be built first, see test.yaml
```
joshgamache, line 2:
```suggestion
# For tests that don't require containers to be built first, see test-code.yaml
```

I assume this was a leftover comment meant to be `test-code.yaml`, since `test.yaml` has been deleted?

.github/workflows/test-backend.yaml
```diff
@@ -0,0 +1,26 @@
+# This file is used for tests that use the backend container
+# For tests that don't require containers to be built first, see test.yaml
```
joshgamache, line 2:
```suggestion
# For tests that don't require containers to be built first, see test-code.yaml
```

I assume this was a leftover comment meant to be `test-code.yaml`, since `test.yaml` has been deleted?

joshgamache (approved):
Since I've been running the `cas-bciers` chart/build PR off of this, I haven't run into a single issue with the workflows. So I'd say that's a good test. It's robust and will be easy to add to.

I've flagged that we'll be able to make use of `nx show projects --affected` to determine which of these targets to run or not in [#264](https://github.com/orgs/bcgov/projects/123/views/1?pane=issue&itemId=67876377). The breakout work you've done here to split the builds (especially the `nx-build-app` action) will make that trivial, excellent!

Went through the code and it looks great. Flagged a few comments, but the changes are trivial. Approved! 🌮

.github/workflows/test-e2e.yaml
marcellmueller: Thanks for catching these!

PR #1759, opened 2024-06-06
docs: model and testing docs for v2

BCerki:
card: https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=62182422

Most of the documentation of the card was already completed. I added a file for models and a little more info about testing to meet the AC.

### DIFF
--- a/docs/registration_part_2/backend-structure.md
+++ b/docs/registration_part_2/backend-structure.md
@@ -62,3 +62,7 @@ class UserDataAccessService:
 ```
 
 If significant changes are made to the service layers, consider creating `v1` and `v2` sub-folders for the service layers as well.
+
+## Tests
+
+The existing testing commands run both `v1` and `v2` tests. See [test.md](../backend/test.md) for more details.
--- a/docs/registration_part_2/data-models.md
+++ b/docs/registration_part_2/data-models.md
@@ -19,6 +19,7 @@ If data is no longer necessary for v2 but existed in v1:
 
 - Set the column to nullable and blank in the model file
 - Run migrations
+- Check if that field is being referenced anywhere in the codebase. If it is, we must remove that reference to avoid breaking the app
 - No test updates needed--we don't test if fields are mandatory or not
 
 ## Updating New Tables and Columns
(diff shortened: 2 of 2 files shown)


docs/registration_part_2/backend-structure.md
```diff
@@ -62,3 +62,7 @@ class UserDataAccessService:
 ```
 
 If significant changes are made to the service layers, consider creating `v1` and `v2` sub-folders for the service layers as well.
+
+## Tests
+
+The existing testing commands run both `v1` and `v2` tests. See [test.md](../backend/test.md) for more details.
```
BCerki, line 68: @Sepehr-Sobhani , is this correct? It seemed to work like this when I tried it

docs/registration_part_2/data-models.md
```diff
@@ -0,0 +1,29 @@
+# Data Model Guidelines
+
+## Overview
+
+Ideally, our Django models will support both v1 and v2. We'll achieve this mainly with use of `null=True`, but in some cases more may be required.
+
+## Adding New Tables and Columns
+
+If you need to add something new, you can directly edit the model files (in `bc_obps/registration/models`):
```
BCerki, line 9: I ~think this is the right directory once Sepehr's PR gets merged

docs/registration_part_2/backend-structure.md
Sepehr-Sobhani: Yes, `make pythontest` looks for all python tests starts with `test_`.

docs/registration_part_2/data-models.md
```diff
@@ -0,0 +1,29 @@
...
+- Create and run migrations
+- Add tests to the v2 test folder. The testing commands in [test.md](../backend/test.md) run the tests for both v1 and v2
+
+## Removing Columns
+
+If data is no longer necessary for v2 but existed in v1:
+
+- Set the column to nullable and blank in the model file
+- Run migrations
+- No test updates needed--we don't test if fields are mandatory or not
```
Sepehr-Sobhani, line 22: It would be good to check if that field is being referenced anywhere in the codebase. If it is, we must remove that reference to avoid breaking the app.

Sepehr-Sobhani (approved):
Looks great! Just a one-liner suggestion.

PR #1760, opened 2024-06-06
1322 confirmation message

BCerki:
card: https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=59180231

This PR:
- shows a different confirmation message if the operation already has a BORO ID
- no changes needed to the email service--emails were already being sent conditionally based on application status (see cas-registration/bc_obps/service/operation_service.py)

To test:
- Make a new operation. You should get the original message.
- Edit an approved operation. You should get the new message.


Sepehr-Sobhani (approved):
Neat 🤩

PR #1761, opened 2024-06-07
1686 vitest grid components

BCerki:
Tests for https://github.com/orgs/bcgov/projects/122/views/2?filterQuery=&pane=issue&itemId=64781153

The in-progress auth PR is making a bunch of changes to import statements, so in this PR, I've got the v2 tests in reg part 1. You can run them with `yarn reg1:test` or to navigate to ~/cas-registration/bciers/apps/registration and run `yarn vitest`

### DIFF
--- a/bciers/apps/registration/app/components/operations/Operations.tsx
+++ b/bciers/apps/registration/app/components/operations/Operations.tsx
@@ -13,7 +13,7 @@ export default async function Operations({
   role,
 }: {
   searchParams: OperationsSearchParams;
-  role: FrontEndRoles;
+  role: string | undefined;
 }) {
   // Fetch operations data
   const operations: {
--- a/bciers/apps/registration/app/components/operations/Operations.tsx
+++ b/bciers/apps/registration/app/components/operations/Operations.tsx
@@ -5,27 +5,7 @@ import OperationDataGrid from "./OperationDataGrid";
 import { FrontEndRoles } from "@/app/utils/enums";
 import { OperationRow, OperationsSearchParams } from "./types";
 import { actionHandler } from "@/app/utils/actions";
-
-// 🛠️ Function to fetch operations
-export const fetchOperationsPageData = async (
-  searchParams: OperationsSearchParams,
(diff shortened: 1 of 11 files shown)


bciers/apps/registration/app/components/operations/Operations.tsx
```diff
@@ -5,27 +5,7 @@ import OperationDataGrid from "./OperationDataGrid";
...
-    );
-    return {
-      rows: pageData.data,
-      row_count: pageData.row_count,
-    };
-  } catch (error) {
-    throw error;
-  }
-};
+import fetchOperationsPageData from "./fetchOperationsPageData";
```
BCerki, line 8: Had to refactor the component file to facilitate testing

bciers/apps/registration/app/components/operations/OperationsPage.tsx
```diff
@@ -21,6 +22,7 @@ export default async function OperationsPage({
    * getServerSession requires passing the same object you would pass to NextAuth
    */
   const session = await getServerSession(authOptions);
+  // @ts-ignore
```
BCerki, line 25: I ~think the auth PR should fix the session types. If it's not in before this PR, I'll add a bullet to its card to remove this ignore

bciers/libs/shared/testConfig/src/mocks.ts
shon-button, line 1: Since Registration1 will use next-auth v4 and all other apps will use auth v5, Registration1 will continue to use `getServerSession` but other apps will use `auth`; so, suggest localizing `const getServerSession = vi.fn();` to a mock file in Regitration1 folder structure.

shon-button (approved):

bciers/libs/shared/testConfig/src/mocks.ts
BCerki: Trying to move this caused import problems too, so I'll implement the suggestion in the last PR for this card once the auth PR (which should fix the imports) is in

PR #1766, opened 2024-06-07
fix: terraform job's template to be time-bound

joshgamache:
Fix to address the issue identified by Sepehr in the Terraform Job deployment. _Should_ have been counted against time-based quota rather than long-running.


Sepehr-Sobhani (approved):
🙏

Issue #1774, opened 2024-06-07
Add django superuser to DEV namespace[5]


Sepehr-Sobhani:
###Description of the Tech Debt###

Adding a Django admin superuser to the bciers DEV namespace to ease granting access to internal users.

###Tech Debt Triage#
The purpose of our technical debt triage process is to analyze technical debt to determine risk level of the technical debt and the value in tackling that technical debt.

Risk Value Scoring:

| Level  | Value |
| ------ | ----- |
| High   | 3     |
| Medium | 2     |
| Low    | 1     |

| Technical Debt - Risk Types                                                                                                   | Level | Value |
| ----------------------------------------------------------------------------------------------------------------------------- | ----- | ----- |
| Business Area Risk - Risk of business area visibility / damage to user experience                                             |       |   1    |
| Developer Fault Risk - How likely will this tech debt cause a future error related to coding on top of it                     |       |   1    |
| System Fault Risk - Risk of system errors or application downtime                                                             |       |  1     |
| Time Scale Risk - Compound risk effect if left alone. How much more difficult to fix or dangerous will this become over time? |       |    1   |
| Time Sink Risk - How much will this tech debt slow the development process down                                               |       |  1     |
| TOTAL SCORE:                                                                                                                  |       |   5    |

Sepehr-Sobhani:
@dleard Do you have any concerns about this?

dleard:
> @dleard Do you have any concerns about this?

As long as we make sure the user is only created in dev I have no issues

hannavovk:
Thanks @Sepehr-Sobhani!

Sepehr-Sobhani:
@hannavovk The work is complete, but I still need to make some adjustments to make it fully functional. I'll make sure to reach out to you once it's ready.

PR #1786, opened 2024-06-10
Job to add django admin in dev

Sepehr-Sobhani:
[ISSUE 1774](https://github.com/bcgov/cas-registration/issues/1774)

### DIFF
--- a/bc_obps/registration/management/commands/create_superuser.py
+++ b/bc_obps/registration/management/commands/create_superuser.py
@@ -1,8 +1,9 @@
 import os
+
 from django.conf import settings
-from django.core.management.base import BaseCommand
 from django.contrib.auth import get_user_model
 from django.core.management import call_command
+from django.core.management.base import BaseCommand
 
 
 class Command(BaseCommand):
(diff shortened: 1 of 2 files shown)


bc_obps/registration/management/commands/create_superuser.py
```diff
@@ -1,19 +1,20 @@
...
-        if settings.DEBUG:
-            superuser_username = os.environ.get('SUPERUSER_USERNAME')
-            superuser_password = os.environ.get('SUPERUSER_PASSWORD')
-            if not superuser_username or not superuser_password:
+        superuser_username = os.environ.get('DJANGO_SUPERUSER_USERNAME')
+        superuser_password = os.environ.get('DJANGO_SUPERUSER_PASSWORD')
+        if superuser_username and superuser_password:
+            User.objects.create_superuser(username=superuser_username, password=superuser_password)
+        else:
+            if settings.DEBUG:  # we don't want to call this in production
```
andrea-williams, line 19: just want to make sure I'm understanding this correctly - the way we ensure that the Django admin panel isn't available in prod is that we won't be setting env vars in prod for DJANGO_SUPERUSER_x and DEBUG will be turned off?

Sepehr-Sobhani: Yes, and also we only run `create_superuser` command in dev.

andrea-williams (approved):

dleard (approved):

PR #1799, opened 2024-06-11
Feat: 256 dynamic config service

dleard:
Building off of the model work done in #1747, this PR adds the `build-form-schema` endpoint to the registration API. It calls a service that:

returns a base_schema and an enum of valid gas types when passed activity & source_type ids
returns a modified schema and an enum of valid methodologies when passed an activity, source_type & gas_type ids
returns a modified schema and a set of extra fields dynamically chosen from the configuration when passed activity, source_type, gas_type and methodology ids

All requests also require a report_date, which is currently hardcoded to default to "2024-04-01". This hardcoded value should be removed once we're pulling the report_date from the report that is calling the endpoint.

All return values also contain mapping objects of gastypes / methodologies to their IDs. This will enable us to match a selection in the formData on the frontend to an ID and pass that ID in the request (or add it to a database record if needed).


dleard:
closing this, the work was migrated to #1922

PR #1806, opened 2024-06-12
DataGrid vitest

BCerki:
card: https://github.com/orgs/bcgov/projects/122/views/2?filterQuery=&pane=issue&itemId=60748637

This PR adds some of the tests described in the card:
- rendering
- sorting
- url changes when paginating

I'm switching over to more high-priority work, so the card will go back into tech debt backlog after this PR is merged.

### DIFF
--- a/bciers/libs/shared/components/src/lib/datagrid/DataGrid.tsx
+++ b/bciers/libs/shared/components/src/lib/datagrid/DataGrid.tsx
@@ -51,7 +51,6 @@ const DataGrid: React.FC<Props> = ({
   const { replace } = useRouter();
   const isRowsEmpty = rows.length === 0;
 
-  // brianna I don't think we need this, the second use effect should do it
   useEffect(() => {
     setIsComponentMounted(true);
 
@@ -59,7 +58,6 @@ const DataGrid: React.FC<Props> = ({
   }, []);
 
   useEffect(() => {
-    // brianna can the check we on if we have props?
     // Don't fetch data if the component is not mounted
     // Since we will grab the first page using the server side props
     if (!isComponentMounted || !fetchPageData) return;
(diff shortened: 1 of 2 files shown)


bciers/libs/shared/components/src/lib/datagrid/DataGrid.tsx
```diff
@@ -51,13 +51,15 @@ const DataGrid: React.FC<Props> = ({
...
 
+  // brianna I don't think we need this, the second use effect should do it
   useEffect(() => {
     setIsComponentMounted(true);
 
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
   useEffect(() => {
+    // brianna can the check we on if we have props?
```
marcellmueller, line 62: I think I understand what you're getting at here though I may be misinterpreting this grammar haha, though if this was:

` if (!props || !fetchPageData) return;`

then it would never run if there was no initial data?

bciers/libs/shared/components/src/lib/datagrid/DataGrid.tsx
```diff
@@ -51,13 +51,15 @@ const DataGrid: React.FC<Props> = ({
   const { replace } = useRouter();
   const isRowsEmpty = rows.length === 0;
 
+  // brianna I don't think we need this, the second use effect should do it
```
marcellmueller, line 54: It's been awhile since I developed this but I'm not sure if the `searchParams` dependency array ran every time or just when there was search params. I think it may work just adding `setIsComponentMounted(true)` in the second useEffect, though there is the possibility that will cause just pass the ` if (!isComponentMounted || !fetchPageData) return;` the first time around.

Having them separate ensures that won't happen and is just clearer and more intentional.

bciers/libs/shared/components/src/lib/datagrid/DataGrid.tsx
BCerki: These are leftover; since I'm switching gears I'm not going to change anything about the component. As for what I was thinking here... I think I was confused by the comment on line 64 about server side props vs intialData

marcellmueller (approved):
Nice tests! Might want to update the data grid comment but other than that this is good to merge unless you wanted to try refactoring the data grid useEffect.

PR #1822, opened 2024-06-13
feat: decouple registration auth from monorepo package

joshgamache:
Linked to Shared Auth. We need to move to Auth.js 5 to accomodate our shared auth strategy. This causes issues with our existing Registration 1 application, which are non-trivial to resolve. As Reg1 is feature-complete, it makes more sense to freeze it as-is, which requires decoupling of the packages from the wider monorepo. 

## Changes 🚧

- Create a yarn workspace in the Registration 1 application. The packages will be pinned to today, allowing the wider monorepo to change and update over time.
- Add script targets for Reg 1. These work for Nx anywhere in the wider repository, or with Yarn in the `bciers/app/registration1` directory.
- Add ephemeral `vitest...timestamp...` files to ignored files. See https://github.com/vitejs/vite/issues/13267#issuecomment-2143958552
 
## To test 🔬

- In the `bciers` directory, run `yarn install`. This will install packages for the Monorepo _and_ Registration1
- Run `nx run registration1:dev`. The development server should start as normal.

## Notes 📝

There are circular dependencies where parts of Registration1 rely on components in the monorepo, which rely on parts of Reg1. These should also be decoupled over time.


shon-button:
@joshgamache 

I ran into issues with local run\build; so, I made a new "debug" [PR](https://github.com/bcgov/cas-registration/pull/1823) based on your branch...this branch runs and builds locally but does not pass CI Nx builds...thoughts?

joshgamache:
Testing branch. Closing.

Issue #1610, opened 2024-05-06
Fix Vitest hot reloading in Nx 


marcellmueller:
Vitest hot reloading was broken in our upgrade to Nx. This was a really nice feature that made writing tests a lot more enjoyable so we should fix it soon. I was going to do it in another PR I was working in though it requires us to upgrade Nx due to a bug that was fixed in a recent release.

Nx PR for more context: https://github.com/nrwl/nx/pull/22355
Nx release with fix: https://github.com/nrwl/nx/releases/tag/18.3.0

nx/vite:test documentation: https://nx.dev/nx-api/vite/executors/test

- [ ] Upgrade Nx 
- [ ] Setup nx vite plugin in `/bciers/apps/registration/project.json`and add new test command `yarn reg:test` with watch flag in `bciers/package.json`.

PR #1832, opened 2024-06-13
1610 fit vitest hot reloading

marcellmueller:
#1610 

This was easier than I expected, I just updated nx to the latest version `v19.3.0` using `nx migrate latest`. I also had to update the `.eslintrc.json` in the components directory as `avoidEscape: true` threw an error as it's not allowed in the nx eslint apparently. 

Note: be sure to run `yarn` and possibly `nx reset`. If you still encounter issues then try `rm -rf .nx` in the `/bciers` directory.


bciers/package.json
```diff
@@ -74,21 +74,21 @@
     "@babel/preset-typescript": "^7.22.11",
     "@mui/material-nextjs": "^5.15.11",
     "@nx-tools/nx-container": "5.2.0",
```
joshgamache, line 76: We can update the `nx-container` plugin that we use, it's had a [version bump to v6](https://github.com/gperdomor/nx-tools/releases/tag/nx-container%406.0.1) to support nx 19 as well

marcellmueller: Thanks for that, I bumped that to match the rest of our nx setup 🙇

joshgamache (approved):
A nice straightforward fix, great to see. 

Tested it locally, though we'll see if these e2e tests sort themselves out. We might want to see if [switching to inferred rather than explicit targets in Nx](https://nx.dev/blog/2024-05-08-nx-19-release#generators-to-convert-to-project-crystal) for the e2e has any better luck?

bciers/apps/registration1/e2e/poms/operations.ts
```diff
@@ -313,7 +313,6 @@ export class OperationsPOM {
               this.buttonConfirmModal,
               this.alertDeclined,
             );
-            await expect(this.messageOperationDeclined).toBeVisible();
             break;
```
marcellmueller, line 316: Removed this as I believe it was erroneously checking for a declined messaged that appeared due to a bug:

This is what appears in `develop` branch currently as the revalidate tag causes rerender:
Screenshot 2024-06-18 at 3 48 47 PM

This is what appears with the revalidate url removed:
Screenshot 2024-06-18 at 3 49 51 PM

Note: the alert message immediately disappears with this updated version of nextjs in this PR without removing the revalidate tag which is why e2e was failing.

bciers/apps/registration1/e2e/poms/operations.ts
```diff
@@ -313,7 +313,6 @@ export class OperationsPOM {
               this.buttonConfirmModal,
               this.alertDeclined,
```
marcellmueller, line 314: This is already checking the proper alert message that is appearing properly now in this PR.

bciers/apps/registration1/e2e/poms/operations.ts
marcellmueller: > Removed this as I believe it was erroneously checking for a declined messaged that appeared due to a rerendering bug. Both alert messages show at the same time, when only the initial alert should show up unless you refresh the page as the other alert is permenant.
> 
> This is what appears in `develop` branch currently as the revalidate tag causes rerender: Screenshot 2024-06-18 at 3 48 47 PM
> 
> This is what appears in this branch with updated nextjs/react and the revalidate url removed. If we don't fix the revalidate tag this message will immediately disappear in this PR due to rerendering which is why e2e was failing. Screenshot 2024-06-18 at 3 49 51 PM



> Removed this as I believe it was erroneously checking for a declined messaged that appeared due to a rerendering bug. Both alert messages show at the same time, when only the initial alert should show up unless you refresh the page as the other alert is permenant.
> 
> This is what appears in `develop` branch currently as the revalidate tag causes rerender: Screenshot 2024-06-18 at 3 48 47 PM
> 
> This is what appears in this branch with updated nextjs/react and the revalidate url removed. If we don't fix the revalidate tag this message will immediately disappear in this PR due to rerendering which is why e2e was failing. Screenshot 2024-06-18 at 3 49 51 PM


To test this check out this branch and:

- run `/bciers && yarn && nx reset` 
-  log in as internal user and go to the Operations grid and click `View details` of a `Pending` Operation
- Approve or Confirm the Operations. You should only see the green personalized alert. If you refresh the page you will see the generic permenant alert.

PR #1837, opened 2024-06-14
1361 facilities form

BCerki:
card: https://github.com/orgs/bcgov/projects/122/views/2?filterQuery=&pane=issue&itemId=59362873

This PR:
- creates a page for the new, editable facilities form
- create a page for the viewable facilities form
- creates facilities form components
- adds light error handling to the single step component
- new api routes and schemas
- updates models
- service, model, endpoint, and vitests (no e2e tests)

To test, go to the operations grid, click into one of the operations, and click the "Add facility" button

Notes:
- I'm struggling with some `mypy` typings, would love some help! (this is why precommit is failing)
- I'm also struggling with the vitests, so the nx check is failing. I hope to have them sorted out by the second round of review
- We don't have a design for the well authorization numbers, so I used the array template we already have.
- Lat and long are decimals and pydantic changes decimals into strings: https://github.com/pydantic/pydantic/issues/7457. I tried transforming the decimals as described in that issue but it didn't work, so I was thinking of leaving it for the edit-facility ticket. It doesn't cause issues creating a new facility, but it will when editing.
- In reg1 we were using the same `page.tsx` for create/edit. This caused some confusion because our parameter was either a uuid (expected) or the word "create" (not intuitive). So I've created a separate `/new/page.tsx` and `/[facilityId]/page.tsx`.
- The single step form component requires the `jsonSchema` to have sections. I've implemented this PR using the sections, but I'm not sure it's the best way to go as I had to write transformation code we wouldn't otherwise need. I think it would be more maintainable to change the single-step form (especially now that we don't need versioning and the green checkmark is sounding more optional) instead of writing this transformation for every form that will use the single-step form.  See inline comments for the areas of transformation

Changes after first round of review:
- everything going in and out of the db is a flat object (this makes services/ninja schemas/etc more reusable across all the apps)
- rjsf form schemas are still nested but there are generic transformation functions in the SingleStepTaskList to nest them (on the way in) and unnest (on the onSubmit)
- lots of updates to tests to accomodate ^


bciers/apps/registration/app/bceidbusiness/industry_user_admin/dashboard/operations/[operationId]/[formSection]/page.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+export default async function Page(params: any) {
+  console.log("params.params.operationId", params.params.operationId);
+  return (
+    <div>
+      TBD
+      <Link
+        href={`/dashboard/operations/${params.params.operationId}/facilities/new`}
+      >
+        <Button variant="contained">Add Facility</Button>
```
Comment (BCerki) [inline bciers/apps/registration/app/bceidbusiness/industry_user_admin/dashboard/operations/[operationId]/[formSection]/page.tsx:14]: This file will get replaced; just stuck a button on this page for now for easier navigation/testing

bc_obps/bc_obps/api.py
```diff
@@ -1,12 +1,20 @@
...
 from registration.api import router as registration_router
+from ninja.errors import ValidationError
 
 # Docs: https://swagger.io/docs/open-source-tools/swagger-ui/usage/configuration/
 # Filtering is case sensitive matching the filter expression anywhere inside the tag.
 api = NinjaAPI(
     title="BCIERS API", docs=Swagger(settings={"filter": True, "operationsSorter": "method", "tagsSorter": "alpha"})
 )
 
+
```
BCerki, line 12: This was taken out at some point--can we put it back in and leave it until we're done form development? It helps debug ninja schema errors

bciers/apps/registration/app/components/facilities/FacilitiesForm.tsx
```diff
@@ -0,0 +1,76 @@
...
+          const endpoint = isCreating ? "registration/facilities" : `tbd`;
+          const pathToRevalidate = isCreating ? "dashboard/facilities" : `tbd`;
+          // brianna bug potential this is going to overwrite with nesting, going to have to do some processing here
+          const body = {
+            ...formData?.section1,
+            ...formData?.section2,
+            ...data.formData.section1,
+            ...data.formData.section2,
+            operation_id: params.operationId,
+          };
```
BCerki, line 50: If we use sections in the json schema, we're going to have to spread the formData at each nesting level, which I think means we can't entirely get away from FE data transformation

bc_obps/registration/schema/v1/facility.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+class FacilitySection2(ModelSchema):
+    street_address: Optional[str] = None
+    municipality: Optional[str] = None
+    province: Optional[str] = None
+    postal_code: Optional[str] = None
+
+    @model_validator(mode='after')
+    def handle_address_data(cls, values: Any) -> Any:
```
BCerki, line 23: This is where I need typing help please :heart:

bc_obps/registration/schema/v1/facility.py
```diff
@@ -0,0 +1,114 @@
...
+        if obj.address:
+            data.update(
+                {
+                    "street_address": obj.address.street_address,
+                    "municipality": obj.address.municipality,
+                    "province": obj.address.province,
+                    "postal_code": obj.address.postal_code,
+                }
+            )
+        return data
```
BCerki, line 114: We wouldn't need all this if we didn't use sections in the `jsonSchema`

bc_obps/registration/schema/v1/facility.py
```diff
@@ -16,3 +20,110 @@ class FacilityFilterSchema(FilterSchema):
...
+    well_authorization_numbers: Optional[list] = None
+
+    @model_validator(mode='after')  # type: ignore
+    def create_address_data(cls, values):
+        payload = {}
+        if hasattr(values, 'street_address') and values.street_address:
+            address_data = values.dict(include={'street_address', 'municipality', 'province', 'postal_code'})
+            payload['address_data'] = address_data
+
+        if hasattr(values, 'well_authorization_numbers') and values.well_authorization_numbers:
```
BCerki, line 75: not sure about this strategy

bc_obps/registration/models/facility.py
```diff
@@ -35,6 +41,19 @@ class Types(models.TextChoices):
         blank=True,
         null=True,
     )
```
BCerki, line 43: I had originally taken out bcghg id and swrs facility id because they're not in the form, but bcghg is in the grid (talking to Zoey about it) so I'm going to leave them both in for now

bc_obps/bc_obps/api.py
Sepehr-Sobhani: Is this still helpful? I thought our `handle_exception` decorator did the same. If not sorry for the hassle.

bc_obps/registration/schema/v1/facility.py
marcellmueller: As we discussed that with Suha yesterday it sounds like the green checks will stay in though if we run into performance issues we will validate on the existing formData on mount and also on submit.  
  
I think reporting will have to use nested data as well and not just for this validation check mark so I appreciate you being the first to sort this out 🥇 It should be a lot easier to do going forward!

bciers/apps/registration/app/components/facilities/FacilitiesForm.tsx
marcellmueller: Curious aren't you doing this in reverse when you load the formData that is returned from the database? I haven't tried this but by default `omitExtraData` is set to true in `FormBase` so it might even clean it for us automatically.

```
const newFormData = {
section1: {
     ...formData
},
section2: {
     ...formData
},
section3: {
     ...formData
},
}

 The only thing with that is we would need to be careful not to have fields named the same at the root of each nested section.

File: bciers/apps/registration/app/components/facilities/Facility.tsx [hunk 8]
```diff
@@ -0,0 +1,51 @@
...
+        schema={
+          operation.type === "Single Facility Operation"
+            ? facilitiesSchemaSfo
+            : facilitiesSchemaLfo
+        }
+        uiSchema={facilitiesUiSchema}
+        formData={facilityFormData}
+        isCreating={isCreating}
+      />
+    </>
```
marcellmueller, line 49:
```suggestion
      <FacilitiesForm
        schema={
          operation.type === "Single Facility Operation"
            ? facilitiesSchemaSfo
            : facilitiesSchemaLfo
        }
        uiSchema={facilitiesUiSchema}
        formData={facilityFormData}
        isCreating={isCreating}
      />
```

This fragment isn't needed

bciers/libs/shared/components/src/lib/form/SingleStepTaskListForm.tsx
```diff
@@ -131,6 +134,7 @@ const SingleStepTaskListForm = ({
           onSubmit={submitHandler}
           liveValidate={isLiveValidate}
         >
+          {error && <Alert severity="error">{error}</Alert>}
```
marcellmueller, line 137:
```suggestion
          <div className="min-h-6">
            {error && <Alert severity="error">{error}</Alert>}
          </div>
```

This should be wrapped in a div that is always there with a minimum height set, so that it doesn't jump. I noticed that the read only form is causing it to jump on submit as well.

There is something going on with the read only template as which is causing it to look buggy on submit when it goes readonly for a second. If you can't figure that out I can take a closer look.

marcellmueller (commented):
Nice work! Feels good having a form in!

bciers/libs/shared/components/src/lib/form/SingleStepTaskListForm.tsx
marcellmueller: Okay it seems like ReadOnly widgets need a min height set. They don't jump when some form data exists but when the form is empty they do.... I will do that in another PR myself thanks.

bciers/apps/registration/app/utils/jsonSchema/facilitiesLfo.ts
```diff
@@ -0,0 +1,78 @@
+import { RJSFSchema } from "@rjsf/utils";
```
marcellmueller, line 1: Do you think it makes sense for `jsonSchema` to live in `app/utils`? Nothing about a schema sounds very util-y to me...I think this is something that happened in early registration1 development for some reason.

Maybe it should live in `/app/data` folder or somewhere else?

bciers/apps/registration/app/components/facilities/FacilitiesForm.tsx
BCerki: I am doing it in reverse, but in the ninja schema instead of in the front end. I wanted to do everything in the ninja schema but can't here because of the nesting.

bc_obps/registration/tests/endpoints/v1/_operations/_operation_id/test_facilities.py
```diff
@@ -113,3 +114,95 @@ def test_facilities_endpoint_list_facilities_with_filter(self):
...
+            'latitude_of_largest_emissions': 5,
+            'longitude_of_largest_emissions': 5,
+            'operation_id': operation_baker().id,
+        }
+        # IRC users can't post
+        for role in ['cas_pending', 'cas_admin', 'cas_analyst']:
+            response = TestUtils.mock_post_with_auth_role(
+                self, role, self.content_type, mock_facility, custom_reverse_lazy("create_facility")
+            )
+            assert response.status_code == 401
```
Sepehr-Sobhani, line 133: Do we also need to check this for unapproved industry users?

bc_obps/service/tests/test_facility_service.py
```diff
@@ -42,3 +52,101 @@ def test_list_facilities_for_industry_user():
...
             [facility_ownership.facility_id in users_facilities_ids for facility_ownership in users_facility_ownerships]
         )
+
+
+class TestGetIfAuthorized:
+    @staticmethod
+    def test_get_if_authorized_cas_user_success():
+        user = baker.make(User, app_role=AppRole.objects.get(role_name="cas_analyst"))
+
+        facility = baker.make(Facility, latitude_of_largest_emissions=5, longitude_of_largest_emissions=5)
```
Sepehr-Sobhani, line 62: `facility_baker`

bc_obps/service/tests/test_facility_service.py
```diff
@@ -42,3 +52,101 @@ def test_list_facilities_for_industry_user():
...
+        operator = operator_baker()
+        baker.make(
+            UserOperator,
+            user_id=user.user_guid,
+            status=UserOperator.Statuses.APPROVED,
+            operator=operator,
+            role=UserOperator.Roles.ADMIN,
+        )
+        owning_operation: Operation = operation_baker(operator.id)
+        facility = baker.make(Facility, latitude_of_largest_emissions=5, longitude_of_largest_emissions=5)
```
Sepehr-Sobhani, line 80: `facility_baker`

bc_obps/service/tests/test_facility_service.py
```diff
@@ -42,3 +52,101 @@ def test_list_facilities_for_industry_user():
...
+                'name': 'zip',
+                'type': 'Single Facility',
+                'latitude_of_largest_emissions': 5,
+                'longitude_of_largest_emissions': 5,
+            },
+            'operation_id': operation_baker().id,
+        }
+
+        FacilityService.create_facility_with_ownership(user.user_guid, payload)
+        assert len(Facility.objects.all()) == 1
```
Sepehr-Sobhani, line 121: nit: we could use `Facility.objects.count()` instead of `len(Facility.objects.all())`

Sepehr-Sobhani (commented):
Great stuff 🤩 Just some nits and suggestions.

bc_obps/service/data_access_service/facility_service.py
```diff
@@ -14,3 +16,19 @@ def get_all_facilities_for_user(cls, user: User) -> QuerySet[Facility]:
...
+    def create_facility(
+        cls,
+        user_guid: UUID,
+        facility_data: DictStrAny,
+    ) -> Facility:
+        facility = Facility.objects.create(
+            **facility_data,
+            created_by_id=user_guid,
+        )
+        return facility
```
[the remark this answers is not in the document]
BCerki: Do we need to do this even when we're setting the created by on line 32?

bciers/apps/registration/app/components/facilities/FacilitiesForm.tsx
```diff
@@ -0,0 +1,79 @@
...
+          );
+          if (response.error) {
+            setError(response.error);
+            // return error so SingleStepTaskList can re-enable the submit button and user can attempt to submit again
+            return { error: response.error };
+          }
+          setConfirmation(true);
+          if (isCreating) {
+            router.replace(
+              `/operations/${params.operationId}/facilities/${response.id}`,
```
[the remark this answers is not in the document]
BCerki: Thanks! Do we always have to use titles to make the breadcrumbs look right? Are they meant to update if we scroll down to section2?

bc_obps/bc_obps/api.py
BCerki: It's helpful when getting lots of 422 (like when making forms!) because you can see which field is the problem. Handle_exception doesn't do this

bc_obps/registration/tests/endpoints/v1/_operations/_operation_id/test_facilities.py
BCerki: Added a test

bc_obps/bc_obps/api.py
Sepehr-Sobhani: We can remove this once we merge this ticket: https://github.com/orgs/bcgov/projects/122/views/2?filterQuery=&pane=issue&itemId=69067820

bc_obps/service/data_access_service/facility_service.py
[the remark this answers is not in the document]
Sepehr-Sobhani: No, we don't have to do that here (the function does the same when creating a new instance). I just suggested that to be similar to other parts of the codebase.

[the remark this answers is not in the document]
BCerki: I threw it in everywhere

Sepehr-Sobhani (approved):
Fantastic work! 🤩

Issue #1812, opened 2024-06-12
ESLint rules are not being enforced across the codebase[9]


Sepehr-Sobhani:
Describe the Bug:

I suspect that the "ignorePatterns": ["**/*"] setting in the bciers folder is causing all files to be ignored. (But I could be wrong 😃)
With this issue ESLint pre-commit hook is always happy.

Probability (how likely the bug is to happen, scored from 1-5):5
(For example, probability of 5 is something like "it happens to all users every time they log in."
Probability of 1 "only happens to certain users when a really specific and unlikely path is followed.")

Effect (how bad the bug is when it does happen, scored from 1-5):4
(For example, effect of 5 is "the entire app crashes and makes it unusable for all users" or "the bug causes the wrong data to be saved, with critical information (e.g. payment) being affected."
Effect of 1 is "It makes some styling look a little bit weird.")

Steps to reproduce the behaviour:

Look at one of the files with unused imports, like: `bciers/apps/registration/app/components/datagrid/models/operationColumns.ts`

PR #1841, opened 2024-06-14
chore: fix eslint pre-commit

marcellmueller:
#1812

Fixes pre-commit eslint, many lining errors that made it through since this was broken and fixes all of the nx project lint setup. 

Fixing the pre-commit was easy, though when we run pre-commit in CI it runs `pre-commit run --all-files` so I had to fix all the linting errors in the project. There are still linting warnings you will see if you run that command.

There were a lot of the same eslint rules being used in different projects, I have moved them all to the global eslint in `/bciers`. Projects still have the ability to define their own custom rules if that isn't working for them! To test that is working change the rules in a project eg: 

In `/bciers/apps/dashboard/.eslintrc.json` add this rule:

```
...
  "overrides": [
    {
      "files": ["*.ts", "*.tsx", "*.js", "*.jsx"],
      "rules": {
        "@typescript-eslint/quotes": [
          "error",
          "single",
          {
            "avoidEscape": true,
            "allowTemplateLiterals": true
          }
        ]
      }
    },
...
```

Then run `nx run dashboard:lint`. To confirm this won't change lint rules for other projects run something like `nx run registration:lint`.

To test that pre-commit is working add a linting error to a file:

`var linterror = 0;` 

Then try and commit it, pre-commit should be working again 🙂


bciers/apps/dashboard/middlewares/withAuthorizationDashboard.tsx
```diff
@@ -24,22 +24,22 @@ const isUnauthenticatedAllowListedPath = (pathname: string): boolean => {
...
-  for (const segment of segments) {
-    // Check if any value in the allowList matches the current segment
-    if (allowList.includes(segment)) {
-      return true; // Return true if match found
-    }
-  }
-
-  return false; // Return false if no match found
-};
+// const isAuthenticatedAllowListedPath = (pathname: string) => {
```
BCerki, line 27: Was this a temporary comment out that needs to go back in?

docs/frontend/nx-monorepo.md
```diff
@@ -70,6 +70,29 @@ A set of helm templates have been created in `{repo}/helm/cas-registration/templ
...
 
+### Creating new projects 🏗
+
+For further information, see the [Nx docs about @nx/next generators](https://nx.dev/nx-api/next/generators/application).
+
+#### Eslint
+
+When creating a new project with Nx, there is some setup that needs to be done to ensure that eslint works correctly with typescript. Nx doesn't set this up by default since there is a performance hit though we are using typescript in all of our projects. More information about this can be found in the [Nx docs](https://nx.dev/latest/react/guides/eslint).
+
+This is done by adding the following to the `apps/{project}/.eslintrc.json` or `libs/shared/{project}/.eslintrc.json` file:
```
BCerki, line 81: Suggest adding how to know which of these two files to edit. Also, if the new project doesn't need any custom rules, can it use the bciers config instead of having its own?

bciers/libs/shared/actions/.eslintrc.json
```diff
@@ -1,6 +1,10 @@
 {
```
BCerki, line 1: Could we cut down on some of the `.eslintrc.json`s and use the main one in bciers instead? `libs/shared` has three and I'm guessing that any different rules they might have ended up with weren't intentional

bciers/apps/dashboard/middlewares/withAuthorizationDashboard.tsx
marcellmueller: Well this was an unused variable so it caused a linting failure once we got that working again. I don't know if it's needed in the future so I just commented it out unlike most of the unused variables.

docs/frontend/nx-monorepo.md
marcellmueller: Well these are generated by nx with the project setup. I think there's an argument to be made to remove them and I thought about it, though instead I moved all the rules to the main `eslintrc` and left these in.  
  
I think there are good reasons to keep these in and allow customization of the rules per project as there may be rules that work better for some projects than others. Though largely I think it's good to keep them in incase we have a team agreement to change a global rule, there may be older projects that it's not worth updating or projects where that rule doesn't make sense. This way we can turn it off.

bciers/libs/shared/actions/.eslintrc.json
marcellmueller: Same as the above comment, generated by nx and good to leave in to tweak in the future. Though for the most part my opinion is that we should keep most of the entire repo on the same rules but with minor tweaks in the future or disabling new rules for projects in maintenance eg registration1.

bciers/apps/dashboard/middlewares/withAuthorizationDashboard.tsx
BCerki: Gotcha. I think this was probably replaced by `isUnauthenticatedAllowListedPath` when @shon-button and @joshgamache were working on auth

docs/frontend/nx-monorepo.md
```diff
@@ -75,6 +75,29 @@ A set of helm templates have been created in `{repo}/helm/cas-registration/templ
...
 
+### Creating new projects 🏗
+
+For further information, see the [Nx docs about @nx/next generators](https://nx.dev/nx-api/next/generators/application).
+
+#### Eslint
+
+When creating a new project with Nx, there is some setup that needs to be done to ensure that eslint works correctly with typescript. Nx doesn't set this up by default since there is a performance hit though we are using typescript in all of our projects. More information about this can be found in the [Nx docs](https://nx.dev/latest/react/guides/eslint).
+
+This is done by adding the following to the `eslintrc.json` file in the project you created in `apps/{project}/.eslintrc.json` or `libs/shared/{project}/.eslintrc.json`:
```
BCerki, line 86: I'm still not sure I follow this (though I think that has more to do with not having more general Nx docs than these docs specifically). Is the following correct?
```suggestion
When creating a new project or library with Nx, there is some setup that needs to be done to ensure that eslint works correctly with typescript. Nx creates an eslint config file by default when creating when new projects or libraries, but Nx doesn't include typescript configuration by default (there's a performance hit). Since we use typescript in all our projects, we need to add configuration for it. More information about this can be found in the [Nx docs](https://nx.dev/latest/react/guides/eslint).

This is done by adding the following to the autogenerated `eslintrc.json` file (`apps/{project}/.eslintrc.json` if you made a new project or `libs/shared/{project}/.eslintrc.json if you made a new library)`:
```

bciers/libs/shared/actions/.eslintrc.json
BCerki: Yes, it would be good to use the same rules when we can

BCerki (approved):
:cartwheeling:

docs/frontend/nx-monorepo.md
marcellmueller: Thanks, this makes sense even if every library is a project in nx (each have their own `project.json`. Though I think this is a good suggestion and makes it easier to understand! 🎉

PR #1719, opened 2024-05-29
217 create operations table

BallardRobinett:
Card: https://github.com/bcgov/cas-reporting/issues/217

Creating a page showing operators a list of their operations, similar to ```OperationsDataGrid``` in registration. Both are relying on the Registration component ```DataGrid```, which has been moved to a shared component library along with other shared components. 

The page has replaced the placeholder page showing the operators API call in the reporting app.

PR #1853, opened 2024-06-17
chore: recreate PR

BallardRobinett:
Recreating this PR because I couldn't figure out why the registration1 tests were failing: https://github.com/bcgov/cas-registration/pull/1719

Card: https://github.com/bcgov/cas-reporting/issues/217

Creating a page showing operators a list of their operations, similar to OperationsDataGrid in registration. Both are relying on the Registration component DataGrid, which has been moved to a shared component library along with other shared components.

The page has replaced the placeholder page showing the operators API call in the reporting app.


pbastia (approved):
Very last thing to do is to re-run yarn to update the lock file in the branch.

PR #1856, opened 2024-06-17
fix: docker build issues

joshgamache:
Addresses build issues in Dockerfiles. Yarn was not pulling files as expected, defaulting to PnP rather than the module folder. 

## Changes 🚧

- Adds any vitest config ephemeral files to `ignore` files.
- Enforce `node_modules` for Yarn in docker build files.

## To test 🔬

- You can run `yarn nx run registration1:container`, while Docker is running, and ensure the build of the container completes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -151,6 +151,9 @@ helm/**/charts
 !.yarn/sdks
 !.yarn/versions
 
-
 # JetBrains
 .idea/
+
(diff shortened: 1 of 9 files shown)


joshgamache:
Rolling these changes into PR #1869, as they serve as a base for that.

PR #1875, opened 2024-06-21
chore: squash migrations into initial & prod_data migrations

dleard:
Just a cleanup of our migrations in the reporting django app


pbastia (approved):
:shipit:

PR #1876, opened 2024-06-21
248 start report

pbastia:
### An API endpoint on POST `/api/reporting/reports` 

Card: https://github.com/bcgov/cas-reporting/issues/248

#### Models created:
- Report
- ReportOperation (for admin operation/operator data for a report)
- ReportFacility (for individual admin facility data for a report)
- ReportingYear (To enforce FK validation for each reporting year, and store metadata associated like reporting windows and program deadlines)

#### Services
- ReportingService with a method to create a report object and pre-populate operation and facility data
- ReportingDataAccessService to handle database queries about reports

#### API routes
- `POST: /api/reporting/reports` creates a report with the above mentioned service and return its ID


joshgamache (approved):
Tests pass and the reports endpoint shows up in Django Ninja with everything as expected. 

Everything looks solid! 🌮

bc_obps/reporting/tests/api/test_reports.py
Sepehr-Sobhani, line 1: We have a helper class named `TestUtils` in `utils/helpers.py` customized for testing endpoints; Could be useful.

bc_obps/reporting/tests/api/test_reports.py
```diff
@@ -0,0 +1,69 @@
...
+from django.http import HttpResponse
+from django.test import Client
+from registration.tests.utils.bakers import operation_baker
+from reporting.models.report import Report
+from reporting.tests.utils.bakers import report_baker, reporting_year_baker
+
+
+@pytest.mark.django_db
+class TestReportsEndpoint:
+    endpointUnderTest = "/api/reporting/reports"
```
Sepehr-Sobhani, line 13: nit: here is python world 🐍 😛

bc_obps/service/data_access_service/report_service.py
```diff
@@ -0,0 +1,12 @@
...
+from reporting.models.report import Report
+
+
+class ReportDataAccessService:
+    @classmethod
+    def report_exists(cls, operation_id: UUID, reporting_year: int) -> Report | None:
+        return (
+            Report.objects.prefetch_related("operation")
+            .filter(operation__id=operation_id, reporting_year=reporting_year)
+            .first()
```
Sepehr-Sobhani, line 11: We should use `.exists()` to align with the method name. Using `.exists()` ensures that we receive a boolean value instead of an empty QuerySet when the queried item does not exist.

bc_obps/service/data_access_service/facility_service.py
```diff
@@ -14,3 +15,7 @@ def get_all_facilities_for_user(cls, user: User) -> QuerySet[Facility]:
             # Industry users can only see operations associated with their own operator
             user_operator = UserOperatorService.get_current_user_approved_user_operator_or_raise(user)
             return queryset.filter(ownerships__operation__operator_id=user_operator.operator_id).distinct()
+
+    @classmethod
+    def get_currently_owned(cls, operation: Operation) -> QuerySet[Facility]:
```
Sepehr-Sobhani, line 20: nit: Could we change the name of this function to be more relevant to its purpose? For example, `get_active_facilities_by_operation`.

bc_obps/service/report_service.py
Sepehr-Sobhani, line 1: nit: To avoid having to refactor this later, as we did in the registration app, I recommend breaking this into separate functions now. like `def _create_report_facilities` and `def _create_report_operation` and ...

bc_obps/service/tests/data_access_service/test_data_access_report_service.py
```diff
@@ -0,0 +1,40 @@
...
+from reporting.models.report import Report
+from reporting.models.reporting_year import ReportingYear
+
+
+class TestDataAccessReportService(TestCase):
+    def test_report_exists(self):
+        from service.data_access_service.report_service import ReportDataAccessService
+
+        operation = operation_baker()
+        reporting_year = baker.make(ReportingYear, reporting_year=1998)
```
Sepehr-Sobhani, line 13: `reporting_year_baker`

bc_obps/service/tests/data_access_service/test_data_access_report_service.py
```diff
@@ -0,0 +1,40 @@
...
+
+        self.assertIsNone(
+            ReportDataAccessService.report_exists(operation_id=operation.id, reporting_year=1998),
+            "Returns None if there is no report for that operation and year",
+        )
+
+        report = baker.make(Report, operation=operation, reporting_year=reporting_year)
```
Sepehr-Sobhani, line 20: `report_baker`

bc_obps/service/tests/test_report_service.py
```diff
@@ -0,0 +1,132 @@
...
+from reporting.models.reporting_year import ReportingYear
+from reporting.tests.utils.bakers import report_baker, reporting_year_baker
+from service.report_service import ReportService
+
+
+class TestReportService(TestCase):
+    from service.report_service import ReportService
+
+    def test_throws_if_operation_doesnt_exist(self):
+        baker.make(ReportingYear, reporting_year=2000)
```
Sepehr-Sobhani, line 22: `reporting_year_baker`

bc_obps/service/tests/test_report_service.py
```diff
@@ -0,0 +1,132 @@
...
+
+        with self.assertRaises(ObjectDoesNotExist) as exceptionContext:
```
Sepehr-Sobhani, line 24: 🐍

Sepehr-Sobhani (commented):
🤩

bc_obps/registration/tests/utils/bakers.py
```diff
@@ -72,16 +80,18 @@ def operator_baker(custom_properties=None) -> Operator:
     return baker.make(Operator, **properties)
 
 
-def operation_baker(operator_id: uuid.UUID = None, *args, **kwargs) -> Operation:
+def operation_baker(operator_id: uuid.UUID = None, **properties) -> Operation:
+    if "type" not in properties:
+        properties["type"] = cycle(["sfo", "lfo"])
```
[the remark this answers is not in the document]
pbastia: `*args` makes little sense as the model baker doesn't take positional arguments

bc_obps/reporting/tests/api/test_reports.py
pbastia: This class is going to see heavy refactor when we add authentication and everything else, so we'll make sure to refactor at that time

bc_obps/service/data_access_service/report_service.py
pbastia: The intent here was to return the report if it exists, so that we can redirect the user directly. The `exists` in the name is only to say that it doesn't blow up if there is no report

bc_obps/service/data_access_service/facility_service.py
pbastia: `active` might mean something different, I'll think of something

PR #1867, opened 2024-06-20
1609 dashboard tiles

marcellmueller:
#1609 and #1653

- **Did not add a new migration yet - Since things are rapidly changing I'm not sure if this is necessary until #1869 is ready and the app is in dev/test**

- **The dashboard content is still very much WIP, there are broken links (eg COAM) and the content is not finalized.** 

- There is an error logging in with a new IDIR unrelated to this PR, we need to fix the profile login flow. Approved IDIR accounts work fine.

### Changes imported from Shon's monster PR:
Includes some changes that were related to the dashboards from Shon's #1878:
- registration landing page is now the dashboard
- all `registration/dashboard` routes are now just `/registration`
- updated breadcrumbs to work with new route updates and show the bciers dashboard when in other apps

### Refactor `Header/Footer/Bread/Main/Layout`

Refactored Header/Footer/Bread/Main/Layout components to a better CSS solution for extending the backgrounds past the max page width. Now the elements sit 'naturally' on top of each other and don't use absolute positioning. This means less fiddling with `Main.tsx` margins for pages that don't have breadcrumbs or have the new `Note` component so less headache, especially with all the upcoming projects that will use this layout.

- Moved the original Layout/Header/Footer/Main/Bread components back to `registration1` so that we can modify the shared ones.
- The max page with is slightly reduced from `1536px` to `1440px` which more closely matches the Figma wireframes. It's now a custom class called `max-w-page` in `globals.css` to easily adjust it in the future.

### Updated the Tiles component
- Added two new icons, `File` and `Pulse`
- All Tiles now require an href as it makes sense to be able to click `Registration` from the BCIERS dashboard and go to the Registration dashboard
- Updated the fixtures, though they are still very WIP

BCEIRS dashboard external user:
Screenshot 2024-06-24 at 2 51 52 PM

BCEIRS dashboard internal user:
Screenshot 2024-06-24 at 2 52 22 PM

Registration dashboard external user:
Screenshot 2024-06-24 at 2 52 01 PM

Registration dashboard internal user:
Screenshot 2024-06-24 at 2 52 27 PM

Issue #1862, opened 2024-06-19
Refine multi-zone links and routing - 11


shon-button:
###Description of the Tech Debt###

The multi-zone feature now requires Nx project apps, other than dashboard, to have link be prefixed with `{project}/`.
There will be some work required to modify some features such as:

- Dashboard links
- Breadcrumbs links
- Profile link
- Routing in middlewares

#### Dev Steps

To update the common dashboard links:
- [ ] In `bc_obps/common/fixtures/dashboard/*` and prefix links with the `/appName/*`. See example `bciers/apps/registration/app/page.tsx`
- [ ]  Load changes to dashboard links using `cd bc_obps && make reset_db`
 
To implement changes in the breadcrumbs under the mulit-zone application patterned as `Dashboard\AppName\*`
**examples:**
![Image](https://github.com/bcgov/cas-registration/assets/120038448/bd4b7cc2-574c-4360-859f-ead0172386e0)
![Image](https://github.com/bcgov/cas-registration/assets/120038448/aeb55d6d-8b42-4b6b-b3d6-98fbb1272624)
- [ ] Remove the dashboard folder within the identity provider\user role folders
- [ ] Provide default props from the app's layout.tsx that include the common dashboard and the name of the app dashboard.  See examples `bciers/apps/registration/app/layout.tsx` and `bciers/apps/registration/app/page.tsx`

#### Acceptance Criteria

- [ ] Common dashboard has expected text and links, as of this time.
- [ ] Registration dashboard has expected text and links, as of this time.
- [ ] Reporting dashboard has expected text and links, as of this time.
- [ ]  Profile links to `registration/profile`
- [ ]  Middleware routing functions as expected with all login ids
- [ ] Breadcrumbs work expected text and links, as of this time.

###Tech Debt Triage#
Risk Value Scoring:

| Level  | Value |
| ------ | ----- |
| High   | 3     |
| Medium | 2     |
| Low    | 1     |

| Technical Debt - Risk Types                                                                                                   | Level | Value |
| ----------------------------------------------------------------------------------------------------------------------------- | ----- | ----- |
| Business Area Risk - Risk of business area visibility / damage to user experience                                             |       |  3     |
| Developer Fault Risk - How likely will this tech debt cause a future error related to coding on top of it                     |       | 3      |
| System Fault Risk - Risk of system errors or application downtime                                                             |       | 3      |
[...]

PR #1884, opened 2024-06-26
1862 - refine multi zone routing

marcellmueller:
PR to bring in some updates from #1862 that weren't brought in with #1867 

View #1867 for some more context on this though what I've brought in this PR:

- Updates `withAuthorization` middlewares
- Adds tests for these middlewares
- Added dev dependency [ts-mockito](https://github.com/NagRock/ts-mockito) to help mock NextResponse. It's a cool library that generates mocks out of types, could come in handy in the future
- Move action handler to `actions` project, update imports


bciers/apps/dashboard/middlewares/withAuthorizationDashboard.test.ts
```diff
@@ -0,0 +1,175 @@
+import { NextURL } from "next/dist/server/web/next-url";
+import { NextFetchEvent, NextRequest, NextResponse } from "next/server";
+import { instance, mock, reset, when } from "ts-mockito";
+import middleware from "../middleware";
```
marcellmueller, line 4: Technically this is testing `bciers/apps/dashboard/middleware.ts` though this currently is the only middleware and all the tests are about this so I thought this made more sense. Also I wasn't sure how to just test this function but found examples online of how to test the NextJS middleware stack.

bciers/apps/registration/app/components/profile/ProfileForm.tsx
```diff
@@ -0,0 +1,152 @@
...
+}
+
+export interface UserProfilePartialFormData {
+  first_name?: string;
+  last_name?: string;
+  position_title?: string;
+  email?: string;
+  phone_number?: string;
+  app_role?: { role_name: string };
+}
```
andrea-williams, line 28: I don't see why we need the `UserProfileFormData` and `UserProfilePartialFormData` interfaces defined again here when they're already defined in bciers/libs/shared/types/src/form/formData.ts ?

andrea-williams (approved):
Can confirm that everything works peachy when I actually set up my local env properly... 🙃

PR #1885, opened 2024-06-26
223 operations table more column

BallardRobinett:
Card: https://github.com/bcgov/cas-reporting/issues/223

Adding a more column to the operations table.


ayeshmcg (approved):
Great work, @BallardRobinett. We collaborated on this during our pairing session to add the more column to the operators table, including adding the MUI icon, updating the column name, and adding the menu. Everything looks good to me. Approving this PR.

PR #1901, opened 2024-07-03
chore: add FuelType model and populate table

BallardRobinett:
Card: https://github.com/bcgov/cas-reporting/issues/180

Adding FuelType model and populating table in the erc schema. Also updated the erd diagram to accommodate changes.

To test:

To test:

    Navigate to /bc_obps/ and run the following commands:
    make install_dev_tools
    make install_poetry_deps
    make start_pg
    make create_db
    make migrate
    make run

    In another terminal navigate to /bc_obps/ and run the shell with the command python manage.py shell

    Run these commands in the shell
    from reporting.models import FuelType
    FuelType.objects.count()

This should return the number of items, 88


bc_obps/reporting/migrations/0007_prod_data.py
```diff
@@ -4,350 +4,562 @@
...
+            "Sub-Bituminous Coal",
+            "Tail gas",
+            "Tires - biomass component",
+            "Tires - non-biomass component",
+            "Trona",
+            "Turpentine",
+            "Vegetable Oil",
+            "Waste Plastics",
+            "Wood Waste",
+            "CO2",
```
pbastia, line 503: Extra `C02`

bc_obps/reporting/models/fuel_type.py
```diff
@@ -0,0 +1,16 @@
+from common.models import BaseModel
+from django.db import models
+
+
+class FuelType(BaseModel):
+    """Reporting fuel type model"""
+
+    name = models.CharField(max_length=1000, db_comment="The name of a fuel type (example: Crude Oil)")
+    unit_type = models.CharField(
```
pbastia, line 9: we can probably just call this `unit`

pbastia (approved):
Very nice 👌 ! Just a couple of minor changes, approving it so you can merge that quickly once it's fixed

pbastia (changes requested):
We're actually missing the tests for the `FuelType` class that was introduced! You can take inspiration from the tests for the other models, there are utility classes that pretty much generate them

pbastia (approved):
slick!

Issue #1824, opened 2024-06-13
Nx - simplify shared component directory structure


marcellmueller:
###Description of the Tech Debt###

As per discussion at team agreements we are going to simplify the shared component directory structure.

Currently the shared components are located at `/bciers/libs/shared/components/src/lib`.

- [ ] Change to `/bciers/libs/components/src`
- [ ] Update `tsconfig` alias paths in `/bciers/tsconfig.base.json`
- [ ] Make sure nothing else is broken

Blocked by: #1361

PR #1911, opened 2024-07-04
nx: simplify shared component directory structure

marcellmueller:
#1824

Changes shared components directory structure from `/bciers/libs/shared/components/src/lib` to `/bciers/libs/components/src`. All other shared projects are in `/bciers/libs/*` now.

To test this just run the app with `yarn dev-all` and make sure everything is working, as well as quickly check `yarn reg1` for registration 1 though as long as this builds and passes CI it will be good.


bciers/apps/registration/vitest.config.mts
```diff
@@ -24,7 +24,7 @@ export default defineConfig({
       provider: "v8",
     },
     setupFiles: [
-      "../../libs/shared/testConfig/src/global.tsx",
+      "../../libs//testConfig/src/global.tsx",
```
BCerki, line 27: There's an extra slash after `libs` in a few places

bciers/libs/types/.eslintrc.json
```diff
@@ -1,8 +1,8 @@
 {
-  "extends": ["plugin:@nx/react", "../../../.eslintrc.json"],
+  "extends": ["plugin:@nx/react", "../../.eslintrc.json"],
   "ignorePatterns": ["!**/*"],
   "parserOptions": {
-    "project": ["libs/shared/actions/tsconfig(.*)?.json"],
+    "project": ["libs//types/tsconfig(.*)?.json"],
```
BCerki, line 5: I think is just a github change display thing, but is it right that this was previously actions and is now types?

BCerki:
> All other shared projects are in `/bciers/libs/*` now.

What does `projects` mean?

marcellmueller:
> > All other shared projects are in `/bciers/libs/*` now.
> 
> What does `projects` mean?

Any `nx project` that was in `bciers/libs/shared/*` ie `actions`, `middlewares` etc.

BCerki (commented):
We still have `src/libs` in some of the other projects (e.g. actions), were you wanting to clean it up there too?

bciers/apps/registration/vitest.config.mts
marcellmueller: Haha thanks for catching these, it was more than just a few. I must have messed up my find and replace 😅

bciers/libs/types/.eslintrc.json
marcellmueller: Hmm yeah I think this is some weird issue with how things were renamed. Looking at both `types` and `actions` eslint in this branch they are pointed at the correct `tsconfig` for each project.

BCerki (approved):
Just needs to pass CI

PR #1914, opened 2024-07-05
Not using mypy cache in pre-commit

Sepehr-Sobhani:


marcellmueller (approved):
Thanks for doing this, I really appreciate it! 🥇 

I deleted mypy cache and made a commit and confirmed that this didn't recreate it.

PR #1915, opened 2024-07-05
243 ReportVersion model

pbastia:
We're missing a `ReportVersion` model object that sits between a Report and the report data, so we can store multiple versions of a report.


BallardRobinett:
when I try to run the db locally and I get to ```make migrate```, its giving me the error:

```Your models in app(s): 'reporting' have changes that are not yet reflected in a migration, and so won't be applied.

  Run 'manage.py makemigrations' to make new migrations, and then re-run 'manage.py migrate' to apply them.```
  
  This is the migration it wants to generate:
  
  ```# Generated by Django 5.0.6 on 2024-07-05 20:53
 
import django.db.models.deletion
from django.db import migrations, models
 
 
class Migration(migrations.Migration):
 
    dependencies = [
        ('reporting', '0008_schema_data'),
    ]
 
    operations = [
        migrations.AlterField(
            model_name='reportversion',
            name='is_latest_submitted',
            field=models.BooleanField(db_comment='True if this version is the latest submitted one', default=False),
        ),
        migrations.AlterField(
            model_name='reportversion',
            name='report',
            field=models.ForeignKey(
                db_comment='The report to which this version applied.',
                on_delete=django.db.models.deletion.CASCADE,
                related_name='report_versions',
                to='reporting.report',
            ),
        ),
    ]```

joshgamache (approved):
Everything passing now locally and a skim through the code looks good me! 🌮

BallardRobinett (approved):
seems to be working on my local too, great work

dleard (approved):
lgtm

Issue #1909, opened 2024-07-04
Create Datepicker Widget


marcellmueller:
#### Describe the task

We will need a datepicker widget for upcoming froms

#### Acceptance Criteria

- [x] Use MUI datepicker https://mui.com/x/react-date-pickers/date-picker/
- [x] Store date in UTC timestamp
- [x] Set time to 9:00am UTC to mitigate incorrect days being stored due to PST difference

#### Additional context

- Add any other context about the task here.
- Or here

PR #1920, opened 2024-07-08
1909 datepicker widget

marcellmueller:
Implements #1909 

To test this set a string form field to use `DateWidget` in the `uiSchema`. For developing this I was using the `facilitiesSfo` schema since that is the only registration 2 form we have. I changed the section1 part of the `uiSchema` located at the bottom of the file to this:

```
  section1: {
    "ui:FieldTemplate": SectionFieldTemplate,
    name: {
      "ui:widget": "DateWidget",
    },
    type: {
      "ui:widget": "ComboBox",
    },
  },
```

Log in as `bc-cas-dev` external user - now we can use the date widget for the `Facilities Name` field at the top of the form located at http://localhost:3000/registration/operations/1ec71316-b073-4ba8-ae6b-2a7713d536e9/facilities/new

Something I struggled with this widget is if the user clicks the text input, should the date picker open? By default it doesn't, though something I considered adding.

Besides the UI the important part of this widget is saving the date at 9am UTC as Andrea suggested. This *should* prevent any issues of the date being a day off due to time zone differences which has been a pain point for previous projects.


shon-button:
For your consideration, rather than "if the user clicks the text input...the date picker opens" I would suggest allowing both direct text entry (on field focus) or using the calendar popup (on icon click).

shon-button:
FYI:
Following your tests steps correctly saved the selected data as seen in the breadcrumbs but, the datewidget field does not display the saved data:
image

marcellmueller:
> FYI: Following your tests steps correctly saved the selected date as seen in the breadcrumbs but, the datewidget field does not display the saved date in YYYY-MM-DD   
  
Thanks! I completely forgot to add a read only version. Also I made a minor update to the global tailwind class `read-only-widget` to stop the form from jumping when you press submit. Now it has a min-height so it won't collapse when there is no content in it, with `box-sizing: border-box` set so that minimum height will also account for the padding.

shon-button (approved):

PR #1922, opened 2024-07-09
Feat/256 form builder service

dleard:


bc_obps/service/form_builder_service.py
```diff
@@ -0,0 +1,196 @@
...
+    elif valid_source_types.count() == 1:
+        rjsf_schema['properties']['sourceTypes'] = {"type": "object", "title": "Source Types", "properties": {}}
+        build_source_type_schema(
+            activity, valid_source_types[0].source_type_id, report_date, rjsf_schema, gas_type_map, methodology_map
+        )
+
+    # If there are multiple source_types for an activity, the user may choose which ones apply. The IDs of the selected source_types are passed as a list in the parameters & we add those schemas to the activity schema.
+    else:
+        for s in valid_source_types:
+            rjsf_schema['properties'][str_to_camel_case(s.source_type.name)] = {
```
dleard, line 164: consider adding a jsonKey column on the source_type table so that we don't have to rely on the camel_case function to generate our keys from a string name

bc_obps/service/form_builder_service.py
```diff
@@ -0,0 +1,196 @@
...
+
+
+class FormBuilderService:
+    @classmethod
+    def build_form_schema(request, activity=None, source_types=[], report_date='2024-04-01'):
+        if report_date is None:
+            raise Exception('Cannot build a schema without a valid report date')
+        if activity is None:
+            raise Exception('ERROR: Cannot build a schema without Activity data')
+        schema = build_schema(activity, source_types, report_date)
```
dleard, line 195: before build_schema is called, we can get the configuration object here and pass it as a parameter, replacing report_date at this point with the set of configs that report_date applies to. We can then replace the above nested date comparisons like so:
`valid_from__valid_from__lte=report_date, valid_to__valid_to__gte=report_date`
->
`valid_from__lte=config.id, valid_to__gte=config.id`

bc_obps/service/form_builder_service.py
```diff
@@ -0,0 +1,196 @@
...
+                    .reporting_fields.all()
+                )
+            except:
+                raise Exception(
+                    f'No configuration found for activity_id {activity} & source_type_id {source_type} & gas_type_id {t.gas_type.id} & methodology_id {u.methodology.id} & report_date {report_date}'
+                )
+            methodology_object = {"properties": {"methodology": {"enum": [u.methodology.name]}}}
+            # For each field related to a methodology, add that field to the methodology object & append to the oneOf branch
+            for f in methodology_fields:
+                property_field = str_to_camel_case(f.field_name)
```
dleard, line 87: same thing here with the `jsonKey` column. Pretty much anywhere we're using the camel_case function we should consider using something more reliable

bc_obps/reporting/models/reporting_field.py
```diff
@@ -10,9 +10,12 @@ class ReportingField(BaseModel):
...
     field_type = models.CharField(max_length=1000, db_comment="Type definition for field.")
+    field_units = models.CharField(
+        max_length=1000, blank=True, null=True, db_comment="Units of measurement relating to the field."
+    )
 
     def serialize(self) -> "dict[str,str]":
-        return {"fieldName": self.field_name, "fieldType": self.field_type}
+        return {"fieldName": self.field_name, "fieldType": self.field_type, "fieldUnits": self.field_units}
 
     class Meta:
```
dleard, line 20: This table could have a unique constraint on field_name, field_type, field_units to help developers as they are adding new records to ensure that we don't end up duplicating things in this table

bc_obps/service/form_builder_service.py
dleard: jsonkey is not going to work here some of the source types have incredibly long names. Suggest exploring using IDs as keys

dleard: or a shortened jsonkey

bc_obps/service/form_builder_service.py
```diff
@@ -0,0 +1,210 @@
...
+        ]['gasType']['enum'] = gas_type_enum
+        st_schema['properties']['units']['items']['properties']['fuels']['items']['properties']['emissions']['items'][
+            'dependencies'
+        ] = gas_type_one_of
+    elif not source_type_schema.has_unit and source_type_schema.has_fuel:
+        st_schema['properties']['fuels']['items']['properties']['emissions']['items']['properties']['gasType'][
+            'enum'
+        ] = gas_type_enum
+        st_schema['properties']['fuels']['items']['properties']['emissions']['items']['dependencies'] = gas_type_one_of
+    else:
```
pbastia, line 116: Is there the possibility of a case where a source type has a unit but not a fuel?

bc_obps/service/form_builder_service.py
```diff
@@ -0,0 +1,210 @@
...
+        st_schema['properties']['emissions']['items']['properties']['gasType']['enum'] = gas_type_enum
+        st_schema['properties']['emissions']['items']['dependencies'] = gas_type_one_of
+    # Add the source_type schema to the schema object being returned by this function
+    rjsf_schema['properties']['sourceTypes']['properties'][
```
pbastia, line 120: Instead of passing in the top-level schema and making this function dependent on the data that is passed in originally, I would return `st_schema` and let the calling method put the returned schema data in the right place.

We could make this function return a tuple so that the caller also gets the name of the source type

bc_obps/service/form_builder_service.py
```diff
@@ -0,0 +1,210 @@
...
+    # If no source_types are passed & there are more than 1 valid source type, only return the activity schema
+    if len(source_types) == 0 and valid_source_types.count() > 1:
+        return json.dumps({"schema": rjsf_schema})
+
+    if valid_source_types.count() > 1:
+        # Create the Source Types object (within which all the selected source_type schemas will be defined)
+        rjsf_schema['properties']['sourceTypes'] = {"type": "object", "title": "Source Types", "properties": {}}
+        # For each selected source_type, add the related schema
+        for source_type in source_types:
+            build_source_type_schema(config, activity, int(source_type), report_date, rjsf_schema, gas_type_map, methodology_map)
```
pbastia, line 185: And this would become:
```python
rjsf_schema['properties']['sourceTypes']
    ['properties'][str_to_camel_case(source_type.name)] = build_source_type_schema(...)
```

pbastia (approved):
Looks awesome! Just a couple of nitpicks. let's send it!

bc_obps/service/form_builder_service.py
```diff
@@ -0,0 +1,210 @@
...
+            # For each field related to a methodology, add that field to the methodology object & append to the oneOf branch
+            for f in methodology_fields:
+                property_field = str_to_camel_case(f.field_name)
+                methodology_object['properties'][property_field] = {"type": f.field_type, "title": f.field_name}
+                if f.field_units:
+                    methodology_object['properties'][f"{property_field}FieldUnits"] = {
+                        "type": "string",
+                        "default": f.field_units,
+                        "title": f"{f.field_name} Units",
+                    }
```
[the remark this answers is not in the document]
dleard: Probably, I've left the question of what to do about UI schema stuff to the next card (implementing gsc page)

bc_obps/service/form_builder_service.py
dleard: I haven't seen one, but we can handle it if it comes up

PR #1930, opened 2024-07-10
273 Pathing fix

joshgamache:
Addresses final part of bcgov/cas-reporting#273. Ensure that the Registration and Reporting applications deploy to and read from the correct paths in OpenShift. 

## Changes 🚧

- Sets NextJS to build in standalone mode.
  - This was found to set the proper paths when next applications fetch their first-fetches and assets, static or otherwise.
- Use standalone as container runtime.
- Update OpenShift routes to match internal service routes.
  - NextJS handles routing to Reg and Rep when deployed locally, but we utilize OpenShift Routes (ingress) to handle routing to services and therefore pods.

## To test 🔬

> I've hand-changed Reporting and Registration's container image target in OpenShift to use this PR, and updated the routes to match.

1. Navigate to https://cas-bciers-frontend-dev.apps.silver.devops.gov.bc.ca/, login with IDIR, then click on the profile link in the header. It should display properly (all styles, etc) show the relevant profile information. 
2. From the logged-in root dashboard, click Registration>Operators.

## Notes 📝

When COAM is containerized/deployed, it will need the same ARG set in the `project.json` and `assetPath`/`basePath` in `next.config` (also, it might need to pull in `next.base.config` as well). Routes will need to follow the same pattern in the helm template deployment.

### DIFF
--- a/helm/cas-bciers/templates/reporting/route.yaml
+++ b/helm/cas-bciers/templates/reporting/route.yaml
@@ -6,7 +6,7 @@ metadata:
 {{ include "cas-bciers.labels" . | indent 4 }}
   annotations:
     haproxy.router.openshift.io/balance: roundrobin
-    haproxy.router.openshift.io/rewrite-target: {{ .Values.registrationFrontend.route.path }}
+    haproxy.router.openshift.io/rewrite-target: {{ .Values.reportingFrontend.route.path }}
 
 spec:
   host: {{ .Values.reportingFrontend.route.host }}
--- a/helm/cas-bciers/templates/reporting/route.yaml
+++ b/helm/cas-bciers/templates/reporting/route.yaml
@@ -6,7 +6,7 @@ metadata:
 {{ include "cas-bciers.labels" . | indent 4 }}
   annotations:
     haproxy.router.openshift.io/balance: roundrobin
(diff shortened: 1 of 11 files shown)


helm/cas-bciers/templates/reporting/route.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
 {{ include "cas-bciers.labels" . | indent 4 }}
   annotations:
     haproxy.router.openshift.io/balance: roundrobin
-    haproxy.router.openshift.io/rewrite-target: /
+    haproxy.router.openshift.io/rewrite-target: {{ .Values.registrationFrontend.route.path }}
```
pbastia, line 9: should that be `reportingFrontend` ?

marcellmueller (approved):
This is such a huge lift, thank you! This doesn't look like it was fun to debug at all, hope you get to pick up something easier for the next little bit 😅

helm/cas-bciers/templates/reporting/route.yaml
joshgamache: Yup, a mistype when rewriting from the testing branch. Fixed with e3c33d7.

Issue #1351, opened 2024-04-10
Admin\Contacts - External User can add a contact


zoeyli-46:
#### Description:
As an External User, I want the ability to add new contacts

[Figma](https://www.figma.com/design/LsDmLDCdnJqI8UahyqvOD8/OBPS-App-Designs?node-id=8953-2004&t=nbQdimhK7lE57KVd-4)

#### Acceptance Criteria:
Given I am an authenticated user with the `industry-user-admin` role (i.e., I have been Approved to have `admin` rights for an Operator)
When I navigate to the "Contacts" table
Then I can add a new contact

Given that I am on the contacts page
When I click the button add a new contact
Then it will open a form where I can fill out the details of the contact
https://www.figma.com/design/LsDmLDCdnJqI8UahyqvOD8/BCIERS-Designs?node-id=8953-2004&t=rBpJCzhxRzIMrjwm-4

Given that I am on the add contacts page
When I selected Yes to "Is the operation representative a BCIERS user" 
Then their information will be pre-populated instead of manually filling it out again

Given that I'm on the add contacts page
When I click submit
Then I can see a confirmation that this contact has been successfully added attn: @nanyangpro 

#### Development Checklist:

- [ ] 

**Definition of Ready** (Note: If any of these points are not applicable, mark N/A)

- [ ] User story is included
- [ ] User role and type are identified
- [ ] Acceptance criteria are included
- [ ] Wireframes are included (if required)
- [ ] Design / Solution is accepted by Product Owner
- [ ] Dependencies are identified (technical, business, regulatory/policy)
- [ ] Story has been estimated (under 13 pts)

·**Definition of Done** (Note: If any of these points are not applicable, mark N/A)

- [ ] Acceptance criteria are tested by the CI pipeline
- [ ] UI meets accessibility requirements
- [ ] Configuration changes are documented, documentation and designs are updated
- [ ] Passes code peer-review
- [ ] Passes QA of Acceptance Criteria with verification in Dev and Test
- [ ] Ticket is ready to be merged to main branch
- [ ] Can be demoed in Sprint Review
- [ ] Bugs or future work cards are identified and created
- [ ] Reviewed and approved by Product Owner

#### Dependencies
Blocking #1607

patrickisaac:
Notes:
- Pat to sync with Zoey on refining/separating these into three tickets after #1350 has been revised

zoeyli-46:
[...]

PR #1940, opened 2024-07-11
1351 add contact

Sepehr-Sobhani:
[ISSUE 1351](https://github.com/bcgov/cas-registration/issues/1351)

**NOTES:**

- Changed default behaviour of `make pythontest` command to not run in parallel when developing locally(This way we can't debug tests)
- Did some cleanup and removed some unused schemas
- Updated some schemas to use `class Meta` instead of `class Config` which is deprecated
- Added a yarn command to run component tests

**TEST:**

- Log in as external user admin -> Contacts -> Add contact -> fill out the form OR use one of the existing users from the dropdown -> submit
- Log in as internal user -> Contacts -> view details of a contact


bciers/apps/administration/app/bceidbusiness/industry_user_admin/contacts/add-contact/page.tsx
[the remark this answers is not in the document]
Sepehr-Sobhani: Very valid. Here I just tried to implement wireframes exactly as is but if that standardizes things across the app, I can change it.

bciers/apps/dashboard/auth/auth.config.ts
```diff
@@ -75,7 +75,7 @@ export default {
   ],
   secret: `${process.env.NEXTAUTH_SECRET}`,
   pages: {
-    error: "../auth/error", // Error code passed in query string as ?error=
+    error: "/auth/error", // Error code passed in query string as ?error=
```
andrea-williams, line 78: @Sepehr-Sobhani just flagging for you that I added this commit (very sneaky of me, I know). I've been having issues trying to get the branch to run locally (not because of your work, just because of our recent cycling of Keycloak secrets). I keep getting an "Invalid client secret" error in my console. But on the frontend, it was trying to redirect me to "localhost:3000/../auth/error" which was then raising another error for the invalid URL. This fixes it.

andrea-williams (commented):
Still reviewing, just pushing up the comments I do have.

bciers/apps/dashboard/auth/auth.config.ts
Sepehr-Sobhani: Thanks for the fix 🙏

bc_obps/service/data_access_service/contact_service.py
```diff
@@ -7,19 +7,25 @@
...
+            id=existing_contact_id,
             defaults={
                 "first_name": updated_data["first_name"],
                 "last_name": updated_data["last_name"],
                 "position_title": updated_data["position_title"],
                 "email": updated_data["email"],
                 "phone_number": updated_data["phone_number"],
-                "business_role": BusinessRole.objects.get(role_name="Operation Registration Lead"),
+                "business_role": updated_data.get(
+                    "business_role", BusinessRole.objects.get(role_name="Operation Registration Lead")
```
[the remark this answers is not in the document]
Sepehr-Sobhani: Just replaced `Operation Registration Lead` with `Operation Representative` but removing this option completely requires a data migration as we have PROD data.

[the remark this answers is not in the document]
andrea-williams: yep that's perfect, I wasn't expecting you to entirely delete `Operation Registration Lead`

bc_obps/service/tests/test_user_service.py
```diff
@@ -0,0 +1,57 @@
...
+class TestUserService:
+    @staticmethod
+    def test_get_user_if_authorized_cas_user_success():
+        admin_user = baker.make(User, app_role=AppRole.objects.get(role_name="cas_analyst"))
+        target_user = user_baker()
+        result = UserService.get_if_authorized(admin_user.pk, target_user.pk)
+        assert result == target_user
+
+    @staticmethod
+    def test_get_user_if_authorized_industry_user_success():
```
andrea-williams, line 26: could we please add another test case here to assert success that when there's multiple users associated with the operator, that all relevant user_operators are fetched? I want to make sure that at some point in the future we don't accidentally use `.first()` instead of `.all()` in the query

bc_obps/service/user_service.py
```diff
@@ -0,0 +1,14 @@
...
+from registration.models.user import User
+from service.data_access_service.user_service import UserDataAccessService
+
+
+class UserService:
+    @classmethod
+    def get_if_authorized(cls, admin_user_guid: UUID, target_user_guid: UUID) -> User:
+        admin_user: User = UserDataAccessService.get_by_guid(admin_user_guid)
+        target_user: User = UserDataAccessService.get_by_guid(target_user_guid)
+        if admin_user.is_industry_user() and admin_user.business_guid != target_user.business_guid:
```
andrea-williams, line 12: I think we'll need another check here to make sure that users with the `cas_pending` role aren't able to get either (they should be considered unauthorized)

bciers/apps/administration/app/components/contacts/types.ts
```diff
@@ -14,3 +16,17 @@ export interface ContactsSearchParams {
...
   sort_order?: string;
 }
+
+export interface ContactFormData {
+  selected_user?: UUID;
+  first_name?: string;
+  last_name?: string;
+  position_title?: string;
+  email?: string;
+  phone_number?: string;
```
andrea-williams, line 26: don't we need their address info here too?

andrea-williams (commented):
As usual, your code is beautiful Sepehr! ✨
Not sure if I should be hitting Approve just yet - has @BCerki has finished doing all the clicks?

BCerki:
> As usual, your code is beautiful Sepehr! ✨ Not sure if I should be hitting Approve just yet - has @BCerki has finished doing all the clicks?

Not yet, I'm waiting for some post-rebase updates

bc_obps/service/tests/test_user_service.py
Sepehr-Sobhani: Sorry I don't get it 😄  I'm pretty sure my brain is not functioning properly right now, but could you please explain more? Thanks!

andrea-williams: ahh sorry! I misread the code. I thought this was getting all users associated with the admin_user's operator. I can see now it's fetching a specific user associated with admin_user's operator. My bad!

BCerki (commented):
Andrea already did the code review, so I've only done manual testing. Notes:
- I was able to fill out the form for industry_user_admin
- I was able to view contacts when logged in with IDIR

- I'm not able to scroll when selecting a contact--is that because we only have 5, or should I be able to see more: (✅ **Sep:** Yes that is because we only have 5 users here, I added more and got scrolls)
image

- do we have a mock data user with an address I can use to check that address pre-populates too?(✅ **Sep:** User data model doesn't have the address so we only populate other fields)

- the AC mentions a confirmation page--was that cut? I don't see one in the design(✅ **Sep:** for now we have Your `Snackbar` widget unless @zoeyli-46 disagrees)

- if I switch the toggle to no after selected a bciers user, the form doesn't clear. I don't see design guidance on what should happen in that case, but it seems like we might end up with duplicate contacts: https://www.loom.com/share/820a0871558f434cbac172acd3cc9e9d
(✅ **Sep:** Yes I agree but that is inevitable as we don't have a unique field in the Contact data model and we eventually end up having duplicate contacts)cc @andrea-williams if further action is needed on this matter.

- if I click the back button after creating a contact, I get `Application error: a client-side exception has occurred (see the browser console for more information).`(✅ **Sep:** I can't replicate this issue 🤔)

- something's not quite right with email and/or phone number validation: https://www.loom.com/share/118444cfcf5742728bc33e55e952ee7c (✅ **Sep:** Thanks for catching this issue, going to fix it now)

- breadcrumbs don't match figma:
image
image
(✅ **Sep:** I mostly tried to implement the common path we are taking in other forms(like facility form) but I'm going to Ask a question in the Developer's channel and we can modify breadcrumb later if need be.)

The AC is mostly done, so I'm happy to approve this PR and then look at a second one for any of the above that needs to be addressed.

BCerki (approved):
Thanks for your work on this!

PR #1941, opened 2024-07-11
chore: run pytest tests in parallel

Sepehr-Sobhani:


.github/workflows/test-backend.yaml
```diff
@@ -17,10 +17,6 @@ jobs:
       - uses: actions/checkout@v4
       - name: dev env setup
         uses: ./.github/actions/dev-env-setup
-      - name: start backend
-        uses: ./.github/actions/run-backend
-        with:
-          django_secret_key: ${{ env.DJANGO_SECRET_KEY }}
```
marcellmueller, line 23: Happy we can optimize this. I guess we can remove `DJANGO_SECRET_KEY: ${{ secrets.DJANGO_SECRET_KEY }}` in the env section near the top of this file.

marcellmueller (approved):
Nice job, CI runs quite a bit quicker!

Locally seems to as well. I'll admit I kind of miss the list of files being tested with the green dots next to them, but I think this keeps it focused on just showing necessary data (ie failures if they happen).

🥇

Sepehr-Sobhani:
> Nice job, CI runs quite a bit quicker!
> 
> Locally seems to as well. I'll admit I kind of miss the list of files being tested with the green dots next to them, but I think this keeps it focused on just showing necessary data (ie failures if they happen).
> 
> 🥇

I'll miss them too 😄 if that's the same feeling for other folks as well, we can always run the pytest in parallel mode, only in pre-commit.

marcellmueller:
> I'll miss them too 😄 if that's the same feeling for other folks as well, we can always run the pytest in parallel mode, only in CI.

Haha it's not a big deal at all, it will be nice having faster tests. Though there was something satisfying about watching it run through the files 🙂

PR #1944, opened 2024-07-11
refactor route conditionals

marcellmueller:
Refactoring our pages which previously using role based conditional rendering to display things such as Notes and buttons based on the users role. 

This will reduce unnecessary overhead since we were wrapping our server component in another component and doing role based conditional rendering. Since we already know which role is viewing the page this is completely unnecessary, less performant, more bug prone and more complex to test.

I'm suggesting we remove that capital P Page component and create an `InternalUserLayout` and `ExternalUserLayout` if there are items at the top of the page that are role dependant. 

**Note:** we don't need to be militant about this, sometimes we will need to use conditional rendering when it makes sense and isn't at the top or bottom of the page where a layout makes sense. For example conditionally rendering a DataGrid column based on role. However we can often avoid pulling the session and role out of that...simply pass something like `isInternalUser={true}` in the `page.tsx` file based on that role.


bciers/apps/administration/app/components/operations/Operations.tsx
```diff
@@ -1,14 +1,53 @@
 import OperationDataGrid from "./OperationDataGrid";
 import { OperationRow, OperationsSearchParams } from "./types";
+import { Suspense } from "react";
+import Loading from "@bciers/components/loading/SkeletonGrid";
 import fetchOperationsPageData from "./fetchOperationsPageData";
+import Note from "@bciers/components/layout/Note";
+import Link from "next/link";
+import { Button } from "@mui/material";
+
+export const InternalUserOperationsLayout = ({
```
marcellmueller, line 10: Should these layouts live in their own file?

bciers/apps/administration/tests/components/contacts/Contacts.test.tsx
```diff
@@ -58,28 +58,4 @@ describe("Contacts component", () => {
       screen.queryByText(/No operations data in database./i),
     ).not.toBeInTheDocument();
   });
-
-  it("renders the correct note message and displays `Add Contact` button for external users", async () => {
```
marcellmueller, line 62: These weren't relevant anymore as there is no conditional rendering in the component. However I did feel slightly weird removing them, ideally they would be tested if we did `page.tsx` tests or just through e2e/screenshot testing.

bciers/apps/administration/app/components/operations/Operations.tsx
shon-button: @marcellmueller  I agree, and would vote yes to "their own file"...potentially beneficially for readability via folder structure

bciers/apps/administration/app/bceidbusiness/industry_user_admin/operations/page.tsx
shon-button, line 1: As mentioned, and for your consideration, perhaps using auth.js  Client `useSession()` or Server `auth()` in the child component, rather than sending the role as a prop, is most secure way manage the security aspect of the app.

PR #1946, opened 2024-07-11
Operations DataGrid Facilities column and Operation Information form

marcellmueller:
This PR was originally the `Add Operation` form but that was changed to be the `Register an Operation` form. I made this the `Operation Information` form and also updated the `Operations Grid` with the new facilities column and action cell text changes.  
  
- Operations Grid has new `Facilities` column with `View Facilities` button
- Operations Grid `Actions` column text is updated to `View Operation Information`
- `View Operation Information` takes you to the `Operation Information` read only form. This currently just pulls the current Operation and will need an update once `Register an Operation` is complete to return all the new data and fields.
- I kept the `ToggleWidget` in here even if it's not usable in this PR since the form is read only
- Updated the primary style of all buttons as that colour has changed in Figma to the lighter shade of blue

To test the ToggleWidget delete the `disabled` prop in the `OperationInformationForm`


bciers/apps/administration/app/components/datagrid/ActionCellFactory.tsx
```diff
@@ -1,29 +1,27 @@
...
+  const renderCell = (params: GridRenderCellParams) => {
+    return (
+      <Link
+        className="no-underline text-bc-link-blue whitespace-normal"
+        href={generateHref(params)}
+        replace={replace}
+      >
+        {cellText}
+      </Link>
+    );
```
marcellmueller, line 22: @Sepehr-Sobhani thanks for this, I borrowed it from your PR and it's working super well 😍 

 What do you think about keeping the styles in this for consistency? I don't think we should have to repeat the same styling every time we use it.  
  
If there is a good reason to want to customize styling we can pass in `className` and add it to the end of the classes, which will allow users to overwrite the default styles:  
`className={`no-underline text-bc-link-blue whitespace-normal ${className}`}

shon-button: @marcellmueller 
Is your "borrowed" code from https://github.com/bcgov/cas-registration/blob/chore/1608-facility-form-edit/bciers/libs/components/src/datagrid/cells/ActionCellFactory.tsx ?
If yes, and you like it; perhaps, put your `bciers/apps/administration/app/components/datagrid/ActionCellFactory.tsx` in the `libs` folder as I did...along with your beauty improvement suggestion about the styles.

bciers/apps/administration/app/components/operations/Operations.tsx
shon-button, line 1: For your consideration, should this be named `OperationsPage` similar to `OperationInformationPage` which might help identify that this is the shared component for the app router route `operations\page.tsx`?

If so, potentially extracting the `Layout` components to their own files might also help in file clarity.

bciers/apps/administration/app/components/datagrid/ActionCellFactory.tsx
marcellmueller: Haha I didn't even notice that, thanks!

bciers/apps/administration/app/components/operations/Operations.tsx
marcellmueller: https://github.com/bcgov/cas-registration/pull/1946/commits/51d82b9be91b46eeb51015f70f74704e3cc8b2e7

Done!

bciers/apps/administration/app/components/datagrid/ActionCellFactory.tsx
Sepehr-Sobhani: Yes, I think it's good to have a way to customize reusable components, just in case.

shon-button (approved):

shon-button:
@marcellmueller 
**`CI E2E` (Reg1) ISSUE:**
Industry User\Operations\View Detail fails to find correct route due to mising path segment ([formSection]
http://localhost:3000/dashboard/operations/827976c0-961a-4b1a-9e1d-c526b625bff6?title=Operation+6
image

Error could be reproduced `locally`:
image

**Suggested Resolution**
I reverted `bciers/apps/registration1/app/components/datagrid/cells/OperationsActionCell.tsx` to reflect `develop/bciers/libs/components/src/datagrid/cells/OperationsActionCell.tsx` which includes logic to manage ading the [FormSection] segment to the path for industry users, and modified `const ActionCell` in `bciers/apps/registration1/app/components/operations/OperationDataGrid.tsx` to include usr role- resolving the issue:
image

Issue #1938, opened 2024-07-11
Customize django migrate command


Sepehr-Sobhani:
#### Describe the task

Create a customized django migrate command that only applies migrations of each app before the latest migration tag.

#### Acceptance Criteria

- [ ] first
- [ ] second
- [ ] third

#### Additional context

- Add any other context about the task here.
- Or here

PR #1947, opened 2024-07-12
1938 customized django migrate command

Sepehr-Sobhani:
[ISSUE 1938](https://github.com/bcgov/cas-registration/issues/1938)

To **test** this command:
- Run `poetry run python manage.py migrate reporting 0004` so that reporting migrations go back to the latest tag
- Run `poetry run python manage.py showmigrations` and ensure that the reporting app migrations after the latest tag are not applied. Something like:
Screenshot 2024-07-11 at 5 08 48 PM
- Create an empty migration tag by running `poetry run python manage.py create_empty_migrations 1_7_0`(simulating a new release)
- Run `poetry run python manage.py custom_migrate`(to apply migrations based on what we want)
- Run `poetry run python manage.py showmigrations` and confirm that migrations after the latest tag are not applied.


dleard (approved):
:100:

PR #1956, opened 2024-07-15
chore: add operation registration form

marcellmueller:
Setup for the `Add or Register an Operation` form

- Set up `OperationRegistrationForm` using `MultiStepFormBase` 
- Set up a basic schema
- Set up `FacilityDataGrid` inside the facilities page - this was very difficult due to re-rendering and shallow routing issues, though it seems to be working well.
- Did some `DataGrid` optimizations by moving functions outside of the component and `memoizing` functions that needed to remain inside it.
- Updated `CheckBoxWidget` styling to match the wireframes - this will affect `Registration1` styling but in my opinion that's okay.
- Some minor `MultiStepHeader` and `MultiStepForm` styling updates
- Move shared fetch functions to `/actions/api` shared directory - Maybe need a better name or organization?
- Set up basic tests to get us started


bciers/apps/registration/app/components/operations/OperationRegistrationForm.tsx
shon-button, line 1: Would this work for you in terms navigate to a specified route with shallow routing?

import { useRouter } from "next/router";


      router.push(nextStepUrl, undefined, { shallow: true });

marcellmueller: No, since next13 app router you need to import from "next/navigation" which doesn't support shallow routing. A lot of people are annoyed 😆 

https://github.com/vercel/next.js/discussions/48110

bciers/libs/actions/src/api/getBusinessStructures.ts
```diff
@@ -0,0 +1,11 @@
+import { actionHandler } from "@bciers/actions";
```
marcellmueller, line 1: I struggled with what to name this. Is`actions/get` better for these? Either way I would like to start putting a lot of these in here to reuse.

bciers/apps/registration/app/components/operations/OperationRegistrationForm.tsx
```diff
@@ -0,0 +1,90 @@
...
+    const nextStepUrl = `/operation/${OPERATION_ID}/${
+      formSection + 1
+    }?title=${OPERATION_NAME}`;
+
+    if (isNotFinalStep) {
+      router.push(nextStepUrl);
+    }
+  };
+
+  const isFacilityDataGrid = formSection === 3;
```
marcellmueller, line 47: This is what we did for previous multi step forms but this will potentially be problematic due to the conditional page in this form. I wonder if there is a better way to figure out which page we are on?

bciers/apps/registration1/tests/shared/DataGrid.test.tsx
```diff
@@ -103,10 +99,14 @@ describe("The DataGrid component", () => {
     act(() => {
       screen.getAllByLabelText("Sort")[0].click();
     });
-    expect(mockReplace).toHaveBeenCalledTimes(2); // two times because we set params twice, once for sort_field, once for sort_order?
```
marcellmueller, line 106: I'm not entirely sure what is happening here but I believe the first call of nextjs app router replace was triggering a hard refresh since it doesn't support shallow routing so it was being called and re-rendering the datagrid twice.

PR #1961, opened 2024-07-17
237 feat: Task List component for reporting pages

pbastia:
A <ReportingTaskList > component with the associated interfaces to render a task list according to the design specs.
Card: https://github.com/bcgov/cas-reporting/issues/237

Usage:
Render the `<ReportingTaskList elements={ ... } />` in any component.
Features supported are:
- expand/collapse section items
- check marks
- active items
- links for individual items

$$ To view
start the backend: `cd bc_obps && make run`
start the frontend: `cd bciers && yarn dev-all`
visit the page at localhost:3000/reporting/demo


bciers/libs/components/src/navigation/reportingTaskList/types.ts
```diff
@@ -0,0 +1,14 @@
+export interface TaskListElement {
+  type: "Section" | "Subsection" | "Page";
+  title: string;
```
dleard, line 3:
```suggestion
  title: string;
  key: string;
```

bciers/apps/reporting/src/app/demo/page.tsx
```diff
@@ -1,3 +1,57 @@
...
+const data: TaskListElement[] = [
+  {
+    type: "Page",
+    title: "main page element",
+  },
+  {
+    type: "Section",
+    title: "Facility 1 info",
+    elements: [
+      { type: "Page", title: "Review information", isChecked: true },
```
dleard, line 13: Is the idea with `isChecked` to update this JSON object when a page is submitted & passes validation? ie: as part of the "submit' action on each page?

pbastia: That's up to the user! Something like that? We might want to hash out the detailed logic with @suhafa.  The AC for this card is "We need to be able to display a green checkmark" 😋

PR #1965, opened 2024-07-17
chore: demo folder in reporting app and auth bypass for folder

pbastia:


dleard (approved):
Nice! I'm able to get directly to the demo page

PR #1966, opened 2024-07-17
196 create review operation page

BallardRobinett:
Card: https://github.com/bcgov/cas-reporting/issues/196

Creating the first page of the new report flow. Changes include:
- Created 2 components for client and server parts of the page
- new API routes for getting a report_operation object from a version_id and saving a report.
- updating the report_operation model
- removing the restriction on the OperationType model to be only 'lfo' or 'sfo'
- created schema for RJSF usage in component


bciers/apps/reporting/src/app/components/operations/OperationReview.tsx
```diff
@@ -0,0 +1,161 @@
...
+}: {
+  formData: any;
+  version_id: UUID;
+}) {
+  const customStepNames = [
+    "Operation Information",
+    "Facilities Information",
+    "Compliance Summary",
+    "Sign-off & Submit",
+  ];
```
pbastia, line 37: Nice! We'll probably refactor these into a common component for all the reporting stuff but for the first page 👌

bciers/apps/reporting/src/app/components/operations/OperationReview.tsx
```diff
@@ -0,0 +1,161 @@
...
+                reporting_activities: formDataObject.reportingActivities || [],
+                regulated_products: formDataObject.regulatedProducts || [],
+              };
+              await actionHandler(endpoint, method, pathToRevalidate, {
+                body: JSON.stringify(body),
+                headers: {
+                  "Content-Type": "application/json",
+                  accept: "application/json",
+                },
+              });
```
pbastia, line 128: We should probably put the handler function separately, just for readabilitiy:
```typescript
const submitHandler = async () => {
... your stuff
}

return (<FormBase onSubmit={submitHandler} ... />)
```

bciers/apps/reporting/src/app/components/operations/OperationReview.tsx
```diff
@@ -0,0 +1,161 @@
...
+          <FormBase
+            schema={operationReviewSchema}
+            uiSchema={operationReviewUiSchema}
+            formData={formData}
+            onSubmit={async (data: { formData?: any }) => {
+              const method = "POST";
+              const endpoint = `reporting/save-report/${version_id}`;
+              const pathToRevalidate = `reporting/save-report/${version_id}`;
+              const formDataObject = JSON.parse(JSON.stringify(data.formData));
+              const body = {
```
pbastia, line 101: We should be able to pass the formData as the request body directly if we name the same thing in the json schema. 
Any reason why we need to provide a default for the fields?

bciers/apps/reporting/src/app/components/operations/OperationReviewFormData.tsx
```diff
@@ -0,0 +1,32 @@
...
+    `reporting/report-operation/${version_id}`,
+    "GET",
+    `reporting/report-operation/${version_id}`,
+  );
+}
+
+export default async function OperationReviewFormData({
+  version_id,
+}: {
+  version_id: UUID;
```
pbastia, line 16: the version id is an integer

bciers/apps/reporting/src/app/components/operations/OperationReviewFormData.tsx
```diff
@@ -0,0 +1,32 @@
...
+  const formData = {
+    operatorLegalName: reportOperation.operator_legal_name,
+    operatorTradeName: reportOperation.operator_trade_name,
+    operationName: reportOperation.operation_name,
+    operationType: reportOperation.operation_type,
+    BCGHGID: reportOperation.operation_bcghgid,
+    BOROID: reportOperation.bc_obps_regulated_operation_id,
+    reportingActivities: reportOperation.reporting_activities,
+    regulatedProducts: reportOperation.regulated_products,
+    operationRepresentative: reportOperation.operation_representative_name,
```
pbastia, line 28: same comment about the formData

bciers/apps/reporting/src/app/utils/jsonSchema/operations.ts
pbastia, line 1: This is not a "utils", I'd put this in `src/data/jsonSchema` or something like that. It could also live next to the component

pbastia (changes requested):
Nice work! There are a few things where we can remove code by just setting up the form/formdata/request body/API schema/backend model to align

bciers/apps/reporting/src/app/components/operations/OperationReview.tsx
BallardRobinett: I just had a default value for when it was missing in the passed data so I could easily tell whats going on. I could remove the defaults now

ayeshmcg (approved):

ayeshmcg (approved):

ayeshmcg (approved):

PR #1967, opened 2024-07-18
chore: fix coam helm deployment.yaml indentation

marcellmueller:
I noticed that this failed in shipit, also updated dependencies as I got an error testing this with `helm template`


joshgamache:
Added the missing network policies for COAM and Administration ingress with a6d8a8a.

joshgamache (approved):
🥇

PR #1968, opened 2024-07-18
chore: fix profile urls

marcellmueller:
Fixing IDIR login for new users. At some point the profile page was moved from `Registration` to `Administration`. The team decided that it was good to stay there. 

To test log in with IDIR using a fresh database, the profile redirect works and the user can fill out their information.


andrea-williams (approved):
Gonna approve this PR, because it does fix one issue.

The issue I'm experiencing with my IDIR session details not being passed to the frontend can be fixed later in a separate PR.

PR #1973, opened 2024-07-19
1964 upload bug

BCerki:
card: https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=71304153

There are two issues here:
1. Files are below the size limit (currently 20mb locally and 1mb in dev/test/prod) but can't be submitted. This is because we had a limit on the body size, not the file size specifically, and sometimes the body has extra stuff in it (probably the dataurl?) that increases the size. This is fixed in this PR: I set a file size limit and increased the body size to give us a buffer.
2. Dev/test/prod has a body size limit of 1mb even though we've got it set much higher in the next config. I ~think the problem may be that the next configs aren't loaded properly in dev/test/prod (it's fine in local). There's a console.log in the next config for testing

### DIFF
--- a/bc_obps/bc_obps/settings.py
+++ b/bc_obps/bc_obps/settings.py
@@ -172,7 +172,8 @@ DEFAULT_AUTO_FIELD = "django.db.models.BigAutoField"
 
 CORS_ALLOWED_ORIGINS = ["http://localhost:3000"]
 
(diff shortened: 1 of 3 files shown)


Sepehr-Sobhani (approved):
🥇

PR #1977, opened 2024-07-19
feat: Add actions column to reporting operations dashboard

joshgamache:
Addresses bcgov/cas-reporting#274. Adds the "Actions" column in the Reporting Dashboard with an action to Start or Continue a report.

## Changes 🚧

- Adds _Action_ column to operations table.
- Adds a _CurrentReportingYear_ context provider.

## To test 🔬

0. Run `make reset_db` in your backend folder if required. 
1. Spin up your local server, log in. 
2. Navigate to `/reporting/operations`. You should see a list of operations. 
3. Click the **Start** action for _Operation 3_. After a few moments, you should be taken to a 404'd page with the address `operations/1/review-operator-data`, with `/1/` being the report number. 
4. Navigate to the previous page, then click the **Start** action for _Operation 5_. After a few moments, you should be taken to a 404'd page with the address `operations/2/review-operator-data`, with `/2/` being the report number. 
5. Navigate to the previous page, then click the **Continue** action for _Operation 3_. After a few moments, you should be taken to a 404'd page with the address `operations/1/review-operator-data`, with `/1/` being the report number from step 3. 

## Notes 📝

`CurrentReportingYear` context should be usable for bcgov/cas-reporting#238, but should be expanded to cover all the needs of that ticket. This ticket was more focussed on supply the year to the Report generation service on the backend.

### DIFF
--- a/bciers/apps/reporting/src/app/components/context/CurrentReportingYearContext.tsx
+++ b/bciers/apps/reporting/src/app/components/context/CurrentReportingYearContext.tsx
@@ -1,0 +1,5 @@
+"use client";
+import { createContext } from "react";
+
+const CurrentReportingYearContext = createContext(2024);
+export default CurrentReportingYearContext;
--- a/bciers/apps/reporting/tests/pages/operationsColumns.test.tsx
+++ b/bciers/apps/reporting/tests/pages/operationsColumns.test.tsx
@@ -6,7 +6,7 @@ describe("operationColumns function", () => {
   it("returns an array of column definitions", () => {
     const columns: GridColDef[] = operationColumns();
 
-    assert(columns.length === 3, "Expected 3 columns");
+    assert(columns.length === 4, "Expected 4 columns");
 
     assert(
       columns[0].field === "bcghg_id",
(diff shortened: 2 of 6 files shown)


bc_obps/service/report_service.py
```diff
@@ -41,13 +41,20 @@ def create_report(cls, operation_id: UUID, reporting_year: int) -> Report:
 
         report_version = ReportVersion.objects.create(report=report)
 
+        # Map operation type from full to abbreviated, as expected by report_operation
+        abbreviated_operation_type = operation.type.lower()
+        if abbreviated_operation_type == "linear facility operation":
+            abbreviated_operation_type = "lfo"
+        elif abbreviated_operation_type == "single facility operation":
+            abbreviated_operation_type = "sfo"
+
```
pbastia, line 50: The `type` field of the `Operation` model is not constrained, so this code has a pretty high chance of breaking too at some point. I'd remove that constraint on our `ReportOperation` model instead

bciers/apps/reporting/src/app/components/context/CurrentReportingYearContext.tsx
```diff
@@ -0,0 +1,5 @@
+"use client";
```
pbastia, line 1: I'm not sure a client component to hold state is the best way to do this. We have a `ReportingYearService` (does the same as this, it's hardcoded to 2024 for now) that could be used to add an extra field on the `api/reporting/reports` API endpoint providing data for the Operations page instead, that would be less code and removing the need for managing this frontend state

bc_obps/service/report_service.py
pbastia: And this way we don't have to test that extra behaviour :p

bciers/apps/reporting/tests/pages/operationsColumns.test.tsx
```diff
@@ -25,13 +25,66 @@ describe("operationColumns function", () => {
...
+      report_status: "draft",
+    };
+
+    const params = {
+      row,
+      value: 1,
+    };
+
+    const cell = columns[2].renderCell(params);
+    expect(cell.props.children).toBe("Continue");
```
pbastia, line 87: We could mock the router and assert that the right link has been clicked / URL called

https://github.com/bcgov/cas-registration/blob/develop/bciers/apps/administration/tests/components/contacts/Contacts.test.tsx#L6

pbastia (approved):
Looks great!! Just a couple of comments

bciers/apps/reporting/src/app/components/context/CurrentReportingYearContext.tsx
joshgamache: I was thinking about this too. I had written up a quick API on top of that service, but decided to go this way. I had assumed that since there would be other places that would use it (ie. bcgov/cas-reporting#238), it might be useful to use this as a start in a single place in the frontend, even if it ultimately fetched against `ReportingYearService` to initialize. Willing to be wrong in that thinking though and just fetch against the service directly where it's needed.

bc_obps/service/report_service.py
joshgamache: I wasn't sure how critical that constraint on `ReportOperation` was, so erred on the side of changing it. Much more straightforward to just align them unconstrained.

PR #1989, opened 2024-07-23
chore: add deploy.sh script for shipit

marcellmueller:
Added a deploy script to fix `reg1` shipit deployment. Now we will be able to deploy both our original `reg1` helm chart as well as the `bciers` helm chart through shipit.

To test this you can add `--dry-run` to the `make install` and `make install_reg1` commands then:

`sudo chmod +x deploy.sh`

Replace `prefix` here with our namespace prefix:
`OBPS_NAMESPACE_PREFIX=prefix ENVIRONMENT=reg1-dev ./deploy.sh`

The terminal should say `Running reg1 make target for dev environment...` and run the `make install_reg1` script.

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+# Get the current git branch
+echo "Branch: $BRANCH"
+# Check if the shipit environment includes 'reg1'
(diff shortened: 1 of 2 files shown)


dleard (approved):
:shipit:

PR #1995, opened 2024-07-24
chore: downgrade to nextjs 14.0.4

dleard:
downgrading from 14.2.3 -> 14.0.4 to see if it fixes our upload size limit issue. The last date at which we were able to upload files larger than 1mb appears to align with the release of 14.2.0 & our upgrade to that version.


dleard:
debugging PR closing

PR #2011, opened 2024-07-25
Command to check migrations with prod data

Sepehr-Sobhani:
Part of [ISSUE 2012](https://github.com/bcgov/cas-registration/issues/2012)


dleard (approved):
:shipit: We tested this together & it works. This is a really nice check to have that enables us to have confidence in our prod migrations

Issue #1942, opened 2024-07-11
Track operational start date in "Add Facility" form


andrea-williams:
#### Describe the task

We need a dynamic label in the "Add Facility" form for "Did this facility begin operations in [preceding year] or [current year]?"

#### Acceptance Criteria

- [x] form label changes dynamically based on calendar year
Example test cases:
- [x] if the current date is Dec 31, 2024, the label says "Did this facility begin operations in 2023 or 2024?"
- [x] if the current date is Jan 1, 2025, the label says "Did this facility begin operations in 2024 or 2025?"

- [x] update `Facility` data model with nullable date field that stores the facility's starting date (when it became operational)
- [x] the datepicker widget only appears for users when they answer "yes" to the question "Did this facility begin operations in blah or blah+1?"
- [x] the selected date is stored in the facility's `starting_date` column of the DB table

#### Additional context

- See wireframe: https://www.figma.com/design/LsDmLDCdnJqI8UahyqvOD8/BCIERS-Designs?node-id=8638-2373&m=dev&t=7nNP6lo0jEvuKl5l-1
- and see comment: https://www.figma.com/design/LsDmLDCdnJqI8UahyqvOD8?node-id=8638-2373#868163190

PR #2013, opened 2024-07-25
1942 start date facility

gdalcengio:
#1942

- feat: Add conditional starting_date field in db and Add Facility form
- style: Reformat from pre commit
- test: Update tests for SFO to include starting date
- chore: [rebase] Display starting date on facility edit page and edit migrations
- test: add test for SFO readonly form
- chore: [rebase] Add starting_date and tests for LFO forms
- chore: Add validation for min/max date for dateWidget + regexp for the related textbox's year field
- chore: Reconcile migrations after rebase


gdalcengio:
note: Couldn't find a way to validate the manual entry of text in the dateWidget, but it restricts based on year and with the actual clicking of the dateWidget the range is correct.

marcellmueller (commented):
Nice work, this is working well! I believe this is the first feature using the date picker widget so looks like you needed to figure out a few things. I think we need a new `format` for dates, as the `lfo` form has this misleading error when a date is typed incorrectly:

Screenshot 2024-07-29 at 10 44 16 AM

Also this isn't directly tied to this PR but I'm not sure why the `Add facility` button isn't doing anything and assume you developed this by just going to the facility form in the url or using the placeholder button. Can we make that work? It took me a bit to figure out how to create a new facility since I've been away from the forms, can we remove this:

Screenshot 2024-07-29 at 10 42 11 AM
&nbsp;
I think we will need to add `operationTitle` query param to the `Add facility` button to make the breadcrumbs work correctly.




`http://localhost:3000/administration/operations/002d5a9e-32a6-4191-938c-2c02bfec592d/facilities/new?operationsTitle=Operation+2`



Screenshot 2024-07-29 at 10 42 22 AM

Anyways this is looking good other than that, let me know if creating a data format is difficult!

gdalcengio:
> Nice work, this is working well! I believe this is the first feature using the date picker widget so looks like you needed to figure out a few things. I think we need a new `format` for dates, as the `lfo` form has this misleading error when a date is typed incorrectly:
> 
> Screenshot 2024-07-29 at 10 44 16 AM
> Also this isn't directly tied to this PR but I'm not sure why the `Add facility` button isn't doing anything and assume you developed this by just going to the facility form in the url or using the placeholder button. Can we make that work? It took me a bit to figure out how to create a new facility since I've been away from the forms, can we remove this:
> 
> Screenshot 2024-07-29 at 10 42 11 AM
> I think we will need to add `operationTitle` query param to the `Add facility` button to make the breadcrumbs work correctly.
> 
> `http://localhost:3000/administration/operations/002d5a9e-32a6-4191-938c-2c02bfec592d/facilities/new?operationsTitle=Operation+2`
> 
> Screenshot 2024-07-29 at 10 42 22 AM
> Anyways this is looking good other than that, let me know if creating a date format is difficult!

Yeah the add facility button wasn't me and it wasn't working after a rebase. It seems to be working with the rebase I'm working on fixing now.

gdalcengio:
Concerns have been fixed and updated

bciers/libs/components/src/form/FormBase.tsx
```diff
@@ -22,6 +27,10 @@ export const customFormatsErrorMessages = {
   phone: "Format should be ### ### ####",
   email: "Please enter a valid email address, e.g. mail@example.com",
   uri: "Please enter a valid website link, e.g. http://www.website.com, https://www.website.com",
+  starting_date_format: "Starting Date format should be YYYY-MM-DD",
```
marcellmueller, line 30: Recommend making this just a generic `date_format` so it can be used for any fields that use the date picker.
```suggestion
  date_format: "Date format should be YYYY-MM-DD",
```

marcellmueller (approved):
Nice work man thanks for doing that, I know it was difficult. Also thanks for showing me that `allOf` works for two formats.

Just one minor suggestion but this is good to go! 🥇

bciers/libs/components/src/form/FormBase.tsx
gdalcengio: Edited!

Issue #1367, opened 2024-04-10
Admin\Operator - External user can view "My Operator"


zoeyli-46:
#### Description:
As an External User, given that I have registered in BCIERS, I can view and update my existing operator information, and fill in any new operator information fields asked in registration part 2
[Figma link](https://www.figma.com/file/LsDmLDCdnJqI8UahyqvOD8/OBPS-App-Designs?type=design&node-id=8638-2726&mode=design&t=KFDAnKMzwJ3b1l14-4)

#### Acceptance Criteria:

Given I am an authenticated user with the `industry-user` or `industry-user-admin` role,
When I click on the My operator tile,
Then I am brought to my operator page
[Operator page](https://www.figma.com/file/LsDmLDCdnJqI8UahyqvOD8/OBPS-App-Designs?type=design&node-id=7763-4336&mode=design&t=KFDAnKMzwJ3b1l14-4) 

Given that I am on my operator page,
When I click edit
Then it turns into editing mode 
[Figma link](https://www.figma.com/file/LsDmLDCdnJqI8UahyqvOD8/OBPS-App-Designs?type=design&node-id=8638-2726&mode=design&t=KFDAnKMzwJ3b1l14-4)

Given that I am on the Operator Information form,
When I select General Partnership or Limited Liability Partnership 
Then the fields for a Partner appear 

Given that I am on the Operator Information form,
When I select “yes” to “Does the operator have one or more parent companies?” 
Then the fields for a parent company appear

Given that I am on the Operator Information form,
When I select “yes” to “Is the parent company registered in Canada?” 
Then I see the field “Parent Company 
CRA Business Number:” 

Given that I am on the Operator Information form,
When I select “no” to “Is the parent company registered in Canada?” 
Then I see the field “Parent Company 
Tax ID Number:” 

Edit-mode of operator information
https://www.figma.com/design/LsDmLDCdnJqI8UahyqvOD8/BCIERS-Designs?node-id=8638-2726&t=oDdZLbbRzQ0FvNal-4

#### Development Checklist:

Updates to `Parent Operator` data model:
- [ ] make `cra_business_number` field nullable
- [ ] add field `tax_id_number` (nullable, varchar)

Create new data model `Partner Company` with fields:
- [ ] `partner_legal_name` (varchar, required)
- [ ] `partner_trade_name` (varchar, nullable)
- [ ] `partner_business_structure` (FK to Business Structures table, required)
- [ ] `partner_cra_business_number` (required)
[...]

PR #2015, opened 2024-07-26
1367 operator form

BCerki:
tweaks to card: #1367 

This PR includes:
- better handling of parent and partner operators (when they're adding during editing, the form opens up with correct default values)
- jsonSchema updates (required and default values)
- better error messaging
- cancel button link
- tests


bc_obps/registration/api/v2/_user_operators/_current/operator.py
```diff
@@ -25,6 +25,7 @@
     tags=V2,
     description="""Retrieves data about the current user-operator and their associated operator.
     Declined user-operators are excluded from the results.""",
+    exclude_none=True,  # To exclude None values from the response (used for parent and partner arrays)
```
marcellmueller, line 28: Oh cool, I just learned that this is a Django feature. Is this to stop null values from causing erroneous validation errors?

bciers/libs/utils/customTransformErrors.ts
```diff
@@ -6,6 +6,9 @@ const customTransformErrors = (
 ) => {
   return errors.map((error) => {
     if (error?.property) {
+      if (error.message === "must be equal to constant") {
+        error.message = undefined; // this is to handle errors generated by dependencies
+      }
```
marcellmueller, line 11: What was causing this? Why are we setting the message to `undefined`, is this just a workaround to stop an erroneous validation error?

marcellmueller (approved):
I've been out of the loop and haven't seen this form! Looks good, I like the snack bar. No nits, just a couple of questions I had! 

🥇

bc_obps/registration/api/v2/_user_operators/_current/operator.py
BCerki: Yes, to avoid validation and display errors. If there's a None value, then the FE receives whatever default value we set in the ninja schema for the field. So we get stuff like parent_operators_array = []/None, which overrides our default jsonSchema values.

bciers/libs/utils/customTransformErrors.ts
BCerki: Without this, if you click "yes" to parent operators and then hit submit without filling out all the required parent operator fields, then we get "must be a constant" validation error on the yes/no field even though it's filled. I think this is because we're using "oneOf" , so it flags the whole subschema as problematic if any one piece is missing. Same thing happens with the partner operators

marcellmueller: Makes sense, thanks for filling me in!

PR #2029, opened 2024-07-30
Show report due date on reporting dashboard

BallardRobinett:
Card: https://github.com/bcgov/cas-reporting/issues/193

Adding the due date for the current reporting period to the reporting dashboard.

Changes:
- updated reporting_year model, reflected in new migration file
- added /reporting-due-date API route
- updated fixture data
- updated erd diagram
- added due date to dashboard page

To Test:
- in ```/bc_obps``` run ```make reset_db``` and ```make run``` to make sure changes are in db
- run frontend locally from ```/bciers``` with ```yarn dev-all```
- log in with BCeID to ```bc-cas-dev``` and navigate to ```/reporting/operations``` in the address bar
- You should see Reports due May 31, 2025 in the upper right side of the screen


bciers/apps/reporting/src/app/components/routes/operations/Page.tsx
```diff
@@ -1,17 +1,24 @@
...
 import { OperationsSearchParams } from "../../operations/types";
+import { getReportingDueDate } from "@reporting/src/app/utils/getReportingDueDate";
 
 export default async function OperationsPage({
   searchParams,
 }: {
   searchParams: OperationsSearchParams;
 }) {
+  const reportingDueDate = await getReportingDueDate();
   const reportingYear = await getReportingYear();
```
pbastia, line 12: Once we've grouped these into a single API call and endpoint, it'll be much more efficient to fetch the data in one go here

bc_obps/reporting/api/reports.py
```diff
@@ -63,3 +63,16 @@ def save_report(
...
+    "/reporting-due-date",
+    response={200: str, custom_codes_4xx: Message},
+    tags=EMISSIONS_REPORT_TAGS,
+    description="""Returns the due date for the current reporting year as `[Month] [Day], [Year]`.""",
+)
+@handle_http_errors()
+def get_reporting_due_date(request: HttpRequest) -> Tuple[Literal[200], str]:
+    due_date = ReportingYearService.get_current_reporting_year().report_due_date
+    formatted_due_date = due_date.strftime("%B %d, %Y")
+    return 200, str(formatted_due_date)
```
pbastia, line 78: A few things:
- Let's group those 2 API calls into 1 that returns JSON data on GET `/reporting-year`
- We can put this in a new file `bc_obps/reporting/api/reporting_year.py`
- We can return an ISO string for date objects (this is the default behaviour I believe)
- We can use django-ninja schemas to select and return the properties we need from the ReportingYear object
- Tests!

bc_obps/reporting/models/reporting_year.py
```diff
@@ -14,6 +14,9 @@ class ReportingYear(BaseModel):
     reporting_window_end = models.DateTimeField(
         blank=False, null=False, db_comment="End of the reporting period for that reporting year, UTC-based"
     )
+    report_due_date = models.DateTimeField(
+        blank=False, null=False, db_comment="Due date for the annual report this year's reporting cycle, UTC-based"
+    )
```
pbastia, line 19: We'll need to amend the model test for this extra field

pbastia (dismissed):
Nice! There's some optimization we can do on the API side to avoid duplicating endpoints and necessary calls to fetch the data. And don't forget the tests!

bciers/apps/reporting/src/app/components/datagrid/models/operationColumns.tsx
```diff
@@ -92,8 +92,11 @@ const ActionCell = (params: GridRenderCellParams) => {
     <Button
       color="primary"
       onClick={async () => {
-        const reportingYear = await getReportingYear();
-        const newReportId = await handleStartReport(OperationId, reportingYear);
+        const reportingYearObj = await getReportingYear();
+        const newReportId = await handleStartReport(
+          OperationId,
+          reportingYearObj.reporting_year,
+        );
```
pbastia, line 99: We can leave that as a tech debt item:
2 API calls in a row to execute logic on the client side is a bit of a code smell
- Either the reporting year is pulled from the page (The operations page already fetches it for its display, we can grab that value from there),
- Or the "start report" API call grabs the current reporting year in its own server implementation

bciers/apps/reporting/src/app/utils/getReportingDueDate.ts
```diff
@@ -0,0 +1,10 @@
+import { actionHandler } from "@bciers/actions";
+
+export const getReportingDueDate = async (): Promise<string> => {
+  const reportingDueDate = await actionHandler(
+    "reporting/reporting-due-date",
+    "GET",
+  );
+
+  return reportingDueDate;
+};
```
pbastia, line 10: This API endpoint ndoesn't exist anymore, we can delete this file

pbastia (approved):
Nice work 🥇 ! Just a couple of comments. If you want to fix the tech debt right away go ahead, otherwise just write a small card and we'll schedule it!

PR #2031, opened 2024-07-31
238 reporting year fixtures

joshgamache:
Fix for deployed version of bcgov/cas-reporting#238. Fixtures existed for reporting years, which loaded locally, but were not part of the common init that would load them when deployed.


BallardRobinett (approved):
looks good

PR #2033, opened 2024-07-31
238 reporting year migrations

joshgamache:
Moves the `reporting`.`reporting_year` fixture mocks from `common`.`load_fixtures` into a `reporting` migration.

### DIFF
--- a/bc_obps/reporting/migrations/0007_prod_data.py
+++ b/bc_obps/reporting/migrations/0007_prod_data.py
@@ -614,6 +614,18 @@ def reverse_init_methodology_data(apps, schema_monitor):
     ).delete()
 
 
+def load_reporting_fixtures(apps, schema_editor):
+    from django.core.management import call_command
+
+    fixture_files = [
+        'reporting/fixtures/mock/reporting_year.json',
+    ]
+
+    # Load the fixtures
+    for fixture in fixture_files:
+        call_command('loaddata', fixture)
+
+
 class Migration(migrations.Migration):
 
     dependencies = [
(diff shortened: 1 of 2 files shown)


bc_obps/reporting/migrations/0010_add_reporting_year_data.py
```diff
@@ -0,0 +1,30 @@
+# Generated by Django 5.0.7 on 2024-07-31 20:47
+
+from django.db import migrations
```
pbastia, line 3: Maybe instead of creating a new migration, we can add it to the existing production data one?

Also no need to delete all existing objects, the table should be empty at that point. Something went wrong if it isn't

pbastia (approved):
We might be able to tack that work onto an existing migration instead, to avoid multiplying them

bc_obps/reporting/migrations/0010_add_reporting_year_data.py
joshgamache: Modified the PR to run the migrations earlier and removed the deletion with 991c629.

bc_obps/reporting/migrations/0007_prod_data.py
```diff
@@ -614,6 +614,18 @@ def reverse_init_methodology_data(apps, schema_monitor):
...
+
+    fixture_files = [
+        'reporting/fixtures/mock/reporting_year.json',
+    ]
+
+    # Load the fixtures
+    for fixture in fixture_files:
+        call_command('loaddata', fixture)
+
+
```
pbastia, line 628: Sorry about the nitpicks! This is missing a reverse function, and we can name it after the reporting year
```
def init_reporting_years(..):
```

PR #2039, opened 2024-08-01
292: facility report data

pbastia:
Small PR for https://github.com/bcgov/cas-reporting/issues/292
- Renames the `ReportFacility` model to `FacilityReport` to better highlight that we're building one report per facility
- `ReportActivity` records now belong to one Facility Report


bc_obps/reporting/models/facility_report.py
```diff
@@ -1,16 +1,31 @@
...
+    A report (each report version) may contain multiple facility reports.
+    '''
+
+    facility = models.ForeignKey(
+        Facility,
+        on_delete=models.DO_NOTHING,
+        null=True,
+        blank=True,
+        db_comment="""The facility record this report was created for, at the time the report was filled out.
+        This may be null in the future if the facility disappears or changes ownership.""",
```
dleard, line 20: Will it be null if the facility changes ownership? I think the relationship will remain in that case since the ID of the facility will remain the same.

dleard (approved):
Looks good! I just left a question about the table comment on report_facility

bc_obps/reporting/models/facility_report.py
pbastia: That makes sense! I didn't really have a good idea of the lifecycle of those things in the registration app. I'll remove the Null/Blank options

PR #2045, opened 2024-08-02
chore: add missing bciers helm test values

marcellmueller:
Adds missing values for the `Administration` and `COAM` apps for test. Currently you can [log into test](https://cas-bciers-frontend-test.apps.silver.devops.gov.bc.ca/) though since these were left out both of these apps are broken.


joshgamache (approved):
A place for everything and everything in it's place.

Issue #1895, opened 2024-07-02
Reg\Register an Operation\Page "Facility Information" - External user can register the operation


zoeyli-46:
#### Description:
As an industry user, given that the operation has been added to my operator, I can register the operation for IRC approval.

#### Acceptance Criteria:
Given that I am on the Operation Information page of the New Registration form,
When I click on the save and continue button,
Then I am brought to the Facility Information page of the form
https://www.figma.com/design/LsDmLDCdnJqI8UahyqvOD8/BCIERS-Designs?node-id=11759-156&t=5omgZedIN5xnnFFl-4

~Given that I selected EIO as my registration purpose 
Then I do NOT see facility page, the facility page is skipped over
But, in the backend the system automatically creates a facility record for them attn: @andrea-williams~
pulled out into ticket #1962  

Given that I am on the Facility Information page of the New Registration form,
When I click Add Facility, 
Then it will display the add facility form. https://www.figma.com/design/LsDmLDCdnJqI8UahyqvOD8/BCIERS-Designs?node-id=12902-157&t=5omgZedIN5xnnFFl-4 

Given that I am on the Facility Information page of the New Registration form,
Then I see a table of all my facilities
If I have not added any facilities and the table is empty, it will have an empty state message: "No facilities to show, add a facility"

Given I clicked view details on a facility row
Then I can see the details of the facility 

#### Development Checklist:

- [ ] each step of the Registration workflow form hits a different API endpoint
- [ ] see notes added to wireframe (in red) for required vs. optional fields and conditionals associated with certain fields
- [ ] for Single Facility Operations (SFOs): facility name & type to be omitted from UI. A `Facility` record should still be created/updated in the database, with the following fields autopopulated (i.e., external user doesn't have to provide any input):
       - facility name = name of operation
       - facility type = 'Single Facility'
- [ ] for SFOs, the "Add Another Facility" button should not appear
- [ ] for Linear Facility Operations (LFOs), facility name and type to be included in UI 
- [ ] Meets the DOD

**Definition of Ready** (Note: If any of these points are not applicable, mark N/A)

- [ ] User story is included
- [ ] User role and type are identified
- [ ] Acceptance criteria are included
- [ ] Wireframes are included (if required)
- [ ] Design / Solution is accepted by Product Owner
- [ ] Dependencies are identified (technical, business, regulatory/policy)
- [ ] Story has been estimated (under 13 pts)

[...]

PR #2046, opened 2024-08-02
1895 operation registration facility information

marcellmueller:
#1895 

### Operation Registration Form - Facility Information page

The main things to take note of in this PR:
- `Single Facility Operation` get the regular SFO form from the Facility form as there will only be one facility for these types. Also we have removed the DataGrid for SFO since it was deemed unnecessary
- `Linear Facility Operations` get the LFO form from the Facility form although this is implemented as an array field to allow the user to add multiple facilities at once.
- The DataGrid has been updated to use shallow routing - this will not trigger a full refresh of the page like the previous `router.replace` that we were using. This was necessary for the DataGrid sorting and filtering to work without re-rendering the entire page, resetting the form and losing the data and causing the page to scroll to the top.

Since this page is in the middle of the `Operation Registration` flow which is still in development you will need to go to some specific URLs to test SFO and LFO forms.

### What to test

- Due to the DataGrid updates see if you can find any buggy behaviour sorting/filtering the DataGrids 
- [The existing Facilities form](http://localhost:3000/administration/operations/837fd32f-2454-4127-943b-84257078ff1c/facilities/add-facility?operations_title=Operation%203) that you can get to from the `FacilitiesDataGrid`. The SFO form has pre-populated fields and the `/registration/facilities` API route was updated to accept an array to make it more reusable for single/multiple facilities.

- The LFO facility array form which also has the DataGrid:
http://localhost:3000/registration/operation/002d5a9e-32a6-4191-938c-2c02bfec592d/3?title=Placeholder+Title
- The SFO facility form (there is no DataGrid):
http://localhost:3000/registration/operation/094e74cd-c80f-4303-9860-48d6b3a89fe0/3?title=Operation+14


andrea-williams:
🥇 gold medal for adding the "What to Test" notes - super helpful!

andrea-williams (approved):

Issue #2060, opened 2024-08-09
RootLayout metadata not working correctly - no page title


marcellmueller:
The `metadata` in the Root Layout isn't working correctly so we do not have any page titles. Looks like this is because it's exported from `RootLayout`, when it needs to be exported from the layout itself.

BCIERS:

<img width="844" alt="Screenshot 2024-08-09 at 7 11 29 AM" src="https://github.com/user-attachments/assets/9f06cdef-c47e-4e3a-8936-ddd843df8c4a">

Registration1: 

<img width="844" alt="Screenshot 2024-08-09 at 7 11 35 AM" src="https://github.com/user-attachments/assets/417bd600-ed7a-43ae-bfe1-0884712ea0dc">

PR #2061, opened 2024-08-09
chore: fix root layout metadata

marcellmueller:
#2060 

The metadata in the `RootLayout` wasn't working correctly as it was exporting from that file instead of being exported from the `layout`. 

To test this first run `yarn dev-all` in `develop` and see that the title isn't working:

Screenshot 2024-08-09 at 7 47 04 AM

Checkout this branch and run `yarn dev-all` and now the title is working:  
Screenshot 2024-08-09 at 7 46 20 AM

### DIFF
--- a/bciers/apps/administration/app/layout.tsx
+++ b/bciers/apps/administration/app/layout.tsx
@@ -8,14 +8,16 @@ import "@bciers/styles/globals.css";
 
 import RootLayout, { rootMetadata } from "@bciers/components/layout/RootLayout";
 
+const title = "Administration";
+
 const defaultLinks = [
   { label: "Dashboard", href: "/" },
-  { label: "Administration", href: "/administration" },
+  { label: title, href: "/administration" },
 ];
 
 export const metadata = {
   ...rootMetadata,
-  title: `${rootMetadata.title} | Administration`,
+  title: `${rootMetadata.title} | ${title}`,
 };
 
 export default function AdministrationLayout({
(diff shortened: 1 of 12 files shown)


shon-button:
@marcellmueller 

🍰 Delicious! 
Thanks for finding and fixing...a solid UX improvement.

Two suggestions, for your consideration ;)

1. Perhaps dry-out the default value for title\zone in all the `layout.tsx` 
e.g.
```
/*
A layout is UI that is shared between routes.
The app directory must include a root app/layout.js.
*/

// eslint-disable-next-line import/extensions
import "@bciers/styles/globals.css";
import RootLayout, { rootMetadata } from "@bciers/components/layout/RootLayout";

const title = "COAM";
const defaultLinks = [
  { label: "Dashboard", href: "/" },
  { label: title, href: "/coam" },
];

export const metadata = {
  ...rootMetadata,
  title: `${rootMetadata.title} | ${title}`,
};

export default function COAMLayout({
  children,
}: {
  children: React.ReactNode;
}) {
  return (
    <RootLayout defaultLinks={defaultLinks} zone={title.toLowerCase()}>
      {children}
    </RootLayout>
  );
}

```
2. Since we do not have an [app/not-found.tsx](https://nextjs.org/docs/app/api-reference/file-conventions/not-found)  in our apps the `404 Page not found` does not respect the  rootMetadata title property:
image

If we added an `app/not-found.tsx` similar to the  `app/error.tsx` (wherein each `app/not-found.tsx` imports from a share component in `bciers/libs/components/src`) then the  rootMetadata title property would be respected.
image

image

shon-button (approved):

PR #2064, opened 2024-08-09
Chore: 295 optimize form builder service

dleard:
Removes an issue where new queries were being fetched from the db in a loop


bc_obps/service/form_builder_service.py
```diff
@@ -40,18 +43,40 @@ def build_source_type_schema(
     gas_type_enum = []
     # Maps of the gas_type & methodology objects will be passed in the return object so we have the IDs on the frontend.
     gas_type_one_of: Any = {"gasType": {"oneOf": []}}
+
+    gas_type_ids = list(map(lambda config_elt: config_elt.gas_type.id, gas_types))
```
Sepehr-Sobhani, line 47: Should be the same as: `gas_types.objects.values_list('gas_type_id', flat=True)`

dleard:
closing in favour of #2067

PR #2066, opened 2024-08-09
chore: update test-setup endpoint with load_only boolean

dleard:
The test-setup endpoint was truncating tables in the dev environment resulting in losing data that we expected to exist after deploy to dev. Truncating the data should be unnecessary here since we reset the db before deploy.

- Added a load_only boolean to the test-setup endpoint & set it to True in the call from helm


Sepehr-Sobhani (approved):
🤩

PR #2068, opened 2024-08-09
chore: replace all JSON.parse() calls with safeJsonParse

dleard:
Getting a null or invalid JSON will currently crash the app. safeJsonParse handles the error more gracefully & just returns an empty set.


BallardRobinett (approved):
Looks good!

PR #2070, opened 2024-08-09
chore: change generic activity for gsc -> gsc_excluding_line_tracing

dleard:
There were some changes to the reporting fields spreadsheet after the initial work was done.
General Stationary Combustion was split into 4 activities.
This PR renames the generic GSC to be the General stationary combustion excluding line tracing activity.


pbastia (approved):
👍 👍

PR #2073, opened 2024-08-12
chore: fix stale form state on edit

marcellmueller:
Fix for stale form data on edit. We ran into this issue with the contacts form as well, though I put the fix in the single step task list as I don' think this was caught when editing was added to that component. 

I believe what is happening is that submitting causes a full re-render and the initial `formData` gets passed back to the component causing this buggy behaviour. The fix is simple - saving the initial form data in `formState`. We are not turning this into a controlled component and updating the form state onChange, though we update the `formState` `onSubmit` which ensures the latest data on re-render.

Also cleaned up the commented code in the single step task list.

### DIFF
--- a/bciers/libs/components/src/form/SingleStepTaskListForm.tsx
+++ b/bciers/libs/components/src/form/SingleStepTaskListForm.tsx
@@ -40,18 +40,12 @@ const SingleStepTaskListForm = ({
 }: SingleStepTaskListFormProps) => {
   const hasFormData = Object.keys(rawFormData).length > 0;
   const formData = hasFormData ? createNestedFormData(rawFormData, schema) : {};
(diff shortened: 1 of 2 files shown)


bciers/libs/components/src/form/SingleStepTaskListForm.tsx
```diff
@@ -40,18 +40,12 @@ const SingleStepTaskListForm = ({
 }: SingleStepTaskListFormProps) => {
   const hasFormData = Object.keys(rawFormData).length > 0;
   const formData = hasFormData ? createNestedFormData(rawFormData, schema) : {};
+  const [formState, setFormState] = useState(formData);
```
marcellmueller, line 43: Just for clarity since I cleaned up all that commented code, this is the actual fix.

Issue #1905, opened 2024-07-04
Reg\Register an Operation\Page "Success" - External user can register the operation


zoeyli-46:
#### Description:
As an industry user, given that the operation has been added to my operator, I can register the operation for IRC approval.

#### Acceptance Criteria:

Given that I am on the submission page
When I click sign and register
Then I am brought to the success page 
And this operation’s status will now be Registered in my operations table.
[Figma](https://www.figma.com/design/LsDmLDCdnJqI8UahyqvOD8/BCIERS-Designs?node-id=14779-436&t=qiVuWLxB21oqZtHM-4)

On the success page, 
When I click Report a Change, 
Then it will take me to the report change form

#### Development Checklist:

- [ ] Checklist item
- [ ] Checklist item
- [ ] Checklist item
- [ ] Meets the DOD

**Definition of Ready** (Note: If any of these points are not applicable, mark N/A)

- [ ] User story is included
- [ ] User role and type are identified
- [ ] Acceptance criteria are included
- [ ] Wireframes are included (if required)
- [ ] Design / Solution is accepted by Product Owner
- [ ] Dependencies are identified (technical, business, regulatory/policy)
- [ ] Story has been estimated (under 13 pts)

·**Definition of Done** (Note: If any of these points are not applicable, mark N/A)

- [ ] Acceptance criteria are tested by the CI pipeline
- [ ] UI meets accessibility requirements
- [ ] Configuration changes are documented, documentation and designs are updated
- [ ] Passes code peer-review
- [ ] Passes QA of Acceptance Criteria with verification in Dev and Test
- [ ] Ticket is ready to be merged to main branch
- [ ] Can be demoed in Sprint Review
- [ ] Bugs or future work cards are identified and created
- [ ] Reviewed and approved by Product Owner

#### Dependencies

- Blocked by #1897 
- Blocking

shon-button:
@patriciarussellCAS

Reachable at:
https://cas-bciers-frontend-dev.apps.silver.devops.gov.bc.ca/registration/register-an-operation

patriciarussellCAS:
Unable to review and test this workflow. Blocked until #2133 is addressed. 
Moved to current sprint.

patriciarussellCAS:
My review notes: 

- Success page works as intended and matches wireframe
- When clicking on "Report a Change" I get a Not Found notification (see screenshot)
- Is this being worked on in another ticket? Do we have confirmation where this button should take the user or is this still being clarified with BA? 
[...]

PR #2075, opened 2024-08-12
1905 operation registration success

marcellmueller:
#1905 

### Operation Restration Success page

Very simple to test at the moment given there is nothing happening during form submission, simply navigate to the final page of the Operation Registration form and click 'Submit': 

http://localhost:3000/registration/operation/test-id/6?title=Placeholder+Title 

#### Note: the `Report a change` page doesn't exist as far as I know so it's just `tbd` currently.

Screenshot 2024-08-12 at 11 34 13 AM


bciers/apps/registration/app/components/operations/registration/FacilityInformationPage.tsx
```diff
@@ -54,7 +54,7 @@ const FacilityInformationPage = async ({
   return (
     <FacilityInformationForm
       facilityId={facilityId}
-      isCreating={!isOperationSfo && !facilityId}
+      isCreating={!isOperationSfo || !facilityId}
```
marcellmueller, line 57: Unrelated fix I am sneaking in here. I introduced this bug which will cause an error if you pick an SFO operation on page 1, page 2 will have an error on submission if there is a facility tied to that one since it will POST instead of PUT.

bciers/libs/components/src/form/widgets/CheckboxWidget.tsx
```diff
@@ -24,7 +24,7 @@ const CheckboxWidget: React.FC<WidgetProps> = ({
           }
         />
       </div>
-      <label>{label}</label>
+      <label htmlFor={id}>{label}</label>
```
marcellmueller, line 27: This is a fix to allow clicking on the label to check the box. A nice UX improvement 😎

Sepehr-Sobhani: Ooooh I love this! I quickly attempted to do this by adding pointer-event on the whole `div` but that failed and I gave up lol. Thanks for finding the pro solution.

bciers/apps/registration/app/data/jsonSchema/operationRegistration/submission.ts
Sepehr-Sobhani, line 1: Thanks for fixing this!

Sepehr-Sobhani (approved):
Neat 🥇 !

PR #2076, opened 2024-08-12
Chore: activity slug & weight

dleard:
Adds 2 fields to the reporting_activity table: 
**slug** for frontend route naming synergy
**weight** for consistent ordering of activity pages


pbastia (approved):
very nice! 
Nit: The migration file could probably be named `0030_slug_and_weight_on_activity_model` or something a little more meaningful.

Issue #1794, opened 2024-06-11
Copy reg1 components to reg


BCerki:
#### Describe the task

We have separate nx apps for registration parts 1 and 2. If components/tests/etc. work the same in part 2 as part 1, then we were importing the part 1 component into files in part 2 folder. If the component needed to be changed for part 2, we created new files in the part 2 folder. Instead, we'd like to have part 2 import only files from within part 2 (so when we delete part 1 we don't have to edit anything in part 2).

#### Acceptance Criteria
- [x] imports referencing `@/app` are contained within `apps/registration1/*`
- [x] imports referencing `apps/registration1/` are contained within `apps/registration1/*` 

#### Dev Checklist
- [x] copy the components in reg1 to admin or reg 
   - [ ] if there are already new components in reg use those instead
   - [ ] if there are any shareable components create in `bciers/libs`

patrickisaac:
Further discussion with overall Moose dev team

PR #2080, opened 2024-08-13
Decouple reg1 from other nx apps

shon-button:
Addressed [1794](https://github.com/bcgov/cas-registration/issues/1794)


BCerki (approved):
Nice work! Thanks for doing this cleanup :cartwheeling:

PR #2081, opened 2024-08-13
301 combustion for line tracing activity page

BallardRobinett:
Card: https://github.com/bcgov/cas-reporting/issues/301

Adding a new activities page 'general stationary combustion solely for the purpose of line tracing'. 

Changes include:
- new frontend page at `/reporting/activities/gsc_solely_for_line_tracing`
- new json schema
- registration and reporting migrations

To test:
- reset db to run new migrations with `make reset_db` from `/bc_obps`
- run frontend with `yarn dev-all` from `/bciers`
- go to `localhost:3000` and login with bceid. I use the bc-cas-dev profile
- enter `localhost:3000/reporting/activities/gsc_solely_for_line_tracing` into the address bar. The activities page should render.
- checking either box should make the corresponding form appear underneath


dleard (changes requested):
Nice Work. A couple minor changes, but it looks really good & works when tested!

dleard (approved):
Awesome! :100:  Just fix the failing python test & this is good to go!

Issue #1952, opened 2024-07-15
Reg\Register an Operation\Page "Operation Information\Opted-in Operation" - External user can register the operation


andrea-williams:
#### Dependencies

- Blocked by #1899 
- Blocked by #2133 
- Blocking https://github.com/bcgov/cas-registration/issues/2094

#### Description:

Brief description of the work involved, including any user research context or key decisions.

In Registration workflow mode, [wireframe](https://www.figma.com/design/LsDmLDCdnJqI8UahyqvOD8/BCIERS-Designs?node-id=13653-2443&t=HA6SzABkE1DQWPQg-4)

#### Acceptance Criteria:

Given an authenticated external user has selected "Opted-in Operation" as the purpose of registration in the Registration workflow
Then they get an extra step at step 3 (after facility information) called Opt-In Application
And they see [this form
](https://www.figma.com/design/LsDmLDCdnJqI8UahyqvOD8/BCIERS-Designs?node-id=13653-2443&t=eV3D2Yqyi2or0X8R-4)

Given an authenticated external user has selected "Opted-in Operation" as the purpose of registration in the Registration workflow
When they get to the "Operation Information" step of the Registration form
Then they see [this](https://www.figma.com/design/LsDmLDCdnJqI8UahyqvOD8/BCIERS-Designs?node-id=13340-2186&m=dev&t=NI9Z4bjk0RMG6DoE-1)
And when they click the "Save and Continue" button, the data they input is saved to our database

#### Development Checklist:

Create new data model `Opted In Operation`:
- [x] FK reference to operation ID
- [x] json that will have Yes/No answers to the following questions

1. Does this operation have emissions that are attributable for the purposes of section 3 of the Act?
2. Is this operation an electricity import operation?
3. Is this operation a part of a single facility operation or a linear facilities operation?
4. Does this operation have emissions that are attributable for the purposes of section 6 of the Act?
5. Is this operation's primary economic activity classified by the following NAICS codes: 11, 22, or 562?
6. Does this operation produce a regulated product listed in Table 1 of Schedule A.1 of the GGERR?
7. Is this operation capable of fulfilling the obligations of a reporting operation and a regulated operation under the Act and the regulations?
8. Will the operator notify the Director as soon as possible if this operation ceases to meet any of the criteria for the designation of the operation as a reporting operation and a regulated operation?

- [ ] Meets the DOD

**Definition of Ready** (Note: If any of these points are not applicable, mark N/A)

- [ ] User story is included
[...]

PR #2096, opened 2024-08-15
1952 Operation Information\Opted-in Operation

Sepehr-Sobhani:
[ISSUE 1952](https://github.com/bcgov/cas-registration/issues/1952)

### Opted-In Operation page

1. Go to the 'Register an Operation' form and choose 'opted-in operation' as the purpose of registration.
2. Enter some dummy data on the facility info page and proceed to the next page.
3. On the Opted-In Operation Page, the 'Save and Continue' button should remain disabled until the user answers all the questions.


bciers/apps/registration/app/data/jsonSchema/operationRegistration/optedInOperation.ts
```diff
@@ -0,0 +1,122 @@
...
+  type: "object",
+  required: [
+    "has_emissions_for_section_3",
+    "is_electricity_import_operation",
+    "is_entire_operation_opted_in_for_designation",
+    "has_emissions_for_section_6",
+    "is_primary_economic_activity_classified_by_naics_11_22_562",
+    "produces_regulated_product_in_ggerr",
+    "is_capable_of_fulfilling_reporting_and_regulated_obligations",
+    "will_notify_director_if_criteria_not_met",
```
Sepehr-Sobhani, line 26: @BCerki You always have a good sense of how to name a variable to be meaningful 😃 . Do you think these variable names reflect the questions [here](https://www.figma.com/design/LsDmLDCdnJqI8UahyqvOD8/BCIERS-Designs?node-id=13653-2641&t=jHA3aEIuF9lJ3mJm-4)? Thank you.

BCerki: I think the way you've got it looks fine. A couple related thoughts:
- `is_electricity_import_operation`--we're already going to know this based on the purpose they selected in step 1, so maybe a question for Zoey about whether we need to include this in the form again
- these all look like acknowledgement/eligibility questions, so if that's indeed the case, we could name them like it's done in `submissionSchema` with a prefix, e.g. meets_section_3_requirements, meets_naics_requirements, meets_entire_operation_requirements, meets_fulfilling_requirments, etc. (or is_eligible_under_section_3, whatever's most accurate)

Sepehr-Sobhani: Thanks for your input, Brianna!

- Good point, I wonder if this serves a different purpose because during registration we only allow users to select 1 option. However, It is good make sure we are not duplicating data.
- Great suggestion.

bciers/apps/registration/app/components/operations/registration/OptedInOperationPage.tsx
```diff
@@ -0,0 +1,39 @@
...
+    formData = await getOptedInOperationDetail(operation);
+  }
+
+  if (formData && "error" in formData) {
+    return (
+      <div>
+        <p>Sorry, Something went wrong. Please try again later!</p>
+      </div>
+    );
+  }
```
marcellmueller, line 27: We already have built in error handling via our `error.tsx` files. I suggest throwing an error here instead ie:

```suggestion

  if (formData && "error" in formData) {
    throw new Error(
      "We couldn't find your operation information. Please ensure you have been approved for access to this operation.",
    );
  }
```

Screenshot 2024-08-22 at 1 09 16 PM

bciers/apps/registration/app/data/jsonSchema/operationRegistration/optedInOperationText.tsx
```diff
@@ -0,0 +1,117 @@
+const optedInOperationPreface = (
+  <div>
+    Complete the following fields to apply to be an Opt-in Operation in the B.C.
+    OBPS. Ensure that you have read through{" "}
+    <a
+      href="https://www2.gov.bc.ca/gov/content/environment/climate-change/industry/bc-output-based-pricing-system"
```
marcellmueller, line 6: Can we put all URLs in the constants file located in `bciers/libs/utils/urls.ts`?

bc_obps/registration/api/v2/_operations/_operation_id/_registration/opted_in_operation.py
```diff
@@ -0,0 +1,44 @@
...
+from registration.api.utils.current_user_utils import get_current_user_guid
+from service.error_service.custom_codes_4xx import custom_codes_4xx
+from registration.decorators import handle_http_errors
+from registration.api.router import router
+from registration.schema.generic import Message
+
+
+@router.get(
+    "/v2/operations/{operation_id}/registration/opted-in-operation-detail",
+    response={200: OperationRegistrationOptedInOperationDetailOut, custom_codes_4xx: Message},
```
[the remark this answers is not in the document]
Sepehr-Sobhani: 🤣

bciers/apps/registration/app/data/jsonSchema/operationRegistration/optedInOperationText.tsx
Sepehr-Sobhani: Good call, I wasn't aware of this file.

PR #2099, opened 2024-08-19
chore: fix indentation in deployment yaml files

dleard:
some bad indents made it into a couple deployment yaml files, preventing helm from deploying properly. This should fix it, I tested the deploy with a --dry-run flag.


marcellmueller (approved):
Thanks so much for this 🙇

Issue #2065, opened 2024-08-09
Admin\Operations - External User sees button "View Facility" or "View Facilities"


zoeyli-46:
Context: SFOs should have a slightly different experience than LFOs 

As an external user, given that the operation is an SFO, then:
- [x] On the operations grid, the button should say "View Facility" instead of "View Facilities"
- [x] Given I have clicked View Facility, the facility page should skip the facilities grid page and directly go to the facility form page

zoeyli-46:
cc @marcellmueller I'm not sure if you would be assigned to this ticket but let me know if these details are sufficient

marcellmueller:
Looks good to me thank you!

shon-button:
@patriciarussellCAS 

Reachable at:
https://cas-bciers-frontend-dev.apps.silver.devops.gov.bc.ca/administration/operations?operations_title=Operation+2

shon-button:
@patriciarussellCAS

Reachable at:
https://cas-bciers-frontend-dev.apps.silver.devops.gov.bc.ca/administration/operations

patriciarussellCAS:
Reviewed in Dev.

PR #2103, opened 2024-08-19
2065 operations sfo changes

marcellmueller:
#2065 

As a follow up to this I've created another issue to ensure that an SFO can only create 1 Facility:
https://github.com/bcgov/cas-registration/issues/2105

To test this go to the Operations DataGrid `Facilities` column:
- LFO Facility stays the same, shows `View Facilities` and links to the Facilities DataGrid
- SFO Facility with existing facility id shows `View Facility` and links to the `Facility` form in read only mode
- SFO Facility without an existing facility id shows `Edit Details` and links to the `Facility` form in edit mode

Screenshot 2024-08-19 at 4 09 20 PM


bc_obps/registration/schema/v2/operation.py
```diff
@@ -39,6 +43,14 @@ class Config:
         ]
         from_attributes = True
 
+    @staticmethod
+    def resolve_sfo_facility_id(obj: Operation) -> Optional[UUID]:
+        # Resolve a single facility id for SFO operations
+        facilities = FacilityDataAccessService.get_current_facilities_by_operation(obj)
+        if obj.type == OperationTypes.SFO.value and len(facilities) > 0:
+            return facilities[0].id
+        return None
```
marcellmueller, line 52: This should be safer once we add some safeguards to ensure an SFO can't create multiple Facilities:

https://github.com/bcgov/cas-registration/issues/2105

bc_obps/registration/enums/enums.py
```diff
@@ -23,3 +23,8 @@ class AccessRequestTypes(Enum):
     ADMIN = "Admin"
     OPERATOR_WITH_ADMIN = "Operator With Admin"
     NEW_OPERATOR_AND_ADMIN = "New Operator And Admin"
+
+
+class OperationTypes(Enum):
+    LFO = "Linear Facility Operation"
+    SFO = "Single Facility Operation"
```
Sepehr-Sobhani, line 30: This makes Dylan unhappy 😆

bc_obps/registration/schema/v2/operation.py
```diff
@@ -39,6 +43,14 @@ class Config:
         ]
         from_attributes = True
 
+    @staticmethod
+    def resolve_sfo_facility_id(obj: Operation) -> Optional[UUID]:
+        # Resolve a single facility id for SFO operations
+        facilities = FacilityDataAccessService.get_current_facilities_by_operation(obj)
+        if obj.type == OperationTypes.SFO.value and len(facilities) > 0:
+            return facilities[0].id
```
Sepehr-Sobhani, line 51: nit: I suggest using `.count()` instead of `len()` and `.first()` instead of `[0]` when we are dealing with Django QuerySets and we don't want to iterate through it.
[Source](https://docs.djangoproject.com/en/5.1/ref/models/querysets/#:~:text=A%20count()%20call%20performs%20a%20SELECT%20COUNT(*)%20behind%20the%20scenes%2C%20so%20you%20should%20always%20use%20count()%20rather%20than%20loading%20all%20of%20the%20record%20into%20Python%20objects%20and%20calling%20len()%20on%20the%20result%20(unless%20you%20need%20to%20load%20the%20objects%20into%20memory%20anyway%2C%20in%20which%20case%20len()%20will%20be%20faster).)

Sepehr-Sobhani (approved):
Bravo 👏 
Just an observation that is not quite related to this PR so feel free to address that or leave it:
- When viewing the read-only facility form associated with an SFO operation, clicking the Cancel button redirects the user to the Facilities data grid. Additionally, there is no breadcrumb navigation on this page.
Screenshot 2024-08-21 at 1 39 33 PM

bc_obps/registration/schema/v2/operation.py
Sepehr-Sobhani: We are pretty sure that we don't pass the `if` statement if `facilities` variable is not truth but seems like mypy is not aware of that :D here is the workaround:
```python
    def resolve_sfo_facility_id(obj: Operation) -> Optional[UUID]:
        # Resolve a single facility id for SFO operations
        facilities = FacilityDataAccessService.get_current_facilities_by_operation(obj)
        if obj.type == OperationTypes.SFO.value and facilities:
            first_facility = facilities.first()
            if first_facility:
                return first_facility.pk
        return None
```

marcellmueller: Thank you so much, really appreciate it!

Also, I've fixed the cancel button for SFO Facilities and fixed the breadcrumbs 🙇

Sepehr-Sobhani: Thanks for tackling all the comments! 🙏

PR #2120, opened 2024-08-22
feat:302-Activity Page-General Stationary Combustion (other than non-compression and non-processing)

ayeshmcg:
Card: https://github.com/bcgov/cas-reporting/issues/302

Adding a new activities page 'General Stationary Combustion, other than non-compression and non-processing'.

Changes include:

new frontend page at gsc_other_than_non_compression
new json schema
registration and reporting migrations
To test:

reset db to run new migrations with make reset_db from /bc_obps
run frontend with yarn dev-all from /bciers
go to localhost:3000 and login with bceid.
enter http://localhost:3000/reporting/reports/1/facilities/1/activities/gsc_other_than_non_compression into the address bar. The activities page should render.
checking either box should make the corresponding form appear underneath


dleard (approved):
Nice work! I tested it & all the methodologies & dependent fields render as they should :)

PR #2125, opened 2024-08-23
chore: 315 - load correct activity schema for gsc line tracing activity

dleard:
One-line fix. Looks like a leftover copy-paste error in the migration


ayeshmcg (approved):
Approved as only one change done

PR #2128, opened 2024-08-26
Reg/Operation form

BCerki:
card: https://github.com/orgs/bcgov/projects/122/views/2?filterQuery=&pane=issue&itemId=64944876

Notes:
- renaming of the ownership models was completed in a different PR
- I didn't test upsert_multiple_operators because I think the other service tests cover it, and we didn't test upsert_partner/parent. Happy to add a test if the reviewer has concerns
- Didn't test custom_create_or_update--Sepehr has plans to tweak it (happy to help with testing it later)
- Sonarcloud security fails are because we used Math.random(). It's to create a key, not for any cryptography, so I think it's ok to ignore
- Sonarcloud duplication on tests is fine to ignore. I'm not sure how I would avoid it for the endpoint files--it's things like the auth decorator, etc., and we need that on all endpoints. Suggestions welcome!

Updates after review:
- Marcel fixed the bug preventing making multiselect fields required, so updated the schemas to required where needed
- made a separate v2 ninja schema for getting an operation--the v2 form is so different from the v1 form now that it was cleaner to have separate ones
- added some mock data to make sure reporting activities were coming in properly
- ^I'd like to do some mock multiple operator data too, but I'll do in a separate PR because this one is a monster. I also noticed some funky rjsf validation in the multiple operator section, so I'll address that in that separate PR too


bc_obps/registration/models/operation.py
```diff
@@ -52,9 +52,25 @@ class Statuses(models.TextChoices):
         NaicsCode,
         on_delete=models.DO_NOTHING,
         null=True,
```
BCerki, line 54: Is the primary code supposed to be nullable?

bc_obps/service/data_access_service/multiple_operator_service.py
```diff
@@ -0,0 +1,17 @@
...
+class MultipleOperatorService:
+    @classmethod
+    def create_or_update(
+        cls, multiple_operator_id: int | None, operation: Operation, user_guid: UUID, data: DictStrAny
+    ) -> MultipleOperator:
+        mo_operator_instance, _ = MultipleOperator.objects.update_or_create(
+            pk=multiple_operator_id,
+            defaults={**data, 'operation': operation},
+        )
+        mo_operator_instance.set_create_or_update(user_guid)
```
BCerki, line 16: do i need to save?

bc_obps/service/tests/test_operation_service_v2.py
```diff
@@ -129,35 +156,20 @@ def test_update_status():
...
-        operation.refresh_from_db()
-        assert approved_user_operator.operator.contacts.count() == 5  # existing contacts should be unaffected
-        assert operation.contacts.count() == 5
-        assert operation.updated_at is not None
-        assert operation.updated_by == approved_user_operator.user
+        operation = baker.make_recipe('utils.operation', operator=random_operator)
+        with pytest.raises(Exception, match=UNAUTHORIZED_MESSAGE):
+            OperationServiceV2.update_status(
+                approved_user_operator.user.user_guid, operation.id, Operation.Statuses.REGISTERED
+            )
```
BCerki, line 172: brianna check this wasn't weird rebase

bc_obps/service/user_operator_service.py
```diff
@@ -157,7 +157,7 @@ def list_user_operators(
                 }
             )
         return UserOperatorPaginatedOut(
-            data=user_operator_list,  # type: ignore[arg-type] # we know this is a list of dictionaries with the correct keys
+            data=user_operator_list,
```
BCerki, line 160: b probably put this back, loocal and ci are different?

bc_obps/registration/models/operation.py
```diff
@@ -47,6 +47,7 @@ class Statuses(models.TextChoices):
         db_comment="The operator who owns the operation",
         related_name="operations",
     )
+    # brianna do we need this? can't we find out by looking at the foreign keys in the multiple operator table?
     operation_has_multiple_operators = models.BooleanField(
```
BCerki, line 51: I think we should be getting this from the foreign keys in the multiple operator table so we don't have to set it separately (and potentially introduced conflicting sources of truth). The multiple op models were never used in reg 1 prod, but this table is, so is it changeable at this point?

bc_obps/registration/models/operation.py
```diff
@@ -94,6 +111,7 @@ class Statuses(models.TextChoices):
         blank=True,
         null=True,
     )
+    # brianna this should be a foreign key instead?
     documents = models.ManyToManyField(
```
BCerki, line 115: This should be a foreign key, not m2m

bc_obps/service/document_service.py
```diff
@@ -1,11 +1,34 @@
...
+        cls, user_guid: UUID, operation_id: UUID, file_data: ContentFile, document_type: str
+    ) -> Document:
+        """This function receives a document and operation id. Operations only have one of each type of document, so this function uses the type to check if an existing document needs to be replaced, or if no document exists and one must be created. This function does NOT set any m2m relationships."""
+        existing_document = DocumentDataAccessService.get_operation_document_by_type(operation_id, document_type)
+        # if there is an existing  document, check if the new one is different
+        if existing_document:
+            # We need to check if the file has changed, if it has, we need to delete the old one and create a new one
+            if not files_have_same_hash(file_data, existing_document.file):
+                # brianna this could have been an update instead of a delete/create?
+                existing_document.delete()
```
BCerki, line 29: This could have been an update instead of a delete/recreate--possible tech debt

bc_obps/service/operation_service_v2.py
```diff
@@ -159,3 +167,123 @@ def register_operation_operation_representative(
...
+        cls.upsert_multiple_operators(operation, multiple_operators_data, user_guid)
+
+        return operation
+
+    @classmethod
+    @transaction.atomic()
+    def register_operation_information(
+        cls, user_guid: UUID, operation_id: UUID | None, payload: OperationInformationIn
+    ) -> Operation:
+        if operation_id:
```
BCerki, line 230: Possible tech debt card--a service (or decorator or something) to check external users' permissions. The endpoint decorator is only role-based, and we often need to ensure a user can only see their own operators/operations/facilities, etc.

bc_obps/service/tests/test_document_service.py
```diff
@@ -0,0 +1,66 @@
...
+            approved_user_operator.user_id, data_url_to_file(MOCK_DATA_URL), 'boundary_map'
+        )
+        operation.documents.set([existing_document.id])
+
+        updated_file = data_url_to_file(MOCK_DATA_URL)
+        document = DocumentService.create_or_replace_operation_document(
+            approved_user_operator.user_id, operation.id, updated_file, 'boundary_map'
+        )
+
+        # brianna how to check that it wasn't deleted and then recreated?
```
BCerki, line 42: ?

bc_obps/service/operation_service_v2.py
```diff
@@ -159,3 +167,126 @@ def register_operation_operation_representative(
...
+
+    #     registration_payload = RegistrationPurposeIn(
+    #         registration_purpose=payload.registration_purpose, regulated_products=payload.regulated_products
+    #     )
+    #     cls.set_registration_purpose(user_guid, operation_id, registration_payload)
+
+    @classmethod
+    @transaction.atomic()
+    def create_or_update_operation_v2(
+        # brianna better to do as a default arg?
```
BCerki, line 183: thoughts?

bc_obps/registration/models/facility.py
```diff
@@ -71,8 +71,8 @@ class Meta:
         verbose_name_plural = "Facilities"
 
     @property
-    def current_owner(self) -> Operation:
+    def current_designed_operation(self) -> Operation:
```
andrea-williams, line 74:
```suggestion
    def current_designated_operation(self) -> Operation:
```

bc_obps/registration/schema/v2/multiple_operator.py
```diff
@@ -0,0 +1,82 @@
...
+    mo_trade_name: str = Field(..., alias="trade_name")
+    mo_cra_business_number: int = Field(..., alias="cra_business_number")
+    mo_bc_corporate_registry_number: str = Field(
+        ..., alias="bc_corporate_registry_number", pattern=BC_CORPORATE_REGISTRY_REGEX
+    )
+    mo_business_structure: str = Field(..., alias="business_structure")
+    mo_physical_street_address: str = Field(..., alias="physical_address.street_address")
+    mo_physical_municipality: str = Field(..., alias="physical_address.municipality")
+    mo_physical_province: str = Field(..., alias="physical_address.province")
+    mo_physical_postal_code: str = Field(..., alias="physical_address.postal_code")
```
andrea-williams, line 61: we don't need physical address for any operator anymore, just mailing address. All the physical address stuff can be removed

bc_obps/service/data_access_service/operation_service.py
```diff
@@ -64,14 +64,46 @@ def create_operation(
...
+        user_guid: UUID,
+        operation_data: DictStrAny,
+        activities: Union[QuerySet[Activity], Iterable[Activity]],
+    ) -> Operation:
+        operation = Operation.objects.create(
+            **operation_data,
+            created_by_id=user_guid,
+        )
+        operation.activities.set(activities)
+
```
andrea-williams, line 87: I thiiiink we need to call `operation.set_create_or_update()` here in L87, but want to double-check with @Sepehr-Sobhani because frankly I've forgotten how to write code.

bc_obps/service/document_service.py
andrea-williams: sorry, I don't understand your comment...?

bc_obps/service/facility_service.py
```diff
@@ -166,7 +167,7 @@ def list_facilities(
     def get_if_authorized(cls, user_guid: UUID, facility_id: UUID) -> Facility:
         """Retrieve a facility if the user is authorized to access it."""
         facility: Facility = FacilityDataAccessService.get_by_id(facility_id)
-        owner: Operation = facility.current_owner
+        owner: Operation = facility.current_designed_operation
```
andrea-williams, line 170: please rename to `current_designated_operation`

bc_obps/service/operation_service_v2.py
```diff
@@ -159,3 +167,126 @@ def register_operation_operation_representative(
...
         return operation
+
+    # @classmethod
+    # @transaction.atomic()
+    # def create_registration_purpose(cls, user_guid: UUID, operation_id: UUID, payload: RegistrationPurposeIn) -> Dict[str, Union[str, UUID]]:
+
+    #     registration_payload = RegistrationPurposeIn(
+    #         registration_purpose=payload.registration_purpose, regulated_products=payload.regulated_products
+    #     )
+    #     cls.set_registration_purpose(user_guid, operation_id, registration_payload)
```
andrea-williams, line 178: do we still need this commented-out code?

bc_obps/service/operation_service_v2.py
andrea-williams: I don't understand the question...

bc_obps/service/tests/test_document_service.py
andrea-williams: should be able to query `existing_document` and `document` to ensure they have the same `created_at` value. If they don't, then yeah it prob was deleted and recreated.

andrea-williams (commented):
Saving my progress so far in this review. Still have 11 files to look at, and haven't run the code locally yet.

bc_obps/service/data_access_service/operation_service.py
Sepehr-Sobhani: Since we are creating an operation here and explicitly passing `created_by_id`, we no longer need to call `set_create_or_update()`. However, if `operation.activities.set(activities)` were outside of this service, we should have called it because we are actually performing an update on the operation instance.

bciers/apps/registration/app/components/operations/registration/OperationInformationForm.tsx
```diff
@@ -1,46 +1,117 @@
...
 
 const OperationInformationForm = ({
-  formData,
+  rawFormData,
   schema,
   step,
   steps,
 }: OperationInformationFormProps) => {
+  const [selectedOperation, setSelectedOperation] = useState("");
+  const [, setError] = useState(undefined);
```
andrea-williams, line 31: we need to be storing the error in a variable, no?

andrea-williams: i.e., `const [error, setError] = useState(undefined);`

bc_obps/service/tests/test_operation_service_v2.py
```diff
@@ -295,3 +310,224 @@ def test_create_or_replace_statutory_declaration():
...
+            tertiary_naics_code_id=3,
+            activities=[1],
+            process_flow_diagram=MOCK_DATA_URL,
+            boundary_map=MOCK_DATA_URL,
+            equipment_list=MOCK_DATA_URL,
+        )
+        operation = OperationServiceV2.create_or_update_operation_v2(
+            approved_user_operator.user.user_guid, None, payload
+        )
+        # brianna why do i have to refresh? just a test thing or a problem?
```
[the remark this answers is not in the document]
andrea-williams: I presume you needed to do the refresh in order to get the expected values for the audit fields - this is consistent with what we see in other test files. Need to do the db refresh because the audit fields are only populated when the record is written to the db. I'm pretty sure the `operation` record that is returned from the create_or_update function only contains the data fields that we explicitly set (not the magic ones).

bc_obps/service/tests/test_operation_service_v2.py
```diff
@@ -295,3 +310,224 @@ def test_create_or_replace_statutory_declaration():
...
+        operation = OperationServiceV2.create_or_update_operation_v2(
+            approved_user_operator.user.user_guid, None, payload
+        )
+        operation.refresh_from_db()
+        assert Operation.objects.count() == 1
+        assert operation.created_by == approved_user_operator.user
+        assert operation.created_at is not None
+        assert MultipleOperator.objects.count() == 2
+        assert MultipleOperator.objects.all().first().bc_corporate_registry_number == 'ghj1234567'
+        assert MultipleOperator.objects.all().last().bc_corporate_registry_number is None
```
andrea-williams, line 446: I don't think the `all()` is needed - should just be able to do
`MultipleOperator.objects.first()` and `MultipleOperator.objects.last()`

andrea-williams (commented):
Finished reading through all the file changes - now to pull the code and try to break things.

Apologies for how long it's taking me to get through this review! Lots of distractions today.

andrea-williams (changes requested):
When I select "OBPS Regulated" as my registration purpose, I should be required to submit a regulated product. I don't even get the field asking me for my regulated product/s.

I also wasn't able to submit my newly created operation. The error in the FE wasn't very helpful, but might be related to the lack of regulated products.

bc_obps/common/models/base_model.py
```diff
@@ -24,3 +26,26 @@ def save(self, *args, **kwargs) -> None:
...
+
+        # If no identifier is provided, create a new instance
+        if identifier is None:
+
+            instance, _ = self.objects.create(**kwargs), True
+        else:
+            # If identifier is provided, update or create the instance
+            instance, _ = self.objects.update_or_create(pk=identifier, defaults=kwargs)
+        # If the model has audit columns, set them
+        from registration.models.time_stamped_model import TimeStampedModel
```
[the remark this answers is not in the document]
BCerki: It's here because I was getting a circular import error. If there's a better way to avoid that, happy to do that instead

bc_obps/registration/schema/v2/multiple_operator.py
BCerki: We don't need at out schema at all at this point because we don't show any of the steps read-only, so I'll remove the whole thing.

bc_obps/service/document_service.py
BCerki: A document update is being handled by deleting the existing document and then creating a new one. For other field updates, we  use the .update() method (or our custom one) and that sets the updated_at etc. columns. So this code is working fine, but it's not a true "update."

bc_obps/service/facility_service.py
BCerki: :scream:

bc_obps/service/operation_service_v2.py
BCerki: These are the args:
```
cls,
        user_guid: UUID,
        operation_id: UUID | None,
        payload: OperationInformationIn,
```
I've done UUID | None instead of making the default for operation_id None. I did this because optional args have to go at the end and we usually give the ids before the payload, but non-python-y me doesn't know if that's a good decision

bc_obps/service/tests/test_document_service.py
BCerki: perfect, thanks!

bciers/apps/registration/app/components/operations/registration/OperationInformationForm.tsx
BCerki: We're not using the `error` variable yet (part of https://github.com/orgs/bcgov/projects/122/views/2?filterQuery=cleanup&pane=issue&itemId=75086601), so ts throws an error if I include and it's not used anywhere. I could leave it in an put in an ignore if you prefer

bciers/apps/registration/app/data/jsonSchema/operationInformation/registrationPurpose.ts
```diff
@@ -0,0 +1,143 @@
...
+        type: "object",
+        readOnly: true,
+      },
+    },
+
+    dependencies: {
+      registration_purpose: {
+        oneOf: registrationPurposes.map((purpose: string) => {
+          return {
+            // required: ["regulated_products"],
```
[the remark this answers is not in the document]
BCerki: I'll uncomment it; Marcel just fixed the bug with required fields and multiselect widgets

bciers/apps/registration/app/data/jsonSchema/operationInformation/registrationPurpose.ts
```diff
@@ -0,0 +1,143 @@
...
+                const: purpose,
+              },
+              ...(purpose !==
+                RegistrationPurposes.ELECTRICITY_IMPORT_OPERATION &&
+                purpose !==
+                  RegistrationPurposes.POTENTIAL_REPORTING_OPERATION && {
+                  regulated_products: {
+                    title: "Regulated Product Name(s)",
+                    type: "array",
+                    // minItems: 1,
```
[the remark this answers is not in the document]
BCerki: Same note, bug is fixed so this can be uncommented

bc_obps/service/tests/test_operation_service_v2.py
BCerki: :exploding_head:

bc_obps/service/operation_service_v2.py
andrea-williams: ah I see. IMO there's nothing wrong with having `operation_id` as the last arg so it can be optional

bciers/apps/registration/app/components/operations/registration/OperationInformationForm.tsx
andrea-williams: ah gotcha, it's fine leaving as-is for now

bc_obps/common/models/base_model.py
```diff
@@ -24,3 +26,26 @@ def save(self, *args, **kwargs) -> None:
         else:
             self.full_clean(exclude=AUDIT_FIELDS)  # validate the model before saving
         super().save(*args, **kwargs)
+
+    @typing.no_type_check
+    def custom_update_or_create(self, user_guid: UUID, **kwargs: dict):
+        # Extract pk or id from kwargs if available
+        pk = kwargs.pop('pk', None)
+        kwargs_id = kwargs.pop('id', None)
```
andrea-williams, line 34: @BCerki  was there any particular reason why you chose to use `pop` here instead of `kwargs.get()`? pop() removed the key-value from the dictionary, which seems unnecessary

bc_obps/common/models/base_model.py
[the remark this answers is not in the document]
andrea-williams: ahh I see, I didn't notice the circular import issue.

What I'd recommend instead is this:
- create a new `utils.py` file inside `bc_obps/common/models`
- in utils.py, create a function along the lines of

```
def set_audit_fields_if_needed(instance, user_guid: UUID):
    if isinstance(instance, TimeStampedModel):
        instance.set_create_or_update(user_guid)
```

- from `base_model.py` we can then replace lines 46-50 with a simple `set_audit_fields_if_needed(instance, user_guid)`

bc_obps/common/models/base_model.py
BCerki: This is @Sepehr-Sobhani 's code, so I'm not sure, sorry! (Brianna isn't actually a python expert :see_no_evil:)

andrea-williams: all good, i re-read the code and now understand why. It's because a few lines lower where we're updating or creating a record with the data stored in `kwargs`, we don't want to overwrite the existing pk or id if one already exists.

P.S. you're gonna have to try harder than that to shake your title of Python Expert

andrea-williams:
After running the branch locally, I was able to create a brand new operation and register it. However, when I try to "Continue Registration" for an existing operation (when signed in as `bc-cas-dev` with the standard mock data fixtures) I first get a view-only version of the operation info page, but then when I click the Edit button, I get error messages for all the missing data fields but the form isn't actually in Edit mode so I can't modify any data.

BCerki:
> After running the branch locally, I was able to create a brand new operation and register it. However, when I try to "Continue Registration" for an existing operation (when signed in as `bc-cas-dev` with the standard mock data fixtures) I first get a view-only version of the operation info page, but then when I click the Edit button, I get error messages for all the missing data fields but the form isn't actually in Edit mode so I can't modify any data.

I don't that's part of this ticket (and I'm not sure if we have a ticket to handle it). Editing an operation is this ticket:  https://github.com/orgs/bcgov/projects/122/views/2?filterQuery=edit&pane=issue&itemId=68731958, but I believe that's when registration has already been completed and we're just editing in admin. My ticket is about editing the data after you select an existing operation, e.g. https://www.loom.com/share/c139c97a095e45f29c55d820d878c496?sid=95e4dee1-2126-4272-a139-f770226d9797

andrea-williams:
Have confirmed (by running the branch locally) that I'm able to create a new operation and register it, and also able to register an existing operation by clicking the "Add and Register an Operation" button and then selecting the existing operation from the dropdown. I suspect the "Continue Registration" action should redirect to the multistep form rather than the Operation details page (what my previous complaint was about), but that's beyond the scope of this ticket. 

Just flagging that mypy fails because of the `UUID | None` arg as discussed somewhere in this review. Otherwise I'm pretty sure this PR is ready to go! Great work Bri! 🚀

bc_obps/registration/api/v2/_operations/current.py
```diff
@@ -22,9 +22,9 @@
     response={200: List[OperationCurrentOut], codes_4xx: Message},
     tags=["V2"],
     auth=authorize('approved_industry_user'),
-    description="""Gets the list of operations associated with the current user's operation.
-    The endpoint ensures that only authorized industry users can get operations belonging to their operator. Unauthorized access attempts raise an error.""",
+    description="""Gets the list of operations associated with the current user's operation that are not yet registered.
```
andrea-williams, line 25:
```suggestion
    description="""Gets the list of operations associated with the current user's operator that are not yet registered.
```

andrea-williams: (I know this wasn't your boo-boo, I just noticed it now and figured I'd take the opportunity to fix it)

bc_obps/service/data_access_service/operation_service.py
```diff
@@ -64,14 +64,46 @@ def create_operation(
...
+    #     activities: Union[QuerySet[Activity], Iterable[Activity]],operation_id: int = None,
+    # ) -> Operation:
+    #     operation_instance, _ = Operation.objects.update_or_create(
+    #         id=operation_id,
+    #         defaults={**operation_data},
+    #     )
+
+    #     operation_instance.activities.set(activities)
+    #     operation_instance.set_create_or_update(user_guid)
+    #     return operation_instance
```
andrea-williams, line 105: is this still needed?

andrea-williams (approved):
Congratulations! You have joined the ranks of Shon as a recognized Dragon Slayer. ⚔️

Seriously, fantastic work Brianna - this was a HUGE piece of work!

Issue #1993, opened 2024-07-24
Admin\Operations - Grid has status column


zoeyli-46:
For External Users, change the "registration status" column to "status"
The status options are
Registered
Continue Registration
Transferred
Closed
Temporarily Shutdown

If an operation that was previously closed or temporarily shutdown, and has been reported as Restarted, the status reverts back to Registered. 
[Figma](https://www.figma.com/design/LsDmLDCdnJqI8UahyqvOD8/BCIERS-Designs?node-id=7763-4636&t=Ci7zKMcvo5ipOiRH-4)

For Internal Users, add a "status" column
The status options are the same as the external user, minus the "continue registration"
Registered
Transferred
Closed
Temporarily Shutdown
~~TBD: attn @zoeyli-46 would there be a need for internal users to view incomplete registration data? i.e., should the "Continue Registration" operations be visible to internal users at all?~~

Internal users are able to view incomplete registration data, and are able to see operations which have a "continue registration" status

[Figma](https://www.figma.com/design/LsDmLDCdnJqI8UahyqvOD8/BCIERS-Designs?node-id=7861-688&t=Ci7zKMcvo5ipOiRH-4)

### Development Details

- we can re-use the `status` column on `Operation` that we had in Registration Part 1, but with modifications. 
- we will continue using the "DRAFT" and "NOT STARTED" status options that we had in Reg Part 1. "NOT STARTED" status applies to operations imported from SWRS where no external user has reviewed the imported data. "DRAFT" status applies to either (1) an operation (either manually created by a user or existing from Part 1) that is part-way through the registration process
- [ ] add options for `Operation.Status` column: "REGISTERED", "TRANSFERRED", "CLOSED", "TEMPORARILY SHUTDOWN"
- in Part 2 we will no longer have a need for the "PENDING", "APPROVED", "DECLINED", or "CHANGES_REQUESTED" statuses that we have in Part 1, but we don't want to delete these statuses at this time because they are in use in production data

#### Development Checklist:

- [x] In `bciers/apps/administration/app/components/datagrid/models/operationColumns.ts` reflect `Status` column similar to `bciers/apps/registration1/app/components/datagrid/models/operationColumns.ts`
- [x] In `bc_obps/registration/models/operation.py` add options for `Operation.Status` column: "REGISTERED", "TRANSFERRED", "CLOSED", "TEMPORARILY SHUTDOWN"
[...]

PR #2144, opened 2024-08-30
1993 admin operations grid has status column

Sepehr-Sobhani:
[ISSUE 1993](https://github.com/bcgov/cas-registration/issues/1993)

**TO TEST:**

Log in as an external and internal user, then navigate to the Operations data grid.


bciers/libs/components/src/datagrid/cells/OperationsActionCell.tsx
```diff
@@ -2,18 +2,18 @@ import Link from "next/link";
...
-      case OperationStatus.NOT_STARTED:
-        actionText = "Start Registration";
-        break;
-      case OperationStatus.DRAFT:
-        actionText = "Continue";
-        break;
-      default:
-        actionText = "View Operation";
+    let actionText = "View Operation";
+    if (!isInternalUser) {
```
marcellmueller, line 8: Nit: I think it's safer to render things based on who the user is rather than who the user is not when reasonable to do so. We've had some visual bugs previously when rendering based on who the user isn't, because it isn't specific and can include edge cases.  
  
Though I think this looks safe and not saying to necessarily change it, just something to consider in the future.

marcellmueller (approved):
Excellent work and nice refactor of the Operations api :1st

PR #2145, opened 2024-08-30
2140 mitigate owasp zap scan rate limit

Sepehr-Sobhani:
Testing this because it is Friday and is a good time to see lots of flaky owasp zap scan failures


Sepehr-Sobhani:
Unsuccessful attempt!

PR #2155, opened 2024-09-04
Use release-it hook instead of make target

Sepehr-Sobhani:


marcellmueller (approved):
This is working well, nice work 🥇

PR #2170, opened 2024-09-05
chore: rename ownership models

BCerki:
card: https://github.com/orgs/bcgov/projects/122/views/2?filterQuery=&pane=issue&itemId=64944876

This PR renames the ownership models. I added 'designated' to the renamed model names to avoid related name clashes and mix-ups with through-tables.

Everything else in the card will be taken care of in another PR.


bc_obps/registration/models/operation_designated_operator_timeline.py
```diff
@@ -0,0 +1,35 @@
...
+    Operation,
+    Operator,
+    TimeStampedModel,
+)
+from simple_history.models import HistoricalRecords
+
+
+class OperationDesignatedOperatorTimeline(TimeStampedModel):
+    operation = models.ForeignKey(Operation, on_delete=models.DO_NOTHING, related_name="designated_operators")
+    operator = models.ForeignKey(Operator, on_delete=models.DO_NOTHING, related_name="operation_designated_operators")
```
Sepehr-Sobhani, line 12: We have used "DO_NOTHING" in many places, and I believe we should reconsider and change it to something more appropriate at some point. When we use "DO_NOTHING," we effectively bypass errors very gracefully. However, if we ever want to retrieve that record, we encounter a database error. I think we should prevent this before actually deleting it. For example, if we delete one of our operations for any reason, the record from this table associated with that operation remains unchanged. However, if we try to retrieve that record and want to fetch the operation of that record, the database will encounter an issue because that operation no longer exists. In short, we need to use "PROTECT" instead of "DO_NOTHING" in this table and in the other table related to the facility to prevent the deletion of the related records.
More about this: https://docs.djangoproject.com/en/5.1/ref/models/fields/#django.db.models.ForeignKey.on_delete

Sepehr-Sobhani (approved):
🥇 Great work! Just a small tweak is required otherwise good to go.

Sepehr-Sobhani:
FYI, we have a tech debt that is quite related to this PR and we will be able to tackle that once this PR is merged: https://github.com/orgs/bcgov/projects/122/views/2?filterQuery=facilit&pane=issue&itemId=66618892

PR #2174, opened 2024-09-09
324 update operation review

ayeshmcg:
Modified Operation Review Information page.

Changes include:

- Adding new component that includes Multistep form with task list.
- Modified Operation Review Page
- Modified Schema
- Added a blank page - Person Responsible
- Modified APIs

To test:

- Go to page /reporting/reports after logging in with BCeId
- Select any operation
- It wil open operation reveiw information page and it the version doesnt exist it will show no version Id.
- If there are any previously added activities or products, it will pre populated otherwise  will show empty.


bc_obps/reporting/api/product_data.py
```diff
@@ -0,0 +1,20 @@
...
+from django.http import HttpRequest
+from typing import Tuple
+from registration.schema.generic import Message
+from ninja.responses import codes_4xx, codes_5xx
+
+
+##### GET #####
+
+
+@router.get(
```
pbastia, line 13: This is something already implemented in registration: `registration/api/v1/regulated_products`
We should reuse / make those endpoints common where we can, instead of maintaining a double implementation

bciers/apps/reporting/src/app/components/operations/OperationReview.tsx
```diff
@@ -1,144 +1,197 @@
...
+        body: JSON.stringify(preparedData),
+        headers: {
+          "Content-Type": "application/json",
+          accept: "application/json",
+        },
+      });
+      router.push(saveAndContinueUrl);
+    } catch (error) {
+      console.error("Submission failed:", error);
+    }
```
pbastia, line 105: Unnecessary error handling, the `actionHandler` already has its own error handling (Sentry, and some logging), and will never throw.

`actionHandler` also returns a `response` object of type [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response) if you want to check the status: `response.ok`, `response.status` are much better bets.

bciers/apps/reporting/src/app/components/operations/OperationReview.tsx
```diff
@@ -1,144 +1,197 @@
...
 interface Props {
   formData: any;
   version_id: number;
+  reportingYear: {
+    reporting_year: number;
+    report_due_date: string;
+    reporting_window_end: string;
+  };
+  allActivities: { id: number; name: string }[];
+  allRegulatedProducts: { id: number; name: string }[];
```
pbastia, line 25: These could be just a `string[]` with the activities/products' names, since that's how the backend keys things (for now)

bciers/apps/reporting/src/app/components/operations/OperationReview.tsx
```diff
@@ -1,144 +1,197 @@
...
+      ...prevSchema,
+      properties: {
+        ...prevSchema.properties,
+        activities: {
+          type: "array",
+          title: "Reporting activities",
+          items: {
+            type: "number",
+            enum: allActivities.map((activity) => activity.id),
+            enumNames: allActivities.map((activity) => activity.name),
```
pbastia, line 129: then these are just the list of names (assumed unique)

bciers/apps/reporting/src/app/components/operations/OperationReview.tsx
```diff
@@ -1,144 +1,197 @@
...
+    if (!formData || !allActivities || !allRegulatedProducts) {
+      return;
+    }
+
+    const activities = formData.activities || [];
+    const products = formData.regulated_products || [];
+
+    const preselectedActivities = activities;
+    const preselectedProducts = products;
+
```
pbastia, line 118: javascript arrays are references, there is no point in duplicating these here

bc_obps/reporting/api/product_data.py
ayeshmcg: @pbastia Fixed this

bciers/apps/reporting/src/app/components/operations/OperationReview.tsx
```diff
@@ -1,144 +1,197 @@
...
+    setLoading(false);
+  }, [allActivities, allRegulatedProducts, formData]);
+
+  if (loading) {
+    return <div>Loading...</div>;
+  }
+
+  if (!formData || formData.error) {
+    return <div>No version ID found(TBD)</div>;
+  }
```
pbastia, line 178: Same here, I would let the generic react error boundary deal with this case. Otherwise we swallow an error and instead render something that may or may not be the root cause of the error

bciers/apps/reporting/src/app/components/operations/OperationReviewFormData.tsx
```diff
@@ -1,20 +1,25 @@
...
 }: {
   version_id: number;
 }) {
-  const reportOperation = await getReportOperation(version_id);
-
-  return <OperationReview formData={reportOperation} version_id={version_id} />;
+  const reportOperation = (await getReportingOperation(version_id)) || null;
+  const allActivities = (await getAllActivities()) || [];
+  const allRegulatedProducts = (await getAllRegulatedProducts()) || [];
+  const reportingYear = (await getReportingYear()) || null;
```
pbastia, line 15: We shouldn't provide default values, this will potentially obscure errors

```suggestion
  const reportOperation = await getReportingOperation(version_id);
  const allActivities = await getAllActivities();
  const allRegulatedProducts = await getAllRegulatedProducts();
  const reportingYear = await getReportingYear();
```

bciers/apps/reporting/src/app/components/routes/operations/Page.tsx
```diff
@@ -1,23 +1,32 @@
...
+  return dayjs(dateString).format("MMM DD,YYYY");
+};
 export default async function OperationsPage({
   searchParams,
 }: {
   searchParams: OperationsSearchParams;
 }) {
   const reportingYearObj = await getReportingYear();
 
+  const reportDueDate = reportingYearObj?.report_due_date
```
pbastia, line 16: `reportingYearObj.report_due_date` is bound to exist here, there is no need to handle the falsy case. This would swallow an error if there is one with the fetch

bciers/apps/reporting/src/app/components/routes/operations/Page.tsx
```diff
@@ -1,23 +1,32 @@
...
+    ? formatDate(reportingYearObj.report_due_date)
```
pbastia, line 17: `formatDate` could be in a `utils` folder in the `libs` package, this way we can reuse it everywhere. It's already duplicated somewhere else

bciers/apps/reporting/src/app/components/routes/operations/Page.tsx
```diff
@@ -1,23 +1,32 @@
...
   return (
     <>
       <div className="flex justify-between items-center">
         <h2 className="text-2xl font-bold">
           Reporting year {reportingYearObj.reporting_year}
         </h2>
-        <h3 className="text-bc-text text-right">
-          Reports due {reportingYearObj.report_due_date}
-        </h3>
+        {reportDueDate && (
```
pbastia, line 25: Same comment about `reportDueDate` existing no matter what

bciers/apps/reporting/src/data/jsonSchema/reviewOperationInformationText.tsx
```diff
@@ -0,0 +1,23 @@
+import InfoIcon from "@mui/icons-material/Info";
+import { Box, Paper, Typography } from "@mui/material";
+export const purposeNote = (
+  <Paper sx={{ p: 2, mb: 3, bgcolor: "#DCE9F6", color: "#313132" }}>
```
pbastia, line 4: the hex color should be coming from `libs/styles/src/colors.tsx` if possible

bciers/apps/reporting/src/data/jsonSchema/reviewOperationInformationText.tsx
```diff
@@ -0,0 +1,23 @@
...
+        The information shown on this page is data entered in registration. You
+        can edit it here but it will only apply to this report. To make edits
+        for all of your reports, please edit this information in registration.
+      </Typography>
+    </Box>
+  </Paper>
+);
+
+export const dateInfo = (
+  <Typography variant="body2" color="#38598A" fontSize="16px" gutterBottom>
```
pbastia, line 17: the hex color should be coming from `libs/styles/src/colors.tsx` if possible

bciers/apps/reporting/src/data/jsonSchema/operations.ts
```diff
@@ -1,71 +1,125 @@
...
+  },
+  purpose_note: {
+    "ui:FieldTemplate": TitleOnlyFieldTemplate,
+    "ui:title": purposeNote,
+  },
+
+  date_info: {
+    "ui:FieldTemplate": TitleOnlyFieldTemplate,
+    "ui:options": {
+      style: { variant: "body2", color: "#38598A", fontSize: "16px" },
```
pbastia, line 81: the hex color should be coming from `libs/styles/src/colors.tsx` if possible

bciers/libs/components/src/form/components/MultiStepTaskListButton.tsx
```diff
@@ -0,0 +1,93 @@
...
+              </Button>
+            )}
+          </>
+        )}
+      </div>
+    </div>
+  );
+};
+
+export default MultiStepTaskListButton;
```
pbastia, line 93: I don't think this is used anywhere?

bciers/apps/reporting/tests/components/OperationReview.test.tsx
```diff
@@ -1,78 +1,65 @@
...
+          activities: [1],
+          regulated_products: [1],
+          operation_representative_name: "John Doe",
+          operation_type: "Test Operation",
+        }}
+        version_id={1}
+        reportingYear={{
+          reporting_year: 2024,
+          report_due_date: "2024-12-31",
+          reportingWindowEnd: "2024-12-31",
```
pbastia, line 53:
```suggestion
          reporting_window_end: "2024-12-31",
```

bciers/apps/reporting/tests/components/OperationReview.test.tsx
```diff
@@ -1,78 +1,65 @@
...
+  it("renders loading state initially", () => {
     render(
-      <OperationReview formData={mockFormData} version_id={mockVersionId} />,
+      <OperationReview
+        formData={null}
+        version_id={1}
+        reportingYear={{
+          reporting_year: 2024,
+          report_due_date: "2024-12-31",
+          reportingWindowEnd: "2024-12-31",
```
pbastia, line 31:
```suggestion
          reporting_window_end: "2024-12-31",
```

bciers/apps/reporting/tests/components/OperationReview.test.tsx
```diff
@@ -1,78 +1,65 @@
...
+        allRegulatedProducts={[{ id: 1, name: "Product 1" }]}
+      />,
     );
 
-    // Check if buttons are rendered
-    expect(
-      screen.getByText(/Sync latest data from registration/i),
-    ).toBeInTheDocument();
-    expect(screen.getByText(/Cancel/i)).toBeInTheDocument();
-    expect(screen.getByText(/Continue/i)).toBeInTheDocument();
```
pbastia, line 76: I think we can keep these tests! The functionality shouldn't have changed after the refactor

pbastia (changes requested):
Nice work! And good thinking about refactoring the page component to house the tasklist, the navigation and the form.

A few things:
- Missing tests for the MultiStepFormWithTaskList / MultiStepTaskListButton
- The MultiStepTaskListButton component isn't used anywhere, was it supposed to be in the  MultiStepFormWithTaskList?
-  I've left a lot of comments about error handling: I think that in general, we can leave it up to Django/React/Sentry to handle the "something went wrong on the network, in the database, or unexpected behaviour" errors. Surrounding code with a try/catch block in random places makes it difficult to get an accurate stack trace and error message once the code is in production.
The only errors we should catch are the ones part of the program flow, that we know will happen for certain cases.

bciers/apps/reporting/src/app/components/operations/OperationReview.tsx
ayeshmcg: I added Ids, as that was required by MultiSelect widget for autocomplete when auto populating existing data

bc_obps/service/activity_service.py
```diff
@@ -28,5 +28,10 @@ def get_initial_activity_data(request, activity_name: str, report_date: str) ->
 
     @classmethod
     def get_all_activities(cls) -> List[Dict[str, Any]]:
-        activities = Activity.objects.all().values("id", "name", "applicable_to")
+        # Fetch activities and sort by weight
+        activities = (
+            Activity.objects.all()
+            .values("id", "name", "weight")  # Only fetch id, name, and weight
```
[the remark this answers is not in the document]
ayeshmcg: yes it isnt being used

bc_obps/service/report_service.py
```diff
@@ -121,11 +119,9 @@ def get_facility_report_by_version_and_id(
 
     @classmethod
     def get_activity_ids_for_facility(cls, facility_id: int) -> List[int]:
-        try:
-            facility = FacilityReport.objects.get(id=facility_id)
-            return list(facility.activities.values_list('id', flat=True))
-        except FacilityReport.DoesNotExist:
-            return []
+        # Fetch the facility and return the list of activity IDs
+        facility = FacilityReport.objects.get(id=facility_id)
```
pbastia, line 123: This is a little confusing. Do we mean to access it by facility ID? or by facility_report id? 
The current call will always return `None` since facility_report_id is of type int and facility_id is a UUID in our system

bc_obps/service/report_service.py
```diff
@@ -140,30 +136,22 @@ def save_facility_report(cls, report_version_id: int, facility_id: int, data: Fa
...
-                    'facility_bcghgid': data.facility_bcghgid.strip(),
-                },
-            )
-
-            # Update ManyToMany fields (activities)
-            if data.activities:
-                facility_report.activities.set(Activity.objects.filter(id__in=data.activities))
+        # Fetch or create a FacilityReport instance
+        facility_report, _ = FacilityReport.objects.update_or_create(
+            id=facility_id,
```
pbastia, line 141: Same, this will never update, we're mixing primary keys

bciers/apps/reporting/src/app/components/operations/PersonResponsible.tsx
```diff
@@ -0,0 +1,51 @@
...
+    isExpanded: true,
+    elements: [
+      { type: "Page", title: "Review Operation information", isChecked: true },
+      { type: "Page", title: "Person responsible", isActive: true },
+      { type: "Page", title: "Review facilities" },
+    ],
+  },
+];
+
+export default function PersonResponsible() {
```
pbastia, line 28: As a note: this is not part of the AC for the card, there is no need for you spend time on that.
But thanks for the effort! :D

pbastia (approved):
Nice!! There's one last piece that needs to be ironed out, between facility_id and report_facility_id. We can pair on that if you'd like.

I'll preemptively approve it

bc_obps/service/report_service.py
dleard: This was changed as part of the fixes I made in #2223

bc_obps/service/report_service.py
dleard: Same thing as above comment, I fixed the uuid - id mismatch here & changed this to a strict update in #2223. We shouldn't need a create or update here as the facility report is created on [report create](https://github.com/bcgov/cas-registration/blob/develop/bc_obps/service/report_service.py#L66), we should expect this to exist already by the time this endpoint is hit

bc_obps/service/report_service.py
```diff
@@ -96,11 +99,60 @@ def save_report_operation(cls, report_version_id: int, data: ReportOperationIn)
...
+        )
+
+        # Update ManyToMany fields (activities)
+        if data.activities:
+            facility_report.activities.set(Activity.objects.filter(id__in=data.activities))
+
+        # Save the updated FacilityReport instance
+        facility_report.save()
+
+        return facility_report
```
dleard, line 158: All of this can be removed. It was split into a `facility_report_service` in #2223

dleard (changes requested):
Just some leftover code to remove. The `facility_report` api/service stuff has been split into its own separate files to separate the concerns from report_service functionality

dleard (approved):
:+1:

PR #2175, opened 2024-09-09
refactor: fetch dynamic schema values in schema files

BCerki:
part of card: https://github.com/orgs/bcgov/projects/122/views/2?filterQuery=&pane=issue&itemId=64944876

Many of our form dropdowns are populated with options retrieved from the database (e.g. NAICS codes options come from the naics_code model). In the past, we've kept our json schemas in `.ts` files and then fetched and added the db options in the component. However, this isn't the most reusable option, and as we now need to use schemas in both admin and reg, I've moved the fetches to the schema file itself. This way, we can import the schema into both administration and registration and not have to refetch and re-edit the schemas in each component. Additionally, I find the schema files more readable because you can see the shape of the data directly in the schema (vs. an `anyOf: []` where you have to go look up what it actually contains in whatever component does the fetching).

This PR just refactors the administration-related schema and files. The registration side will be handled in https://github.com/bcgov/cas-registration/pull/2128.


bciers/apps/administration/tests/components/operations/OperationInformationPage.test.tsx
```diff
@@ -86,7 +88,6 @@ describe("the OperationInformationPage component", () => {
     expect(screen.getByText(/Process Flow Diagram/i)).toBeVisible();
     expect(screen.getByText(/Boundary Map/i)).toBeVisible();
     expect(screen.getByText(/Equipment List/i)).toBeVisible();
-    expect(screen.getByText(/Operation Representative/i)).toBeVisible();
```
BCerki, line 89: This will be handled in the next PR

bciers/apps/administration/tests/components/operations/OperationInformationPage.test.tsx
```diff
@@ -98,9 +99,9 @@ describe("the OperationInformationPage component", () => {
     expect(
       screen.getByRole("heading", { name: "Registration Information" }),
     ).toBeVisible();
-    expect(screen.getByText(/Registration Category/i)).toBeVisible();
-    expect(screen.getByText(/Regulated Product Names/i)).toBeVisible();
-    expect(screen.getByText(/Forcasted Emissions/i)).toBeVisible();
```
BCerki, line 103: This was cut

bciers/apps/administration/tests/components/operations/OperationInformationPage.test.tsx
```diff
@@ -98,9 +99,9 @@ describe("the OperationInformationPage component", () => {
     expect(
       screen.getByRole("heading", { name: "Registration Information" }),
     ).toBeVisible();
-    expect(screen.getByText(/Registration Category/i)).toBeVisible();
-    expect(screen.getByText(/Regulated Product Names/i)).toBeVisible();
```
BCerki, line 102: Only certain purposes have regulated products, and they're not applicable in this test

bciers/apps/administration/app/data/jsonSchema/operationInformation.ts
```diff
@@ -2,303 +2,374 @@ import provinceOptions from "@bciers/data/provinces.json";
...
-            },
+        activities: {
+          type: "array",
+          items: {
+            type: "number",
+            enum: reportingActivities.map(
+              (activity: { id: number; applicable_to: string; name: string }) =>
+                activity.id,
+            ),
+            // enumNames is a non-standard field required for the MultiSelectWidget
```
BCerki, line 87: The MultistepWidget requires enumNames, but it's a non-standard field that could be [deprecated in the next release](In your JSON Schema, you may also specify enumNames, a non-standard field which RJSF can use to label an enumeration. This behavior is deprecated and may be removed in a future major release of RJSF.)
Proposed tech debt card: https://github.com/orgs/bcgov/projects/122/views/2?filterQuery=purpose+&pane=issue&itemId=79158916

marcellmueller (commented):
Nice, I think this is a solid improvement and a good pattern to set. 

Only suggestion I have is the `operationInformation.ts` file is 375 lines long and getting a bit hard to parse. How can we organize this better now?

A couple ideas - let me know if you can think of something better:
- Separate `schema` and `uiSchema` folders. This is how we organized it in CCBC iirc. I didn't mind having them in the same file though I think we are growing past that especially with this example.
- An `operationInformation` folder with sections split out into different files

bciers/apps/administration/app/data/jsonSchema/operationInformation.ts
marcellmueller: Hmm I wonder if we should tackle this sooner than later. I've thought we should standardize how we do some of these since we have `enum/enumNames` and `title/const`.

BCerki: I think the SelectWidget also needs enum, not title/const

marcellmueller (approved):
🥇

Issue #1697, opened 2024-05-28
Admin\Operator - Internal user can view Operators


zoeyli-46:
#### Description:
As an internal user, I can see a list of all the operators 

#### Acceptance Criteria:

Given I am on the dashboard and click operators, or on the registration dashboard and click the operators tile
Then I am brought to the operators grid
[Figma](https://www.figma.com/design/LsDmLDCdnJqI8UahyqvOD8/OBPS-App-Designs?node-id=7861-1060&t=lP4wgjHuBjopAaSN-4) 

#### Development Checklist:

- [x] Create `bciers/apps/administration/app/components/operators/OperatorDataGrid.tsx`  similar to `bciers/apps/administration/app/components/operations/OperationDataGrid.tsx` 
- [x] In  `bciers/apps/administration/app/components/operators/OperatorDataGrid.tsx` include pagination similar to `bciers/apps/administration/app/components/operations/OperationDataGrid.tsx` 
- [x] Create `bciers/apps/administration/app/components/operators/OperatorDataGridPage.tsx` similar to `bciers/apps/administration/app/components/operations/OperationDataGridPage.tsx`
- [x] In `bciers/apps/administration/app/idir/cas_admin/operators/page.tsx` return `/operators/OperatorDataGridPage.tsx`` similar to `bciers/apps/administration/app/bceidbusiness/industry_user/operations/page.tsx`
- [x] In `bciers/apps/administration/app/idir/cas_analyst/operators/page.tsx` return `/operators/OperatorDataGridPage.tsx`` similar to `bciers/apps/administration/app/bceidbusiness/industry_user/operations/page.tsx`
- [x] Create `bciers/apps/administration/tests/components/operators/OperatorDataGridPage.test.tsx` similar to `bciers/apps/administration/tests/components/operations/OperationDataGridPage.test.tsx`
- [x] Create `bciers/apps/administration/tests/components/operators/OperatorDataGrid.test.tsx` similar to `bciers/apps/administration/tests/components/operations/OperationDataGrid.test.tsx`
- [ ] Meets the DOD

OR, because Brianna didn't realize Shon was already doing this:
- [ ] ~we can reuse the endpoint from bc_obps/registration/api/v1/user_operators.py, so no backend work in this card~
- [ ] ~create InternalUserOperatorsDataGridLayout, UserOperatorDataGridPage, UserOperatorDataGrid components in reg 2 (copy/paste over from reg 1 if they exist and then tweak to match wireframes)~
[...]

PR #2177, opened 2024-09-09
1697 operators grid

gdalcengio:
#1697
- feat: Set up and display operators page, grid and data with routing for internal users
- chore: Add search functionality to operator columns
- test: Add tests for front end operator grid components
- test: Add endpoint test for operators v2


bc_obps/registration/tests/endpoints/v2/test_operators_endpoint.py
```diff
@@ -0,0 +1,104 @@
+from registration.constants import PAGE_SIZE
+from registration.tests.utils.helpers import CommonTestSetup, TestUtils
+
+from registration.tests.utils.bakers import operator_baker
+from registration.utils import custom_reverse_lazy
+
+
+class TestOperatorsEndpoint(CommonTestSetup):
+    endpoint = CommonTestSetup.base_endpoint + "v2/operators"
```
Sepehr-Sobhani, line 9: We can update the endpoint here to `custom_reverse_lazy("list_operators_v2")` and then replace all occurrences of `custom_reverse_lazy("list_operators_v2")` in this file with `self.endpoint`. This way, if the endpoint name changes for any reason, we only need to update one variable.

bc_obps/registration/tests/endpoints/v2/test_operators_endpoint.py
```diff
@@ -0,0 +1,104 @@
...
+
+        # Get the default page 1 response
+        response = TestUtils.mock_get_with_auth_role(
+            self, "cas_admin", custom_reverse_lazy('list_operators_v2') + "?legal_name=gouda"
+        )
+        assert response.status_code == 200
+        response_data = response.json().get('data')
+        assert len(response_data) == 1
+        for item in response_data:
+            assert item.get('legal_name') == 'gouda'
```
Sepehr-Sobhani, line 70: We can use an index instead of a for-loop since we only have one item with that name.
```python
response_data[0].get('legal_name') == 'gouda'
```

bc_obps/registration/tests/endpoints/v2/test_operators_endpoint.py
```diff
@@ -0,0 +1,104 @@
...
+        response_data = response.json().get('data')
+        assert len(response_data) == 0
+
+        # Test with a cra_business_number filter
+        response = TestUtils.mock_get_with_auth_role(
+            self, "cas_admin", custom_reverse_lazy('list_operators_v2') + "?cra_business_number=118675309"
+        )
+        assert response.status_code == 200
+        response_data = response.json().get('data')
+        assert len(response_data) == 1
```
Sepehr-Sobhani, line 86: I suggest validating the `cra_business_number` in the response to ensure it aligns with our expectations.

bc_obps/registration/tests/endpoints/v2/test_operators_endpoint.py
```diff
@@ -0,0 +1,104 @@
...
+
+        # Test with multiple filters (with business structure obj.first (BC Corp))
+        response = TestUtils.mock_get_with_auth_role(
+            self,
+            "cas_admin",
+            custom_reverse_lazy('list_operators_v2') + "?cra_business_number=118675309&business_structure=BC",
+        )
+        assert response.status_code == 200
+        response_data = response.json().get('data')
+        assert len(response_data) == 1
```
Sepehr-Sobhani, line 104: Same as the above comment.

bciers/apps/administration/tests/components/operators/OperatorDataGrid.test.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+  //   expect(operationInfoLinks[0]).toHaveAttribute(
+  //     "href",
+  //     "operations/1?operations_title=Operation+1",
+  //   );
+  //   expect(operationInfoLinks[1]).toHaveAttribute(
+  //     "href",
+  //     "operations/2?operations_title=Operation+2",
+  //   );
+  // });
```
Sepehr-Sobhani, line 97: I suggest removing this test as it will pollute the code and let the user who tackles the related ticket to handle that. This way we have a cleaner codebase.

Sepehr-Sobhani (changes requested):
Great work! 🥇 Only a couple of minor tweaks are required.

bc_obps/service/data_access_service/operator_service.py
```diff
@@ -10,6 +10,10 @@ class OperatorDataAccessService:
     def get_operator_by_id(cls, operator_id: UUID) -> Operator:
         return Operator.objects.get(id=operator_id)
 
+    @classmethod
+    def get_all_operators(cls) -> QuerySet[Operator]:
+        return Operator.objects.all()
+
```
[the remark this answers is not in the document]
andrea-williams: for now I think we can assume that yes, Declined operators should be excluded. 

For Gabe's context - in Reg 1 internal users wanted the ability to review new operators that were created and approve/decline them. They've since decided they don't need to do that, so Reg 2 doesn't have the concept, but in our prod database we have a few operators that are Declined.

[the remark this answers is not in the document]
andrea-williams: actually just checked Metabase - we don't have any Declined operators in prod. But I guess it's theoretically possible that one could come along between now and 2025

bciers/apps/administration/app/components/operators/OperatorDataGridPage.tsx
```diff
@@ -0,0 +1,40 @@
...
+  const operators: {
+    rows: OperatorRow[];
+    row_count: number;
+  } = await fetchOperatorsPageData(searchParams);
+  if (!operators) {
+    return <div>No user operator data in database.</div>;
+  }
+
+  const isAuthorizedAdminUser =
+    role?.includes("cas") && !role?.includes("pending");
```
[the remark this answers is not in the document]
andrea-williams: I think we do need this - `cas_pending` is a user role, and those users shouldn't be allowed access to operator data

bc_obps/registration/tests/endpoints/v2/test_operators_endpoint.py
andrea-williams: @gdalcengio what a very un-Italian choice of cheese. Tsk tsk.

bciers/apps/administration/app/components/operators/OperatorDataGridPage.tsx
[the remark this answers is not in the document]
Sepehr-Sobhani: We are not using this variable anywhere. Additionally, our FE authorization won't allow access to this page for any cas_pending user.

bc_obps/registration/tests/endpoints/v2/test_operators_endpoint.py
gdalcengio: @andrea-williams I took it from operations endpoint and it is a very gouda cheese, but you're right. Will change it.

bciers/apps/administration/app/components/operators/OperatorLayouts.tsx
Sepehr-Sobhani, line 1: Based on the wireframe, we still need the following code to be on top of the grid:
```typescript
<Note>	
  <b>Note:</b> View all the operators here.	
</Note>	
<h1>Operators</h1>
```
We only need this for internal users, so there's no need to create a separate file for external users.

Sepehr-Sobhani (changes requested):

Sepehr-Sobhani (approved):
Neat 🤩

Issue #1602, opened 2024-05-03
MultiSelectWidget doesn't support required field


marcellmueller:
Describe the Bug:

This was caught during testing as the only field we use this widget for, `regulated_products` is not a required field. To make an array field required in RJSF you need to set `minItems: 1` in the schema. Doing this currently breaks the `MultiSelectWidget`. 

This will be a low priority to fix though adding this bug issue here to document it as it could cause a headache for someone as we build out reporting and registration part 2.

marcellmueller:
Moving this bug into the sprint backlog as it is blocking a required field in #1899

patriciarussellCAS:
@marcellmueller could we chat about this sometime next week? I'm reviewing the Operation Information page in registration and no longer see the regulated products dropdown option anymore. Is this related?

marcellmueller:
> @marcellmueller could we chat about this sometime next week? I'm reviewing the Operation Information page in registration and no longer see the regulated products dropdown option anymore. Is this related?

Looking into this, there may have been a bug introduced. We can chat on Monday!

marcellmueller:
@patriciarussellCAS This is good to test in dev now. Thanks to Brianna for fixing the bug with the regulated operations field in a recent PR 🙇

patriciarussellCAS:
Thanks @marcellmueller (and @BCerki !) - This works for me, however I see that it's displaying the regulated product names field for Reporting Operations. 
I have posted in the SME channel to confirm if this should be the case, as my understanding is that the regulated product name field should ONLY apply to Regulated Operations, Opt-ins and New Entrants. I will let you know once I have that confirmed.

BCerki:
> Thanks @marcellmueller (and @BCerki !) - This works for me, however I see that it's displaying the regulated product names field for Reporting Operations. I have posted in the SME channel to confirm if this should be the case, as my understanding is that the regulated product name field should ONLY apply to Regulated Operations, Opt-ins and New Entrants. I will let you know once I have that confirmed.

[...]

PR #2185, opened 2024-09-11
multiselect required

marcellmueller:
#1602 

To test this:
- Go to the first page of the `Register an Operation` form - `Operation Information`
- Select `Reporting Operation` in the Registration Purpose dropdown
- The `Regulated Product Name(s)` field will appear which is a `MultiSelectWidget`. Click `Save and Continue` to trigger validation.
Screenshot 2024-09-12 at 8 19 29 AM


bciers/apps/registration1/tests/utils/customTransformErrors.test.ts
```diff
@@ -108,6 +108,19 @@ const phoneFormatError = [
   },
 ];
 
+const arrayMinItemsError = [
+  {
+    name: "minItems",
+    property: ".array_field",
+    message: "must NOT have fewer than 1 items",
```
marcellmueller, line 115: I changed this default error because all of our messages start with a capital letter and I felt like the NOT was yelling at us.

bciers/apps/registration/app/components/operations/registration/OperationInformationPage.tsx
```diff
@@ -42,25 +42,31 @@ export const createOperationInformationSchema = (
...
+      purpose !== RegistrationPurposes.ELECTRICITY_IMPORT_OPERATION &&
+      purpose !== RegistrationPurposes.POTENTIAL_REPORTING_OPERATION;
     return {
-      // required: ["regulated_products"],
+      // This works:
+      // required: [isRegulatedProducts && "regulated_products"],
+      //
+      // This triggers validation error even when dynamic field isn't displayed:
+      required: ["regulated_products"],
       properties: {
```
marcellmueller, line 54: When testing this I uncommented the `minItems` and `required` field and initially thought everything was working as expected. However I noticed that the `required` validation would still be triggered if the field wasn't displayed.

To reproduce this:
- run the app
- go to `Register an Operation` form and on the first page `Operation Information` select `Potential Reporting Operation` as the `Registration Purpose` and select an Operation in the `Select your Operation` dropdown.
- Click `Save and continue`
- Nothing happens, you can open up your browser console and see the validation error for `regulated_products`

However if you try comment out this required and uncomment the conditional required above it works. I'm not used to this error and am not sure if it's due to this fix or something with our schema.

BCerki: It's an existing problem with the schema. The `map` adds `required: ["regulated_products"]` to all of the `oneOf` options, even the ones that don't actually have a `regulatedProducts` property, e.g.
image

bciers/apps/registration/app/components/operations/registration/OperationInformationPage.tsx
```diff
@@ -42,25 +42,31 @@ export const createOperationInformationSchema = (
   );
 
   const oneOfOptions = purposes.map((purpose) => {
+    const isRegulatedProducts =
```
BCerki, line 45: Maybe `needsRegulatedProducts` or `hasRegulatedProducts`?

bciers/apps/registration/app/components/operations/registration/OperationInformationPage.tsx
```diff
@@ -42,25 +42,31 @@ export const createOperationInformationSchema = (
...
+      purpose !== RegistrationPurposes.ELECTRICITY_IMPORT_OPERATION &&
+      purpose !== RegistrationPurposes.POTENTIAL_REPORTING_OPERATION;
     return {
-      // required: ["regulated_products"],
+      // This works:
+      // required: [isRegulatedProducts && "regulated_products"],
```
BCerki, line 50: More details below,  yes, we should use `required: [isRegulatedProducts && "regulated_products"]`

bciers/libs/components/src/form/widgets/MultiSelectWidget.test.tsx
```diff
@@ -2,10 +2,14 @@ import { userEvent } from "@testing-library/user-event";
...
 import FormBase from "@bciers/components/form/FormBase";
-import { checkNoValidationErrorIsTriggered } from "@/tests/helpers/form";
+import {
+  checkNoValidationErrorIsTriggered,
+  checkComboBoxWidgetValidationStyles,
+} from "@/tests/helpers/form";
 
 const multiSelectFieldLabel = "MultiSelectWidget test field";
 const multiSelectLabelRequired = `${multiSelectFieldLabel}*`;
+const minItemsMessage = "Must not have fewer than 1 items";
```
BCerki, line 12:
```suggestion
const expectedMinItemsMessage = "Must not have fewer than 1 items";
```

BCerki (approved):
Nice work, bug is fixed! I have a couple nits and put an explanation for what's going on with the required fields, you're right, it was a schema issue and your conditional required works well. I'd also suggest searching the codebase for anything that uses the MultiSelectWidget and uncommenting/adding `minItems: 1`. That widget is used in at least one other schema. :cartwheeling:

bciers/apps/registration/app/components/operations/registration/OperationInformationPage.tsx
marcellmueller: Ah that makes sense, thanks so much for figuring this out. I wasn't completely confident I had fixed it.

bciers/apps/registration/app/components/operations/registration/OperationInformationPage.tsx
marcellmueller: I don't feel strongly at all however I think prefixing boolean variables with `is` is a common practice we have been using throughout our codebase. This keeps it in line with most of our other conditional variables! 

That said I don't mind changing it if you think otherwise.

Issue #2092, opened 2024-08-15
Admin\ContactForm - Field `existing_bciers_user` displays after creating a new contact


shon-button:
Describe the Bug:

Admin\ContactForm field `existing_bciers_user` displays after creating a new contact but should only display on new form.
![image](https://github.com/user-attachments/assets/29724cbe-5a2b-43e0-8992-a8be6d724b93)

Seemingly, the after `ContactForm` create/submit the `ContactPage` parameter `schema={createContactSchema(userOperatorUsers, isCreating)}` still computes as create.
![image](https://github.com/user-attachments/assets/8ab85a64-090e-40dc-9fce-3a984d0ea1ff)

Probability (how likely the bug is to happen, scored from 1-5):
(For example, probability of 5 is something like "it happens to all users every time they log in."
Probability of 1 "only happens to certain users when a really specific and unlikely path is followed.")

Effect (how bad the bug is when it does happen, scored from 1-5):
(For example, effect of 5 is "the entire app crashes and makes it unusable for all users" or "the bug causes the wrong data to be saved, with critical information (e.g. payment) being affected."
Effect of 1 is "It makes some styling look a little bit weird.")

Steps to reproduce the behaviour:

Screenshots:

Additional information:

- Browser: [e.g. chrome, safari]
- Version: [e.g. 22]
- Additional context

Sepehr-Sobhani:
@shon-button I'm curious why the following line of code in `createContactSchema.tsx` isn't working?
```typescript
...
  } else {
    delete localSchema.properties.section1.properties.existing_bciers_user;
    delete localSchema.properties.section1.allOf;
  }
...
```

BCerki:
@shon-button , I'm not able to reproduce this: https://www.loom.com/share/5d010892a279493a8c7df730780b5127. Maybe it got fixed in some of the work @Sepehr-Sobhani  was doing with the Operation Representative form?

Sepehr-Sobhani:
@BCerki To replicate this issue, create a new contact. After submitting the form, you will still see 'existing_bciers_user'(in read-only format) but it disappears upon refreshing the page.

patriciarussellCAS:
Are we still using the "Is this contact a user in BCIERS" toggle? I feel like I heard this was taken out, but may be mis-remembering.

patriciarussellCAS:
I'm unable to recreate this issue - has this been resolved? @Sepehr-Sobhani or @BCerki

BCerki:
[...]

PR #2187, opened 2024-09-12
Fix: user toggle

BCerki:
card: https://github.com/bcgov/cas-registration/issues/2092

The bug wasn't just the user toggle; the page was displaying everything for creating instead of editing. E.g. the note, the title "Add Contact":
image

This was happening because we were going to the `add-contact` route after submission. `add-contact` has no `contactId`, and whether to show create/edit is conditional on the existence of `contactId`. This was the log: image I've left the console.logs in the first commit of this PR if you want to look.

I was able to fix it by redirecting using `router.replace` instead of a `window.replace`.


shon-button (approved):
Sublime

Issue #2140, opened 2024-08-29
OWASP Zap scan - mitigate flakey rate limited error


marcellmueller:
###Description of the Tech Debt###

The OWASP Zap scan in our CI regularly fails because of rate limiting. We have gotten quite used to it but this is wasting a lot of time as we have to regularly re-run the job. We should fix this save time, avoid toil and ensure that we can deploy to dev automatically once a PR is merged into main and passes all checks again.

Some ideas to mitigate this:

- Use a [retry action](https://github.com/marketplace/actions/retry-step) and make this set retry once or twice before it fails
- Only run OWASP zap on merge to main? There are some risks to this however, such as merging in unsafe code and then being blocked going to prod until we fixed it.

Sepehr-Sobhani:
Not tested but we might be able to use github actions `strategy` and [handle the failure](https://docs.github.com/en/enterprise-cloud@latest/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow#handling-failures) to re-run the action multiple times:

Something like:
```yaml
jobs:
  zap-scan:
    runs-on: ubuntu-latest
    strategy:
      fail-fast: false
      matrix:
        include:
          - attempt: 1
          - attempt: 2
          - attempt: 3
     ....
continue-on-error: ${{ matrix.attempt != 3 }}
```

PR #2188, opened 2024-09-12
2140 retry owasp zap

marcellmueller:
#2140 

This method seems to be commonly recommended as a way to re-run workflows on failure:
https://github.com/orgs/community/discussions/67654#discussioncomment-8038649
https://medium.com/@aarne.laur/retry-failed-github-actions-8661e7601c66
https://dev.to/logto/automatically-rerun-your-github-workflow-after-failure-26hi

However I discovered that you can't manually trigger workflows that don't exist yet in the default branch so I get a 404 error:
https://stackoverflow.com/a/78765950

Screenshot 2024-09-12 at 10 45 58 AM

Tried the same command with a workflow that exists in `develop`:
Screenshot 2024-09-12 at 11 01 35 AM

### DIFF
--- a/.github/workflows/zap-owasp.yaml
+++ b/.github/workflows/zap-owasp.yaml
@@ -41,14 +41,14 @@ jobs:
           issue_title: OWASP Baseline - Backend
           fail_action: false
 
-  # retry-on-failure:
-  #   needs: zap-scan
-  #   if: failure() && fromJSON(github.run_attempt) < 3
-  #   runs-on: ubuntu-latest
-  #   steps:
-  #     - env:
-  #         GH_REPO: ${{ github.repository }}
-  #         GH_TOKEN: ${{ github.token }}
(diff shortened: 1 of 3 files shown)


.github/workflows/zap-owasp.yaml
```diff
@@ -38,3 +40,15 @@ jobs:
           cmd_options: "-a -d -T 5 -m 2"
           issue_title: OWASP Baseline - Backend
           fail_action: false
```
pbastia, line 42: would this get triggered if the failure is the secondary rate limit? We could pass the "retry-on-failure" directly here

marcellmueller: Oh by creating a `retry-on-failure` action? Might be a good idea anyways to reuse if there are other flaky workflows in the future.

Sepehr-Sobhani (approved):

PR #2193, opened 2024-09-12
2140 retry workflow

marcellmueller:
#2140

This took me awhile to figure out since the secondary rate limiting error was seemingly causing the job to fail, however it wasn't triggering the `failure()` condition. So after some trial and error I got the retry working well. The `!cancelled()` condition had to be added as the retry workflow was being called if I pushed an update which cancels the in progress pr.

If you want to test this it's pretty easy, though a little time consuming depending on how many tries it takes to trigger the rate limiting error. Just retry the `zap-scan` job. It will fail eventually, then you will see the retry job being called. It will also appear as a second workflow in our Action tab:

Screenshot 2024-09-13 at 11 45 41 AM

After that completes you will see the failed workflow restart:
Screenshot 2024-09-13 at 11 45 26 AM

### DIFF
--- a/.github/workflows/zap-owasp.yaml
+++ b/.github/workflows/zap-owasp.yaml
@@ -12,8 +12,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      # - name: test failure
-      #   run: exit 1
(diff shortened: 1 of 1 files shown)


pbastia (approved):
💯

PR #2199, opened 2024-09-13
Chore/346 updates to fuels

dleard:
- Unit Type & Fuel Name fields are now dropdowns
- Fuel Name enum is populated from list of fuels in the form_builder
- FuelType model values updated to reflect what IRC has determined the list of fuels to be
- Classifications added to fuel records


bc_obps/service/form_builder_service.py
```diff
@@ -11,6 +17,16 @@ def str_to_camel_case(st: str) -> str:
...
 
 
+def get_fuel_list() -> List[str]:
+    """
+    Return the roles that are considered as authorized CAS users (excluding cas_pending).
+    """
+    try:
+        return list(FuelType.objects.all().values_list("name", flat=True))
+    except Exception:
+        return []
```
pbastia, line 27: What exception would we expect here? We should probably let our `@handle_http_errors()` decorator deal with it at the API level instead of swallowing whatever might have gone wrong

bc_obps/reporting/tests/api/test_build_form_schema.py
```diff
@@ -65,6 +65,15 @@ def test_returns_source_type_schema(self):
...
             ]['items']['properties']
         )
+        # Fetched list of fuels & added to enum
+        assert (
+            len(
+                response_object['schema']['properties']['sourceTypes']['properties'][source_type_key]['properties'][
+                    'units'
+                ]['items']['properties']['fuels']['items']['properties']['fuelName']['enum']
+            )
+            > 0
```
pbastia, line 75:
```suggestion
            ==  FuelType.objects.count()
```

bc_obps/service/form_builder_service.py
pbastia: This method should probably be part of the data access service, and could implement a caching mechanism (See RegulatedProductDataAccessService or NaicsCodeDataAccessService)

pbastia (approved):
Very nice! 🥇  Just a couple of small comments, data access should be part of the services layer

PR #2201, opened 2024-09-13
fix/reg1 nx affected e2e

marcellmueller:
Some issues cropped up from the e2e PR when there are no nx affected projects. Changes in this PR:

- `registration1` runs every time to be safe since it is in production (we should run the full e2e suite on merge to main in the future)
- fixes the `e2e-finalize` conditional so it doesn't get skipped when other projects are skipped
- adds separate Happo projects for each app. If projects were skipped the other screenshots were showing as deleted. This will also make it easier to figure out where the issues are if we have diffs.
- Adds `happo-skip` helper to call the Happo api and skip projects that aren't in `nx affected`. This way we can add them  as a required status check.

Adding `happo-skip` was slightly more involved than expected since it's not part of the `happo-e2e` cli:
https://github.com/happo/happo-e2e/issues/27

The `cas-coam` Happo project was manually skipped here running `happo-script.js` locally:
Screenshot 2024-09-16 at 4 14 42 PM

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -120,7 +120,8 @@ jobs:
         test-registration-e2e,
         test-registration1-e2e,
       ]
-    # This wasn't wasn't being triggered if any tests were skipped
+    # Run this job only if all the e2e tests are successful or skipped but not on failure.
+    # This was needed as just the `needs` condition was skipping the entire job if any of the e2e tests were skipped.
     if: |
       always() &&
       (needs.test-administration-e2e.result == 'success' || needs.test-administration-e2e.result == 'skipped') &&
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -93,7 +93,6 @@ jobs:
       image_url: "ghcr.io/bcgov/cas-rep-frontend"
     secrets: inherit
   test-registration1-e2e:
-    # if: ${{contains(needs.check-nx-affected.outputs.NX_AFFECTED_E2E, 'registration1')}}
     needs: [build-backend, build-registration1, check-nx-affected]
     uses: ./.github/workflows/test-registration1-e2e.yaml
     secrets: inherit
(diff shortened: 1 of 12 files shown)


.github/workflows/main.yaml
```diff
@@ -121,6 +120,14 @@ jobs:
         test-registration-e2e,
         test-registration1-e2e,
       ]
+    # This wasn't wasn't being triggered if any tests were skipped
```
BCerki, line 123: I don't totally follow this comment. It seems like it refers to old code that you've since replaced? Maybe a comment more like "we run this job only if tests are successful or skipped, without this condition, [insert bad things] happen because [reasons]"

bciers/libs/e2e/src/utils/happo-skip.js
```diff
@@ -0,0 +1,52 @@
...
+
+const apiKey = process.env.HAPPO_API_KEY;
+const apiSecret = process.env.HAPPO_API_SECRET;
+const project = process.env.HAPPO_PROJECT;
+const commitSha = process.env.COMMIT_SHA;
+
+// This helper will skip the current commit/project in happo
+// https://happo.io/docs/api
+
+async function skipHappo() {
```
BCerki, line 35: this is cool

BCerki (approved):
This looks great, thanks for putting it together! A couple of inline comments, and re this bullet in the PR note: `registration1 runs every time to be safe since it is in production (we should run the full e2e suite on merge to main in the future)`:
- thanks for including reg1
- flagging that in our last dev discussion, the plan was to run happo on PRs but skip it for shipit so that we don't have to manually approve flaky screenshots just to get to dev. I tagged you in Teams re the PR for that
- yarn audit will probably fail for you if you need to rebase. I've fixed it [here](https://github.com/bcgov/cas-registration/pull/2215) but I'd rather put your PR in first so I can run e2e on mine (I tried playwright locally and had errors on both develop and my branch, so unsure if the update broke anything).

.github/workflows/main.yaml
marcellmueller: That's a good point, edited the comment for clarity:
https://github.com/bcgov/cas-registration/pull/2201/commits/b01fa505e5383cbb56172527abc4ebc91e3c472d

PR #2202, opened 2024-09-13
fix/regulated products

marcellmueller:
Follow up PR for #1602 

I'm not sure what changed but Patricia mentioned the `Regulated Products` field wasn't appearing in her review. I verified that it wasn't working anymore, however this definitely worked when I put it up for review. Anyways this fixes the conditional field.


bciers/apps/registration/app/components/operations/registration/OperationInformationPage.tsx
```diff
@@ -46,8 +46,14 @@ export const createOperationInformationSchema = (
...
 
+    const required = isRegulatedProducts
+      ? {
+          required: ["registration_purpose"],
+        }
+      : {};
+
     return {
-      required: [isRegulatedProducts && "regulated_products"],
+      ...required,
```
marcellmueller, line 56: I was split on this solution or just setting `required` to `required: [isRegulatedProducts ? "regulated_products" : ""]` seemed to work. Though I opted for leaving the required field out altogether if not needed.

Issue #2136, opened 2024-08-28
Reg\Register an Operation\Page "Operation Representative" - External user can create new Contacts by using existing user information


Sepehr-Sobhani:
#### Describe the task

This ticket covers the missing part of the ticket: https://github.com/orgs/bcgov/projects/122/views/2?filterQuery=&pane=issue&itemId=69560403

#### Acceptance Criteria:

Given that I am on the preceding page (either new entrant, opt in, or facility information), When I click on the save and continue button,
Then I am brought to the Operation Representative page of the form
https://www.figma.com/design/LsDmLDCdnJqI8UahyqvOD8/BCIERS-Designs?node-id=18211-1609&t=4RQUxOy5Cqbs60a7-4

Given that I am on the Operation Representative page as an External user,
When the operation has existing operation representatives,
Then I see a list of operation representatives

Given that I am on the Operation Representative page as an External user,
When the operation has existing operation representatives,
Then I can see a button to add new operation representatives

Given that I am on the Operation Representative page as an External user,
When I click the button to add new operation representatives
Then I see a form to add the required info

**OPTION 1:**
Given that I am on the Operation Representative page as an External user,
When I fill out the form and click the `Save Operation Representative` button
Then I see the new operation representative being added to the operation(see: https://www.figma.com/design/LsDmLDCdnJqI8UahyqvOD8/BCIERS-Designs?node-id=18250-6024&t=OZlPtCxjAD0tYi1s-4)

**OPTION 2:**
Given that I am on the Operation Representative page as an external user, 
When I select an existing contact from the dropdown,
Then the form should be populated with the information of the selected contact(see:https://www.figma.com/design/LsDmLDCdnJqI8UahyqvOD8/BCIERS-Designs?node-id=18250-6794&t=OZlPtCxjAD0tYi1s-4(**NOTE**: First Name, Last Name and Email fields of existing contact are not editable)

Given that I am on the Operation Representative page as an external user, 
When I fill out the form for existing contact and click the `Save Operation Representative` button
Then I see the new operation representative being added to the operation(see: https://www.figma.com/design/LsDmLDCdnJqI8UahyqvOD8/BCIERS-Designs?node-id=18250-6024&t=OZlPtCxjAD0tYi1s-4)

**NOTE:**
Removing existing operation representatives will be handled in ticket: https://github.com/orgs/bcgov/projects/123/views/16?pane=issue&itemId=81400225

[...]

PR #2204, opened 2024-09-14
2136 Reg\Register an Operation\Page "Operation Representative"

Sepehr-Sobhani:
[ISSUE 2136](https://github.com/bcgov/cas-registration/issues/2136)

**NOTES:**

- Added 'uuid' prefix to all registration endpoints using 'operation_id' as the path parameter to prevent conflicts with similar endpoint names.
- Updated existing tests related to the Operation Representative step and added new ones.
- Refactored the logic in handling submit button text in 'MultiStepButtons.tsx'.
- Added 'ui:inline' prop to 'ReadOnlyMultiSelectWidget.tsx' to display the value inline.
- Added a new 'beforeForm' prop to 'MultiStepBase.tsx', allowing us to add other components and forms before the main form. This enables us to have multiple forms on the same page and add new records while staying on the same page.
- Removing operation representatives will be handled by this ticket: https://github.com/orgs/bcgov/projects/123/views/16?pane=issue&itemId=81400225

**TO TEST:**
- Log in as `bc-cas-dev` and follow the AC in the Card: [# 2136](https://github.com/bcgov/cas-registration/issues/2136)


andrea-williams (commented):
FINALLY finished reading through the code! will pull the branch to run locally and try to break it (😈) after lunch

andrea-williams (approved):
Ran the code locally and confirmed:

- with an existing unregistered operation, I was able to select an existing contact and assign them as operation rep. The operation then showed up in the "Places Assigned" list for the contact
- with the same existing operation, I was able to create a new contact and also assign them as operation rep. The new contact then appeared in the Contacts grid, and had the appropriate "Places Assigned"

I noticed that the list of operation reps doesn't appear in the Operation Details page, so at the moment there's no way to edit the list of operation reps once the operation has been registered. However, I think that's beyond the scope of this ticket so I'm approving. 

Great work!

Sepehr-Sobhani:
> Ran the code locally and confirmed:
> 
> * with an existing unregistered operation, I was able to select an existing contact and assign them as operation rep. The operation then showed up in the "Places Assigned" list for the contact
> * with the same existing operation, I was able to create a new contact and also assign them as operation rep. The new contact then appeared in the Contacts grid, and had the appropriate "Places Assigned"
> 
> I noticed that the list of operation reps doesn't appear in the Operation Details page, so at the moment there's no way to edit the list of operation reps once the operation has been registered. However, I think that's beyond the scope of this ticket so I'm approving.
> 
> Great work!

Thank you @andrea-williams !
Suggestion implemented, and the handling of Op Reps in the operation details page will be tackled in the ticket: https://github.com/bcgov/cas-registration/issues/2327

PR #2214, opened 2024-09-17
Chore/349 autofill dependent fuel fields

dleard:
On fuel selection, the classification & unit fields are filled automatically from data in the fuel_type database table.
- A fuelType object has been added to all schemas that contain fuels (object contains: fuelName, classification, units)
- A custom rjsf field for the fuelType object has been added to handle all these fields as one
- The form_builder has been updated to handle the change in the schema structure

To test this:
Go to any activity page that has fuels in the schema & add a fuel from the dropdown. After selection, the classification & unit fields should be automatically updated. They should also be read-only.


joshgamache (approved):
Tests passing, and clickthrough had everything showing up where expected and read-only. Took a look through the code and everything looks great there! 👍

PR #2215, opened 2024-09-17
chore: update next

BCerki:
Don't merge until we have reg1 e2e tests running in CI again--good to go now, #2201 is merged

No card, yarn audit in CI failed


marcellmueller:
Looks like registration1 tests were running in this PR. I got worried about this as doing something like updating nextjs should definitely trigger `nx affected` and looks like all e2e are running in this.

Anyways thanks for updating next! Appreciate it.

PR #2234, opened 2024-09-19
chore: activities shown in custom order & then alphabetical

dleard:
IRC requested that activities be ordered in this way:
General Stationary Combustion
Fuel by Mobile Combustion
Alphabetical after the above activities

This PR updates the weights to be equal for all non GSC / mobile combustion activities & updates the api service to order by weight & then alphabetical by name.

You can see this ordering implemented on the facility review page (screenshot below)

Screenshot from 2024-09-19 15-26-13


ayeshmcg (approved):
@dleard  Looks great. Activities list appear to be in oder on Facilities review page as well on the Review Operation Data page

PR #2235, opened 2024-09-19
Feat: save activity form data

pbastia:


bc_obps/reporting/service/report_activity_save_service.py
```diff
@@ -0,0 +1,224 @@
...
+    user_guid: uuid.UUID
+
+    def __init__(self, report_version_id: int, facility_id: uuid.UUID, activity_id: int, user_guid: uuid.UUID):
+        self.user_guid = user_guid
+        self.activity = Activity.objects.get(pk=activity_id)
+        self.facility_report = FacilityReport.objects.select_related('report_version', 'report_version__report').get(
+            report_version_id=report_version_id, facility_id=facility_id
+        )
+
+    @transaction.atomic()
```
pbastia, line 42: @Sepehr-Sobhani what's the best practice about atomic transactions? Should that be specified only at the API level? We risk nesting transactions otherwise, I don't know the performance penalty for that

Sepehr-Sobhani: Based on my knowledge having it at the top level of the function should be enough (service layer) but we can also manage nested transactions with context manager: https://docs.djangoproject.com/en/5.1/topics/db/transactions/#:~:text=and%20as%20a%20context%20manager%3A

Note: I wasn't able to find a way to automate testing atomic transactions so I always do it manually.

Issue #2105, opened 2024-08-20
Admin\Operations - Single Facility Operation should only be able to create 1 facility


marcellmueller:
#### Describe the task

The current implementation of facilities features comes with the risk that an SFO could create multiple Facilities. We should implement some safeguards that only allow an SFO to create 1 facility.

#### Acceptance Criteria

Just some thoughts - feel free to modify, add or remove some of these

- [x] Currently we are bypassing the `Facilities` DataGrid and linking an SFO to the Facility form. However the could still go to the facilities grid via URL, so we should hide the `Add Facility` button.
- [x] Add a check in the backend create facility POST route that won't allow creation of a second Facility for an SFO

#### Additional context

- Add any other context about the task here.
- Or here

patriciarussellCAS:
Does this need to be refined and estimated before getting pulled in? And added to the sprint?

patrickisaac:
Sweet! Thanks @gdalcengio! I'm going to add your estimate, remove the backlog refinement label and pull it in our current sprint. A reminder that when you do move items from `backlog` or `next sprint backlog` to `In Progress`, be sure to select our current sprint too! Unfortunately, GH doesn't automatically pull things into our current sprint when we move items to `In Progress`

PR #2236, opened 2024-09-20
fix: Remove add facility button for SFO and include check on backed POST to ensure SFO has 1 facility

gdalcengio:
#2105

### DIFF
--- a/bc_obps/service/facility_service.py
+++ b/bc_obps/service/facility_service.py
@@ -182,7 +182,7 @@ class FacilityService:
         cls.check_user_access(user_guid, operation)
 
         # Validate that SFO can only make one facility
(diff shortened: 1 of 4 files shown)


marcellmueller (approved):
Nice work, this is working well! 🥇 

For your consideration: if it's not too much hassle I'd consider adding 1 small backend test to test the updated condition, I think this is a relatively important safeguard. Though I'll leave that up to you.

gdalcengio:
> Nice work, this is working well! 🥇
> 
> For your consideration: if it's not too much hassle I'd consider adding 1 small backend test to test the updated condition, I think this is a relatively important safeguard. Though I'll leave that up to you.

Added! Thanks for bringing my attention to it

PR #2260, opened 2024-09-20
2256 error handling

BCerki:
card: https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=80555060

This PR:
- adds broader error handling to MultiStepBase (previously it was only handling `response.error` errors)
- moved onSubmit error handling from individual form components to MultiStepBase (other error handling, e.g. fetching to populate a dropdown, is still handled in the form components and then the error is passed to MultiStepBase)
- ~move step 1 handling to the MultiStepBase so we don't have to do extra handling in the OperationInformationForm. Step 1 of _any_ workflow will always need to be handled specially because we won't ever have an id before create, but I didn't write a generic case--premature optimization evil and all that~ created new prop for handling step 1
- moves success confirmation to MultiStepBase so that we don't have to do extra error handling in the RegistrationSubmission. Also because a success confirmation is a pretty standard form feature that makes sense in a generic component
- moves isSubmitting logic to MultiStepBase
- removes baseUrl param from our forms because the wireframes don't show a title in the workflow. It's still in the MultiStep component
- Sonarcloud warnings are duplication in test files, I think they're ok to ignore

To test, raise an exception in each endpoint of the registration workflow, and you should see your exception appear on the front end when you hit the endpoint. You can also raise exceptions in `/v2/operations/{uuid:operation_id}` etc. to test that errors in the form components are passed to the multistepbase correctly

Updates after revisions:
- removed showing Submitting... when `isSubmitting` is true--it was only randomly applied, and a loading spinner will be applied in [another ticket ](https://github.com/bcgov/cas-registration/issues/2290)
- no special step 1 handling in MultiStepBase
- success confirmation back to the last step (not necessarily generic to always show a confirmation)
- new prop to optionally put the successful response in state to make it available for the parent component if needed (currently used in first and last steps)
- removed try/catch from MultiStepBase per dev agreement session (let sentry/error boundary handle it)

### DIFF
--- a/bc_obps/registration/api/v2/_operations/_operation_id/_registration/operation.py
+++ b/bc_obps/registration/api/v2/_operations/_operation_id/_registration/operation.py
@@ -1,11 +1,9 @@
 from typing import Literal, Tuple
 from uuid import UUID
 from django.http import HttpRequest
-from registration.schema.v1.operation import OperationUpdateStatusOut
 from registration.schema.v2.operation import (
     OperationInformationIn,
     OperationUpdateOut,
-    OperationRegistrationSubmissionIn,
 )
 from service.operation_service_v2 import OperationServiceV2
 from registration.constants import V2
@@ -34,25 +32,3 @@ def register_edit_operation_information(
     request: HttpRequest, operation_id: UUID, payload: OperationInformationIn
 ) -> Tuple[Literal[200], Operation]:
     return 200, OperationServiceV2.register_operation_information(get_current_user_guid(request), operation_id, payload)
-
-
-@router.patch(
-    "/v2/operations/{uuid:operation_id}/registration/submission",
-    response={200: OperationUpdateStatusOut, custom_codes_4xx: Message},
-    tags=V2,
-    description="""Updates the status of an operation to 'Registered' when the registration is submitted.
-    The endpoint ensures that only authorized industry users can update operations belonging to their operator.""",
-    auth=authorize('approved_industry_user'),
-)
-@handle_http_errors()
-def operation_registration_submission(
-    request: HttpRequest, operation_id: UUID, payload: OperationRegistrationSubmissionIn
-) -> Tuple[Literal[200], Operation]:
-    # Check if all checkboxes are checked
-    if not all(
-        [payload.acknowledgement_of_review, payload.acknowledgement_of_information, payload.acknowledgement_of_records]
-    ):
-        raise Exception("All checkboxes must be checked to submit the registration.")
-    return 200, OperationServiceV2.update_status(
-        get_current_user_guid(request), operation_id, Operation.Statuses.REGISTERED
-    )
--- a/bc_obps/registration/api/v2/_operations/_operation_id/_registration/operation.py
+++ b/bc_obps/registration/api/v2/_operations/_operation_id/_registration/operation.py
@@ -31,4 +31,5 @@ from registration.schema.generic import Message
 def register_edit_operation_information(
     request: HttpRequest, operation_id: UUID, payload: OperationInformationIn
 ) -> Tuple[Literal[200], Operation]:
+    # raise Exception('d')
     return 200, OperationServiceV2.register_operation_information(get_current_user_guid(request), operation_id, payload)
--- a/bciers/apps/registration/app/components/operations/registration/FacilityInformationForm.tsx
+++ b/bciers/apps/registration/app/components/operations/registration/FacilityInformationForm.tsx
@@ -65,7 +65,6 @@ const FacilityInformationForm = ({
   step,
   steps,
 }: FacilityInformationFormProps) => {
-  const [error, setError] = useState(undefined);
   const [formState, setFormState] = useState(formData ?? {});
   const [isSubmitting, setIsSubmitting] = useState(false);
   // Get the list of sections in the LFO schema - used to unnest the formData
@@ -126,15 +125,11 @@ const FacilityInformationForm = ({
             formSectionListLFo,
             operation,
           );
-
+      // errors are handled in MultiStepBase
       const response = await actionHandler(endpoint, method, "", {
         body: JSON.stringify(body),
       });
-      if (!response || response?.error) {
-        setError(response.error);
-        setIsSubmitting(false);
-        return { error: response.error };
-      }
+      return response;
     },
     [operation, isOperationSfo, formSectionListLFo, isCreating, facilityId],
   );
@@ -143,12 +138,10 @@ const FacilityInformationForm = ({
     <MultiStepBase
       allowBackNavigation
       baseUrl={`/register-an-operation/${operation}`}
-      baseUrlParams="title=Placeholder+Title"
       cancelUrl="/"
       formData={formState}
       onChange={handleFormChange}
       onSubmit={handleSubmit}
-      error={error}
       schema={schema}
       step={step}
       steps={steps}
--- a/bciers/apps/registration/app/components/operations/registration/OperationInformationForm.tsx
+++ b/bciers/apps/registration/app/components/operations/registration/OperationInformationForm.tsx
@@ -30,6 +30,7 @@ const OperationInformationForm = ({
   const router = useRouter();
   const [selectedOperation, setSelectedOperation] = useState("");
   const [error, setError] = useState(undefined);
+
   const nestedFormData = rawFormData
     ? createNestedFormData(rawFormData, schema)
     : {};
@@ -71,37 +72,37 @@ const OperationInformationForm = ({
       {
         body,
       },
-    );
-    // errors are handled in MultiStepBase
+    ).then((resolve) => {
+      if (resolve?.error) {
+        return { error: resolve.error };
+      } else if (resolve?.id) {
+        // this form step needs a custom push (can't use the push in MultiStepBase) because the resolve.id is in the url
+        const nextStepUrl = `/register-an-operation/${resolve.id}/${step + 1}`;
+        router.push(nextStepUrl);
+        return resolve;
+      }
+    });
     return response;
   };
   const handleSelectOperationChange = async (data: any) => {
     const operationId = data.section1.operation;
-    try {
-      setSelectedOperation(operationId);
-      const operationData = await getOperationV2(operationId);
-      if (operationData?.error) {
-        setError("Failed to fetch operation data!" as any);
-      }
-      // combine the entered data with the fetched data
-      const combinedData = { ...data, section2: operationData };
-      setFormState(combinedData);
-      setKey(Math.random());
-    } catch (err) {
+    setSelectedOperation(operationId);
+    const operationData = await getOperationV2(operationId);
+    if (operationData?.error) {
       setError("Failed to fetch operation data!" as any);
     }
+    // combine the entered data with the fetched data
+    const combinedData = { ...data, section2: operationData };
+    setFormState(combinedData);
+    setKey(Math.random());
   };
+
   return (
     <MultiStepBase
       key={key}
       cancelUrl="/"
       formData={formState}
       onSubmit={handleSubmit}
-      firstStepExtraHandling={(response) => {
-        // Since our form's route includes the operation's id, which doesn't exist until after the first step, we need to pass in a custom function that uses the response to generate a redirect url
-        const nextStepUrl = `/register-an-operation/${response.id}/${step + 1}`;
-        router.push(nextStepUrl);
-      }}
       schema={schema}
       step={step}
       steps={steps}
--- a/bciers/apps/registration/app/components/operations/registration/OperationInformationForm.tsx
+++ b/bciers/apps/registration/app/components/operations/registration/OperationInformationForm.tsx
@@ -99,9 +99,7 @@ const OperationInformationForm = ({
       onSubmit={handleSubmit}
       firstStepExtraHandling={(response) => {
         // Since our form's route includes the operation's id, which doesn't exist until after the first step, we need to pass in a custom function that uses the response to generate a redirect url
-        const nextStepUrl = `/register-an-operation/${response.id}/${
-          step + 1
-        }$`;
+        const nextStepUrl = `/register-an-operation/${response.id}/${step + 1}`;
         router.push(nextStepUrl);
       }}
       schema={schema}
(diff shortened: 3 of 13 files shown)


bc_obps/registration/api/v2/_operations/_operation_id/_registration/operation.py
```diff
@@ -36,41 +32,3 @@ def register_edit_operation_information(
     request: HttpRequest, operation_id: UUID, payload: OperationInformationIn
 ) -> Tuple[Literal[200], Operation]:
     return 200, OperationServiceV2.register_operation_information(get_current_user_guid(request), operation_id, payload)
-
-
-@router.put(
```
BCerki, line 41: These endpoints were duplicated, cleaning them up in this PR

bciers/apps/registration/app/components/operations/registration/OperationInformationForm.tsx
```diff
@@ -28,7 +28,7 @@ const OperationInformationForm = ({
   steps,
 }: OperationInformationFormProps) => {
   const [selectedOperation, setSelectedOperation] = useState("");
-  const [, setError] = useState(undefined);
+  const [error, setError] = useState(undefined);
```
BCerki, line 31: The component sets errors in handleSelectOperationChange

bciers/apps/registration/app/components/operations/registration/FacilityInformationForm.tsx
```diff
@@ -143,12 +138,10 @@ const FacilityInformationForm = ({
     <MultiStepBase
       allowBackNavigation
       baseUrl={`/register-an-operation/${operation}`}
-      baseUrlParams="title=Placeholder+Title"
```
BCerki, line 146: The wireframes don't show anything besides `Register an Operation` in the breadcrumbs, so we don't need a title placeholder atm

bciers/apps/registration/app/components/operations/registration/RegistrationSubmissionForm.tsx
```diff
@@ -46,43 +44,33 @@ const RegistrationSubmissionForm = ({
...
-        />
-      )}
+      <MultiStepBase
+        allowBackNavigation
+        baseUrl={`/register-an-operation/${operation}`}
+        cancelUrl="/"
+        formData={formState}
+        onSubmit={handleSubmit}
+        schema={
+          isSubmitting
```
BCerki, line 60: I'm considering moving the `isSubmitting` logic to MultiStepBase. Some forms have it at the moment and some don't. Thoughts?

bciers/apps/registration/app/components/operations/registration/OptedInOperationForm.tsx
```diff
@@ -55,20 +54,15 @@ const OptedInOperationForm = ({
...
   return (
     <MultiStepBase
       allowBackNavigation
       baseUrl={`/register-an-operation/${operation}`}
-      baseUrlParams="title=Placeholder+Title"
       cancelUrl="/"
-      error={error}
       formData={formState}
       onChange={handleChange}
       onSubmit={handleSubmit}
```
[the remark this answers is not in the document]
marcellmueller: Well I think this is a good idea to handle internally either way so we don't have to do this every time we reuse MultiStepBase. If you look inside `FormBase` we also handle it there to stop form data reset on submit. Though we may also need to have `isSubmitting` state in components where we use `MultiStepBase` for other reasons.

In regards to not showing the read only schema I think the comment could have been more helpful here though I just tested it and I think this is outdated. Before the rich text field titles were added, when you submitted the form it looked very buggy. The titles would disappear and you would just see a list of `No` momentarily, so we added that to stop that behaviour. At the time it was a nice improvement.

Though I just tried it again and that is fixed. Seeing it shift to the read only form still isn't ideal, though something that may be fixed if we remove the read only schema and the radio buttons will just be disabled.

bciers/libs/components/src/form/MultiStepBase.tsx
```diff
@@ -54,29 +55,54 @@ const MultiStepBase = ({
...
+        const nextStepUrl = `${baseUrl}/${step + 1}${
+          baseUrlParams ? `?${baseUrlParams}` : ""
+        }`;
+        router.push(nextStepUrl);
+        return;
+      }
+      // Last step
+      if (!isNotFinalStep) {
+        setIsSuccessfullySubmitted(true);
+      }
```
marcellmueller, line 100: This is supposed to be a general use form base component, we shouldn't have routes such as `register-an-operation` hardcoded in it. I get that this is trying to reduce duplication but I find this kind of confusing and it makes this component less reusable. I'm definitely pro automating a lot of this though some things make sense to handle in each form component.

Suggest keeping `baseUrl` prop required and remove most of this refactor except perhaps the last step block.

bciers/apps/registration/app/components/operations/registration/OptedInOperationForm.tsx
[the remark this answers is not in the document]
BCerki: I was able to move all the `isSubmitting` logic to MultiStepBase. This revealed some weird behaviour when switching steps: https://www.loom.com/share/99f835e99463426fb3a855205c7b43d9. I'll make a bug card

bciers/libs/components/src/form/MultiStepBase.tsx
BCerki: As discussed, the first step does need to be handled differently, I went with a prop.

bciers/libs/components/src/form/MultiStepBase.tsx
```diff
@@ -86,7 +109,11 @@ const MultiStepBase = ({
     setIsEditMode(true);
   };
 
-  return (
+  return isSuccessfullySubmitted ? (
+    successComponent
+  ) : isSubmitting ? (
+    <div>Submitting...</div>
+  ) : (
```
marcellmueller, line 116: Why are we doing this in here? It feels like we are trying to solve problems that don't exist, the MultiStepBase component is starting to look overly complicated and hard to follow. Now that we have split up the forms into individual pages we can keep logic specific to those pages in those components.

This also dictates that we have to display a different component on submit (we may want to redirect to a page or do something else).

bciers/apps/registration/app/components/operations/registration/OperationInformationForm.tsx
```diff
@@ -97,9 +97,15 @@ const OperationInformationForm = ({
       cancelUrl="/"
       formData={formState}
       onSubmit={handleSubmit}
+      firstStepExtraHandling={(response) => {
+        // Since our form's route includes the operation's id, which doesn't exist until after the first step, we need to pass in a custom function that uses the response to generate a redirect url
+        const nextStepUrl = `/register-an-operation/${response.id}/${step + 1}`;
+        router.push(nextStepUrl);
+      }}
```
marcellmueller, line 104:
```suggestion
      onSubmit={async (data: any) => {
        const response = await handleSubmit(data);
        const nextStepUrl = `/register-an-operation/${response.id}/${step + 1}`;
        router.push(nextStepUrl);
      }}
```

bciers/libs/components/src/form/MultiStepBase.tsx
```diff
@@ -54,29 +56,50 @@ const MultiStepBase = ({
...
+        return;
+      }
+      // Last step
+      if (!isNotFinalStep) {
+        setIsSuccessfullySubmitted(true);
+      }
+    } catch (err) {
+      // Handle any errors beyond response.error
+      setError("An unexpected error occurred. Please try again.");
+    } finally {
```
marcellmueller, line 101:
```suggestion
  const submitHandler = async (data: any) => {
    // Set isSubmitting to true to disable submit buttons and prevent multiple form submissions
    setIsSubmitting(true);
    // Clear any old errors
    setError(undefined);
    try {
      const response = await onSubmit(data);
      if (response?.error) {
        setError(response?.error);
        return response;
      }

      if (baseUrl) {
        const nextStepUrl = `${baseUrl}/${step + 1}${
          baseUrlParams ? `?${baseUrlParams}` : ""
        }`;
        router.push(nextStepUrl);
        return response;
      }
      return response;
    } catch (err) {
      // Handle any errors beyond response.error
      setError("An unexpected error occurred. Please try again.");
    } finally {
      setIsSubmitting(false);
    }
  };
```

For your consideration: I think this is getting overly complicated and hard to follow and will be difficult to maintain in the future. I think we should just not handle navigation if the `baseUrl` doesn't exist, and return the response to the parent component if they need that information.

marcellmueller (commented):
Great work on this PR, there are a lot of solid improvements! I do have some concerns about a couple of additions though and have added some suggestions. 

Though I'm mindful that there may have been some issues that I've missed with my suggestions since I haven't been immersed in this as you have. Happy to hop on a call and discuss it! Let me know what you think, thanks :)

bciers/libs/components/src/form/MultiStepBase.tsx
BCerki: > Now that we have split up the forms into individual pages we can keep logic specific to those pages in those components.

There isn't much logic that actually is specific to individual pages, which is why I've moved some things into MultiStepBase. Error handling and `isSubmitting` apply to all form pages. Showing a loading message `<div>Submitting...</div>` while `isSubmitting` is true also seemed general and like it should apply to all steps (previously it was on about half of them, and not the ones I'd expect like the ones with attachments).

Re `successComponent`, previously it was conditionally shown in the last step based on the value of `isSubmitting`. Since `isSubmitting` is now being managed by MultiStepBase, and showing a success message is a general form submission thing, it makes sense to me to have it in MultiStepBase. It could be a redirect to a different page instead, but this was how it was originally set up so I didn't change it.

bciers/libs/components/src/form/MultiStepBase.tsx
BCerki: > return the response to the parent component if they need that information
What do you mean by this? I've only ever done this with state, and that's not working here (maybe because it takes too long to get the response?) Or do you mean like your suggestion above, which does work:
```
onSubmit={async (data: any) => {
        const response = await handleSubmit(data);
        const nextStepUrl = `/register-an-operation/${response.id}/${step + 1}`;
        router.push(nextStepUrl);
      }}
```

marcellmueller: Yeah that's what I meant, in your MultiStepBase changes we just do a plain return. In the suggestion I made we `return reponse` and can access it in the parent component as shown here.

bciers/libs/components/src/form/MultiStepBase.tsx
marcellmueller: Showing `Submitting...` in place of the entire component is really jarring, and especially noticeable after page 1 with file uploads. I believe we were asked about adding a loading spinner to the `Save and continue` button.

Looking at the `RegistrationSubmissionForm` component it is really clean and simple, I don't see the need to further complicate `MultiStepBase`. We can await the `onSubmit` response like in the example and `setIsSubmitted` and do what we need to in that component if `reponse.ok` or whatever.

bciers/libs/components/src/form/MultiStepBase.tsx
marcellmueller: Sorry I think I misunderstood you, though I think we can pass `handleSubmit` and to a `.then` to get the response from in the parent?

bciers/libs/components/src/form/MultiStepBase.tsx
```diff
@@ -54,30 +56,35 @@ const MultiStepBase = ({
...
-    // and allow user to attempt to re-submit the form
+    const response = await onSubmit(data);
     if (response?.error) {
-      setIsSubmitting(false);
-    } else if (isNotFinalStep && baseUrl && baseUrlParams) {
+      setError(response?.error);
+      return;
+    }
+
+    if (setOnSubmitSuccessfulResponse) setOnSubmitSuccessfulResponse(response);
```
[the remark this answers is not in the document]
marcellmueller: Why can't we use a useEffect without the extra prop? We should always have access to the response in the parent component.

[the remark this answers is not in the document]
BCerki: > We should always have access to the response in the parent component.
How do we have access to this without state? We need the response from the `MultiStepBase`'s `onSubmit` (not the response from the parent's, because that doesn't include error handling)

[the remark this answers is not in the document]
marcellmueller: Well we had it before this refactor, so currently this seems like a step down. There are many reasons why we would want the response in the parent (ie page 1 and submission). I'm working on a POC where we get the response in the parent component, though while I get a successful response I'm not getting the errors for some reason. Will let you know if I figure this out!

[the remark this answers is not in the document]
marcellmueller: Hey! I created a simple POC for the submission page with console logs showing that we can access the response in the MultiStepBase and the parent component. Let me know what you think, perhaps there is something I didn't consider. Would love to get on a call and pair on this later :)

https://github.com/bcgov/cas-registration/pull/2292/files

bciers/apps/registration/app/components/operations/registration/OperationInformationForm.tsx
```diff
@@ -71,26 +72,32 @@ const OperationInformationForm = ({
...
         body,
       },
-    );
-    // handling the redirect here instead of in the MultiStepBase because we don't have the operation information until we receive the response
-    const nextStepUrl = `/register-an-operation/${response.id}/${
-      step + 1
-    }${`?title=${response.name}`}`;
-    router.push(nextStepUrl);
+    ).then((resolve) => {
+      console.log("!!!!!resolve", resolve);
```
marcellmueller, line 76: ✂️

bciers/apps/registration/app/components/operations/registration/RegistrationSubmissionForm.tsx
```diff
@@ -63,19 +63,10 @@ const RegistrationSubmissionForm = ({
         <MultiStepBase
           allowBackNavigation
           baseUrl={`/register-an-operation/${operation}`}
```
marcellmueller, line 65: Is `baseUrl` being used for the final page?

marcellmueller (approved):
Awesome work, thanks for doing this and being receptive to my pestering review suggestions 😅 I enjoyed the back and forth and think this is a solid improvement! 

Just a couple clean up comments!

bciers/apps/registration/app/components/operations/registration/RegistrationSubmissionForm.tsx
BCerki: Nope!

PR #2232, opened 2024-09-19
Story/facilities grid add status/2018

andrea-williams:
card: https://github.com/orgs/bcgov/projects/122/views/2?filterQuery=2018&pane=issue&itemId=72458135

PR #2286, opened 2024-09-25
Add status column to facility grid

BCerki:
https://github.com/orgs/bcgov/projects/123/views/16?pane=issue&itemId=80445790

This PR:
- Continues the work started in https://github.com/bcgov/cas-registration/pull/2232. I made a new PR because I was doing experimental troubleshooting on filtering
- Previously the facility grid showed data from the `Facility` model. Now it shows data from `FacilityDesignatedOperationTimeline`
- Because we're filtering on stuff that's not in `FacilityDesignatedOperationTimeline` (like facility name), the filter names changed
- Fixed some typos in tests
- New bakers
- We're no longer using some of `facility_service` because nowhere in the app currently requires a list of all facilities (the grids are all the facilities belonging to a specific operation). However, I left the service functions in because we'll use them when we build cas_admin's facility tile
- Tests for services and endpoints (the original tests had most of the sorting/filtering in endpoints, I left them there)
- Vitests for grids


bc_obps/registration/models/facility_designated_operation_timeline.py
```diff
@@ -23,6 +35,7 @@ class FacilityDesignatedOperationTimeline(TimeStampedModel):
     class Meta:
         db_table_comment = "A table to connect facilities and operations"
         db_table = 'erc"."facility_designated_operation_timeline'
+        # brianna how is status set?
```
BCerki, line 38: Not really related to this PR, but this constraint looks like it checks that only one facility can have a null end date. It doesn't seem tied to status though (unless I'm missing something? a trigger somewhere?), so we could have multiple facilities showing a status of `Active` without throwing a db error

andrea-williams: you are correct @Sepehr-Sobhani ! And thank you for mocking me - I still haven't caught up on all the emails I missed while I was out sick.

Sepehr-Sobhani, line 38: leftover!

bc_obps/registration/schema/v1/facility_designated_operation_timeline.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class FacilityDesignatedOperationTimelineFilterSchema(FilterSchema):
+    # NOTE: we could simply use the `q` parameter to filter by name and ... but,
+    # due to this issue: https://github.com/vitalik/django-ninja/issues/1037 mypy is unhappy so I'm using the `json_schema_extra` parameter
+    # If we want to achieve more by using the `q` parameter, we should use it and ignore the mypy error
+    facility__bcghg_id: Optional[str] = Field(None, json_schema_extra={'q': 'facility__bcghg_id__icontains'})
+    facility__name: Optional[str] = Field(None, json_schema_extra={'q': 'facility__name__icontains'})
+    facility__type: Optional[str] = Field(None, json_schema_extra={'q': 'facility__type__icontains'})
+    status: Optional[str] = Field(None, json_schema_extra={'q': 'status__icontains'})
```
Sepehr-Sobhani, line 23: I believe using double underscores (dunders) is very specific to Python. Since we are working with these value on the front end, I recommend using a single underscore. Alternatively, we can use camelCase like Giraffe does, but it's not consistent with the approach we've been using in Moose from the beginning.

bc_obps/service/facility_designated_operation_timeline_service.py
```diff
@@ -0,0 +1,44 @@
...
+
+class FacilityDesignatedOperationTimelineService:
+    @classmethod
+    def get_timeline_by_operation_id(
+        cls, user: User, operation_id: UUID
+    ) -> QuerySet[FacilityDesignatedOperationTimeline]:
+        if user.is_industry_user():
+            UserOperatorService.get_current_user_approved_user_operator_or_raise(user)
+        return (
+            FacilityDesignatedOperationTimeline.objects.all()
```
Sepehr-Sobhani, line 23: nit: no need to use `.all()`

bc_obps/service/tests/test_facility_designated_operations_timeline_service.py
```diff
@@ -0,0 +1,94 @@
...
+            irc_user, timeline_for_selected_operation[0].operation.id
+        )
+
+        assert expected_facilities.count() == 10
+
+    @staticmethod
+    def test_get_timeline_by_operation_id_for_unapproved_user_industry_user():
+        industry_user = baker.make_recipe('utils.industry_operator_user')
+        facility_designated_operation_timeline = baker.make_recipe('utils.facility_designated_operation_timeline')
+        with pytest.raises(Exception):
```
Sepehr-Sobhani, line 32: nit: A good practice would be to add the exception error as the second argument of the `.raises` method, like this: `pytest.raises(Exception, match="<error message>")`

Sepehr-Sobhani (approved):
Looks great 🤩 Requested a minor change, but I’m marking the PR as approved so you won’t be blocked.

bc_obps/registration/schema/v1/facility_designated_operation_timeline.py
BCerki: It looks like we need the dunders for the filter schema to work. If I change to a single underscore on the FE, then the URL params end up with a single underscore and filtering breaks. I could maybe change some of the params logic to add the underscore back in, but I think that might be more confusing. Thoughts?

Sepehr-Sobhani: Filter schema only needs the dunder for `json_schema_extra` (dunder is the way Django reaches the related object), but I think we can change the name to whatever we want, like `whatever_name: Optional[str] = Field(None, json_schema_extra={'q': 'facility__name__icontains'}). 
If this is not the case, please ignore my comment :)

BCerki: That's what I had thought too, but it wasn't working :( Thanks for your help!

PR #2288, opened 2024-09-25
chore: update auth.config call to user-profile

shon-button:
### 🚀 Impact:
- Updated auth.config to limit API call to `user-profile` only if token is missing `full_name`

### 🔬 Local Testing:

 ```
cd bc_obps
make reset_db
make run
```  
 ```
cd client && yarn dev-all

```

3. Verify the `bc_obps` terminal has correct number of calls to `user-profile` from the auth.config 
BEFORE
image
AFTER
image

### DIFF
--- a/bciers/apps/dashboard/auth/auth.config.ts
+++ b/bciers/apps/dashboard/auth/auth.config.ts
@@ -100,16 +100,18 @@ export default {
 
           token.identity_provider = account.providerAccountId.split("@")[1];
         }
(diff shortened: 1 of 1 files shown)


marcellmueller (approved):
Thanks, this is a solid improvement! I went from 9 calls to the backend to 5 loading the BCIERS dashboard. Though it's interesting that we get different results sometimes between runs/users 🤔

Either way happy for this improvement, thanks for the quick PR! 😃

PR #2292, opened 2024-09-26
poc error

marcellmueller:
poc

To test this go through the Registration workflow and look at the browser console after submitting page 1 `Operation Information` and the final `Registration Submission pages`. We are throwing an error on the submission page in the API that can be removed if you want to test successful submission.

Again, I might be missing something here so happy to jump on a call and discuss or else hear any other ideas/criticisms of this! Thanks.


BCerki:
Heya, thanks for this! I was trying to avoid any error handling anywhere except MultiStepBase, but what you've got here is just a couple lines and I think that works better than useEffect. I'll pull some of this code over into the PR later today

PR #2300, opened 2024-09-27
Fix e2e on merge to develop

marcellmueller:
Our new BCIERS e2e tests run fine in pull requests but Happo shows they aren't connecting to the backend on merge to develop. This was causing a back and forth of Happo failures - merging to develop created a new erroneous baseline while we would have to approve the correct ones to merge to main. It seems a bit counter intuitive but I think that is what was causing diffs appear in our PRs but be 'auto' approved - because they were the same diffs reviewed/approved in the previous PR that merged. 

I am somewhat certain the Happo issue will right itself after we merge to main and create a proper baseline there. Though not 100%. However during testing I confirmed if I create actual new diffs (I put some new text on the dashboard) they will show up with the red X as expected.

This took me ages to figure out but for some unknown reason it seems like the runners are slower on GitHub push (which is a merge) over on pull request. I've added health checks to our docker compose backend containers that check that `http://127.0.0.1:8000/api` is available before moving to the next step.

Also in this PR I have removed the e2e matrix stuff which allows us to remove the Happo finalize job that ties runs together. This will save quite a bit of time - about a couple of minutes for each run. 

For reviewing purposes I left some testing code in I will remove before main. As you can see in the Happo screens each run is connecting to the db as expected and showing each dashboard.

### Reminders for me before merge:
- Remove `test-happo-diffs` branch from `main.yaml` and uncomment `pull_request`
- Remove console.log in `/components/src/index.ts` which triggers nx affected for all
- Update GitHub required status checks (remove Happo finalize)

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -4,9 +4,9 @@ name: main
 
 on:
   push:
-    branches: [develop, main]
-  pull_request:
-    branches: [develop, main]
+    branches: [develop, main, test-happo-diffs]
+  # pull_request:
+  #   branches: [develop, main]
   workflow_dispatch:
 
 concurrency:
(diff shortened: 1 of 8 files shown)


.github/workflows/main.yaml
```diff
@@ -4,9 +4,9 @@ name: main
 
 on:
   push:
-    branches: [develop, main]
-  pull_request:
-    branches: [develop, main]
+    branches: [develop, main, test-happo-diffs]
+  # pull_request:
+  #   branches: [develop, main]
```
marcellmueller, line 9: Just for testing/reviewing purposes, will undo before merging thank you 🙇

bciers/libs/components/src/index.ts
```diff
@@ -6,3 +6,5 @@ export { default as Bread } from "./navigation/Bread";
 export * from "./theme/ThemeRegistry";
 export * from "./theme/EmotionCache";
 export * from "./theme/theme";
+
+console.log("trigger nx affected for ci");
```
marcellmueller, line 10: Just for testing/reviewing purposes, will undo before merging thank you 🙇

Issue #2203, opened 2024-09-13
Can't submit the facility form for an LFO [8]


BCerki:
Describe the Bug:

When you click "Save and Continue," the button goes grey but you're never redirected to the next step: https://www.loom.com/share/dab33f8a692f470599df2bd106433e26. Operation 2 is our mock LFO.

SFOs seem to work fine.

Probability (how likely the bug is to happen, scored from 1-5): 5?
(For example, probability of 5 is something like "it happens to all users every time they log in."
Probability of 1 "only happens to certain users when a really specific and unlikely path is followed.")

Effect (how bad the bug is when it does happen, scored from 1-5): 3
(For example, effect of 5 is "the entire app crashes and makes it unusable for all users" or "the bug causes the wrong data to be saved, with critical information (e.g. payment) being affected."
Effect of 1 is "It makes some styling look a little bit weird.")

Steps to reproduce the behaviour:

Screenshots:

Additional information:

- Browser: [e.g. chrome, safari]
- Version: [e.g. 22]
- Additional context

marcellmueller:
I looked into this because it was previously working and it's the small bugfix for the DataGrid empty rows that caused this to break. I'm not exactly why this one line fix caused it though from my work on the Facility DataGrid issue I know that the `router.push` in the form page change sometimes causes the DataGrid to refresh due to the query param change and stops the page transition.

PR #2319, opened 2024-10-01
fix/facilty lfo save and continue

marcellmueller:
#2203 

This was a strange one to debug as sometimes it would work fine repeatedly, others it would fail every time. I'm still not entirely sure of what is causing it, though I have a fix (I think 🤞 ).

The easiest way to replicate the bug is:
- Go to Register an Operation form
- Fill out the form and select `Operation 2` since it is an LFO
- Click Save and Continue
- Page 2 of the form should show the Facility DataGrid since it's an LFO 
- This part was important for me to reproduce it _most_ times - open browser dev tools (I used Chromium), click Network tab -> throttle -> slow3g

Screenshot 2024-10-01 at 4 03 54 PM

- Click Save and Continue
- Most of the time you will remain on the Facility Information page 😢 

To test this fix follow the same steps, you should be taken to the next page of the form every time.

### DIFF
--- a/bciers/apps/registration/app/components/operations/registration/FacilityInformationForm.tsx
+++ b/bciers/apps/registration/app/components/operations/registration/FacilityInformationForm.tsx
@@ -125,11 +125,11 @@ const FacilityInformationForm = ({
             formSectionListLFo,
             operation,
           );
-      // errors are handled in MultiStepBase
       const response = await actionHandler(endpoint, method, "", {
         body: JSON.stringify(body),
       }).then((resolve) => {
         if (resolve?.error) {
+          // errors are handled in MultiStepBase
           return { error: resolve.error };
         } else {
           // Using window.location.href instead of MutliStepBase router.push as there were rerender issues due to
--- a/bciers/apps/registration/app/components/operations/registration/FacilityInformationForm.tsx
+++ b/bciers/apps/registration/app/components/operations/registration/FacilityInformationForm.tsx
@@ -128,7 +128,18 @@ const FacilityInformationForm = ({
       // errors are handled in MultiStepBase
       const response = await actionHandler(endpoint, method, "", {
         body: JSON.stringify(body),
+      }).then((resolve) => {
+        if (resolve?.error) {
+          return { error: resolve.error };
+        } else {
+          // Using window.location.href instead of MutliStepBase router.push as there were rerender issues due to
+          // Facility Grid listerning to params change on this page which sometimes hijacked the route change
+          window.location.href = `${
+            window.location.origin
+          }/registration/register-an-operation/${operation}/${step + 1}`;
+        }
       });
+
       return response;
     },
     [operation, isOperationSfo, formSectionListLFo, isCreating, facilityId],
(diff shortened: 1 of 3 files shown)


bciers/apps/registration/app/components/operations/registration/FacilityInformationForm.tsx
```diff
@@ -128,7 +128,18 @@ const FacilityInformationForm = ({
...
+        if (resolve?.error) {
+          return { error: resolve.error };
+        } else {
+          // Using window.location.href instead of MutliStepBase router.push as there were rerender issues due to
+          // Facility Grid listerning to params change on this page which sometimes hijacked the route change
+          window.location.href = `${
+            window.location.origin
+          }/registration/register-an-operation/${operation}/${step + 1}`;
+        }
       });
```
marcellmueller, line 141: Generally it's always preferable to use Next routing for internal links however in this case it fixed routing issues so I think it's okay to use on this one special page!

bciers/apps/registration/app/components/operations/registration/FacilityInformationForm.tsx
```diff
@@ -128,7 +128,18 @@ const FacilityInformationForm = ({
       // errors are handled in MultiStepBase
```
BCerki, line 128:
```suggestion
```
Since it looks like you've moved error handling here

BCerki (commented):
Fix works, thanks! I suggest adding an assertion in `apps/registration/tests/components/operations/registration/FacilityInformationForm.test.tsx` (or the page component if that makes more sense) to make sure the redirect fires

bciers/apps/registration/app/components/operations/registration/FacilityInformationForm.tsx
marcellmueller: Thanks, I just moved this comment to inside the error block. I didn't move error handling here, but it's important to check for an error and return it to MultiStepBase to handle if one occurs.

bciers/apps/registration/tests/components/operations/registration/FacilityInformationForm.test.tsx
```diff
@@ -435,4 +435,43 @@ describe("the FacilityInformationForm component", () => {
...
+    window = Object.create(window);
+    const origin = "http://localhost:3000";
+    const path = `/registration/register-an-operation/002d5a9e-32a6-4191-938c-2c02bfec592d`;
+    Object.defineProperty(window, "location", {
+      value: {
+        href: `${origin}${path}/2`,
+        origin,
+      },
+      writable: true,
+    });
```
marcellmueller, line 449: I ran into difficulties spying on `window.location.href` specifically (window.location.get) seems to work fine so I ended up with this.

bciers/libs/components/src/form/MultiStepBase.test.tsx
```diff
@@ -246,7 +246,7 @@ describe("The MultiStepBase component", () => {
 
     await user.click(saveAndContinueButton);
 
-    waitFor(() => {
+    await waitFor(() => {
```
marcellmueller, line 249: 🤞

bciers/apps/registration/tests/components/operations/registration/FacilityInformationForm.test.tsx
BCerki: Wow, that's a lot more complicated than spying on the router, thanks for figuring it out

BCerki (approved):

Issue #353, opened 2023-11-29
Change `UserOperator` data model id to UUID


Sepehr-Sobhani:
###Description of the Tech Debt###

In the route specified below, the `user_operator_id` created is returned to the front-end for use as a query parameter. However, exposing sensitive data in query parameters is not considered a secure practice. Therefore, it is advisable to enhance security by altering the ID of the `UserOperator` data model to a UUID format. This modification helps prevent potential security vulnerabilities or attacks.

`@router.post("/user-operator/operator", response={200: RequestAccessOut, codes_4xx: Message})`

###Tech Debt Triage#
The purpose of our technical debt triage process is to analyze technical debt to determine risk level of the technical debt and the value in tackling that technical debt.

Risk Value Scoring:

| Level  | Value |
| ------ | ----- |
| High   | 3     |
| Medium | 2     |
| Low    | 1     |

| Technical Debt - Risk Types                                                                                                   | Level | Value |
| ----------------------------------------------------------------------------------------------------------------------------- | ----- | ----- |
| Business Area Risk - Risk of business area visibility / damage to user experience                                             |       |     1  |
| Developer Fault Risk - How likely will this tech debt cause a future error related to coding on top of it                     |       |       1|
| System Fault Risk - Risk of system errors or application downtime                                                             |       |    1   |
| Time Scale Risk - Compound risk effect if left alone. How much more difficult to fix or dangerous will this become over time? |       |   1    |
| Time Sink Risk - How much will this tech debt slow the development process down                                               |       |    1   |
| TOTAL SCORE:                                                                                                                  |       |   5    |

Sepehr-Sobhani:
Addressed in: https://github.com/bcgov/cas-registration/pull/939

PR #2371, opened 2024-10-15
353 methodology fields not flat with emission

acatchpole:
Card = reporting [#353](https://github.com/bcgov/cas-reporting/issues/353)

 - Updated form_builder_service to have methodology field/data nested inside its own object
 - Updated carbonates_use.json : removed methodology fields from the schema as it is getting added by the form builder
 - Updated activity uiSchemas to account for the new schema structure. Also addressed errors that were showing from some fuelTypes fields not matching between schema and uiSchema.

### Testing
To test locally: Go to -> http://localhost:3000/reporting/reports/6/facilities/f486f2fb-62ed-438d-bb3e-0819b51e3aff/review and make sure that the relevant existing activities are selected:
 - ALL activities dealing with ```General stationary combustion```
 - ```Fuel combustion by mobile equipment```
 - ```Carbonate use```
 - ```Hydrogen production```
 - ```Pulp and paper production```
 - ```Refinery fuel gas combustion```

Click ```Save & Continue```

For each of the activities, see that:
 - the form renders as expected. 
 - dependant fields appear as expected
 - there are no errors related to uischema
 - if you put some mock data in and submit, all methodology-related fields will be nested inside a methodology object.

### DIFF
--- a/bc_obps/reporting/json_schemas/2024/carbonates_use/carbonates_use.json
+++ b/bc_obps/reporting/json_schemas/2024/carbonates_use/carbonates_use.json
@@ -9,16 +9,6 @@
         "title": "Carbonate data",
         "type": "object",
         "properties": {
-          "methodology": {
-            "title": "Methodology",
-            "type": "string",
-            "enum": [
-              "Calcination Fraction",
-              "Mass of Output Carbonates",
-              "Alternative Parameter Methodology",
-              "Replacement Methodology"
-            ]
-          },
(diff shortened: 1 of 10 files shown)


dleard (approved):
Really great work! Exactly what was needed. :100: 
@pbastia Once this has been merged we can update the save service to properly send the methodology data to the report_methodology table. I've added a card to our backlog for this

PR #2373, opened 2024-10-15
fix: activity form state should not persist when switching activities

dleard:
The formState was being persisted when navigating to different activity pages.
The dependency array for the useEffect() was throwing warnings across different activity because it was changing its shape depending on how many source types each activity has.

This PR should fix those issues. Though I think that useEffect() is probably doing too much & we should look into potentially splitting it up into different components. I think the way it is set up currently may have an effect on the [Load data ticket](https://github.com/bcgov/cas-reporting/issues/345)


pbastia (approved):
Nice!
I agree with your comment on the useEffect doing too much. I think there are quite a few cases where we trigger a re-render no matter what, so that code could be taken out of the effect/dependencies array. Let's revisit that with the loading data feature

PR #2382, opened 2024-10-17
chore:added sync button

ayeshmcg:
Card: https://github.com/bcgov/cas-reporting/issues/364
Changes:-

- Added Sync button
- Added Sync button handler

### DIFF
--- a/bciers/apps/reporting/src/data/jsonSchema/personResponsible.ts
+++ b/bciers/apps/reporting/src/data/jsonSchema/personResponsible.ts
@@ -1,7 +1,10 @@
 import { TitleOnlyFieldTemplate } from "@bciers/components/form/fields";
 import FieldTemplate from "@bciers/components/form/fields/FieldTemplate";
 import { RJSFSchema, UiSchema } from "@rjsf/utils";
-import { infoNote } from "@reporting/src/data/jsonSchema/personResponsibleInfoText";
+import {
+  infoNote,
+  SyncContactsButton,
+} from "@reporting/src/data/jsonSchema/personResponsibleInfoText";
 import SectionFieldTemplate from "@bciers/components/form/fields/SectionFieldTemplate";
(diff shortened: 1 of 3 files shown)


bciers/apps/reporting/src/data/jsonSchema/personResponsible.ts
```diff
@@ -44,6 +51,9 @@ export const personResponsibleUiSchema: UiSchema = {
   assign_info: {
     "ui:FieldTemplate": SectionFieldTemplate,
   },
+  sync_button: {
+    "ui:FieldTemplate": SyncContactsButton,
```
pbastia, line 55: the RJSF docs do a good job of what belongs in a Field, a Template, and a Widget: https://rjsf-team.github.io/react-jsonschema-form/docs/advanced-customization/custom-templates/

Using a widget to encapsulate rendering + behaviour is probably much easier. If needed the sync handler can be passed in to a constructor for the widget (something like `"ui:widget":SyncContactsWidget(onSync)`), but there _might_ be a way of making it act within RJSF the whole time

pbastia (approved):
Nice work! I'm approving it, but we're defintiely mis-using the `ui:options` field to pass behaviour instead of just rendering options.
The comments I've added can be put in a tech debt card

bciers/apps/reporting/src/data/jsonSchema/personResponsible.ts
ayeshmcg: @pbastia 
Yes, I agree, and I will open a tech-debt ticket to address this. The ultimate plan was to encapsulate the SyncButton in a widget, which will simplify the implementation and avoid misusing ui:options for passing handler functions. This approach will also be helpful as we need to add the SyncButton in another location (the Operation Review Page), so creating a widget will make this more scalable and maintainable.

Thanks for bringing this up!

PR #2386, opened 2024-10-17
Feat/208 additional report data page

ayeshmcg:
Card: https://github.com/bcgov/cas-reporting/issues/208
Changes: 
- Added Additional Reporting Data Page: New page for collecting additional reporting data.
- Schema and UI Schema: Defined schemas to validate and structure the new data.
- APIs/Services:
   - Get Registration Purpose: Fetch registration context.
   - Save Report Additional Data: Handle data submission.
   - Get Report Additional Data: Retrieve saved data for review.
- Model for Additional Reporting Data: Created a model for data storage and retrieval.

To test:- go to http://localhost:3000/reporting/reports/3/additional-reporting-data
to test Electricity generated field, click on Operation 17 on http://localhost:3000/reporting/reports


bc_obps/service/report_additional_data.py
```diff
@@ -0,0 +1,47 @@
...
+
+        # Always return the result as a dictionary with an array (empty or populated)
+        return {"registration_purposes": purposes_list}
+
+    @staticmethod
+    def save_report_additional_data(version_id: int, data: ReportAdditionalDataIn) -> ReportAdditionalData:
+        # Fetch the report version
+        report_version = ReportVersion.objects.filter(id=version_id).first()
+        if report_version is None:
+            raise ValueError("ReportVersion with this ID does not exist.")
```
dleard, line 33: When we know we should get 1 or 0 objects, it's better to use .get rather than .filter().first(). We can replace this bit with a get by pk and allow it to 404 error if no report_version is found rather than throwing our own error here.

```suggestion
        report_version = ReportVersion.objects.get(pk=version_id)
```

bciers/apps/reporting/src/app/utils/getReportAdditionalData.ts
```diff
@@ -0,0 +1,13 @@
+import { actionHandler } from "@bciers/actions";
+
+export async function getReportAdditionalData(version_id: number) {
+  const response = await actionHandler(
+    `reporting/report-version/${version_id}/report-additional-data`,
+    "GET",
+  );
+
+  console.log("res", response);
```
dleard, line 9: Remove console log

dleard (approved):
Really great work on the conditional dependencies! I found one small change to make and this is good to go

PR #2389, opened 2024-10-17
Feat/333 apply emission categories

dleard:
Applies an emission category (or categories) on save to a report_emission record based on the rules from the Emission Summary excel sheet in our SME channel files.

DB:
- New emission_category & emission_category_mapping tables
- m2m relationship added to report_emission table
- Initial data for the new tables

API:
- Added new function (apply_emission_categories) to the report_activity_save_service to apply categories on save to the report_emission table

FE:
- Added an API call to the backend call the save API service

To test this:
- Go to any activity page that we've developed (GSC, Mobile Fuel, Pulp & Paper..)
- Enter data & save
- Check in the DB that at least one record was created in the erc.report_emission_emission_categories table
- Verify using the Emission Summary excel sheet that it is the correct category/categories


ayeshmcg:
Great work @dleard , Everything looks good. Also we reviewed this PR together. 
Just one thing, 
on clicking submit should it take user to next page or disable, because on clicking twice on submit button it gives error in fromt end

ayeshmcg:
n

dleard:
> Great work, @dleard! Everything looks good, and we also reviewed this PR together.
> 
> Just one suggestion/question: When the user clicks Submit, should the button navigate to the next page or be disabled after the first click?
> 
> Currently, clicking the Submit button multiple times triggers an error on the frontend. Here's the error message: An error occurred while fetching reporting/report-version/3/facilities/f486f2fb-62ed-438d-bb3e-0819b51e3aff/activity/3/report-activity: Unexpected token '<', " <!doctype "... is not valid JSON
> 
> This happens because the POST API request results in a 500 error: POST /api/reporting/report-version/3/facilities/f486f2fb-62ed-438d-bb3e-0819b51e3aff/activity/3/report-activity HTTP/1.1" 500

This error is happening because we haven't implemented the loading of saved data into the forms yet, which means we don't have an id for the activity/source type/unit... records to send with the request. Since we don't have an id to pass, the `update_or_create` functions in the save service always try to `create`, which throws a uniqueness error since we're trying to create new records with the same data where one already exists. This will be handled by our [Load Activity Data ](https://github.com/bcgov/cas-reporting/issues/345) ticket. When we've loaded the data from the database, we'll have an id to pass & the `update_or_create` functions will update rather than create.

ayeshmcg (approved):

PR #2397, opened 2024-10-21
203 production information

pbastia:
Production data page.

To test:

- Add products to the facility report
- Access the production data page with a URL like:

http://localhost:3000/reporting/reports/1/facilities/f486f2fb-62ed-438d-bb3e-0819b51e3aff/production-data

### Note:
This PR also refactors how to generate the tasklist with the Activities (step nr.2 in our multiform header).
New:
- a `components/taskList` folder
- The `x_stepName.ts` file exports a method to build the required task list elements based on the required data. These are meant to be extended and modified as we flesh out the app


bciers/apps/reporting/src/app/components/products/ProductionData.tsx
```diff
@@ -0,0 +1,55 @@
...
+    "Jan 1",
+    "Dec 31",
+    allowedProductNames,
+  );
+
+  const orderedActivities = await getOrderedActivities(
+    report_version_id,
+    facility_id,
+  );
+  const taskListElements = await getFacilitiesInformationTaskList(
```
ayeshmcg, line 35: getFacilitiesInformationTaskList does not return a promise, so we can consider removing await here 
  const taskListElements = getFacilitiesInformationTaskList(
    report_version_id,
    facility_id,
    orderedActivities,
  );

bciers/apps/reporting/src/app/components/taskList/2_facilitiesInformation.ts
```diff
@@ -0,0 +1,35 @@
...
+            type: "Page",
+            title: activity.name,
+            link: `/reports/${versionId}/facilities/${facilityId}/activities?activity_id=${activity.id}`,
+          }) as TaskListElement,
+      ),
+    },
+    { type: "Page", title: "Non-attributable Emissions" },
+    { type: "Page", title: "Emissions Summary" },
+    {
+      type: "Page",
```
ayeshmcg, line 27: The current page is not marked as isActive here, are we planning to work on TaskList in the other card?

bciers/libs/actions/src/api/postProductionData.ts
```diff
@@ -0,0 +1,15 @@
+import { actionHandler } from "../actions";
+import { ProductData } from "../../../types/src/form/productionData";
```
ayeshmcg, line 2: "Consider updating the import statement to use the alias path for consistency and clarity:

import { Product, ProductData } from "@bciers/types/form/productionData";

bciers/apps/reporting/src/app/components/products/ProductionData.tsx
ayeshmcg: A question: Shouldn't the selected Regulated Products from the operation review page display here? It currently only shows fixture data, or may be I am missing something, We can pair up on that to understand it better.

ayeshmcg (commented):
Great work Pierre, this will help unblock other task which require Reporting task list.

bc_obps/reporting/models/report_product.py
```diff
@@ -0,0 +1,74 @@
...
+        db_comment="The facility report this production information belongs to",
+    )
+
+    product = models.ForeignKey(
+        RegulatedProduct,
+        on_delete=models.PROTECT,
+        related_name='report_products',
+        db_comment="The product this production information is about",
+    )
+    annual_production = models.FloatField(
```
dleard, line 34: Question: Should we consider using DecimalField here? Or does the slight imprecision of floats not matter with these values? They will be used in compliance calculations later

bciers/apps/reporting/src/data/jsonSchema/productionData.tsx
```diff
@@ -0,0 +1,138 @@
...
+  "ui:classNames": "form-heading-label",
+  product_selection_title: {
+    "ui:FieldTemplate": TitleOnlyFieldTemplate,
+    "ui:classNames": "mt-2 mb-5 emission-array-header",
+  },
+  product_selection: {
+    "ui:widget": "checkboxes",
+    "ui:options": {
+      label: false,
+    },
```
ayeshmcg, line 111: I would use our custom CheckboxWidget for product_selection.

ayeshmcg (commented):
Great work @pbastia , product list is displayed as required, just added a small suggestion regarding checkbox widget.

ayeshmcg (approved):

bc_obps/reporting/models/report_product.py
pbastia: @dleard after thinking about it, I would suggest leaving it as a float and bringing this to the BA. Decimal fields constrain how many decimal places we can accomodate, what if some products are in small quantity (I report 0.0034723 units) and others are very large quantities?
Floats will let us handle that for now, and we can change it later before release if we get more precise requirements

Issue #1700, opened 2024-05-28
Admin\Operator - Internal User can view "Operator Information"


zoeyli-46:
Partially blocked by #1699 

#### Description:
As an internal user, I can see the information for an operator

#### Acceptance Criteria:
Given I am on [operator details page](https://www.figma.com/design/LsDmLDCdnJqI8UahyqvOD8/BCIERS-Designs?node-id=14652-3185&t=EdbpdHgRnvGxWEpV-4) handled in ticket #1699 
When I click the Operator Information tile
Then I am brought to the operator information page 
* same as external user view, but in view-only mode, with no edit button
[Figma](https://www.figma.com/design/LsDmLDCdnJqI8UahyqvOD8/BCIERS-Designs?node-id=18028-8029&t=EdbpdHgRnvGxWEpV-4)

#### Development Checklist:

- [x] Create page in folder  `bciers/apps/administration/app/idir/cas_admin/operators/[operatorId]/page.tsx` to ` return <OperatorPage />;`  
- [x] Also(?), Create page in folder `bciers/apps/administration/app/idir/cas_analyst/operators/[operatorId]/page.tsx` to ` return <OperatorPage />;`  
- [x] implement logic in `OperatorForm` for `SingleStepTaskListForm` property  `allowEdit={role === FrontEndRoles.INDUSTRY_USER_ADMIN}`
- [x] Ensure `bc_obps/common/fixtures/dashboard/?` fixture created in #1699 for operator details has link to `/administration/operators/stub-id` 
- [x] Implement function to replace the static href, `/administration/operators/stub-id`, with dynamic href of the Operator ID obtained from the app router parameters (eg. https://github.com/bcgov/cas-registration/pull/1943/files#diff-75b8bf405e1f0dbab3d2218568846b7d4de7a5c78e86105644124064be3d025c)
- [x] modify `OperatorForm` vitests to reflect internal view-only mode, with no edit button
- [x] Delete page `bciers/apps/administration/app/idir/cas_admin/operators/user-operator/[id]/[formSection]/page.tsx`- [ ] Delete page `bciers/apps/administration/app/idir/cas_analyst/operators/user-operator/[id]/[formSection]/page.tsx`
- [ ] Meets the DOD

**Definition of Ready** (Note: If any of these points are not applicable, mark N/A)

- [ ] User story is included
- [ ] User role and type are identified
- [ ] Acceptance criteria are included
- [ ] Wireframes are included (if required)
[...]

Issue #1699, opened 2024-05-28
Admin\Operator - Internal User can view "Operator details"


zoeyli-46:
#### Description:
As an internal user, I can see the details for an operator

#### Acceptance Criteria:

Given I am on the [operator grid](https://www.figma.com/design/LsDmLDCdnJqI8UahyqvOD8/BCIERS-Designs?node-id=7861-1060&t=EdbpdHgRnvGxWEpV-4)
When I click on view details for an operator
Then I am brought the operator details page for that operator
[Figma](https://www.figma.com/design/LsDmLDCdnJqI8UahyqvOD8/BCIERS-Designs?node-id=14652-3185&t=b1Z2o0PV01yDoPhP-4)

Given I am on the operator details page,
Then the title says "[Operator Name] Operator Details
So that I know which operator I am viewing details for

#### Development Checklist:

- [x] create new dashboard for [Operator Name] Operator Details
- [x] add tiles for each of Information, Operations, Contacts, Users (Events can be ignored for now)
- [x] add routing protections so that only approved internal users are able to access this dashboard
- [x] Meets the DOD

**Definition of Ready** (Note: If any of these points are not applicable, mark N/A)

- [ ] User story is included
- [ ] User role and type are identified
- [ ] Acceptance criteria are included
- [ ] Wireframes are included (if required)
- [ ] Design / Solution is accepted by Product Owner
- [ ] Dependencies are identified (technical, business, regulatory/policy)
- [ ] Story has been estimated (under 13 pts)

·**Definition of Done** (Note: If any of these points are not applicable, mark N/A)

- [ ] Acceptance criteria are tested by the CI pipeline
- [ ] UI meets accessibility requirements
- [ ] Configuration changes are documented, documentation and designs are updated
- [ ] Passes code peer-review
- [ ] Passes QA of Acceptance Criteria with verification in Dev and Test
- [ ] Ticket is ready to be merged to main branch
- [ ] Can be demoed in Sprint Review
- [ ] Bugs or future work cards are identified and created
- [ ] Reviewed and approved by Product Owner

#### Dependencies
#1697 
- Blocked by
- Blocking

andrea-williams:
for clarity:
at the moment, when an internal user clicks on the "View Details" link next to an operator name in the operators grid, they only see this:

![Image](https://github.com/user-attachments/assets/ca616f70-1a46-4c33-9ef6-bb145e7bb26e)

instead of the Operator Details dashboard as shown in the linked wireframe

PR #2403, opened 2024-10-22
1700 internal operator page

BCerki:
Partially addresses: https://github.com/orgs/bcgov/projects/123/views/16?pane=issue&itemId=80445876&issue=bcgov%7Ccas-registration%7C1700

This PR:
- deletes placeholder files
- creates the internal view of the operator form (readonly, no edit button)
- types operatorId as a uuid instead of a string
- setting up the routing is blocked by #1699, so will leave #1700 open but blocked. Once 1699 is complete, will do another PR to hook everything together

Happo looks like flake to me

### DIFF
--- a/bciers/apps/administration/tests/components/operators/OperatorForm.test.tsx
+++ b/bciers/apps/administration/tests/components/operators/OperatorForm.test.tsx
@@ -305,7 +305,12 @@ describe("OperatorForm component", () => {
   });
   it("renders the empty operator form when creating a new operator", async () => {
     render(
-      <OperatorForm schema={testSchema} formData={{}} isCreating={true} />,
+      <OperatorForm
+        schema={testSchema}
+        formData={{}}
+        isCreating={true}
+        isInternalUser={false}
+      />,
     );
     // form fields and headings
     expectHeader(formHeaders);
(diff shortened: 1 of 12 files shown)


bciers/apps/administration/tests/components/operators/OperatorForm.test.tsx
```diff
@@ -683,4 +568,134 @@ describe("OperatorForm component", () => {
       ).toBeVisible();
     });
   }, 60000);
+  it("loads existing readonly Operator form data for an internal user", async () => {
```
BCerki, line 571: Only change to this test is mocking the session and moving it last (partly to keep the internal/external tests together, partly so I don't have to reset the mock every time)

bciers/apps/administration/tests/components/operators/OperatorPage.test.tsx
```diff
@@ -52,6 +56,9 @@ describe("Operator component", () => {
   it("renders the operator form with form data", async () => {
     // Mock the session data
     mockUseSession();
+    auth.mockReturnValueOnce({
+      user: { app_role: "industry_user_admin" },
+    });
```
BCerki, line 61: I ~think I have to mock both of these because `auth` is used in the server component and `useSession` in the client component it renders. Suggestions for a cleaner way always welcome!

PR #2406, opened 2024-10-23
Temporarily exclude ESLint from auditing

Sepehr-Sobhani:


dleard (approved):
Thanks! Can you add a mention of this exclusion in the card we have for updating esLint?

Sepehr-Sobhani:
> Thanks! Can you add a mention of this exclusion in the card we have for updating esLint?

Done!

PR #2408, opened 2024-10-23
2141 Internal user can assign BORO ID

BCerki:
card: https://github.com/orgs/bcgov/projects/123/views/16?pane=issue&itemId=80445898&issue=bcgov%7Ccas-registration%7C2141. See the comments for more details on the AC

This PR:
- endpoint and service to assign a boro id based on biz area conditions
- pytests
- add boro id to the admin operation form
- add an arg to `createOperationInformationSchema` because now the admin/registration versions of it are different
- boroid widget that uses `formContext`, and `formContext` optionally added to single step tasklist
- basic error handling for the widget specifically. To test it, raise an error in endpoint `operation_boro_id`
- vitests

For testing, the field will look different for each of these cases in both the internal/external:
- EIO
- status isn't registered
- neither of the above but no BORO ID issued
- BORO ID issued

Some of our mock operations are set up with the above conditions, or you can make your own and register them.

Happo looks like flake to me.


bc_obps/service/operation_service_v2.py
```diff
@@ -396,3 +397,22 @@ def check_conditions() -> Generator[Tuple[Callable[[], bool], str], None, None]:
...
+        if operation.bc_obps_regulated_operation:
+            raise Exception('Operation already has a BORO ID.')
+        if is_eio:
+            raise Exception('EIOs cannot be issued BORO IDs.')
+        if operation.status != Operation.Statuses.REGISTERED:
+            raise Exception('Operations must be registered before they can be issued a BORO ID.')
+
+        operation.generate_unique_boro_id()
+        operation.save(update_fields=['bc_obps_regulated_operation'])
+        # brianna do we want to do this, or better to set the verification columns?
```
BCerki, line 416: Do we want to set_create_or_update, or the verification columns? Verification columns were set in reg1 but not sure if that still applies

bciers/libs/components/src/form/widgets/BoroIdWidget.test.tsx
Sepehr-Sobhani, line 1: nit: I believe that many of these tests can be sync since they don't involve any async actions, except for the last two. I'm curious if we are applying the same approach in our other tests (which I suspect we are) and how this might affect the performance of the tests. 🤔

Sepehr-Sobhani (approved):
Looks awesome! 🤩 Just double-check your comment here: https://github.com/bcgov/cas-registration/pull/2408/files#r1815185657 with Andrea, Everything else looks good, so I'm approving to keep things moving forward!

bciers/libs/components/src/form/widgets/BoroIdWidget.test.tsx
BCerki: I copy/pasted the test setup from another test, so yes, we probably are doing this in other tests (at least the ones I wrote, because I'm a copy/paster). Chatgpt tells me it's bad because:
Implications:
Extra Overhead: The testing framework will treat the test as an asynchronous operation, introducing minor but unnecessary overhead by wrapping the test in a Promise. While this overhead is typically small, it can accumulate if many tests are written this way.

Misleading Code: Marking a function as async implies there is asynchronous behavior, even when there isn't. This can mislead other developers (or even yourself) who read the code, making them think that the test involves asynchronous operations when it doesn’t.

Test Behavior: Even if the function body is completely synchronous, returning a Promise may cause unexpected issues, especially if you're using done callbacks or other testing patterns that rely on detecting whether a test is synchronous or asynchronous.

Error Handling Differences: In an async test, throwing an error inside the function body will cause the returned Promise to reject, and the testing framework will treat it as a failed test. In contrast, a synchronous test throws the error directly. This behavior might obscure certain issues or create subtle differences in how errors are reported.

bciers/apps/administration/app/data/jsonSchema/operationInformation/administrationOperationInformation.ts
```diff
@@ -21,7 +22,7 @@ export const createAdministrationOperationInformationSchema = async (
   const administrationOperationInformationSchema: RJSFSchema = {
     type: "object",
     properties: {
-      section1: await createOperationInformationSchema(),
+      section1: await createOperationInformationSchema(Apps.ADMINISTRATION),
```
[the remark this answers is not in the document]
BCerki: I thought it was more readable to use the app name, but no strong feelings

bc_obps/service/tests/test_operation_service_v2.py
```diff
@@ -767,3 +768,68 @@ def test_do_not_raise_exception_if_data_complete_opt_in():
...
+            operator=approved_user_operator.operator,
+            status=Operation.Statuses.REGISTERED,
+        )
+        baker.make(BcObpsRegulatedOperation, operation=operation, id='21-0001')
+        baker.make(
+            RegistrationPurpose,
+            registration_purpose=RegistrationPurpose.Purposes.POTENTIAL_REPORTING_OPERATION,
+            operation=operation,
+        )
+        OperationServiceV2.generate_boro_id(approved_user_operator.user.user_guid, operation.id)
```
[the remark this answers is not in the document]
BCerki: 😱

bc_obps/service/operation_service_v2.py
andrea-williams: excellent point. Since the "Registered" status is now set automatically, the verification columns on the operation aren't relevant in Reg2. However, for the `BC Obps Regulated Operation` model, we're only tracking `issued_at` and not `issued_by`. IMO we should add an `issued_by` column to this model so that we can track which internal user issued the BORO id. But I'll leave it up to you @BCerki to decide whether you want to include that as part of this PR, or we can create a separate ticket for that work to be done later.

BCerki: I can include it here, it'll be quick. So only set issued by, not verified or updated?

andrea-williams: cool, thanks Bri! We'll have a nice separation of audit columns this way actually - the `Operation` model's created and updated fields will only be affected by external users, and the `BcObpsRegulatedOperation` model will have `issued_by` and `issued_at` columns that are only triggered by internal users.

so yes, for this user story, an internal user issuing a BORO ID should only change the audit cols on the `BORO` model, not the Operation model.

bc_obps/service/operation_service_v2.py
```diff
@@ -404,3 +406,25 @@ def check_conditions() -> Generator[Tuple[Callable[[], bool], str], None, None]:
...
+        ).exists()
+        if operation.bc_obps_regulated_operation:
+            raise Exception('Operation already has a BORO ID.')
+        if is_eio:
+            raise Exception('EIOs cannot be issued BORO IDs.')
+        if operation.status != Operation.Statuses.REGISTERED:
+            raise Exception('Operations must be registered before they can be issued a BORO ID.')
+
+        operation.generate_unique_boro_id()
+        if operation.bc_obps_regulated_operation is None:
```
BCerki, line 424: Had to add this to make mypy happy

bc_obps/registration/tests/endpoints/v2/_operations/_operation_id/test_boro_id.py
```diff
@@ -0,0 +1,35 @@
...
+            )
+            assert response.status_code == 401
+            assert response.json()['detail'] == "Unauthorized"
+
+    def test_authorized_role_can_patch(self):
+        operation = baker.make_recipe('utils.operation', status=Operation.Statuses.REGISTERED)
+
+        response = TestUtils.mock_patch_with_auth_role(
+            self,
+            "cas_admin",
```
andrea-williams, line 29: any particular reason why "cas_analyst" role wasn't included in this test?

andrea-williams (approved):
just one question about a missing user role in the tests.

Otherwise looks great, and I appreciate your willingness and speed in adding that `issued_by` col so quickly that I'll be erecting a statue in your honour.

bc_obps/registration/tests/endpoints/v2/_operations/_operation_id/test_boro_id.py
BCerki: In all our permission configs, `cas_analyst` and `cas_admin` are always included together, so there's no way one of them could access an endpoint and another couldn't. (cas-registration/bc_obps/common/permissions.py). But that may not be a great strategy if we do start adding differences :grimacing: I'll add `cas_analyst` to this test

PR #2410, opened 2024-10-23
Chore/372 updated additional report data

ayeshmcg:
Card: https://github.com/bcgov/cas-reporting/issues/372
Changes: 

- fixed check for registration purspose
- fixed tests
- Moved Electricity generated to separate section.

### DIFF
--- a/bciers/apps/reporting/src/app/bceidbusiness/industry_user_admin/reports/[version_id]/additional-reporting-data/page.tsx
+++ b/bciers/apps/reporting/src/app/bceidbusiness/industry_user_admin/reports/[version_id]/additional-reporting-data/page.tsx
@@ -1,26 +1,14 @@
 import { Suspense } from "react";
 import Loading from "@bciers/components/loading/SkeletonForm";
-import AdditionalReportingData from "@reporting/src/app/components/additionalInformation/AdditionalReportingData";
-import { getRegistrationPurpose } from "@reporting/src/app/utils/getRegistrationPurpose";
-
+import AdditionalReportingData from "@reporting/src/app/components/additionalInformation/additionalReportingData/AdditionalReportingData";
 interface PageProps {
   params: { version_id: number };
 }
 
 export default async function Page({ params }: PageProps) {
-  const versionId = params.version_id;
-  const registrationPurposes =
-    (await getRegistrationPurpose(versionId))?.registration_purposes || [];
-  const includeElectricityGenerated = registrationPurposes.includes(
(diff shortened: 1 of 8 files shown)


pbastia (changes requested):
Nice work! Very nice use of JsonSchema features.
I think we can do a small refactor to push some of the work from the `useEffect` into a straight method call in a wrapping server-side component

bciers/apps/reporting/src/data/jsonSchema/additionalReportingData/additionalReportingData.ts
```diff
@@ -10,107 +10,160 @@ export const additionalReportingDataSchema: RJSFSchema = {
...
+        properties: {
+          electricity_generated: {
+            type: "string",
+            title: "Electricity Generated",
+          },
+        },
+      },
+    },
+  };
+}
```
pbastia, line 169: we can export this as a constant and compute it only once (I'm sure it's not huge, but there's definitely a performance penalty for destructuring dictionaries).
Also, we should make functions pure where we can, it's good coding practice
```ts
function buildSchemaWithElectricityGenerated(additionalReportingDataUiScheam: RJSFSchema){
...return the logic 
}

export const additionalReportingDataWithElectricityGeneratedUiSchema : RJSFSchema = 
    buildSchemaWithElectricityGenerated(additionalReportingDataUiSchema)
```

bciers/apps/reporting/src/app/bceidbusiness/industry_user_admin/reports/[version_id]/additional-reporting-data/page.tsx
```diff
@@ -1,15 +1,26 @@
...
+
   return (
     <Suspense fallback={<Loading />}>
-      <AdditionalReportingData versionId={params.version_id} />
+      <AdditionalReportingData
+        versionId={versionId}
+        includeElectricityGenerated={includeElectricityGenerated}
+      />
     </Suspense>
   );
```
Comment (pbastia) [inline bciers/apps/reporting/src/app/bceidbusiness/industry_user_admin/reports/[version_id]/additional-reporting-data/page.tsx:25]: This code is now duplicated between the multiple page in the different roles.
We should put that in an `AdditionalReportingData` server-side component instead, next to the other one

We've been (loosely) following this pattern in other places:
`AdditionalReportingData.tsx` : server-side component with initial data fetching/computing logic
`AdditionalReportingDataForm.tsx`: client-side component with the form and the interaction

pbastia (changes requested):
Nice getting rid of the effect!
Just one more thing about code duplication

PR #2411, opened 2024-10-24
2404  align apps FE

shon-button:


bciers/apps/reporting/src/tests/components/dataGrid/model/operations/operationColumns.test.tsx
```diff
@@ -151,7 +151,7 @@ describe("operationColumns function", () => {
     await user.click(screen.getByText("Continue"));
 
     expect(useRouter().push).toHaveBeenCalledWith(
-      `reports/${row.report_version_id}/review-operator-data`,
+      `reports/${row.report_version_id}/review-operator-data?reports_title=${row.name}`,
```
pbastia, line 154: As much as possible Iike to hardcode the expected values in tests. This way we make the test slightly more sensitive to a change in string format, or catch side-effects transforming the initial value we pull data from.

```suggestion
      `reports/15/review-operator-data?reports_title=Operation with report`,
```

bciers/libs/components/src/datagrid/cells/ChangeUserOperatorStatusColumnCell.tsx
```diff
@@ -21,21 +21,18 @@ const handleUpdateStatus = async (
...
+    "PUT",
+    "",
+    {
+      body: JSON.stringify({
+        role: roleUpdate,
+        status: statusUpdate,
+      }),
+    },
+  );
+  return response;
```
pbastia, line 35: We can return the call directly without an await here too

```suggestion
  return actionHandler(
    `registration/user-operators/${userOperatorId}/update-status`,
    "PUT",
    "",
    {
      body: JSON.stringify({
        role: roleUpdate,
        status: statusUpdate,
      }),
    },
  );
```

pbastia (approved):
This is some top notch refactoring 🥇 🥇  ! Thanks so much for the hard and thorough work.
Just left a couple of minor nitpicks without consequence

PR #2416, opened 2024-10-24
activity page- Open Pit Coal Mining

ayeshmcg:
Card: https://github.com/bcgov/cas-reporting/issues/339
Changes:-

- Adds backend config/migration for Open Pit coal mining activity.
- Adds frontend page and components for Open Pit coal mining activity.
- Added components for activity-specific pages and dynamic field rendering.
- Fixed typo(Emission and Equivalent Emission in Hydrogen Production and Carbonates use)

To test:-
- make reset_db to apply latest migrations.
- spin up the front, backend and postgres.
- Login, then navigate to http://localhost:3000/reporting/reports , select operation 3 -> press save and continue to go to review Facilities page-> select Open pit coal mining(from the list of activities) and press save & continue. 
- The Gas type dropdown options should change options available in the methodology field.
- The Methodology dropdown options spawn additional fields, depending on the option selected.

### DIFF
--- a/bc_obps/reporting/json_schemas/2024/carbonates_use/carbonates_use.json
+++ b/bc_obps/reporting/json_schemas/2024/carbonates_use/carbonates_use.json
@@ -9,8 +9,8 @@
         "title": "Carbonate data",
         "type": "object",
         "properties": {
-          "emissions": {
-            "title": "Emissions",
+          "emission": {
+            "title": "Emission",
             "type": "number"
           },
           "gasType": {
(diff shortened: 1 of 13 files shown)


pbastia (approved):
Looks awesome! 🥇 
I don't think this is anywhere in the wireframes, I would just put the unit as a `ReadOnlyWidget` and place it right next to the "Quantity of coal" field. For some reason it shows up disconnected, something doesn't seem to be working right with the ui schema order

image

PR #2423, opened 2024-10-29
Feat/142 facility emission summary

dleard:
Adds the facility emission summary page & necessary API endpoints/services.

Backend:
- emission_category API endpoint /emission-summary to fetch all the aggregated summary data
- emission_cataegory_service to do the calculations. 
  - Has individual functions to aggregate by emission category
  - Has a special function to determine reporting_only emissions
  - Has a function to return the results of all the individual aggregation functions

Frontend:
- Added a page route for the emission summary
- Added a component to render the summary
- Schema/UI Schema for the read-only form

To test:
- Start a report & save several emissions of data in an activity form
- Go to the `reporting/reports/{version_id}/facilities/{facility_id}/summary` page & see the aggregated results
- You can manipulate the data in the database to change which categories an emission belongs to in order to see the summary display different data


bciers/apps/reporting/src/app/components/facility/FacilityEmissionSummary.tsx
```diff
@@ -0,0 +1,94 @@
...
+  return (
+    <Box sx={{ p: 3 }}>
+      <div
+        className="container mx-auto p-4"
+        data-testid="facility-emission-summary"
+      >
+        <MultiStepHeader stepIndex={1} steps={customStepNames} />
+      </div>
+      <div className="w-full flex">
+        <ReportingTaskList elements={tasklistData} />
```
pbastia, line 70: We could fetch the activities in the FacilitiesEmissionSummaryData server-side component, and use the real task list in `2_facilitiesInformation.ts`

pbastia (approved):
Just needs updating the task list! Awesome

PR #2428, opened 2024-10-29
Chore/equivalent emissions

dleard:
Adds a gwp column & data to the gas_type table.
Auto-calculates the equivalent emission (CO2e) & saves it to the json data on saving a report_emission record.

To test this:
open an activity page & save a couple emissions
check the report_emission records in the database, "equivalentEmission" values should have been automatically added to the json data with a value of: emission * (gwp for the gas type reported)


ayeshmcg (approved):
@dleard 
Good work 
Equivalent emission are being calculated automatically by GWP and added to report_emission table

PR #2429, opened 2024-10-29
367 methodology

gdalcengio:
#367

- feat: Integrate save_methodology func in save report activity file and add fk migration associated
- chore: Fix up id from previous commit to allow multiple methodologies in the database
- test: Add methodology to save report tests and edit emission tests to include methodology info
- test: fix tests from last commit and add report methodology related tests


bc_obps/reporting/service/report_activity_save_service.py
```diff
@@ -222,4 +226,29 @@ def save_emission(
         report_emission.set_create_or_update(self.user_guid)
         EmissionCategoryMappingService.apply_emission_categories(report_source_type, report_fuel, report_emission)
 
+        self.save_methodology(report_emission, emission_data['methodology'])
+
         return report_emission
+
+    def save_methodology(
+        self,
+        report_emission: ReportEmission | None,
```
pbastia, line 235: the database model enforces the foreign key to ReportEmission to not be null, so the `None` case doesn't exist
```suggestion
        report_emission: ReportEmission,
```

bc_obps/reporting/service/report_activity_save_service.py
```diff
@@ -222,4 +226,29 @@ def save_emission(
...
+        methodology_data: dict,
+    ) -> ReportMethodology:
+        json_data = exclude_keys(methodology_data, ['methodology'])
```
pbastia, line 238: we should exclude the `id` key here too

bc_obps/reporting/service/report_activity_save_service.py
```diff
@@ -222,4 +226,29 @@ def save_emission(
...
+        methodology = Methodology.objects.get(name=methodology_data['methodology'])
+
+        report_methodology_id = methodology_data.get('id')
+        report_methodology, _ = ReportMethodology.objects.update_or_create(
+            id=report_methodology_id,
+            create_defaults={
+                "methodology": methodology,
+                "json_data": json_data,
+                "report_version": self.facility_report.report_version,
+                **({"report_emission": report_emission} if report_emission else {}),
```
pbastia, line 248: No `None` case here to handle either
```suggestion
                "report_emission": report_emission,
```

bc_obps/reporting/tests/service/test_report_activity_save_service/data.py
```diff
@@ -29,9 +29,9 @@
                                     "test_emission_number": 12345,
                                     "test_emission_bool": True,
                                     "test_emission_str": "test",
-                                    "methodology": "Default EV",
-                                    "methodology_field_1": 123,
-                                    "method_field_2": "Some description",
+                                    "methodology": {
+                                        "methodology": "Default HHV/Default EF",
```
pbastia, line 33: We should still pass some random additional fields to the methodology here (or the real ones for that one)

bc_obps/reporting/tests/service/test_report_activity_save_service/test_save_report_emission.py
```diff
@@ -117,14 +123,27 @@ def test_update_emission(self):
...
+                "methodology": {
+                    "id": 9003,
+                    "methodology": "Default EF",
+                    "unitFuelCo2DefaultEmissionFactor": 3,
+                    "unitFuelCo2DefaultEmissionFactorFieldUnits": "kg/GJ",
+                },
+            },
         )
 
         assert report_emission.id == updated_return_value.id
```
pbastia, line 149: we're still missing an assertion/test to make sure the `save_methodology` method was called with the right parameters (see the `test_save_fuel` for an example

bc_obps/reporting/tests/service/test_report_activity_save_service/test_save_report_methodology.py
```diff
@@ -0,0 +1,95 @@
...
+            report_activity.report_version.id,
+            report_activity.facility_report.facility.id,
+            report_activity.activity.id,
+            test_infrastructure.user.user_guid,
+        )
+
+        # test for no report emission, should not be allowed due to model constraints
+        with self.assertRaises(ValidationError):
+            service_under_test.save_methodology(None, {"methodology": "Default HHV/Default EF"})
+
```
pbastia, line 27: We're basically testing django with this statement, which is already tested (or we hope so, but it's not our responsibility to make sure it works as intended)

pbastia (approved):
Top notch work! Just a couple of comments on tests and types, mostly removing some code.
I'll approve it, let me know if you want me to double-check again before merging

PR #2430, opened 2024-10-30
Feat/336 non attributable emissions page

ayeshmcg:
Card: https://github.com/bcgov/cas-reporting/issues/336
Changes:- 

- Added Non Attributable Emissions Page: New page for collecting non attributable emissions data.
- Create CheckboxGroup Widget
- Modified NestedArrayFieldTemplate
- Schema and UI Schema: Defined schemas to validate and structure the new data.
- APIs/Services:
    - Get Gas type
    - Get Emission Category
    - Save Non Attributable Emissions Data
    - Get Non Attributable Emissions Data saved data for review.

- Model for Non Attributable Emissions Data: Created a model for data storage and retrieval.


bc_obps/reporting/service/report_non_attributable_service.py
```diff
@@ -0,0 +1,30 @@
...
+    def get_report_non_attributable_by_version_id(cls, report_version_id: int) -> List[ReportNonAttributableEmissions]:
+        return list(ReportNonAttributableEmissions.objects.filter(report_version__id=report_version_id))
+
+    @classmethod
+    def save_report_non_attributable_emissions(
+        cls, version_id: int, data: ReportNonAttributableIn
+    ) -> ReportNonAttributableEmissions:
+        emission_category = EmissionCategory.objects.get(category_name=data.emission_category)
+
+        report_non_attributable = ReportNonAttributableEmissions.objects.create(
```
dleard, line 19: Isn't this going to create a new record even if a record is being edited? Should something like `update_or_create..` with the id of the record be used here instead?

bciers/apps/reporting/src/app/components/reportInformation/nonAttributableEmissions/NonAttributatbleEmissions.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { UUID } from "crypto";
+import { getAllGasTypes } from "@reporting/src/app/utils/getAllGasTypes";
+import { getAllEmissionCategories } from "@reporting/src/app/utils/getAllEmissionCategories";
+
+interface NonAttributableEmissionsProps {
+  versionId: number;
+  facilityId: UUID;
+}
+
+export default async function NonAttributableEmissions({
```
dleard, line 13: I think this component is missing some functionality where it should load any existing non-attributable records. When I save data & then navigate back to the /non-attributable page, it doesn't load the data that has been saved. Likely needs an additional api endpoint/service to pull existing data & parse it into the shape of the formData.

bciers/apps/reporting/src/data/jsonSchema/nonAttributableEmissions/nonAttributableEmissions.ts
```diff
@@ -0,0 +1,107 @@
...
+                      type: "string",
+                      enum: gasTypes.map((gas) => gas.chemical_formula),
+                    },
+                    uniqueItems: true,
+                  },
+                  emission_category: {
+                    type: "string",
+                    title: "Emission Category (Optional)",
+                    enum: emissionCategories.map(
+                      (category) => category.category_name,
```
dleard, line 96: I seem to remember someone mentioning that the Gas Type & Emission Category fields on this page are actually not optional, but I can't find any figma comment or mention of it in the card. Could you confirm with @zoeyli-46 to find out whether or not these fields should be optional?

dleard (changes requested):
Thanks Ayesha! The form layout looks great. I have a few comments around return types in the api/service & avoiding the use of the 'Any' type. I tested the form manually & the save functionality works great, but when I navigate back to the form after saving it doesn't load the data I just saved. Let me know if you have questions about any of the comments!

ayeshmcg:
@dleard  Addessed all the requested changes

bc_obps/reporting/service/report_non_attributable_service.py
```diff
@@ -0,0 +1,33 @@
...
+
+from reporting.models import ReportNonAttributableEmissions, GasType, EmissionCategory
+
+from reporting.schema.report_non_attributable_emissions import ReportNonAttributableIn
+
+
+class ReportNonAttributableService:
+    @classmethod
+    def get_report_non_attributable_by_version_id(cls, report_version_id: int) -> List[ReportNonAttributableEmissions]:
+        return list(ReportNonAttributableEmissions.objects.filter(report_version__id=report_version_id))
```
dleard, line 11: Now that I'm looking at this service, I'm realizing that we're going to need to add `facility_id` to the `report_non_attributable_emissions model` & use that in combination with the `version_id` to fetch these records. Otherwise, when we implement this page for LFOs we won't be able to differentiate which facility these non-attributable emissions belong to with only the `version_id` foreign key

dleard (changes requested):
One more change needed here. The non-attributable emissions needs facility_id as well as version_id

ayeshmcg:
@dleard  Added Facility report id as wel to the model

dleard (approved):
:100: Thanks Ayesha, this looks great! Ready to merge

PR #2446, opened 2024-11-05
Feat/345 load activity data from database

pbastia:
- report-activity GET endpoint serializes the `ReportActivity`, `ReportUnit`, `ReportFuel`, `ReportEmission` and `ReportMethodology` in a format that the activity form can display and understand
- report-activity POST endpoint makes use of the serialization endpoint to return saved data with ids and computed fields
- Updates the `ActivityForm` to update the form data with the saved IDs and computed data

### DIFF
--- a/bciers/apps/reporting/src/app/components/activities/ActivityForm.tsx
+++ b/bciers/apps/reporting/src/app/components/activities/ActivityForm.tsx
@@ -1,5 +1,4 @@
 "use client";
-import FormBase from "@bciers/components/form/FormBase";
 import { customizeValidator } from "@rjsf/validator-ajv8";
 import { useEffect, useState } from "react";
 import { actionHandler } from "@bciers/actions";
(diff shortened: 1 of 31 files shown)


bciers/apps/reporting/src/app/components/activities/ActivityForm.tsx
```diff
@@ -11,6 +10,10 @@ import { FuelFields } from "./customFields/FuelFieldComponent";
 import { FieldProps } from "@rjsf/utils";
 import { getUiSchema } from "./uiSchemas/schemaMaps";
 import { UUID } from "crypto";
+import { withTheme } from "@rjsf/core";
+import formTheme from "@bciers/components/form/theme/defaultTheme";
+
+const Form = withTheme(formTheme);
```
pbastia, line 16: `FormBase` was introducing some unintended behaviours by also holding a copy of the formData in the state, so this uses the RJSF component directly

dleard (approved):
Looks great! Love how those serializers work. Just one note on a potential performance optimization.
I've made a [bug card](https://github.com/bcgov/cas-reporting/issues/395) for the issue this PR unearthed around the persistence of formState that should be handled shortly after this goes in

Issue #2376, opened 2024-10-16
Backend integration test for registration workflow


Sepehr-Sobhani:
#### Describe the task

We are currently testing each registration endpoint and service individually. However, we need to create a test that covers all steps together to ensure that no mistakes occur between steps.

#### Acceptance Criteria

- [x] Add a comprehensive backend test that covers all endpoints involved in the registration process.

#### Additional context

- Add any other context about the task here.
- Or here

Issue #2337, opened 2024-10-07
Reg\Reg workflow\Operation info - No regulated products for Reporting operations


andrea-williams:
#### Describe the task

Currently the "Regulated Products" dropdown appears when the selected registration purpose is "Reporting Operation" - this dropdown should not appear, because reporting operations don't have regulated products.

#### Acceptance Criteria

- [x] the "regulated products" dropdown appears as a required field when the selected registration purpose is one of "OBPS Regulated", "New Entrant", or "Opt-in"
- [x] the "regulated products" dropdown DOES NOT appear in the form at all when the selected registration purpose is any of the options not listed in the first AC

#### Additional context

- Add any other context about the task here.
- Or here

patriciarussellCAS:
Is this already being addressed in #2224 ?

patriciarussellCAS:
Tagging @BCerki as this is similar to the ticket you just created. Maybe we could merge them?

BCerki:
> Tagging @BCerki as this is similar to the ticket you just created. Maybe we could merge them?

They're not particularly related in the code. This will be quick though if you want to additionally bring it in

vesselak:
tested successfully in dev.

PR #2452, opened 2024-11-06
2376 Integration test for registration workflow

Sepehr-Sobhani:
[ISSUE 2376](https://github.com/bcgov/cas-registration/issues/2376)

**NOTES:**
- Added a missing test for the `user_has_access` function (for Operation)
- Corrected the endpoint name in the test.
- Populated the `submission_date` field when the user submits an operation.
- Added an integration test covering all registration purposes workflows.

### DIFF
--- a/bc_obps/registration/tests/integration/test_operation_registration.py
+++ b/bc_obps/registration/tests/integration/test_operation_registration.py
@@ -1,0 +1,273 @@
+from datetime import timedelta
+
+import pytest
+from django.db.models import Q
+from model_bakery import baker
+from registration.models import RegistrationPurpose, FacilityDesignatedOperationTimeline, Operation
+from registration.tests.constants import MOCK_DATA_URL
+from registration.tests.utils.bakers import bc_obps_regulated_operation_baker
+from registration.tests.utils.helpers import CommonTestSetup, TestUtils
+from registration.utils import custom_reverse_lazy
+
+
+class TestOperationRegistration(CommonTestSetup):
+    def setup_method(self):
+        super().setup_method()
+        self.approved_user_operator = baker.make_recipe('utils.approved_user_operator', user=self.user)
+        self.bcghg_id = baker.make_recipe('utils.bcghg_id')
+        self.swrs_facility_id = 1003
+        self.boro_id = bc_obps_regulated_operation_baker()
+        # add some random contacts so we can test that they are not overridden
+        random_contacts = baker.make_recipe('utils.contact', _quantity=5)
+        self.operation = baker.make_recipe(
+            'utils.operation',
+            created_by=self.user,
+            operator=self.approved_user_operator.operator,
+            bcghg_id=self.bcghg_id,
+            swrs_facility_id=self.swrs_facility_id,
+            bc_obps_regulated_operation=self.boro_id,
+            contacts=random_contacts,
+        )
+        self.created_at = self.operation.created_at
+        self.updated_at = None
+        self.operation_representative_id = None
+
+    def _set_operation_information(self, purpose: RegistrationPurpose.Purposes):
+        operation_information_payload = {
+            "registration_purpose": purpose,
+            "regulated_products": [1],
+            "name": f"{purpose} name",
+            "type": "LFO",
+            "naics_code_id": 1,
+            "secondary_naics_code_id": 2,
+            "tertiary_naics_code_id": 3,
+            "activities": [1],
+            "boundary_map": MOCK_DATA_URL,
+            "process_flow_diagram": MOCK_DATA_URL,
+        }
+        response = TestUtils.mock_put_with_auth_role(
+            self,
+            "industry_user",
+            self.content_type,
+            operation_information_payload,
+            custom_reverse_lazy("register_edit_operation_information", kwargs={'operation_id': self.operation.id}),
+        )
+        if response.status_code != 200:
+            raise Exception(response.json())
+        self.operation.refresh_from_db()
+
+    def _set_facilities(self):
+        facility_payload = [
+            {
+                "name": "Test Facility 1",
+                "type": "Single Facility",
+                "latitude_of_largest_emissions": "5",
+                "longitude_of_largest_emissions": "5",
+                "operation_id": self.operation.id,
+            },
+            {
+                "street_address": "123 street",
+                "municipality": "city",
+                "province": "AB",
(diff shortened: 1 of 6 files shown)


bc_obps/registration/tests/integration/test_operation_registration.py
```diff
@@ -0,0 +1,273 @@
+from datetime import timedelta
```
BCerki, line 1: Do we need the empty init file above?

bc_obps/registration/tests/integration/test_operation_registration.py
```diff
@@ -0,0 +1,273 @@
...
+from registration.tests.utils.helpers import CommonTestSetup, TestUtils
+from registration.utils import custom_reverse_lazy
+
+
+class TestOperationRegistration(CommonTestSetup):
+    def setup_method(self):
+        super().setup_method()
+        self.approved_user_operator = baker.make_recipe('utils.approved_user_operator', user=self.user)
+        self.bcghg_id = baker.make_recipe('utils.bcghg_id')
+        self.swrs_facility_id = 1003
```
BCerki, line 18: Are we still using swrs_facility_id in reg2 now that we have a Facility model? I don't remember seeing it on any of the forms

bc_obps/registration/tests/integration/test_operation_registration.py
```diff
@@ -0,0 +1,273 @@
...
+        self.boro_id = bc_obps_regulated_operation_baker()
+        # add some random contacts so we can test that they are not overridden
```
BCerki, line 20: Might be good to add a facility to test for no overwrites too--probably many users will already have facilities in the system from the SWRS import

bc_obps/registration/tests/integration/test_operation_registration.py
```diff
@@ -0,0 +1,273 @@
...
+            created_by=self.user,
+            operator=self.approved_user_operator.operator,
+            bcghg_id=self.bcghg_id,
+            swrs_facility_id=self.swrs_facility_id,
+            bc_obps_regulated_operation=self.boro_id,
+            contacts=random_contacts,
+        )
+        self.created_at = self.operation.created_at
+        self.updated_at = None
+        self.operation_representative_id = None
```
BCerki, line 33: I think operation rep would already be None since none of the setup appears to add it?

BCerki (approved):
This looks awesome, Sepehr, thanks for putting it together!

bc_obps/registration/tests/integration/test_operation_registration.py
```diff
@@ -0,0 +1,273 @@
...
+        self.operation = baker.make_recipe(
+            'utils.operation',
+            created_by=self.user,
+            operator=self.approved_user_operator.operator,
+            bcghg_id=self.bcghg_id,
+            swrs_facility_id=self.swrs_facility_id,
+            bc_obps_regulated_operation=self.boro_id,
+            contacts=random_contacts,
+        )
+        self.created_at = self.operation.created_at
```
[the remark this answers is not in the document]
Sepehr-Sobhani: Yes, you're correct!

bc_obps/registration/tests/integration/test_operation_registration.py
```diff
@@ -0,0 +1,273 @@
...
+        self.updated_at = None
+        self.operation_representative_id = None
+
+    def _set_operation_information(self, purpose: RegistrationPurpose.Purposes):
+        operation_information_payload = {
+            "registration_purpose": purpose,
+            "regulated_products": [1],
+            "name": f"{purpose} name",
+            "type": "LFO",
```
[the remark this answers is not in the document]
Sepehr-Sobhani: Good point, Thanks!

bc_obps/registration/tests/integration/test_operation_registration.py
Sepehr-Sobhani: It aims to prevent future naming conflicts if we create a test with the same name.

bc_obps/registration/tests/integration/test_operation_registration.py
Sepehr-Sobhani: Good point!

bc_obps/registration/tests/integration/test_operation_registration.py
Sepehr-Sobhani: Yes, it's just a variable that captures that value for later comparison. I apologize for the confusion; I've added a comment to prevent this in the future.

bc_obps/registration/tests/integration/test_operation_registration.py
Sepehr-Sobhani: I think we can’t do that since we’re testing the SFO type, unless I create an additional set of tests for the LFO. 🤔

Sepehr-Sobhani: Added another parameterized test for operation types.

bc_obps/registration/tests/integration/test_operation_registration.py
```diff
@@ -0,0 +1,273 @@
...
+            contacts=random_contacts,
+        )
+        self.created_at = self.operation.created_at
+        self.updated_at = None
+        self.operation_representative_id = None
+
+    def _set_operation_information(self, purpose: RegistrationPurpose.Purposes):
+        operation_information_payload = {
+            "registration_purpose": purpose,
+            "regulated_products": [1],
```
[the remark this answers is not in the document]
BCerki: I think this requirement changed since our existing implementation, and as far as I know, #2337 is the new desired behaviour (@andrea-williams?). As far as this test goes, I'd suggest getting the mock data and assertions as close as possible to desired behaviour, but if that makes the test fail, just add a bullet in 2337 to correct it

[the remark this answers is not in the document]
andrea-williams: confirming Brianna's statement - reporting operations won't have a regulated product (ever). Our existing implementation that made regulated products required for reporting operations was a boo-boo.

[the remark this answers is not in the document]
Sepehr-Sobhani: Thank you both! 🙏 
@andrea-williams I'm not going to change the implementation of the `set_registration_purpose` service, as it might cause conflicts with your PR.

PR #2455, opened 2024-11-08
chore: programatically get poetry version from file to avoid drift

dleard:
The poetry version defined in the path of the Dockerfile drifted from what was defined in .tool-versions. This will get the version from .tool-versions to avoid this in the future.


pbastia (approved):
Thank you!

PR #2468, opened 2024-11-13
Chore: add informative error to superuser command

Sepehr-Sobhani:


pbastia (commented):
we could use `curl --fail-with-body` in the job template to let us abort if the superuser couldn't be created

PR #2477, opened 2024-11-15
Task/update tile titles registration external user/2350

andrea-williams:
Reminder to devs when running this locally to test - you have to `make reset_db` for the updated tiles to take effect. Just adding this as a reminder because I consistently forget and then get confused why it isn't working.

**Registration dashboard:**
I updated the tile descriptions to match the updated wireframes, and set the Report an Event tile to always be hidden for the time being, since this page won't be available to external users as part of Reg2 MVP.
 
Screenshot 2024-11-14 at 4 55 34 PM

**Main dashboard:**
Hid the link to Report an Event from the "Registration" tile (for the same reason as above).

Screenshot 2024-11-14 at 4 55 44 PM

### DIFF
--- a/bc_obps/common/fixtures/dashboard/bciers/external.json
+++ b/bc_obps/common/fixtures/dashboard/bciers/external.json
@@ -144,10 +144,6 @@
                     "value": false
                   }
                 ]
-              },
-              {
-                "title": "Report an Event",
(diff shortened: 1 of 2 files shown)


BCerki (approved):
:cartwheeling:

PR #2481, opened 2024-11-16
chore: conditionally add superuser envs

dleard:
I think there's also an issue with the superuser being in an init container & how we drop our database beforehand (which means the django auth admin tables don't exist).

This PR is just to fix the side effect of having envs in test/prod that reference a secret that isn't created in those environments, causing a container init config error on deploy. I created a stand-in secret in test to get around that for now. This PR makes those envs conditional on the dev environment.

### DIFF
--- a/helm/cas-bciers/templates/backend/deployment.yaml
+++ b/helm/cas-bciers/templates/backend/deployment.yaml
@@ -34,6 +34,16 @@ spec:
             - "create_superuser"
           env:
             {{- include "cas-bciers.backendEnvVars" . | nindent 12 }}
+            - name: DJANGO_SUPERUSER_USERNAME
+              valueFrom:
+                secretKeyRef:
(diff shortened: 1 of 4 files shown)


helm/cas-bciers/templates/backend/deployment.yaml
```diff
@@ -34,6 +34,16 @@ spec:
             - "create_superuser"
           env:
             {{- include "cas-bciers.backendEnvVars" . | nindent 12 }}
+            - name: DJANGO_SUPERUSER_USERNAME
```
pbastia, line 37: I think it's also missing the "if -dev" toggle here

pbastia: ..nevermind, it's further up!

pbastia (approved):

PR #2487, opened 2024-11-19
369: Add unit to regulated product

acatchpole:
[Card](https://github.com/bcgov/cas-reporting/issues/369)

#### Changes:
 - Created migration that:
    -  adds ```unit``` and ```is_regulated``` fields to ```regulated_product``` table. (**Note**: the ticket says the field should be "regulated" and i made the boolean field "is_regulated")
    - populates new fields based off of [this file](https://bcgov.sharepoint.com/:x:/r/teams/00608-CPRInternal/Shared%20Documents/Carbon%20Pricing%20Transition%20SMEs/Registration%20App/Registration%20Pt.%202%20Fields.xlsx?d=w8cf871009f8d43c19d1d79c69fa259e8&csf=1&web=1&e=NdGXiC)
    - adds 2 new ***unregulated*** products (as listed in file above) to the ```regulated_product``` table
 - Updated relevant models in the front and backend to support these new fields.
 - Unit data is now included in existing API call
 - Units now display correctly in the Production Data form
 - Removed regulated products 'Mining: critical mineral, other than copper' and 'Smelting: critical mineral, other than aluminum and lead-zinc' per request from business (they are not used by any existing operation)

#### To Verify:
(mostly stolen from SH)
 ```
cd bc_obps
make reset_db
make run
```  
 ```
cd bciers && yarn dev-all
```

1. Navigate to `http://localhost:3000`
2. Click button `Log in with Business BCeID`
3. Sign in to Keycloak using `bc-cas-dev`
4. Click title ```Submit Annual Report(s)```
5. Click ```Start``` next to Operation 3 (an SFO)
6. Proceed through the 3 tabs populating Operation Information and hitting ```Save & Continue```
7. On 'Review Facility Information' page, hit ```Save & Continue```
8. On the side tasklist, click title ```Production Data```
9. Click the checkboxes beside any of the listed products and see that ***the 'Unit' field is populated***
10. Put some test data into the first 3 fields of a product and hit ```Save & Continue``` and see that ***the 'Unit' field is still populated*** (necessary b/c the unit value now comes from a different structure [see [here](https://github.com/bcgov/cas-registration/blob/7f049ad3872a3d30acc7ec3941f4e3ff548597a1/bciers/apps/reporting/src/app/components/products/ProductionDataForm.tsx#L39) where data source is different depending on if there is report_product data available])


bc_obps/registration/migrations/0057_regulatedproduct_add_isregulated_and_unit.py
```diff
@@ -0,0 +1,191 @@
...
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('registration', '0056_remove_historicaltransferevent_future_designated_operator_and_more'),
+    ]
+
+    def update_regulated_product_units_and_is_regulated(apps, schema_editor):
```
acatchpole, line 12: should i have a test to validate all this data being added?

dleard: I usually add tests to verify the number of items I've added. And if there are different types of data (ie regulated vs unregulated) I count those too & make sure the correct number of those have been added.
Here's an example of tests I added for emission categories: 
https://github.com/bcgov/cas-registration/blob/develop/bc_obps/reporting/tests/models/test_emission_category.py

dleard (approved):
Nice work! Looks good to me :shipit:

Issue #373, opened 2023-12-05
Enable no explicit any typescript eslint rule


marcellmueller:
#### Describe the task

As discussed in our dev team agreements we are going to enable the typescript eslint rule `no-explicit-any`. The benefits of this are enforcing rules to have a mostly typed front end and streamline PR reviews as this would be caught by pre-commit.

https://typescript-eslint.io/rules/no-explicit-any/

#### Acceptance Criteria

- [ ] Enable `no-explicit-any` rule

#### Additional context

- Add any other context about the task here.
- Or here

PR #2491, opened 2024-11-19
373 cement production

gdalcengio:
#373

This pr creates adds necessary configuration for Cement Production reporting activity

- feat: Add cement production reporting activity migration, data, and json
- chore: Tweak previous migration and add UISchema
- test: add test for cement production configuration

Steps to test:

- Make reset_db & make run
- Login as BCEID
- Nav to an SFO operation and set reporting activity to cement production
- Emission should have gas type as C02
- Field "Number of times in the reporting year that missing data procedures were followed" should exist
- Should have 5 methodologies and CEMS should have no reporting fields


dleard:
This one is a little more complicated than the other activities. "Month" is not really a field, but a header separation for a set of fields. Each set of fields in the excel sheet (Clinker production, emission factor...) needs to be reported for each month (ditto for quarter). So I think we'll need to create a custom schema for this where each month is an object that has a set of fields & try to display that with a custom widget in a way that doesn't make the page super long. Happy to chat about how to implement this together.

shon-button:
@gdalcengio 

Excellent 💥

A few suggestions, for your consideration:

1. Instead of adding new methodology to `bc_obps/reporting/migrations/0008_prod_data.py` add them within the associated migration `bc_obps/reporting/migrations/0063_cement_production.py`
2. For `CO2\Calcination Emmissions`:
1. Perhaps order the months before quarters as in the BCIER Fields excel file
2. Perhaps style the fields in a similar way as the other methodology fields; so without your fancy, and very nice collapsing section and with vertical stacking instead of horizontal.  
Something like: 
image

[Example code](https://github.com/bcgov/cas-registration/blob/373-cement-production-field-sort/bciers/apps/reporting/src/app/components/activities/uiSchemas/cementProductionUISchema.ts)

shon-button (approved):

bciers/apps/reporting/src/app/components/facility/TaskListElements.ts
```diff
@@ -0,0 +1,25 @@
+import { TaskListElement } from "@bciers/components/navigation/reportingTaskList/types";
```
shon-button, line 1: @gdalcengio 

Is this required?

bciers/libs/components/src/form/fields/CollapsibleDefinitionFieldTemplate.tsx
```diff
@@ -0,0 +1,75 @@
+"use client";
```
shon-button, line 1: @gdalcengio 

fyi: since we do not have required fields on the activity pages; perhpas, this can be paired down to:

```
"use client";

import { useState } from "react";
import { Paper, Card, Grid, CardHeader, CardActions, IconButton, Collapse } from "@mui/material";
import ExpandLessIcon from "@mui/icons-material/ExpandLess";
import ExpandMoreIcon from "@mui/icons-material/ExpandMore";

function CollapsibleDefinitionFieldTemplate({ label, children }) {
  const [expandCollapse, setExpandCollapse] = useState(true);

  return (
    <Paper sx={{ marginBottom: "1rem" }}>
      <Card>
        <Grid container spacing={1} sx={{ justifyContent: "space-between" }}>
          <Grid item xs={"auto"}>
            <CardHeader
              title={label}
              titleTypographyProps={{ variant: "h6" }}
            />
          </Grid>
          <Grid item xs={1}>
            <CardActions sx={{ justifyContent: "flex-end" }}>
              <IconButton onClick={() => setExpandCollapse(!expandCollapse)}>
                {!expandCollapse ? <ExpandMoreIcon /> : <ExpandLessIcon />}
              </IconButton>
            </CardActions>
          </Grid>
        </Grid>
        <Collapse in={expandCollapse}>
          {children}
        </Collapse>
      </Card>
    </Paper>
  );
}

export default CollapsibleDefinitionFieldTemplate;

```

gdalcengio:
> @gdalcengio
> 
> Excellent 💥
> 
> A few suggestions, for your consideration:
> 
> 1. 🚩 Instead of adding new methodology to `bc_obps/reporting/migrations/0008_prod_data.py` add them within the associated migration `bc_obps/reporting/migrations/0063_cement_production.py` (example: `bc_obps/reporting/migrations/0055_liquefied_natural_gas.py`)
> 2. For `CO2\Calcination Emmissions`:
> 
> * Perhaps order as months then quarters - as in the BCIER Fields excel file
> * Perhaps order month fields as in the BCIER Fields excel file
> * Perhaps order quarter fields as in the BCIER Fields excel file
> * Perhaps style the fields in a similar way as the other methodology fields; so **without** your fancy, very nice collapsing section and **with** vertical stacking instead of horizontal.
>   **Something like:**
> 
> ### [Example code here...](https://github.com/bcgov/cas-registration/blob/373-cement-production-field-sort/bciers/apps/reporting/src/app/components/activities/uiSchemas/cementProductionUISchema.ts)
> image

I can get these suggestions done. However, for the last point. I made it three columns and collapsable after a conversation with @dleard discussing how in the review page, it would take up quite a bit of space. With 3 columns and collapsing boxes, the user can review and fill out the form without needing to scroll so much.

bciers/apps/reporting/src/app/components/facility/TaskListElements.ts
gdalcengio: nope! Thanks for catching this. Must've been reintroduced during a rebase

PR #2500, opened 2024-11-20
Refactor custom error codes

Sepehr-Sobhani:


andrea-williams (approved):
I noticed you deleted all the `codes_5xx` - does `custom_codes_4xx` also handle 500 errors?

Sepehr-Sobhani:
> I noticed you deleted all the `codes_5xx` - does `custom_codes_4xx` also handle 500 errors?

No, we only handle 4xx errors in `handle_exception`, so we never return 5xx errors to the user.

PR #2503, opened 2024-11-21
2453 boro id restrictions

BCerki:
card: https://github.com/orgs/bcgov/projects/123/views/16?pane=issue&itemId=86401827&issue=bcgov%7Ccas-registration%7C2453

This PR:
- conditionally shows the boro id widget button only for regulated operations
- backend conditions on not creating boro ids for non-regulated operations
- updated mock data so we have Operation 19 eligible for BORO ID (needs to be registered and have an appropriate purpose)
- updated vitests and pytest

### DIFF
--- a/bc_obps/service/operation_service_v2.py
+++ b/bc_obps/service/operation_service_v2.py
@@ -415,14 +415,10 @@ class OperationServiceV2:
     @classmethod
     def generate_boro_id(cls, user_guid: UUID, operation_id: UUID) -> Optional[BcObpsRegulatedOperation]:
         operation = OperationService.get_if_authorized(user_guid, operation_id)
-        is_regulated_operation = operation.registration_purpose in [
-            Operation.Purposes.OBPS_REGULATED_OPERATION,
-            Operation.Purposes.NEW_ENTRANT_OPERATION,
-            Operation.Purposes.OPTED_IN_OPERATION,
-        ]
+
         if operation.bc_obps_regulated_operation:
             raise Exception('Operation already has a BORO ID.')
-        if not is_regulated_operation:
+        if not operation.is_regulated_operation:
             raise Exception('Non-regulated operations cannot be issued BORO IDs.')
         if operation.status != Operation.Statuses.REGISTERED:
             raise Exception('Operations must be registered before they can be issued a BORO ID.')
--- a/bc_obps/service/operation_service_v2.py
+++ b/bc_obps/service/operation_service_v2.py
@@ -415,11 +415,15 @@ class OperationServiceV2:
     @classmethod
     def generate_boro_id(cls, user_guid: UUID, operation_id: UUID) -> Optional[BcObpsRegulatedOperation]:
         operation = OperationService.get_if_authorized(user_guid, operation_id)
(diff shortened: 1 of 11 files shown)


bc_obps/service/operation_service_v2.py
```diff
@@ -415,11 +415,15 @@ def check_conditions() -> Generator[Tuple[Callable[[], bool], str], None, None]:
     @classmethod
     def generate_boro_id(cls, user_guid: UUID, operation_id: UUID) -> Optional[BcObpsRegulatedOperation]:
         operation = OperationService.get_if_authorized(user_guid, operation_id)
-        is_eio = operation.registration_purpose == Operation.Purposes.ELECTRICITY_IMPORT_OPERATION
+        is_regulated_operation = operation.registration_purpose in [
+            Operation.Purposes.OBPS_REGULATED_OPERATION,
+            Operation.Purposes.NEW_ENTRANT_OPERATION,
+            Operation.Purposes.OPTED_IN_OPERATION,
+        ]
```
Sepehr-Sobhani, line 422: I believe we may need this check in other areas in the future, so it would be great to move it to a function under the Operation data model(or a util file for the Operation data model). Additionally, it would be helpful to implement a constraint in the data model to prevent generating Boro IDs for non-eligible purposes.

bciers/apps/registration/app/components/operations/registration/enums.ts
```diff
@@ -27,6 +27,19 @@ export enum RegistrationPurposes {
...
+  RegistrationPurposes.OPTED_IN_OPERATION,
+];
+
+export const nonRegulatedOperationPurposes: ReadonlyArray<RegistrationPurposes> =
+  [
+    RegistrationPurposes.REPORTING_OPERATION,
+    RegistrationPurposes.ELECTRICITY_IMPORT_OPERATION,
+    RegistrationPurposes.POTENTIAL_REPORTING_OPERATION,
+  ];
+
```
Sepehr-Sobhani, line 42: Are we using this anywhere?

Sepehr-Sobhani (commented):
Great work!

This isn't directly related to this PR, but I wanted to flag it here in case you'd like to address it. Otherwise, I can create a bug ticket for it: I wasn't able to register a new operation. On the first page, we attempt to fetch all registration purposes using the `getRegistrationPurposes()` function, which calls an endpoint that no longer exists.

bciers/apps/registration/app/components/operations/registration/enums.ts
BCerki: Nope!

bc_obps/service/operation_service_v2.py
BCerki: I made it a property on the `operation` model and moved the checks that were previously in the endpoint into the property. The status check is still in the endpoint because reg 1 had different rules

Sepehr-Sobhani: Good catch!

BCerki: Actually, it's the same situation for checking if an operation is regulated. Reg1 doesn't use registration purposes, so if I put the check on the db, I'll break reg1

Sepehr-Sobhani (approved):
🤩 Thanks again!

PR #2519, opened 2024-11-25
392 button tech debt

gdalcengio:
-392

To Test:

- Login as bc-cas-dev
- start operation 3 report
- Freely navigate back or Save & Continue through the form
- Buttons should all work and disable if the page is read only, they should be at the bottom and change display based on saving state or redirecting state
- *Note: allocation of emissions page is not completed at the time of submitting this PR

### DIFF
--- a/bciers/apps/reporting/src/app/components/operations/personResponsible/PersonResponsible.tsx
+++ b/bciers/apps/reporting/src/app/components/operations/personResponsible/PersonResponsible.tsx
@@ -41,14 +41,13 @@ const PersonResponsible = ({ version_id }: Props) => {
   const [operationType, setOperationType] = useState(null);
 
   const [schema, setSchema] = useState<RJSFSchema>(personResponsibleSchema);
-  const router = useRouter();
 
   const queryString = serializeSearchParams(useSearchParams());
   const continueUrl =
     operationType === "Linear Facility Operation"
       ? `/reports/${version_id}/facilities/lfo-facilities${queryString}`
       : `/reports/${version_id}/facilities/${facilityId}/review${queryString}&facilities_title=Facility`;
-  const backUrl = `/reports/${version_id}/review-operator-data${queryString}`
+  const backUrl = `/reports/${version_id}/review-operator-data${queryString}`;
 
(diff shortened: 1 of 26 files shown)


bciers/apps/reporting/src/app/components/operations/personResponsible/PersonResponsible.tsx
```diff
@@ -44,10 +44,11 @@ const PersonResponsible = ({ version_id }: Props) => {
   const router = useRouter();
 
   const queryString = serializeSearchParams(useSearchParams());
-  const saveAndContinueUrl =
+  const continueUrl =
     operationType === "Linear Facility Operation"
       ? `/reports/${version_id}/facilities/lfo-facilities${queryString}`
       : `/reports/${version_id}/facilities/${facilityId}/review${queryString}&facilities_title=Facility`;
```
dleard, line 50: There is an issue here with this path. When `queryString` is null, we get:
`.../review&facilities_title=Facility`, which doesn't parse properly because we don't have the `?` before the query params.

A potential solution (which would need to be investigated for side effects) is to change the [serializeSearchParams](https://github.com/bcgov/cas-registration/blob/develop/bciers/libs/utils/src/serializeSearchParams.ts#L11) utility function to return `?` when there are no params & append `&` to the end of the querystring when there are.

We'd have to look at where `serializeSearchParams` is used & make sure everything still works as expected where it's implemented

dleard: To reproduce the bug, start a report & get to the Person Responsible page, once you enter data there & click Save&Continue, you'll get a 404 error because the params are missing the `?` to start them off

pbastia (approved):
This is really good! I only removed an unnecessary timeout in the Activity Form.

I absolutely hate how we have to do this trick with a `useRef` and an effect to submit with a context, but it looks like it's the only easy solution.
I think an alternative, slightly cleaner solution would be to inspect the event fired by the submit button, by digging through the event internals and refactoring our FormBase / submit handler APIs, but that's much more of an uplift.

PR #2527, opened 2024-11-26
2522 separate v1 v2 endpoints

BCerki:
https://github.com/orgs/bcgov/projects/123/views/16?pane=issue&itemId=88755084&issue=bcgov%7Ccas-registration%7C2522

This PR:
- All v1 endpoints now start with `v1`, and v2 ones have no prefix
- For the most part, I was able to just copy over a v1 endpoint to make a v2 version. A couple exceptions:
- - The `operators` endpoint did different things in v1 vs. v2. In v1, it's for the Select Operator workflow (called in OperatorSearchWidget), but in v2 it's for getting the grid data to show in Administration. I've created a new endpoint in v2 called operators/search, and added formContext to the widget to know which one to call. We can ditch the formContext after retiring reg1 (unless the reviewer has a better idea for how to handle this)
- - We have two endpoints that get info about user_operator (user-operators/current and user-operators/current/operator) but they both had the same `def` name, so changed that slightly
- Tests are also copy/pasted over
- Changed the permission configs slightly to make it easier to keep track of which version uses which configs
- New make command to reset the reg data (had to do this a lot while testing)
- I only got rid of "v2" in the endpoints. There are still "v2"s in the services and tags. It would be nice to clean this up at some point to avoid confusion, but it's not needed for having separate v1/v2 endpoint permission, so I left it out of this PR as it's already a :japanese_ogre: 
- I manually tested everything in reg1 and 2 to make sure it still works, and the e2e tests should do a good job with reg1, but the reviewer may want to do manual testing too--this PR touches so many things :scream:


bc_obps/registration/api/v2/e2e_test_setup.py
```diff
@@ -0,0 +1,35 @@
...
+                call_command('truncate_dev_data_tables')
+                return HttpResponse("Test setup complete.", status=200)
+            if load_only:  # only load the data
+                call_command('load_fixtures', workflow)
+                return HttpResponse("Test setup complete.", status=200)
+            call_command('truncate_dev_data_tables')
+            call_command('load_fixtures', workflow)
+            return HttpResponse("Test setup complete.", status=200)
+        except Exception as e:
+            return HttpResponse(f"Test setup failed. Reason:{e}", status=500)
```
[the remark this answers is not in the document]
BCerki: I brought this over from reg1 so we could just axe the whole v1 foler

bc_obps/registration/api/v2/_user_operators/_current/operator.py
```diff
@@ -1,40 +1,36 @@
...
 from registration.schema.generic import Message
 from registration.api.router import router
-
-
-# "current" refers to the current user-operator (we look up the current user-operator via the current user)
-# this endpoint retrieves data about both the user-operator and the operator
+from service.data_access_service.user_service import UserDataAccessService
 
 ## GET
+# brianna
```
Sepehr-Sobhani, line 15: leftover!

bc_obps/common/permissions.py
```diff
@@ -103,7 +99,11 @@ def get_permission_configs(permission: str) -> Optional[Union[Dict[str, List[str
             'authorized_app_roles': ["industry_user"],
             'authorized_user_operator_roles': ["admin"],
         },
-        "authorized_irc_user": {
+        "v1_authorized_irc_user_and_industry_admin_user": {
+            'authorized_app_roles': all_authorized_app_roles,
+            'authorized_user_operator_roles': ["admin"],
+        },
+        "v1_authorized_irc_user": {
```
Sepehr-Sobhani, line 106: I’m curious about the reason for adding `v1_` to these permission groups. 🤔

Sepehr-Sobhani (commented):
Great job! Our endpoints are much more understandable now, even though there’s a bit of duplication.

Would it be possible to adjust the endpoint tags to separate V1 and V2? If it’s a quick fix, that would be awesome! If not, no worries—we can add it to our tech debt for later.

bc_obps/common/permissions.py
BCerki: I was thinking that reg2 would have totally different cas user groups. Upon reflection, everyon can view everything, so the `GET`s could still use these. I'll ditch the prefix

bc_obps/registration/tests/endpoints/v1/_facilities/test_facility_id.py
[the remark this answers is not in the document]
BCerki: Definitely! I think this might be some github funny business, I had moved tests not deleted them. I'll double check everything

BCerki:
> Great job! Our endpoints are much more understandable now, even though there’s a bit of duplication.
> 
> Would it be possible to adjust the endpoint tags to separate V1 and V2? If it’s a quick fix, that would be awesome! If not, no worries—we can add it to our tech debt for later.

Tags unfortunately don't look like a quick fix--V2 weren't used super consistently

Sepehr-Sobhani:
> > Great job! Our endpoints are much more understandable now, even though there’s a bit of duplication.
> > Would it be possible to adjust the endpoint tags to separate V1 and V2? If it’s a quick fix, that would be awesome! If not, no worries—we can add it to our tech debt for later.
> 
> Tags unfortunately don't look like a quick fix--V2 weren't used super consistently

Sounds good! I have created this ticket: https://github.com/bcgov/cas-registration/issues/2547
Feel free to add to it.

Sepehr-Sobhani (approved):
This is amazing—thank you so much for tackling it! 🎖️ 

One last favor: could you update the documentation where we explain the API structure for our endpoints (`backend-structure.md`)?

BCerki:
> This is amazing—thank you so much for tackling it! 🎖️
> 
> One last favor: could you update the documentation where we explain the API structure for our endpoints (`backend-structure.md`)?

Yes, definitely!

PR #2530, opened 2024-11-27
Reporting production data: fix frontend issues

pbastia:
### Addressing some issues with the production data form:

- Production Methodology should be a dropdown menu with 2 options: OBPS Calculator or "other"
> If they select "other" there needs to be a description box for them to add the details

- Production Methodology should be called "Production Quantification Methodology"
- the second "quantity of storage" listed should say "at the END of the compliance period [Dec 31]" but current says at the beginning.

### Work done:
- Added `production_methodology_description` field to the ReportProduct model
- Added a checkconstraint to enforce the presence of a description only if the methodology selected is 'other'
- Frontend `allOf` json_schema rule to match the backend


dleard (approved):
:shipit: Looks good to me! Addresses all the comments in the card from PO review.

PR #2532, opened 2024-11-27
Fix test data after removing registration purpose table

Sepehr-Sobhani:


dleard (approved):
Thanks for updating this!

PR #2546, opened 2024-12-03
chore:fixed navigation for report new entrant

ayeshmcg:
Card: https://github.com/bcgov/cas-reporting/issues/414
Changes:- 
- Fixed Navigation from additional reporting data page
- FIxed initial form/data rendering of additional reporting data page


bc_obps/reporting/api/report_additional_data.py
```diff
@@ -52,8 +50,9 @@ def save_report_additional_data(
 @handle_http_errors()
 def get_report_additional_data_by_version_id(
     request: HttpRequest, report_version_id: int
-) -> Tuple[Literal[200], ReportAdditionalDataOut]:
-    report_person_responsible = ReportAdditionalDataService.get_report_person_responsible_by_version_id(
+) -> tuple[Literal[200], ReportAdditionalData]:
```
pbastia, line 53: This needs to be a django ninja schema. Was there something missing from `ReportAdditionalDataOut` ?

bc_obps/reporting/api/report_additional_data.py
```diff
@@ -52,8 +50,9 @@ def save_report_additional_data(
...
+    report_additional_data = ReportAdditionalDataService.get_report_report_additional_data_by_version_id(
```
pbastia, line 54: I'd say the name is redundant here, the "by version id" is already exposed by the name of the parameter.
I'd suggest `ReportAdditionalData.get_report_additional_data` to match the `save` function name

bciers/apps/reporting/src/app/components/additionalInformation/additionalReportingData/AdditionalReportingData.tsx
```diff
@@ -2,22 +2,68 @@ import { Suspense } from "react";
...
+interface AdditionalReportingDataProps {
+  versionId: number;
+}
+const transformToFormData = (
+  data: {
+    capture_emissions: any;
+    emissions_on_site_use: any;
+    emissions_on_site_sequestration: any;
+    emissions_off_site_transfer: any;
+    electricity_generated: any;
```
pbastia, line 24: shouldn't these be string or number types?

bciers/apps/reporting/src/app/components/additionalInformation/additionalReportingData/AdditionalReportingData.tsx
```diff
@@ -2,22 +2,68 @@ import { Suspense } from "react";
...
+  registrationPurpose: string,
+) => ({
+  captured_emissions_section: {
+    capture_emissions: data.capture_emissions,
+    capture_type: [
+      data.emissions_on_site_use && CAPTURE_TYPES.ON_SITE_USE,
+      data.emissions_on_site_sequestration &&
+        CAPTURE_TYPES.ON_SITE_SEQUESTRATION,
+      data.emissions_off_site_transfer && CAPTURE_TYPES.OFF_SITE_TRANSFER,
+    ].filter(Boolean),
```
pbastia, line 35: I'm not sure what this is supposed to do, but there's a few issues I can spot:
- `data.emissions_on_site_use` will evaluate to `false` if the value is present but 0 or ""
- while `Boolean` is the constructor for the object type derived from `boolean` and a function (so it doesn't trip up typescript) it won't work as an array filter predicate

pbastia (changes requested):
Nice work!!
I've just left a couple of questions

dleard (approved):
Looks good! Task list issue where 'new entrant' is incorrectly displayed to be handled in https://github.com/bcgov/cas-reporting/issues/335

bc_obps/service/report_additional_data.py
```diff
@@ -1,24 +1,73 @@
...
+
+        # Prepare the additional data section if the conditions are met
+        additional_data_section = (
+            {"electricity_generated": report_additional_data.electricity_generated}
+            if registration_purpose == "OBPS Regulated Operation" and report_additional_data.electricity_generated
+            else None
+        )
+
+        # Construct and return a ReportAdditionalDataOut instance
+        return ReportAdditionalDataOut(
```
pbastia, line 47: All this needs to be done in a django ninja schema

Your schema will also be a lot easier to test than this (I don't think there are tests for this service) since we won't have to test the field transformations, just the fact that it's setup properly

PR #2550, opened 2024-12-04
chore: refactor async app route to factory

shon-button:


pbastia (changes requested):
Thanks for tackling this, quite a bit of an uplift! 🥇 
I've left some comments about the use of the `defaultPageFactory` that apply to most pages refactor - we'll need to make sure:
- that the Next app page params (version_id, facility_id, etc) align with the props of the page components, otherwise the page factory won't know what to map
- that we handle the `seachParams` case in a consistent way where they're needed (either flatten them, or add them in their own prop)

PR #2552, opened 2024-12-05
chore: added contact page to open in new tab

ayeshmcg:
Card: N/A
A quick PR , so that when user clicks on contact on Person Responsible page, it opens it in a new tab.


pbastia (approved):
👍

PR #2553, opened 2024-12-06
Feat/287 allocation of emissions

shon-button:
[Addresses](https://github.com/bcgov/cas-reporting/issues/287)

### 🚀 Impact:

Provides the functionality to allocate emissions to reported emission production data.

### ✏️ Notes
- This PR does not include tests.  Tests (and other potential tech debt) will be addressed in [Allocation of Emissions Tests](https://github.com/bcgov/cas-reporting/issues/431)

### 🔬 Local Testing:

1. From terminal command, start the api server:
 ```
cd bc_obps
make reset_db
make run
```  
2.  From terminal command, start the app development server:
 ```
cd bciers && yarn dev-all
```  

1. Navigate to http://localhost:3000
2. Click button "Log in with Business BCeID"
3. Navigate to http://localhost:3000/reporting/reports
4. On row `Operation 3`, click button `Start`
5. On page `Review operation information` , under 'Reporting activities', add `Carbonate use`.
6.  Click button `Save and Continue`
7. On page 'Person responsible', choose one of the options in 'Select contact if the are already a BCIERS user'.
8. Click button 'Save and Continue'.
9. On page 'Review facility information', click the check mark beside 'Carbonate use'.
10. Click button 'Save and Continue'.
11. Complete and save the activity forms: `General stationary combustion excluding line tracing` and 'Carbonate use'
*Note the 'Equivalent Emission' values after you have entered some test data*
13. Click button `Submit`
14. Change url to: http://localhost:3000/reporting/reports/3/facilities/f486f2fb-62ed-438d-bb3e-0819b51e3aff/production-data
15.  Complete the form: `Production Data`, making sure that at least 2 different products are selected.
16. Click button `Save and Continue`.
**Expected results:** 
  -  `Allocation of Emissions` page displays under step `Report Information`, task `Allocation of Emissions`
  - Category products display products reported on form: `Production Data`
  - Total emissions displays total as expected for categories: `Stationary fuel combustion emissions` and `Industrial process emissions`.
  - `Totals in tCO2e\Total emissions attributable for reporting` displays total as expected 
  - Button `Save and Continue` is disabled
  - Message displays `All emissions must be allocated to 100% before saving and continuing`
18. Complete form's required fields
19. Enter the category product allocated emissions values to match the category's total emissions
**Expected results:** 
  -  Category `Total Allocated` value reflects category product field(s) value(s) as a percent of the category `Total Emissions` value
  - `Totals in tCO2e` products matches category product field(s) value(s) 
[...]

PR #2562, opened 2024-12-10
289 allocation of emissions bugs

shon-button:
### ✏️ Notes
- This PR addresses bugs/tech debt from #2553 
- The issue "Save'" and "Save and Continue" buttons do the same thing (which bypasses the 100% allocation check) was introduced in #2519...and is 
**Fixed** in  #2567 (also applied to this branch):

### 🐛 Issues addressed:

Backend:
- Records are being saved to the report_emission_allocation table in the database when the allocated value is 0. 
- Carry the emission category Ids in the schema & use them in the service
- (new) Update `reporting.models.report_product_emission_allocation\allocated_quantity` field `max_digits` from 10 to 20

Frontend:
- Carry the emission category Ids in form data as hidden fields (and use them in the payload)
- The details box when methodology is other is too small & the font is different
- Field `Totals in tCO2e\Total emissions attributable for reporting` is of inconsistent width with the rest of the page. 
- (new) Category sections display header `emission_category_name`

### 🔬 Local Testing:

 ```
cd bc_obps
make reset_db
make run
```  
 ```
cd bciers && yarn dev-all
```  

1. Navigate to http://localhost:3000
2. Click button "Log in with Business BCeID"
3. Navigate to http://localhost:3000/reporting/reports
4. On row `Operation 3`, click button `Start`
5. Complete the report pages until http://localhost:3000/reporting/reports/3/facilities/f486f2fb-62ed-438d-bb3e-0819b51e3aff/allocation-of-emissions
**Expected results:** 
  -  `Allocation of Emissions` page displays under step `Report Information`, task `Allocation of Emissions`
image
  - Category product names display as title for each emission allocation section
  - Products emissions reported on form `Production Data` display as expected in product category `Total Emissions`
image
  - `Totals in tCO2e\Total emissions attributable for reporting` displays total as expected 
  - Button `Save` is enable
  - Button `Save and Continue` is disabled
  - Message displays `All emissions must be allocated to 100% before saving and continuing`
image
6. Allocate some emissions across category products
**Expected results:** 
- Allocation divisions and totlas are reflected as expected in `Totals in tCO2e` 
image
7. Complete required fields, click `Save` 
**Expected results:** 
- POST request is successful 
- Page does not route to another page
8.  Enter all emission allocations
**Expected results:** 
  - Button `Save and Continue` is enabled
  - Message does not display `All emissions must be allocated to 100% before saving and continuing`
 9. Click Button `Save and Continue`
**Expected results:** 
  - POST request is successful and page routes to ` Additional Reporting Data`
image


bciers/apps/reporting/src/app/components/facility/FacilityEmissionAllocationPage.tsx
```diff
@@ -11,7 +11,7 @@ export default async function FacilityEmissionAllocationPage({
 }) {
   const orderedActivities = await getOrderedActivities(version_id, facility_id);
   const initialData = await getEmissionAllocations(version_id, facility_id);
-
+  console.log(initialData);
```
dleard, line 14: left over debugging line

dleard (approved):
Nice work! This makes things much more smooth & no extraneous data is getting into the database. 

One thing I'd check with PO / BA is the behaviour of the "Save" button when 100% of emissions are not yet allocated. There is a message at the bottom that says you cannot "save & continue" without 100% allocation, which is good & that button is greyed out. However, the "Save" button looks live but when clicked, nothing gets saved. If the preference of the PO/BA is to not allow saving data at all until 100% of emissions are allocated then this button should also be greyed out, if they want to be able to save the data as-is at that point, then we'll need to wire that button to actually save when clicked.

^ That can be handled in another ticket when we get clarity on what the expectation is there around saving data when 100% of emissions have not been allocated.

bciers/apps/reporting/src/app/components/facility/FacilityEmissionAllocationPage.tsx
dleard: Oops, I just went ahead and merged it immediately after approval. My bad. I'll just take this out in my current PR

shon-button:
> Nice work! This makes things much more smooth & no extraneous data is getting into the database.
> 
> One thing I'd check with PO / BA is the behaviour of the "Save" button when 100% of emissions are not yet allocated. There is a message at the bottom that says you cannot "save & continue" without 100% allocation, which is good & that button is greyed out. However, the "Save" button looks live but when clicked, nothing gets saved. If the preference of the PO/BA is to not allow saving data at all until 100% of emissions are allocated then this button should also be greyed out, if they want to be able to save the data as-is at that point, then we'll need to wire that button to actually save when clicked.
> 
> ^ That can be handled in another ticket when we get clarity on what the expectation is there around saving data when 100% of emissions have not been allocated. Screenshot: Screenshot from 2024-12-16 13-35-19

@dleard 

When you "However, the "Save" button looks live but when clicked, nothing gets saved" did you complete the required field `Methodology`?  If not then `Save` will trigger the field validation errors; otherwise; the data should save, as it did for me (according to the test steps).

image

dleard:
Ahh, that might have been it. I do forget to update that field sometimes and jump right into allocating. I'll check again, thanks!

PR #2586, opened 2024-12-12
chore:fixed activity page bug

ayeshmcg:
Card: https://github.com/bcgov/cas-reporting/issues/439
Changes:- fixed url for buildSchema 
Since we were using join and when use .join() without a separator, the default separator is a comma (,), so adding a separator.


bciers/apps/reporting/src/app/components/activities/ActivityForm.tsx
```diff
@@ -89,7 +89,7 @@ export default function ActivityForm({
   const fetchSchemaData = async (sourceTypeIds: string[]) => {
     const sourceTypeQueryString = sourceTypeIds
       .map((id) => `&source_types[]=${id}`)
-      .join();
+      .join("");
```
dleard, line 92: :heart:

dleard (approved):
Nice find!

PR #2597, opened 2024-12-16
Chore/compliance summary tests

dleard:
- Unit tests for the static functions in the compliance summary service
- Multiple test cases for the class function in the compliance summary service
- Refactors the service to not depend on the ninja schemas for data shaping
- Fixes a bug in the compliance service where the Reduction Factor & Tightening Rate were being applied in the wrong place


bc_obps/reporting/tests/service/test_compliance_service/test_compliance_service_static.py
```diff
@@ -0,0 +1,323 @@
...
+        operation.naics_code_id = 1
+        operation.save()
+        report.reporting_year_id = 2024
+        report.save()
+        operation = Operation.objects.get(pk=report_version_2.report.operation.id)
+        report = Report.objects.get(pk=report_version_2.report_id)
+        operation.naics_code_id = 22
+        operation.save()
+        report.reporting_year_id = 2024
+        report.save()
```
pbastia, line 29: For shorter syntax we can use the `update` queryset method:
```py
Report.objects.filter(pk=...).update(reporting_year_id=2024)
```

pbastia (approved):
Very nice! I left a few nitpicks but this is good to go

PR #2600, opened 2024-12-17
feat: a report cannot be submitted if the verification statement is m…

pbastia:
Done:
- an API endpoint calling a report submission service looking for errors, and setting a report to "submitted"
- a backend validation method looking for a verification statement
- frontend error display


bciers/apps/reporting/src/app/components/signOff/reportValidationMessages.ts
shon-button, line 1: Love this!
**Suggestion:** I might place it in `bciers/apps/reporting/src/app/utils/reportValidationMessages.ts`

bciers/apps/reporting/src/app/components/signOff/SignOffPage.tsx
```diff
@@ -53,6 +46,14 @@ export default function SignOffPage({ version_id }: HasReportVersion) {
 
   const handleSubmit = async () => {
```
shon-button, line 47: **Suggestion:** Refactored the `handleSubmit` function to set setSubmitButtonDisabled on error and on submit.

shon-button (approved):

bc_obps/service/report_submission_service.py
```diff
@@ -0,0 +1,35 @@
...
+    """
+
+    @staticmethod
+    def validate_report(version_id: int) -> None:
+        """
+        Future implementation could create a specific exception housing all the issues a report would have
+        Django-ninja could then have a special way of parsing that error with a custom error code.
+        """
+        try:
+            ReportAttachment.objects.get(
```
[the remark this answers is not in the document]
pbastia: I think for the demo we'll keep the logic simple, I've added this work to https://github.com/bcgov/cas-reporting/issues/450

bciers/apps/reporting/src/app/components/signOff/SignOffPage.tsx
pbastia: I'd argue that it would prevent them from re-submitting without unclicking and clicking some checkboxes, if someone else uploaded the file in the meanwhile.
I think we can safely let a user retry, since the error was thrown by the server and not by the current page

PR #2603, opened 2024-12-17
Fix grid link

BCerki:
No card

This PR:
- changes the Link tag to an a because even with passHref, Link seemed flaky

### DIFF
--- a/bciers/libs/components/src/datagrid/cells/operations/OperationsActionCell.tsx
+++ b/bciers/libs/components/src/datagrid/cells/operations/OperationsActionCell.tsx
@@ -4,16 +4,16 @@ import { OperationStatus } from "@bciers/utils/src/enums";
 const OperationsActionCell = (isInternalUser: boolean) => {
   const renderCell = (params: GridRenderCellParams) => {
     let actionText = "View Operation";
-    let url = `../administration/operations/${params.row.id}?operations_title=${params.row.name}`;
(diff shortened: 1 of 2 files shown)


bciers/libs/components/src/datagrid/cells/operations/OperationsActionCell.tsx
```diff
@@ -1,11 +1,10 @@
 import { GridRenderCellParams } from "@mui/x-data-grid";
 import { OperationStatus } from "@bciers/utils/src/enums";
-import Link from "next/link";
 
 const OperationsActionCell = (isInternalUser: boolean) => {
   const renderCell = (params: GridRenderCellParams) => {
     let actionText = "View Operation";
-    let url = `/operations/${params.row.id}?operations_title=${params.row.name}`;
+    let url = `../administration/operations/${params.row.id}?operations_title=${params.row.name}`;
```
shon-button, line 7: @BCerki 
**Suggestion:** use absolute paths (/path) rather than upward path (../path) 

I believe the preferred pattern of navigating to a route(app) relative to the root of the current zone is to use absolute paths (/path) - as we do in the dashboard fixture links.

I think the use of upward path (../path) will only work within the current app's directory structure; so, if we ever want to route outside of bciers  directory structure this pattern would fail.

shon-button (approved):

PR #2618, opened 2024-12-18
Chore/431 allocation tests

shon-button:
Frontend and Backend tests for Allocation of Emissions
   - Also some minor changes to the Allocation model to bring its constraints in line with similar models

### DIFF
--- a/bc_obps/reporting/tests/service/test_report_emission_allocation_service.py
+++ b/bc_obps/reporting/tests/service/test_report_emission_allocation_service.py
@@ -1,5 +1,6 @@
 from decimal import Decimal
 from django.test import TestCase
+from reporting.models.emission_category import EmissionCategory
 from reporting.schema.report_product_emission_allocation import ReportProductEmissionAllocationsSchemaIn
 from reporting.models.report_emission import ReportEmission
 from reporting.models.report_fuel import ReportFuel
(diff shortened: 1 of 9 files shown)


bc_obps/reporting/tests/service/test_report_emission_allocation_service.py
```diff
@@ -0,0 +1,855 @@
...
+        )
+        report_product_2 = make(
+            ReportProduct,
+            id=4,
+            report_version=self.test_infrastructure.report_version,
+            facility_report=self.test_infrastructure.facility_report,
+            product_id=29,  # "Sugar: solid"
+        )
+
+        self.mock_get_response = {
```
dleard, line 98: @Sepehr-Sobhani @pbastia any thoughts on ways we could possibly cut down on the size of the mocked get repsonse / post payloads in this file? This service returns/ingests a pretty large object

dleard: maybe a class or function to generate the data objects not being tested (ie: where allocated emissions are 0)?

pbastia: @dleard I think this is something that will be improved when we decouple the service return types from the API schemas. Using the API data types here forces us to use those large JSON objects, whereas we could use data structures much simpler or easy to manipulate for the service use case

dleard (approved):

PR #2635, opened 2024-12-20
chore: remove double counting of reporting-only emissions in compliance summary service

dleard:
Compliance summary was considering emissions from _all_ emission categories when calculating the total allocated, but should have been only considering basic categories. This was highlighted by manual testing of expected values vs what the app is calculating, finding a .02% difference in the credited emissions. 
After the work in this PR, the expected values from manual calculations vs the app show a .00000004% difference in credited emissions when using the same values. This can be attributed to slight differences in rounding after the 4th decimal point, which will be handled later pending a BA decision around precision.


pbastia (approved):
🤯

acatchpole (approved):

PR #2637, opened 2024-12-20
Fix emission allocation bugs identified in UAT

acatchpole:
card: https://github.com/bcgov/cas-reporting/issues/468

#### Bug 1: Unable to submit when all emissions have been 100% allocated
The problem: This bug was due to a precision issue. During testing, a user had all emissions showing as 100% allocated, but was prevented from saving by a validation error indicating an incorrect allocation. In fact, the validation was working as intended. There was an over-allocation of emissions. BUT the amount over was .000049% of the total. And because the percentage display only uses 2-digit precision, it still showed as just 100.00%.

The fix: This problem really only exists for values that are very slightly off 100%. If it is more substantially off, then it would not show 100.00% and the problem allocation would be obvious. So i have added some edge case handling. For values that are >100 and <100.01, display 100.01. And for values <100 and >99.99, display 99.99. This is only for displaying to the users and does not affect allocation amounts or validation. It just makes it more obvious to users where an allocation error is.

image1 <correct allocation to 100.00%>
image

image2 <an over-allocation of ~.000001%>
image

#### Bug 2: The totals of the totals allocated to products does not equal the total amount of emissions
The problem: Excluded emission allocations were being improperly included in the fields showing the total emissions allocated to a product which led to a double counting of emissions. This error was not present for the overall total emissions value, and so a discrepancy was seen.
The fix: omit excluded emission allocations from the calculation of total emissions allocated to each product.

### To Test locally:
see #2553 and https://teams.microsoft.com/l/message/19:81544b60877c4a5eb4fc7946ecbcb525@thread.tacv2/1733854218662?tenantId=6fdb5200-3d0d-4a8a-b036-d3685e359adc&groupId=29821293-3244-455d-beae-e01d27762cfe for instructions on adding reportable and excluded emissions to be allocated

### DIFF
--- a/bciers/apps/reporting/src/app/components/facility/FacilityEmissionAllocationForm.tsx
+++ b/bciers/apps/reporting/src/app/components/facility/FacilityEmissionAllocationForm.tsx
@@ -42,6 +42,19 @@ interface FormData {
   allocation_methodology: string;
   allocation_other_methodology_description: string;
 }
+// Function that makes sure the percentage does not show 100 when it is not exactly 100
+const handlePercentageNearHundred = (value: number) => {
+  let res;
+  if (value > 100.0 && value < 100.01) {
+    res = 100.01;
+  } else if (value < 100.0 && value > 99.99) {
+    res = 99.99;
+  } else {
+    res = value;
+  }
+
+  return parseFloat(res.toFixed(4));
+};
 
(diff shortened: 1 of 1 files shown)


dleard (approved):
Nice work. I like the solution for the ~100% issue!

PR #461, opened 2023-12-21
Feat: 427 BC OBPS Regulated Operation ID

Sepehr-Sobhani:
[ZH CARD 427](https://app.zenhub.com/workspaces/team-moose-sprint-board-60d216175a2689000ec774d0/issues/gh/bcgov/cas-registration/427)

PR #2642, opened 2024-12-28
Fix: form submit blocked after error

pbastia:
### Work done:
- Fixes the forms holding onto errors and prevending resubmissions when the server returned an error once.
- Refactored the navigation logic in a `NavigationForm` component, to avoid having to implement the funky `useEffect` everywhere it's needed
- Updated the various forms to do display server errors

### Notable change:
- `submitHandler` needs to return a boolean to indicate whether the submission succeeded, server-side

Fixes:
- #461 the facility item improperly showing up on the operation review task list

### DIFF
--- a/bciers/apps/reporting/src/app/components/activities/ActivityForm.tsx
+++ b/bciers/apps/reporting/src/app/components/activities/ActivityForm.tsx
@@ -1,22 +1,16 @@
 "use client";
-import { customizeValidator } from "@rjsf/validator-ajv8";
-import { useEffect, useRef, useState } from "react";
+import { useEffect, useState } from "react";
 import { actionHandler } from "@bciers/actions";
-import { Alert } from "@mui/material";
-import ReportingTaskList from "@bciers/components/navigation/reportingTaskList/ReportingTaskList";
 import { TaskListElement } from "@bciers/components/navigation/reportingTaskList/types";
 import { FuelFields } from "./customFields/FuelFieldComponent";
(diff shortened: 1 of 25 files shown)


bciers/libs/components/src/form/NavigationForm.tsx
shon-button, line 1: `NavigationForm` seems to be causing FE errors:

image

bciers/apps/reporting/src/app/components/activities/ActivityForm.tsx
shon-button, line 1: `ActivityForm` can get stuck in an error loop when toggling checkmark field `General stationary combustion of waste without production of useful energy` and clicking `Submit`

1. Check field `General stationary combustion of waste without production of useful energy` 
image
2. Click `Submit` (returns error)
image
3. Uncheck field `General stationary combustion of waste without production of useful energy` 
image
4. Click `Submit`  (returns error)
image

shon-button:
Form `Save & Continue` can get stuck in an error loop when an initial error occurs and is ~not cleared?

Example: `VerificationForm` 
image

bciers/libs/components/src/form/NavigationForm.tsx
pbastia: I think this is a side-effect of having a `key` to the form itself to control re-rendering, good catch

pbastia:
@shon-button thanks for the review! I fixed the issues, but wasn't able to reproduce the last one on the verification form. It looks like some form elements didn't render on your screenshot, very odd

Issue #2624, opened 2024-12-19
Can't submit new contact or new operation


pbastia:
#### Describe the Bug:

A user can't submit a new contact or a new operation in the registration portal.

Developers:
There are errors thrown by RJSF on those forms, that the anyOf in the schema is not being properly used.

Reporting might have a linked error: https://github.com/bcgov/cas-reporting/issues/434

#### Impact: 5

<img src="https://www.getzephyr.com/sites/default/files/riksk%20matrix.jpg" alt="Bug Matrix" width="250" height="250"/>

#### Steps to reproduce the behaviour:

1. Login as an industry user with no prior operation/contact (for instance `bc-cas-dev-secondary` or `bc-cas-dev-three`)
2. Create a new operator
3. Create a new contact or operation
4. Hit submit
5. Nothing happens

#### Screenshots:
New Operation:
![Image](https://github.com/user-attachments/assets/141edc83-fd21-4efc-96b6-7bc0566a2ee1)

New Contact:
![Image](https://github.com/user-attachments/assets/b2c14087-7821-462f-9f57-507190a701e4)

#### Additional information:

Browser: [e.g. chrome, safari]
Version: [e.g. 22]

#### Additional Context
- Same behavior in both TEST and DEV environment

pbastia:
@fviduya 👀

fviduya:
Updated the issue description, added screenshots, issue exists on both DEV and TEST environment.

Sepehr-Sobhani:
I couldn’t replicate the issue locally or in the DEV environment. Maybe it’s already resolved? 🤔

fviduya:
Hmm, I was able to replicate it still on both dev and test envs. I used bc-cas-dev-secondary in DEV and created an operator (successful), then created an operation (unsuccessful, logs on console). On TEST env, I used bc-cas-test and got the same result.

patriciarussellCAS:
Is this a separate issue to ticket 463 in Reporting?

pbastia:
@patriciarussellCAS it's very likely the same. This one also mentions the Operations form, the cause is very likely the same

Sepehr-Sobhani:
> Hmm, I was able to replicate it still on both dev and test envs. I used bc-cas-dev-secondary in DEV and created an operator (successful), then created an operation (unsuccessful, logs on console). On TEST env, I used bc-cas-test and got the same result.

[...]

PR #2643, opened 2024-12-31
fix: populate anyOf only if we have options

Sepehr-Sobhani:
[ISSUE 2624](https://github.com/bcgov/cas-registration/issues/2624)


pbastia (approved):
Nice! We might want to add a test that we can render the page properly with an empty array

Sepehr-Sobhani:
> Nice! We might want to add a test that we can render the page properly with an empty array

@pbastia Thank you for reviewing this PR! We already had tests for the contact form, and we should have caught that error earlier if we had been using the correct schema to render the component. By using the appropriate schema, an error was raised, and I was able to fix it. I added an extra test for the operation form to ensure it's working as expected.

Issue #2414, opened 2024-10-24
Operations grid should show more details for BORO ID


BCerki:
#### Describe the task

The operation grid has a column for BORO ID. If an ID is assigned, its shown, otherwise "N/A" is shown. "N/A" isn't always accurate, so instead, we should show:
- The BORO ID if it exists
- "Pending" for any operation that's not an EIO, Reporting Operation, or Potential Reporting Operation
- "N/A" for EIOs, Reporting Operations, Potential Reporting Operations (they'll never get BORO IDs)

#### Acceptance Criteria

- [x] add conditions to `bciers/apps/administration/app/components/datagrid/models/operationColumns.ts`
- [x] vitests
- [ ] third

#### Additional context

- Add any other context about the task here.
- Or here

BCerki:
@zoeyli-46 , did you just want "N/A" replaced everywhere with "Pending"? "N/A" still makes sense to me for EIOs. Should anything else show based on any other conditions?

zoeyli-46:
N/A for EIOs is perfect, thanks Brianna!!

patriciarussellCAS:
Hey @BCerki sorry for not catching your question on this earlier. BORO ID's are only assigned to Regulated Operations - so this includes OBPS Regulated Operations, New Entrants and Opt-ins. 
Those who do NOT get a BORO ID include all the non-regulated operations - so this includes Reporting Operations, Potential Reporting Operations and EIOs. Let me know if you have any questions! Thanks.

BCerki:
> Hey @BCerki sorry for not catching your question on this earlier. BORO ID's are only assigned to Regulated Operations - so this includes OBPS Regulated Operations, New Entrants and Opt-ins. Those who do NOT get a BORO ID include all the non-regulated operations - so this includes Reporting Operations, Potential Reporting Operations and EIOs. Let me know if you have any questions! Thanks.

OK, thanks! I'm going to make a task ticket to update the "Issue BORO ID" button. That ticket only talked about EIOs, so the way it's implemented allows issusing a BORO ID to Reporting Operations, Potential Reporting Operations as well. Ticket: https://github.com/orgs/bcgov/projects/123/views/16?pane=issue&itemId=86401827&issue=bcgov%7Ccas-registration%7C2453

patriciarussellCAS:
Thanks @BCerki !

patriciarussellCAS:
Review notes: I can confirmed that these details are correct now. 
[...]

PR #2644, opened 2024-12-31
2414 operations grid shows more details for boro id

Sepehr-Sobhani:
[Issue 2414](https://github.com/bcgov/cas-registration/issues/2414)

### DIFF
--- a/bciers/apps/administration/app/components/datagrid/models/operations/operationColumns.ts
+++ b/bciers/apps/administration/app/components/datagrid/models/operations/operationColumns.ts
@@ -25,10 +25,17 @@ const operationColumns = (
           operation__bc_obps_regulated_operation: boroId,
           operation__registration_purpose: purpose,
         } = params?.row || {};
-        return (
-          boroId ||
-          (regulatedOperationPurposes.includes(purpose) ? "Pending" : "N/A")
-        );
+
+        switch (true) {
+          case !purpose:
+            return ""; // Return empty string if operation purpose is not defined
+          case !!boroId:
+            return boroId; // Return 'boroId' if it exists
+          case regulatedOperationPurposes.includes(purpose):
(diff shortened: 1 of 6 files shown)


bciers/apps/administration/app/components/datagrid/models/operations/operationColumns.ts
```diff
@@ -19,8 +20,16 @@ const operationColumns = (
       field: "operation__bc_obps_regulated_operation",
       headerName: "BORO ID",
       width: 120,
-      valueGetter: (params) =>
-        params.row?.operation__bc_obps_regulated_operation ?? "N/A",
+      valueGetter: (params) => {
```
BCerki, line 23: fancy

BCerki (commented):
Looks great!

A couple notes:
- If an operation isn't registered yet and doesn't have a purpose, we show "N/A". I'm not sure that makes sense because before registration, we don't know if the operation is eligible for a BORO ID or not. Maybe the column should be blank instead? @zoeyli-46?
- I noticed this^ with Operation 18 in the mock data, but that operation actually does have a status of registered, it's just missing a purpose. Do we still have a mock data PR in progress? It looks like a few of the mock operations have registration statuses/purposes that can't actually happen. We can always do this in a separate ticket if there's nothing currently open about mock data
- there's already a [ticket](https://github.com/orgs/bcgov/projects/123/views/15?filterQuery=-status%3Adone+filter&pane=issue&itemId=88018467&issue=bcgov%7Ccas-registration%7C2498) for this, but FYI sorting and filtering by "Pending" or "N/A" doesn't work: 
image

zoeyli-46:
Answer: show BORO ID column as blank if not yet registered

fviduya:
Yes, I've already added that as part of the mock ticket.

Issue #2563, opened 2024-12-10
Users shouldn't be able to add places assigned on the contacts form


BCerki:
#### Describe the Bug:

The places only field on the Contacts Page should be readonly. Users shouldn't be able to edit or delete places. Fix might be using the readonly widget/theme.
![Image](https://github.com/user-attachments/assets/92b76fa0-1f9e-4f50-b3da-7ead59a5ec43)

#### Probability:

(How likely the bug is to happen, scored from 1-5. For example, probability of 5 is something like "it happens to all users every time they log in." whereas probability of 1 "only happens to certain users when a really specific and unlikely path is followed.")

#### Impact:

(How bad the bug is when it does happen, scored from 1-5. For example, effect of 5 is "the entire app crashes and makes it unusable for all users" or "the bug causes the wrong data to be saved, with critical information (e.g. payment) being affected." whereas effect of 1 is "It makes some styling look a little bit weird.")

<img src="https://www.getzephyr.com/sites/default/files/riksk%20matrix.jpg" alt="Bug Matrix" width="250" height="250"/>

#### Steps to reproduce the behaviour:

1. Step #1
2. Step #2
3. Step #3

#### Additional information:

Browser: [e.g. chrome, safari]
Version: [e.g. 22]

#### Additional Context

-

patriciarussellCAS:
Review notes: I can still add and delete a contact in places. Should i NOT see the "add" button here? 
![Image](https://github.com/user-attachments/assets/2c5b96d8-7393-48e4-9854-d0b220a6191d)

vesselak:
Tested with bc-cas-dev 

![Image](https://github.com/user-attachments/assets/34a8e9d1-98e6-4652-b6d6-fc2e874b36cd)

Places assigned is read-only

![Image](https://github.com/user-attachments/assets/356f8cae-d80b-4e24-9e86-07740088c3bf)

PR #2648, opened 2025-01-03
2563 users should not be able to add places assigned on the contacts form

Sepehr-Sobhani:
[ISSUE 2563](https://github.com/bcgov/cas-registration/issues/2563)

### DIFF
--- a/bciers/apps/administration/app/data/jsonSchema/contact.ts
+++ b/bciers/apps/administration/app/data/jsonSchema/contact.ts
@@ -1,7 +1,7 @@
 import { RJSFSchema } from "@rjsf/utils";
 import provinceOptions from "@bciers/data/provinces.json";
 import SectionFieldTemplate from "@bciers/components/form/fields/SectionFieldTemplate";
-import { InlineArrayFieldTemplate } from "@bciers/components/form/fields";
+import { ArrayFieldTemplate } from "@bciers/components/form/fields";
 
 const section1: RJSFSchema = {
   type: "object",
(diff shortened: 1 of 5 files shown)


bciers/apps/administration/app/data/jsonSchema/contact.ts
```diff
@@ -1,7 +1,7 @@
 import { RJSFSchema } from "@rjsf/utils";
 import provinceOptions from "@bciers/data/provinces.json";
 import SectionFieldTemplate from "@bciers/components/form/fields/SectionFieldTemplate";
-import { InlineArrayFieldTemplate } from "@bciers/components/form/fields";
+import { ArrayFieldTemplate } from "@bciers/components/form/fields";
```
BCerki, line 4: Why is it better to use this one?

BCerki (approved):
Looks great!
Could add a regression vitest

bciers/apps/administration/app/data/jsonSchema/contact.ts
Sepehr-Sobhani: I'm not entirely sure why it was originally used, but these two field templates are quite similar. When I used them interchangeably, they appeared identical. The only noticeable difference was the note prop, which I assume we would also need in the ArrayField. That said, I aimed to minimize the usage of InlineArrayFieldTemplate and instead reuse ArrayFieldTemplate.

Additionally, the removable prop on the ArrayFieldTemplate perfectly aligned with the requirements of this feature. I avoided modifying InlineArrayFieldTemplate to replicate the same removable prop behaviour, as it seemed unnecessary.

Issue #2040, opened 2024-08-02
Reg\Transfer Entity\Cronjob to apply transfers on effective dates


andrea-williams:
#### Describe the task

~Blocked by: https://github.com/bcgov/cas-registration/issues/2427~

Create a cronjob that runs on a daily schedule in Openshift. It should then handle each event effective for the current date and make the appropriate changes in our database. 

#### Acceptance Criteria

To implement this functionality, we need to follow the approach used in the `cas_cif` application:

- [x] Create a django admin command that calls `process_due_transfer_events` function

- [x] Define a CronJob (e.g., process-due-transfer-events) in a Helm chart template to encapsulate the task logic. We need to call the django command in this CronJob

- [x] Integrate the CronJob into an Airflow DAG within the appropriate DAG file.

- [x] **Pierre**: we'll also need to add a helm dependency on the `cas-airflow-dag-trigger` chart with a `download-dags` and a dag configuration in the make install target (cif is a great example again). This makes sure that airflow always has the latest dag version from the deployed commit

For reference, a similar CronJob named insert-backup-test-timestamp is already implemented and utilized in the `cas_cif_dags.py` file. Following this pattern, we can schedule the new functionality to invoke the process_due_transfer_events function on a daily basis.

#### Additional context

- Add any other context about the task here.
- Or here

patrickisaac:
Refinement Notes:
- @Sepehr-Sobhani to further refine and have additional conversation with Pierre/Dylan

Sepehr-Sobhani:
> Refinement Notes:
> 
> * [@Sepehr-Sobhani](https://github.com/Sepehr-Sobhani) to further refine and have additional conversation with Pierre/Dylan

Refined!

Sepehr-Sobhani:
Hi @pbastia / @dleard !

Could you please review the AC for this ticket? I want to ensure everything is covered.

Thanks!

PR #2655, opened 2025-01-07
2040 transfer event cronjob

Sepehr-Sobhani:
[ISSUE 2040](https://github.com/bcgov/cas-registration/issues/2040)


dleard (approved):
:100:

PR #2662, opened 2025-01-09
chore: give industry_user create access to contacts

BCerki:
card: https://github.com/orgs/bcgov/projects/123/views/16?pane=issue&itemId=91260610&issue=bcgov%7Ccas-registration%7C2605

This PR:
- adds a `page.tsx` for add-contacts for the industry_user (reporter) role
- removes conditionality that only allowed admin industry users to add contacts


Sepehr-Sobhani (approved):
🤩

PR #2667, opened 2025-01-09
Put operator and operation in contacts grid

BCerki:
card: https://github.com/orgs/bcgov/projects/123/views/16?pane=issue&itemId=88763967&issue=bcgov%7Ccas-registration%7C2525

This PR:
- Puts the operator in the contacts grid. It's a m2m field so I've used the same strategy with `distinct` as we used in the transfers grid
- Enabled sorting and filtering for this colu,n--I didn't see any reason not to
- add more readable names to the baker recipes to help me debug, and I left them because they'll probably be helpful in the future too

### DIFF
--- a/bc_obps/service/tests/data_access_service/test_data_access_contact_service.py
+++ b/bc_obps/service/tests/data_access_service/test_data_access_contact_service.py
@@ -10,6 +10,7 @@ from registration.tests.utils.bakers import (
 )
 from registration.models.user_operator import UserOperator
 from service.data_access_service.contact_service import ContactDataAccessService
+from model_bakery import baker
 
 pytestmark = pytest.mark.django_db
 
@@ -17,9 +18,9 @@ pytestmark = pytest.mark.django_db
 class TestDataAccessContactService:
     @staticmethod
     def test_list_contacts_for_irc_user():
-        contact_baker(_quantity=10)
-        cas_admin = user_baker({'app_role': AppRole.objects.get(role_name='cas_admin')})
-        assert ContactDataAccessService.get_all_contacts_for_user(cas_admin).count() == 10
+        baker.make_recipe('utils.contact', _quantity=10)
+        user = baker.make_recipe('utils.cas_admin')
+        assert ContactDataAccessService.get_all_contacts_for_user(user).count() == 10
 
     @staticmethod
     def test_list_contacts_for_industry_user():
--- a/bc_obps/service/tests/test_contact_service_v2.py
+++ b/bc_obps/service/tests/test_contact_service_v2.py
@@ -1,0 +1,33 @@
+from registration.schema.v1.contact import ContactFilterSchema
+import pytest
+from service.contact_service_v2 import ContactServiceV2
+from model_bakery import baker
+
+pytestmark = pytest.mark.django_db
+
+
+class TestListContactService:
+    @staticmethod
+    def test_list_contacts():
+
+        user = baker.make_recipe('utils.cas_admin')
+        contact1 = baker.make_recipe('utils.contact')
+        contact2 = baker.make_recipe('utils.contact')
+
+        operator1 = baker.make_recipe('utils.operator')
+
+        operator2a = baker.make_recipe('utils.operator')
+        operator2b = baker.make_recipe('utils.operator')
+
+        # contact 1 is associated with one operator, count = 1
+        operator1.contacts.set([contact1])
+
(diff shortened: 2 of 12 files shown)


bc_obps/service/tests/data_access_service/test_data_access_contact_service.py
```diff
@@ -10,14 +10,60 @@
...
+        # contact 3 is for an operator and two of its operations count = 4
+        operator3.contacts.set([contact3])
+        operation3a.contacts.set([contact3])
+        operation3b.contacts.set([contact3])
+
+        # contact 4 belongs to two operators count = 6
+        operator4a.contacts.set([contact4])
+        operator4b.contacts.set([contact4])
+
+        # contact5 belongs to two operators and operations count = 10 (comes back as 5 records with each combination of operator/operation)
```
BCerki, line 61: This one hurt my brain a little

bc_obps/service/tests/data_access_service/test_data_access_contact_service.py
```diff
@@ -10,14 +10,60 @@
...
+        # contact 3 is for an operator and two of its operations count = 4
+        operator3.contacts.set([contact3])
+        operation3a.contacts.set([contact3])
+        operation3b.contacts.set([contact3])
+
+        # contact 4 belongs to two operators count = 6
+        operator4a.contacts.set([contact4])
+        operator4b.contacts.set([contact4])
+
+        # contact5 belongs to two operators and operations count = 10 (comes back as 4 records with each combination of operator/operation)
```
BCerki, line 61: This one hurt my brain a little, does this make sense to the reviewer?

bc_obps/service/tests/test_contact_service_v2.py
```diff
@@ -0,0 +1,33 @@
...
+
+        operator2a = baker.make_recipe('utils.operator')
+        operator2b = baker.make_recipe('utils.operator')
+
+        # contact 1 is associated with one operator, count = 1
+        operator1.contacts.set([contact1])
+
+        # contact 2 belongs to two operators count = 3
+        operator2a.contacts.set([contact2])
+        operator2b.contacts.set([contact2])
```
andrea-williams, line 27: hmmm this has raised a red flag for me (I know that this isn't something you introduced, turns out we've had our db structured this way for a long time). I don't think our app should allow for the same contact to be assigned to multiple operators - that could mean that Operator2a modifies/deletes contact data and then Operator2b either sees updated contact data or loses data.

we've got M2M fields configured on the Operator model for both contacts and documents, when they should IMO be Many-to-1 relationship. One operator shouldn't ever be able to access another operator's documents or contacts. Do you agree?

BCerki: Yes, I agree. I can't think of why documents was m2m; that just seems like a mistake. For contacts, I think we might have originally set them up as m2m because we thought a consultant might work with multiple operators. But I don't think that makes sense with how the app works anymore.

andrea-williams: created tech debt ticket https://github.com/bcgov/cas-registration/issues/2680

andrea-williams (commented):
As discussed in call - width of email address column, and IDs of contacts are causing links to wrong contact pages.

bciers/apps/administration/app/components/contacts/ContactsDataGrid.tsx
```diff
@@ -46,13 +46,22 @@ const ContactsDataGrid = ({
     [SearchCell, isExternalUser],
   );
 
+  // the mui grid requires a unique id, and since we can multiple rows with the same contact for different operators, we can't use the contact's id alone
+  function getRowId(row: ContactRow) {
```
BCerki, line 50: @Sepehr-Sobhani , FYI I just learned about [this prop](https://mui.com/x/react-data-grid/row-definition/?srsltid=AfmBOoqOYFCcKdcLzSipS6QXGGlQITIcWO6fE-YNcPONj-GYtu3LEGk0#:~:text=By%20default%2C%20the,where%20it%27s%20located.). We could have used it in the transfers grid instead of autogenerating a random uuid

Sepehr-Sobhani: That's great, Thanks for sharing this! I can replace it with the existing code since I'm working on that grid too 😃

andrea-williams (approved):
Looks 🔥!

PR #2670, opened 2025-01-10
fix: Add minimum to most number fields and functions to validate numbers in activityForm oneOf arrays

gdalcengio:
#470


dleard (approved):
:100: Nice work!

PR #2676, opened 2025-01-11
Refactor operation review page to avoid relying on client side effects

pbastia:


dleard:
PR has become outdated. Work will be complete when the https://github.com/bcgov/cas-reporting/issues/533 ticket gets picked up

PR #2679, opened 2025-01-13
Replace operation rep

BCerki:
https://github.com/orgs/bcgov/projects/123/views/16?pane=issue&itemId=88758186&issue=bcgov%7Ccas-registration%7C2524

Notes: 
- I've used a url param to decide whether or not to show the alert on the contact page. As far as I could find online (e.g. https://medium.com/@awayyao/keep-page-state-between-page-navigation-in-nextjs-13-using-app-router-schema-f6eacd0300ad), the next router doesn't store any state or history. window.history doesn't store the previous url, and document.referer only gives the origin.
- Updated the mock data so that contacts are more representative of how the app works (e.g., operations can only have contacts that belong to their operators; all registered operations have to have a contact)
- because of ^, I noticed you can get to the operation administration page before the operation is registered if you go via the URL. I added a check in the endpoint
- I wrote a couple comments in the card about where I haven't followed the card exactly

### DIFF
--- a/bc_obps/registration/schema/v2/operation.py
+++ b/bc_obps/registration/schema/v2/operation.py
@@ -112,6 +112,10 @@ class OperationInformationIn(ModelSchema):
         fields = ["name", 'type']
 
 
+class OperationInformationInUpdate(OperationInformationIn):
+    operation_representatives: List[int]
+
+
 class OptedInOperationDetailOut(ModelSchema):
     class Meta:
         model = OptedInOperationDetail
@@ -153,6 +157,11 @@ class OperationOutV2(ModelSchema):
     date_of_first_shipment: Optional[str] = None
     new_entrant_application: Optional[str] = None
     bcghg_id: Optional[str] = Field(None, alias="bcghg_id.id")
+    operation_representatives: Optional[List[int]] = []
+
+    @staticmethod
+    def resolve_operation_representatives(obj: Operation) -> List[int]:
+        return list(obj.contacts.filter(business_role='Operation Representative').values_list('id', flat=True))
 
     @staticmethod
     def resolve_multiple_operators_array(obj: Operation) -> List[MultipleOperator]:
@@ -176,7 +185,17 @@ class OperationOutV2(ModelSchema):
 
     class Meta:
         model = Operation
-        fields = ["id", 'name', 'type', 'opt_in', 'regulated_products', 'status', 'activities', 'opted_in_operation']
+        fields = [
+            "id",
+            'name',
+            'type',
+            'opt_in',
+            'regulated_products',
+            'status',
+            'activities',
+            'opted_in_operation',
+            'contacts',
+        ]
         from_attributes = True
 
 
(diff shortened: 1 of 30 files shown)


bc_obps/service/operation_service_v2.py
```diff
@@ -198,6 +199,22 @@ def create_or_update_operation_v2(
 
         # set m2m relationships
         operation.activities.set(payload.activities)
+        if isinstance(payload, OperationInformationInUpdate):
```
BCerki, line 202: This (vs if payload.operation_representatives) was to make mypy happy

bc_obps/service/operation_service_v2.py
```diff
@@ -330,12 +347,6 @@ def update_operation(
             payload,
             operation_id,
         )
-
-        if payload.regulated_products:
```
BCerki, line 334: This is already handled in `create_or_update_operation_v2`

bciers/apps/administration/tests/components/operations/OperationInformationForm.test.tsx
```diff
@@ -162,7 +190,7 @@ const formData = {
   naics_code_id: 1,
   secondary_naics_code_id: 2,
   operation_has_multiple_operators: true,
-  activities: [1, 2, 3, 4, 5],
+  activities: [1, 2],
```
BCerki, line 193: We only mocked two activities

Sepehr-Sobhani (approved):
This is dope! 🔥 I’ve left a few small suggestions, but nothing major.

bc_obps/registration/schema/v2/operation.py
```diff
@@ -194,7 +194,7 @@ class Meta:
             'status',
             'activities',
             'opted_in_operation',
-            'contacts',
+            # 'contacts',
```
Sepehr-Sobhani, line 197: leftover!

bc_obps/service/contact_service_v2.py
```diff
@@ -48,17 +33,12 @@ def list_contacts_v2(
...
     @classmethod
-    def get_if_authorized_v2(cls, user_guid: UUID, contact_id: int) -> Optional[Contact]:
-        user = UserDataAccessService.get_by_guid(user_guid)
-        user_contacts = ContactDataAccessService.get_all_contacts_for_user(user)
-        contact = user_contacts.filter(id=contact_id).first()
-        if user.is_industry_user() and not contact:
-            raise Exception(UNAUTHORIZED_MESSAGE)
-        return contact
+    def get_by_id(cls, contact_id: int) -> Contact:
+        return Contact.objects.get(id=contact_id)
```
Sepehr-Sobhani, line 37: We already have this in `ContactDataAccessService`.

Sepehr-Sobhani (approved):

bc_obps/service/contact_service_v2.py
BCerki: :facepalm:

Issue #2338, opened 2024-10-07
Admin\Contacts - make all users who have an operator contacts by default


andrea-williams:
#### Describe the task

Based on discussion with business area & PR, all external users should automatically be made a Contact. 
As an industry user, when I go to my contacts list, anyone who is a user in my operator will show up in my contacts list.

#### Acceptance Criteria

- [x] add logic so that whenever an external user's access request for an operator is approved, a new Contact record is created using the available info from the user's profile. This should include adding FK references for the contact record and operator record in the `operator_contacts` table. (This will require making some fields in the Contact table, such as address, optional).
- [x] add the new contact to the operator
- [x] remove the toggle "Is this contact a user?" from the form for creating a new Contact

#### Additional context

- the address field will be optional on the Contact table in the database, but should remain a required field on the frontend
- separate ticket for writing data migration script that will create new contact records for all existing external users in prod who don't already have corresponding contact records: #2339

zoeyli-46:
@BCerki "As an industry user, when I go to my contacts list, anyone who is a user in my operator will show up in my contacts list." I added this to the description

BCerki:
If contacts should be deleted if a user's access is revoked (?), we could add that to https://github.com/orgs/bcgov/projects/123/views/16?filterQuery=repo%3Abcgov%2Fcas-registration+-status%3Adone+delete+contact&pane=issue&itemId=80445851&issue=bcgov%7Ccas-registration%7C1606

andrea-williams:
I suspect that we'll need to include an option for hard-deleting a user/contact (thus deleting their personal info) but I don't know that for sure, and don't know whether it needs to be included in MVP. 

@patriciarussellCAS or @vesselak likely know more about this than me.

BCerki:
To test this:
- login as bc-cas-dev
- go to the contacts grid and note bc-cas-dev-secondary is not there
- go to access requests and approve bc-cas-dev-secondary
- go back to the contacts grid and bc-cas-dev-secondary is now in the grid.

fviduya:
I have tested this in the DEV environment, [here is the test document for reference](https://bcgov-my.sharepoint.com/:w:/g/personal/francesca_viduya_gov_bc_ca/Eb93Q62oL3pCskH2AG_kLL0B1oGB6_ypL2Hql81MsfrQOQ?e=S1fFl4).

I can confirm that
- [x] Upon approval in ```Users and Access Requests```, a contact is created under the Contacts page for the approved user
[...]

PR #2710, opened 2025-01-18
Make all users contacts by default

BCerki:
card: https://github.com/bcgov/cas-registration/issues/2338


bc_obps/service/user_operator_service_v2.py
```diff
@@ -109,3 +115,80 @@ def list_user_operators_v2(
...
+
+        user_operator.status = payload.status  # type: ignore[attr-defined]
+        updated_role = payload.role
+
+        if user_operator.status in [UserOperator.Statuses.APPROVED, UserOperator.Statuses.DECLINED]:
+            user_operator.verified_at = datetime.now(ZoneInfo("UTC"))
+            user_operator.verified_by_id = admin_user_guid
+
+            if user_operator.status == UserOperator.Statuses.DECLINED:
+                # Set role to pending for now but we may want to add a new role for declined
```
BCerki, line 148: I brought this over from the v1 version of the service

bc_obps/registration/tests/endpoints/v2/_user_operators/_user_operator_id/test_user_operator_update_status.py
```diff
@@ -38,14 +35,14 @@ def test_industry_user_approves_access_request(self, mocker):
...
         )
-        # Assert that the email notification was called
-        mock_send_operator_access_request_email.assert_called_once_with(
-            AccessRequestStates.APPROVED,
-            AccessRequestTypes.OPERATOR_WITH_ADMIN,
-            operator.legal_name,
-            subsequent_user_operator.user.get_full_name(),
-            subsequent_user_operator.user.email,
-        )
+        # # Assert that the email notification was called
```
BCerki, line 38: @andrea-williams , the email tests fail now that I've created a v2 version of the service. Does that make sense?

andrea-williams: hmmm I'm not sure. Will investigate this more tmr.

bc_obps/service/user_operator_service_v2.py
```diff
@@ -109,3 +115,80 @@ def list_user_operators_v2(
...
+                operator_business_guid = None
+            if operator_business_guid != admin_user.business_guid:
+                raise PermissionError("Your user is not associated with this operator.")
+
+        user_operator.status = payload.status  # type: ignore[attr-defined]
+        updated_role = payload.role
+
+        if user_operator.status in [UserOperator.Statuses.APPROVED, UserOperator.Statuses.DECLINED]:
+            user_operator.verified_at = datetime.now(ZoneInfo("UTC"))
+            user_operator.verified_by_id = admin_user_guid
```
andrea-williams, line 145: could use `set_verification_columns()` here instead

bc_obps/registration/tests/endpoints/v2/_user_operators/_user_operator_id/test_user_operator_update_status.py
andrea-williams: tbh I'm feeling suspicious of this. I played around with a code a bit locally and was getting the same failure where it's not being called, but I can't think why. @Sepehr-Sobhani do you have time to take a quick look and see if you can spot what's going on?

andrea-williams: if it's not a quick debug/fix, I'm happy to leave it for now and call it a later problem, since we're not ready to implement the email notifications for Reg2 yet anyway.

Sepehr-Sobhani: Instead of `service.user_operator_service.send_operator_access_request_email` use `service.user_operator_service_v2.send_operator_access_request_email` and it will work.

andrea-williams (approved):
Ran locally. Logged in as bc-cas-dev and confirmed that after approving bc-cas-dev-secondary as a user, they automatically appeared in the Contact table when they weren't there before.

Would recommend adding testing notes in either the PR or the card that basically explain that ^. Feedback received from POs is that they sometimes have to spend a long time figuring out how to test a PR before they can approve it, so testing notes like this could probably help.

PR #2656, opened 2025-01-07
Feat: 383 change report type backend

pbastia:
Frontend:
- API call when the user confirms the modal to change report type
- Redirection to the new report version

Backend:
Upon report type change, the system actually removes an entire report version, and recreates a new one with a different ID
- API endpoint
- Service capabilities
- Model update

PR #2717, opened 2025-01-21
Feat/383 report type change rebased

dleard:
Continued from #2656.
Fixed merge conflicts & updated some incorrectly set on_delete=PROTECT statements on some models that should cascade delete when a version is deleted.


dleard:
closes #2656

ayeshmcg (approved):

PR #2718, opened 2025-01-21
509 alumina emission doesn't have default

gdalcengio:
[509](https://github.com/bcgov/cas-reporting/issues/509)


ayeshmcg (approved):
Nice job, @gdalcengio 
The emission form is now displaying by default as expected.

PR #2720, opened 2025-01-22
352 filter out Facilities crumb from breadcrumb component

gdalcengio:
[352](https://github.com/bcgov/cas-reporting/issues/352)


ayeshmcg (approved):
Good work @gdalcengio

Issue #2502, opened 2024-11-21
Test the user role / permission matrix


pbastia:
#### Describe the task

After adding new roles in (#2493 , #2494, #2497 )
Blocked by #2522 

We need to test that the user roles can actually only perform the operations allowed.

This could be:
- an additional test, verifying each endpoint with each role name
- an extension of the API authorization test

#### Acceptance Criteria

- [ ] Add a new test to `TestEndpointPermissions` that validates each role has access only to its permitted endpoints as specified in `endpoints_to_test`. The test should iterate over all roles and verify they do not have access to endpoints outside their allowed list.
Validation Across All Roles and Endpoints

- [ ] Implement the test to dynamically evaluate every endpoint against every role, ensuring that unauthorized access attempts fail for each role and endpoint combination not explicitly defined in `endpoints_to_test`.

patrickisaac:
Refinement Notes:
- @Sepehr-Sobhani to further refine ticket

Sepehr-Sobhani:
@patrickisaac refined!

patriciarussellCAS:
@fviduya

patriciarussellCAS:
Once we can confirm that the user roles in the system have been set up properly, we can get the BA to do a round of UAT in -test to confirm everything works as they planned. This will be required for the FRCR.

PR #2722, opened 2025-01-22
chore: test user role permission

fviduya:
This test 
- verifies that roles do NOT have access to endpoints outside of their allowed list
- ensures that unauthorized access attempts fail for each role and endpoint combination

#2502


fviduya:
I was still getting 400, 422, and 404 for some endpoints, I found that it happens when I was checking for 200 response. I also was having an issue with this endpoint: submit_report_version, the response only has this ```response={200: int, 400: Message},``` should it also have ```custom_codes_4xx: Message``` instead of 400?

fviduya:
I had to change the filename of endpoints file because pythontests was failing due to the filename. Would it be better to have the file transferred to utils folder?

Sepehr-Sobhani (approved):
🤩 Great job!

PR #2729, opened 2025-01-23
485 initial rls setup - PART 1

Sepehr-Sobhani:


bc_obps/rls/middleware/rls.py
```diff
@@ -0,0 +1,43 @@
...
+                raise RuntimeError(
+                    "CurrentUserMiddleware must be placed before RlsMiddleware in MIDDLEWARE settings."
+                )
+
+    def __call__(self, request: HttpRequest) -> HttpResponse:
+        user: Optional[User] = getattr(request, 'current_user', None)
+        if user:
+            self._set_user_context(user.user_guid)
+        else:
+            logger.info("Anonymous user detected, skipping user context setup", exc_info=True)
```
dleard, line 34: Just a note for later: when we get to actually setting roles, we'll want to still set a role here. We might need an "Unauthenticated" role to set in this case.

dleard (approved):
Nice work! Just left a note for later around an 'Unauthenticated' role

PR #2735, opened 2025-01-25
Feat/og extraction ncnp activity

dleard:
Adds all the configs & schema data for activity `Non-compression and non-processing activities that are oil and gas extraction and gas processing activities`

To test this, create a report with this activity selected & go to the activities page.
**Note:** Because the activity name is long, it gets cut off in the dropdown & so there are 2 that say `Non-compression and non-processing activities...` Choose the second one.


ayeshmcg (approved):
@dleard Excellent work with an activity page with these many source types a

PR #2737, opened 2025-01-27
chore: upgrade vite

BCerki:
no card, fixing yarn audit


BCerki:
Handled in a different PR

PR #2758, opened 2025-01-30
chore: wrap text in chips in MultiSelectWidget

gdalcengio:
[430](https://github.com/bcgov/cas-reporting/issues/430)

### DIFF
--- a/bciers/libs/components/src/form/widgets/MultiSelectWidget.tsx
+++ b/bciers/libs/components/src/form/widgets/MultiSelectWidget.tsx
@@ -109,10 +109,17 @@ const MultiSelectWidget: React.FC<WidgetProps> = ({
             <Chip
               {...getTagProps}
               key={option.id}
-              label={option.label}
+              label={
+                <div style={{ whiteSpace: "normal", padding: ".5rem 0" }}>
+                  {option.label}
+                </div>
+              }
               {...getTagProps({
                 index,
               })}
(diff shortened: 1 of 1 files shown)


bciers/libs/components/src/form/widgets/MultiSelectWidget.tsx
shon-button, line 1: @gdalcengio 
Looks great...one point, for your consideration, instead of using inline styles for the Chip label, a better practice might be to use Material-UI's `sx` prop to provide a more theme-aware approach and to remove the need for  unnecessary `div` inside the label prop.

Perhaps, something like:
```
label={option.label}
              sx={{
                height: "auto", // Allow chip to grow if needed
                maxWidth: "150px", // Prevents overflow (adjust as needed)
                "& .MuiChip-label": {
                  whiteSpace: "normal", // Allows text wrapping
                  wordBreak: "break-word", // Breaks long words if needed
                  overflowWrap: "break-word", // Ensures proper word breaks
                  display: "block", // Forces multiline wrapping
                },
              }}
```

shon-button (approved):

bciers/libs/components/src/form/widgets/MultiSelectWidget.tsx
gdalcengio: Didn't realize this was an option, thank you!

PR #2759, opened 2025-01-30
Feat/set audit fields via trigger

dleard:
There are a **ton** of files changed in this PR. _Most_ of the changes are either adding triggers to models by inheriting the Meta of the Timestamped Model, removing the set_create_or_update() function calls that have been replaced by the triggers or updating tests.

The "meat" of the work done in this PR to apply triggers is in this one file:
- registration/models/time_stamped_model.py

Most of the relevant tests are in these files:
- conftest.py (one file in each moduel's "test" directory)
- common/tests/utils/helpers.py
- registration/tests/models/test_timestamped_model.py

Bug fix in:
- service/user_operator_service_v2.py

This could be manually tested in the python shell by setting the my.guid value to an existing user's in postgres with cursor. Or by logging into the app & creating/updating anything (report, operation, operator, etc..) & looking at the created_by/updated_by values in the db


bc_obps/reporting/tests/service/test_report_activity_save_service/infrastructure.py
```diff
@@ -45,14 +44,12 @@ def build(cls):
...
         )
+        t.facility_report.refresh_from_db()
         t.report_version = t.facility_report.report_version
         t.user = make_recipe('registration.tests.utils.industry_operator_user')
-        t.configuration = make_recipe(
-            "reporting.tests.utils.configuration",
-            valid_from=datetime(1999, 1, 1, tzinfo=timezone.utc),
-            valid_to=datetime(1999, 12, 31, tzinfo=timezone.utc),
+        t.configuration = Configuration.objects.get(
+            valid_from__lte=t.facility_report.created_at, valid_to__gte=t.facility_report.created_at
```
pbastia, line 52: I'm not a big fan of tests that depend on what time it is when they run...But as things are right now I don't think it's worth getting into mocking database time for that

bc_obps/reporting/tests/service/test_report_attachment_service.py
```diff
@@ -42,23 +42,6 @@ def test_set_attachment_raises_if_attachment_type_not_in_database(self):
...
-
-        assert ReportAttachment.objects.count() == 1
-
-        r = ReportAttachment.objects.first()
-        assert r.attachment_type == "verification_statement"
-        assert r.attachment_name == "test_file.txt"
-        assert r.attachment is not None
-        assert r.created_by == self.user
-        assert r.created_at is not None
-
```
pbastia, line 61: This test is still needed, we can just remove lines 59 and 60

pbastia (approved):
Just a couple of nitpicks. Amazing work!

PR #2765, opened 2025-01-30
chore:fixed the bug for regulated products

ayeshmcg:
Card: https://github.com/bcgov/cas-reporting/issues/524
Changes:-
Added Regulated Products on saving reports


pbastia (approved):
Very nice! We should also write a small test for this change, I don't think that method is tested at all. I'm approving it; conditional to having that test in place

PR #2771, opened 2025-01-31
Task/change reg purpose frontend/2169

andrea-williams:
Notes on Brianna's additions:
- The errors @andrea-williams was encountering were related to mocking. Even though the mocks were set up correctly, vitest wasn't picking them up (logged them in the code and they came out as real functions instead of spys). Vitest can be touchy about mocking the same file/function [more than once](https://teams.microsoft.com/l/message/19:81544b60877c4a5eb4fc7946ecbcb525@thread.tacv2/1733871611295?tenantId=6fdb5200-3d0d-4a8a-b036-d3685e359adc&groupId=29821293-3244-455d-beae-e01d27762cfe&parentMessageId=1733871611295&teamName=External%3A%20CAS%20-%20Clean%20Growth%20Digital%20Service%20Team&channelName=Developers&createdTime=1733871611295&ngc=true), and I think our two `fetchFormEnums` helpers (one in admin, one in reg) were colliding with each other 
- To address ^, I moved all the mocks that were used in `fetchFormEnums` to the shared bciers mocks/global files
- I also combined the two separate `fetchFormEnums` into a single function in the shared bciers file. The admin/reg returned different mock values, so I added an `App` arg so I could keep the different mock values without refactoring any tests


BCerki:
Completed in a different PR, closing

PR #2772, opened 2025-01-31
Feat: 343 activity ng transmission non compression

pbastia:


dleard (approved):
Love the tests/uiSchema stuff! One small change needed to a source type & this is good to go.

PR #2775, opened 2025-01-31
EXAMPLE ONLY, DON'T MERGE

BCerki:


BCerki:
Closing, bug is no longer reproducible

PR #2783, opened 2025-02-03
507 bug emission allocation fails with other

acatchpole:
Card: https://github.com/bcgov/cas-reporting/issues/507

#### Part 1: Allocations not saving when methodology == 'other'

Updated model to expect 'Other'
Updated tests to use 'Other'
Added test to verify that 'Other' works and 'other' does not

#### Part 2: Missing methodology on page but error still shows "need 100% allocation"

Added validations and errors for methodology and methodology_description fields

### DIFF
--- a/bc_obps/reporting/tests/models/test_report_product_emission_allocation.py
+++ b/bc_obps/reporting/tests/models/test_report_product_emission_allocation.py
@@ -28,7 +28,7 @@ class ReportProductEmissionAllocationModelTest(BaseTestCase):
             report_product=report_product,
             emission_category=emission_category,
             allocated_quantity=Decimal('300.4151'),
-            allocation_methodology='other',
+            allocation_methodology='Other',
             allocation_other_methodology_description='Test description',
         )
         cls.field_data = [
@@ -56,7 +56,7 @@ class ReportProductEmissionAllocationModelTest(BaseTestCase):
 
         with pytest.raises(
             ValidationError,
-            match="A value for allocation_other_methodology_description must be provided if the allocation_methodology is 'other'",
+            match="A value for allocation_other_methodology_description must be provided if the allocation_methodology is 'Other'",
         ):
             make(
                 ReportProductEmissionAllocation,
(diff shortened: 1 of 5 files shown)


bc_obps/reporting/tests/models/test_report_product_emission_allocation.py
```diff
@@ -69,6 +69,21 @@ def test_allow_null_description_if_methodology_is_not_other(self):
...
+                ReportProductEmissionAllocation,
+                report_version=facility_report.report_version,
+                facility_report=facility_report,
+                report_product=report_product,
+                emission_category=emission_category,
+                allocated_quantity=Decimal('300.4151'),
+                allocation_methodology='other',
+                allocation_other_methodology_description='test',
+            )
+
```
pbastia, line 86: since 'other' is not in the choices anymore, this will fail on a django model error. It's not really something we need to test - that is covered by the unit tests of the django framework.

we can safely remove this code

bciers/apps/reporting/src/app/components/facility/FacilityEmissionAllocationForm.tsx
```diff
@@ -52,6 +52,43 @@ const validateEmissions = (formData: FormData): boolean => {
...
+    newErrors.push(errorMismatch);
+  }
+  if (!validateMethodology(formData)) {
+    newErrors.push(errorMethodology);
+  }
+  if (!validateMethodologyOther(formData)) {
+    newErrors.push(errorMethodologyOther);
+  }
+
+  return newErrors;
```
pbastia, line 89: Very nice error handling and user feedback!

bciers/apps/reporting/src/app/components/facility/FacilityEmissionAllocationForm.tsx
```diff
@@ -97,10 +133,9 @@ export default function FacilityEmissionAllocationForm({
...
   // 🔄 Check for allocation mismatch on page load to prevent submit
   useEffect(() => {
-    if (!validateEmissions(formData)) {
-      setErrors([errorMismatch]);
-    }
-    setSubmitButtonDisabled(!validateEmissions(formData));
+    const newErrors = validateFormData(formData);
+    setErrors(newErrors);
+    setSubmitButtonDisabled(newErrors.length > 0);
   }, [formData]);
```
pbastia, line 139: this one might be a question for @zoeyli-46 : Any of our other components show errors when the user tries to click 'submit', but in this case, the errors will show instantly, including on a blank form. Is this the expected behaviour?

pbastia (commented):
Very nice! I've left a question that is more of a UX question than a developer's one

pbastia (approved):

Issue #344, opened 2023-11-25
Error in UserOperator form when mailing address the same as physical address


andrea-williams:
Describe the Bug: In cases where an operator's mailing address is the same as its physical address, the physical address data is not being propagated to the mailing address fields in the DB

Probability (how likely the bug is to happen, scored from 1-5): 4
(For example, probability of 5 is something like "it happens to all users every time they log in."
Probability of 1 "only happens to certain users when a really specific and unlikely path is followed.")

Effect (how bad the bug is when it does happen, scored from 1-5): 3
(For example, effect of 5 is "the entire app crashes and makes it unusable for all users" or "the bug causes the wrong data to be saved, with critical information (e.g. payment) being affected."
Effect of 1 is "It makes some styling look a little bit weird.")

Steps to reproduce the behaviour: (assuming you're running the app locally with user-operator fixtures already in your db)

1. in browser, go to `dashboard/select-operator/user-operator/1`
2. if it isn't already, click the "Yes" radio button to indicate that the operator's mailing address is the same as its physical address
3. ensure that all other required fields are filled in
4. click "Submit"

Screenshots:

UI:
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOVVIQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--9544f8bcf898bdde0a10263f5ba831a8b40fd874/Screenshot%202023-11-24%20at%205.17.56%E2%80%AFPM.png)

Backend server logs:
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOVlIQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--6194b6b77bcc5e2ba01dd907b437867f40154713/Screenshot%202023-11-24%20at%205.18.36%E2%80%AFPM.png)

Additional information:

- Browser: [e.g. chrome, safari]
- Version: [e.g. 22]
- Additional context

BCerki:
I wasn't able to reproduce this

Sepehr-Sobhani:
No longer valid!

PR #2791, opened 2025-02-04
Reporting Activity: LNG Activities #344

pbastia:


dleard (approved):
same minor change as the last one & then this is good to go

PR #2806, opened 2025-02-07
feat: add gpg ssh setup guide

nataliepiguliak:


Sepehr-Sobhani (approved):
Great job!

Could you add this document as a link in the `README.md` file? Also, please update the Developer Environment Setup path in the README, as it currently points to a non-existent location. Since we now have setup instructions for two different operating systems, please ensure the correct paths are referenced.

PR #2807, opened 2025-02-07
Feat: 459 submitted reports are immutable

pbastia:
Work done:
- Added function to generate a trigger for a model, + tests
- Added triggers to relevant models
- Test that all cascading models from a report version have the proper triggers
- Test that each model can't be updated if their report version is submitted

Housekeeping:
- A few models were missing an 'id' PKEY and were incorrectly using their report_version as pkey. A couple of migrations to remove those models and recreate them properly were necessary


bc_obps/reporting/models/triggers.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+def get_sql_relation_alias_generator() -> Iterator[str]:
+    i = 0
+    while True:
+        i += 1
+        yield f"rel{str(i)}"
+
+
+def immutable_report_version_trigger(
```
pbastia, line 12: This generates SQL to apply as a trigger.
The entire expected SQL can be found in the test file for this method

dleard (approved):
Looks great! I'm going to go ahead and merge this one. 
I did a bit of poking around manually and there's a couple things we may want to handle in a follow up PR:
- I wasn't able to change anything in any models with the trigger, but I could still delete records. We may want to add the trigger to the delete action as well
- This one may not be a big issue, but I was able to update values in intersection models that were created by django as they did not receive the trigger (ie: report_emission_emission_categories).

PR #2809, opened 2025-02-08
486 grant permissions registration app

Sepehr-Sobhani:
[ISSUE 486](https://github.com/bcgov/cas-reporting/issues/486)

**NOTES:**
- Added grants for the registration app.
- Grants access to the public schema when developing locally to enable record creation in the Silk Profiler database.
- Some tests from Reg1 are skipped or deleted because the logic from Reg1 does not align with the grants, so I didn't want to spend time fixing them.
- Applying RLS whenever we run the migrate command (post-migrate signal).
- Added grants to historical m2m fields manually (the Django Simple History package autogenerates these tables, and we lack control over the schema they are generated in).
- Passed `user_guid` directly to some functions to enable allowed internal users to create contacts (internal users should be able to create a contact but not update one).
- Passed `user_guid` directly to some functions to enable allowed internal users to create `bcghg_id` and `boro_id` (internal users should be able to create them but not update them. We need to pass `user_guid` to populate `issued_by_id` during creation).

**TEST:(Developer only)**
- Set `RLS_FLAG = True` in `settings.py` file
- Change `if settings.RLS_FLAG is False:` to `if settings.RLS_FLAG is True:` in `bc_obps/rls/models.py`


BCerki (commented):
No questions about the code since the walkthrough was thorough. The following came up in manual testing:
- I'm no longer able to approve/deny access requests as cas_director: 
image
- I had trouble doing this transfer as a cas_analyst: 
image
Then when I changed the effective transfer date and tried to submit, I got this: 
image, so I think the error is beyond create_transfer. If I try to edit a transfer to put the effective date in the future, that also fails with the same error.

-This is unrelated to your PR I think, but if I edit this transfer: 
image
And add Facility 7, I get an error, but the form goes read-only and it looks like it worked: 
image
. On refresh Facility 7 is gone

Sepehr-Sobhani:
> No questions about the code since the walkthrough was thorough. The following came up in manual testing:
> 
> * I'm no longer able to approve/deny access requests as cas_director:
>   image
> * I had trouble doing this transfer as a cas_analyst:
>   image
>   Then when I changed the effective transfer date and tried to submit, I got this:
>   image, so I think the error is beyond create_transfer. If I try to edit a transfer to put the effective date in the future, that also fails with the same error.
> 
> -This is unrelated to your PR I think, but if I edit this transfer: image And add Facility 7, I get an error, but the form goes read-only and it looks like it worked: image . On refresh Facility 7 is gone


Thank you for reviewing this PR!

I've added two missing grants for the Facility and Operation data models, which I believe were causing related transaction errors. Since a transaction error can block others, this issue may have led to errors in other forms as well. My changes should resolve these issues, but please let me know if anything still needs fixing.

I also updated the endpoint exception handler to return Internal Server Error instead of the transaction error. Additionally, I wrapped the event creation process in a transaction atomic block to ensure an event isn't created if it can't be processed.

dleard (approved):
Great work! The grants & supporting arcthitecture look good. Approved on my end, but let's wait to merge until Brianna confirms the issues she found have been addressed

bc_obps/service/tests/test_operation_service_v2.py
```diff
@@ -1260,9 +1260,10 @@ def test_raises_exception_if_operation_is_non_regulated():
             status=Operation.Statuses.REGISTERED,
             registration_purpose=Operation.Purposes.ELECTRICITY_IMPORT_OPERATION,
         )
+        cas_director = baker.make_recipe('registration.tests.utils.cas_director')
 
         with pytest.raises(Exception, match="Non-regulated operations cannot be issued BORO ID."):
-            OperationServiceV2.generate_boro_id(approved_user_operator.user.user_guid, operation.id)
+            OperationServiceV2.generate_boro_id(cas_director.user_guid, operation.id)
```
BCerki, line 1266: good call

BCerki (approved):
Nice work! :cartwheeling:

PR #2824, opened 2025-02-12
Remove packages used for debugging when running tests

Sepehr-Sobhani:
This also uses a single conftest file for all tests(less repetition)


BCerki (approved):
:cartwheeling:

Issue #2169, opened 2024-09-05
Reg\Select Registration Purpose


andrea-williams:
Blocking #1962 and #2224 

#### Describe the task

We need to handle the scenario where an external user partially or fully submits an operation registration form, and then wants to change their selected registration purpose (RP), since this could impact the data associated with the registration.

The Reg Purpose may be changed in the multistep registration workflow (before the "Submit" button is clicked), or after submission, from the Operation Details page.

This requires writing a lot of backend manual logic to delete data when it becomes no longer relevant because the RP has changed.

#### Dev Checklist

- [x] write functionality to delete data that is no longer relevant when the RP is changed
- [x] make the RP dropdown in Operation details page editable
- [ ] create new UI component that warns user about pending data deletion (see Miro wireframe)
- [x] requires extensive testing to ensure that only and all the correct data is deleted, and that no additional data is deleted that should be kept
- [x] make sure that if user selects "Reporting Operation" or "Potential Reporting Operation", the "OBPS Regulated Operation" purpose is NOT applied

#### Acceptance Criteria

- [x] if an external user wants to change their registration purpose after they've saved (but not submitted) a Registration record, they are prompted to confirm that they want to change their RP, and the Registration record they had saved will be deleted (see [wireframe for Reporting](https://www.figma.com/design/LsDmLDCdnJqI8UahyqvOD8/BCIERS-Designs?node-id=23393-5554) - Registration module should match this)
- [x] if an external user wants to change their registration purpose after they've submitted (i.e., their operation is already registered, so they're editing from the Operation Details page), ensure that no longer relevant data fields are deleted from database. See [Miro table](https://miro.com/app/board/uXjVN6uBj9Y=/?moveToWidget=3458764598469935516&cot=14) for which fields would no longer be relevant

#### Additional context

- See Miro board: https://miro.com/app/board/uXjVN6uBj9Y=/?moveToWidget=3458764598469935002&cot=14
- and this board: https://miro.com/app/board/uXjVLUOERmg=/?share_link_id=931450344603

shon-button:
AW:

Changing Registration Purpose
for Moose and Russell, Patricia ENV:EX - I've been thinking a bit more about that dreaded topic, changing registration purpose. Here's our options as I see them:
 
[...]

PR #2830, opened 2025-02-12
2169 changing reg purpose

BCerki:
card: #2169


BCerki:
Just an exprimental PR, closing

PR #2831, opened 2025-02-12
2169 changing reg purpose

BCerki:
card #2169


BCerki:
Closing, completed in https://github.com/bcgov/cas-registration/pull/2834 instead

PR #2847, opened 2025-02-14
chore: disable triggers before loading fixtures

Sepehr-Sobhani:


dleard (approved):
Nice! I didn't know pgtrigger did that :+1:

Sepehr-Sobhani:
> Nice! I didn't know pgtrigger did that 👍

Yeah, pgtrigger has good stuff! We can also ignore triggers inline: https://django-pgtrigger.readthedocs.io/en/4.13.3/ignoring_triggers/

PR #2850, opened 2025-02-14
Feat/487 reporting app grants

dleard:


bc_obps/reporting/models/rls_configs/activity_json_schema.py
```diff
@@ -0,0 +1,14 @@
...
+
+class Rls:
+    role_grants_mapping = {
+        RlsRoles.INDUSTRY_USER: [RlsOperations.SELECT],
+        RlsRoles.CAS_DIRECTOR: [RlsOperations.SELECT],
+        RlsRoles.CAS_ADMIN: [RlsOperations.SELECT],
+        RlsRoles.CAS_ANALYST: [RlsOperations.SELECT],
+        RlsRoles.CAS_VIEW_ONLY: [RlsOperations.SELECT],
+    }
+    grants = generate_rls_grants(role_grants_mapping, ReportingTableNames.ACTIVITY_JSON_SCHEMA)
```
pbastia, line 14: As a note for some potential refactor: That call could be generic for every model, since the table name can be derived from the meta class of the model. The grant mappings should be sufficient as configuration

bc_obps/reporting/models/rls_configs/activity_json_schema.py
```diff
@@ -0,0 +1,14 @@
...
+from rls.utils.helpers import generate_rls_grants
+
+
+class Rls:
+    role_grants_mapping = {
+        RlsRoles.INDUSTRY_USER: [RlsOperations.SELECT],
+        RlsRoles.CAS_DIRECTOR: [RlsOperations.SELECT],
+        RlsRoles.CAS_ADMIN: [RlsOperations.SELECT],
+        RlsRoles.CAS_ANALYST: [RlsOperations.SELECT],
+        RlsRoles.CAS_VIEW_ONLY: [RlsOperations.SELECT],
```
pbastia, line 12: As another note for some potential refactor: we could introduce a `RlsRoles.ALL` "meta-role" to shorten this and avoid refactoring if we needed to add a role at a later point

bc_obps/reporting/models/rls_configs/facility_report.py
```diff
@@ -0,0 +1,34 @@
...
+from rls.utils.helpers import generate_rls_grants, generate_m2m_rls
+
+
+class Rls:
+    role_grants_mapping = {
+        RlsRoles.INDUSTRY_USER: [
+            RlsOperations.SELECT,
+            RlsOperations.INSERT,
+            RlsOperations.UPDATE,
+            RlsOperations.DELETE,
```
pbastia, line 12: Another comment for future refactoring, a meta-operation `RlsOperations.ALL` for readability and simplicity's sake. Pretty unlikely that we would add an element in the CRUD operations 😄

bc_obps/reporting/models/rls_configs/report.py
```diff
@@ -0,0 +1,19 @@
...
+from rls.utils.helpers import generate_rls_grants
+
+
+class Rls:
+    role_grants_mapping = {
+        RlsRoles.INDUSTRY_USER: [
+            RlsOperations.SELECT,
+            RlsOperations.INSERT,
+            RlsOperations.UPDATE,
+            RlsOperations.DELETE,
```
pbastia, line 12: Do we need "DELETE" here? would reports be archived if we ever needed to re-create one?

bc_obps/rls/utils/grant.py
```diff
@@ -18,7 +19,11 @@ class RlsGrant:
     """
 
     def __init__(
-        self, role: RlsRoles, grants: List[RlsOperations], table: RegistrationTableNames, schema: Schemas = Schemas.ERC
+        self,
+        role: RlsRoles,
+        grants: List[RlsOperations],
+        table: RegistrationTableNames | ReportingTableNames,
```
pbastia, line 25: This makes the Rls app dependent on the registration and reporting apps - effectively a circular dependency in our app tree.
I'd suggest type `str` instead - since this class can build a grant for any table

bc_obps/rls/utils/helpers.py
```diff
@@ -8,7 +9,7 @@
 
 def generate_rls_grants(
     role_grants_mapping: Dict[RlsRoles, List[RlsOperations]],
-    table: RegistrationTableNames,
+    table: RegistrationTableNames | ReportingTableNames,
```
pbastia, line 12: Same comment about circular dependency

pbastia (approved):
Amazing work! Just a comment about a circular dependency. 
I think the Rls app should be made generic enough that we could eventually break it out into its own pip package to share with the rest of the community!

bc_obps/reporting/models/rls_configs/report.py
dleard: Won't they need delete for actions that fully delete a report? Like changing registration purpose?

pbastia: I guess you're right! Although we wouldn't be tossing away entire reports if they changed registration purpose, maybe just versions? Anyways, it's their data, so we can give them that option

PR #2859, opened 2025-02-18
Use global exception handler

Sepehr-Sobhani:
This removes the need to use the `@handle_http_errors()` decorator on each API endpoint.


andrea-williams (approved):
🤩

PR #2866, opened 2025-02-18
feat: fix carbon registry link

nataliepiguliak:


bc_obps/common/fixtures/dashboard/compliance/external.json
```diff
@@ -25,13 +25,17 @@
           },
           {
             "title": "B.C. Carbon Registry",
-            "href": "/compliance/bc-carbon-registry",
+            "href": "https://carbonregistry.gov.bc.ca/br-reg/public/bc/home.jsp#/login",
+            "target": "_blank",
+            "rel": "noopener",
```
pbastia, line 30: These attributes are not directly set to an `<a>` tag, so they wouldn't have any effect.
You can check out the `bciers/libs/components/src/navigation/Tile.tsx` file to see how it'll render. If we need those, I'd suggest:
- either making maybe an "external: true" option to indicate we need the target/rel attributes set properly
- or look whether the link's href starts with `http` to set these accordingly?

pbastia (changes requested):

bc_obps/common/fixtures/dashboard/compliance/external.json
nataliepiguliak: done

pbastia (approved):
Looks great! 👌

Issue #2770, opened 2025-01-31
Admin - Write & practice run migration script for Operation statuses


andrea-williams:
**Blocking:** #2041 

#### Describe the task

Write a SQL script and perform a practice run for #2041 

#### Acceptance Criteria

- [x] write a Django migration that satisfies the AC described in #2041 
- [x] run your script against a copy of the prod db pulled into your local env
- [x] when satisfied that the script performs correctly with no errors, delete your local copy of the prod data

#### Definition of Done

- [ ] first
- [ ] second
- [ ] third

#### Additional context

- Add any other context about the task here.
- Or here

BCerki:
Work completed in https://github.com/bcgov/cas-registration/issues/2041

PR #2868, opened 2025-02-19
Task/migrate operation reg1 statuses/2770

andrea-williams:
Card: #2770 

## How to Test/Run:
1. If the script file you're testing is in a /migrations folder, move it to some other folder so that it won't be automatically included in `check_migrations_with_prod_data`
2. Follow the steps listed in bc_obps/common/management/commands/check_migrations_with_prod_data.py (regarding making changes to your local bc_obps/.env file.
3. You'll also need to comment out lines 73-74 and 80-81 in check_migrations_with_prod_data, so that you can retain a copy of the prod data locally while you test.
4. Sign in to Openshift with the CLI tool
5. Run `poetry run python manage.py check_migrations_with_prod_data --pod <pod_name>`, where pod_name is the name of a postgres pod in our Openshift PROD env
6. Once you've successfully pulled the prod data into your local "obps" database, run `poetry run python manage.py migrate reporting` to apply all Django migrations except for the migration included in this PR. The `poetry run python manage.py showmigrations` command is also useful for seeing which migrations have been applied.
7. Move the script file you're testing back into the appropriate /migrations folder.
8. Now run `poetry run python manage.py migrate registration`, which will run the migration included in this PR.  You'll be able to see the print statements describing what data was affected by the latest migration.

For testing purposes, the migration in this PR contains assertions and print statements for testing purposes and to demonstrate that the migration is working as expected. These assertions and print statements will be deleted from the migration before merging (it's also why the pre-commit hook is failing).

REMEMBER TO DELETE YOUR LOCAL COPY OF THE PROD DATA ONCE YOU'VE FINISHED TESTING!


bc_obps/registration/migrations/0079_migrate_reg1_operation_statuses.py
```diff
@@ -0,0 +1,120 @@
...
+
+    print("\n\n\nDeclined operations - before migration:")
+    for op in declined_operations_before:
+        print(f'{op.id}: archived_by_id={op.archived_by_id}, archived_at={op.archived_at}')
+
+    pending_operations_updated = Operation.objects.filter(status="Pending").update(status="Draft")
+    approved_operations_updated = Operation.objects.filter(status="Approved").update(status="Draft")
+    changes_requested_operations_updated = Operation.objects.filter(status="Changes Requested").update(status="Draft")
+    declined_operations_archived = Operation.objects.filter(status="Declined").update(
+        archived_by_id='c3bc1b69-15de-44ac-b03b-982bf3163406', archived_at=datetime.datetime.now(datetime.timezone.utc)
```
BCerki, line 62: Is it OK to have a real uuid in a public repo? (I know we have some in our mock data too, but I think we were trying to avoid where possible.) Possibly we could select it from the User model by some other field?

BCerki (commented):
Looks good! A couple thoughts:
- running `poetry run python manage.py check_migrations_with_prod_data`, runs all the migrations so in the testing instructions, you have to additionally comment out line 69-70 of `check_migrations_with_prod_data` if you want to apply the migrations up to 0078 and play around with the before data
- if there's anything we'll have to do when we do the migration for real (e.g. remove reg1 statuses from the model choices if we're doing that), we can add to the real migration card so we remember
- Thank you for providing a detailed example I can steal for mine >:)

bc_obps/registration/migrations/0079_migrate_reg1_operation_statuses.py
andrea-williams: I don't think it's a major security risk but probably not the best idea either. My original thinking was to avoid including her full name in the repo... but it's her github username anyway, so she's basically doxxed herself (as have I). I'll fix this.

Issue #2750, opened 2025-01-28
Remove statuses from designated timeline tables


dleard:
blocking #1362 

#### Description of the Tech Debt
Through discussion in #2590, we have decided that the status columns on the designated timeline tables are redundant & can be removed. The "state" of an operation or facility can be derived in other ways & this gives the designated timeline tables a single responsibility.

#### Development Checklist:

- [x] Remove status column from `operation_designated_operator_timeline` model
- [x] Remove status column from `facility_designated_operation_timeline` model
- [x] Remove references in other places that set these columns / in tests

fviduya:
I have tested this in the DEV environment, [here is the test document for reference](https://buttoninc-my.sharepoint.com/:w:/g/personal/francesca_button_is/EZzbaLHKk-xIgLTqnjeW4psBCiqm0FYdvhlRfiey89z2Hw?e=ree5zw).

I can verify that:
- [x] External users can still create an operation and facility 
- [ ] The `Status` column for operation and facility shows `Active` for operations and facilities with no end date - **The column is currently blank, created #3010 to investigate the issue** 
- [x] When transferring a facility, and using a past date as the effective date, the transfer reflects correctly
- [x] When transferring an operation , and using a past date as the effective date, the transfer reflects correctly
- [x] When transferring a facility and using a future date as the effective date, the transfer does not go through right away 
- [x] When transferring an operation and using a future date as the effective date, the transfer does not go through right away

Moving to Pending PO for review of test and notes. Thank you!

vesselak:
Reviewed the QA documentation and accepting the ticket with the deficiency noted in #3010

PR #2872, opened 2025-02-19
Task/remove statuses from timeline tables/2750

andrea-williams:
Card: #2750


BCerki (commented):
Nice work! Yay deleting code :man_dancing: 

Some notes:
- I think we should keep the test assertions as they are, and if a mock record was being assigned a status of TRANSFERRED, it gets an end date instead. We're not using statuses anymore but it's still relevant that people see/don't see transferred records. The timeline bakers assign an end_date by default so we might have to change that
- I get an error when I try to make a transfer:
image
`django.core.exceptions.FieldError: Cannot resolve keyword 'status' into field. Choices are: archived_at, archived_by, archived_by_id, created_at, created_by, created_by_id, end_date, id, operation, operation_id, operator, operator_id, sfo_facility_id, sfo_facility_name, start_date, updated_at, updated_by, updated_by_id
`

bc_obps/service/tests/data_access_service/test_data_access_operation_designated_operator_timeline.py
```diff
@@ -22,25 +21,31 @@ def test_unapproved_user_exception():
...
+            status=Operation.Statuses.REGISTERED,
             _quantity=20,
         )
+        for operation in registered_operations:
+            baker.make_recipe(
+                'registration.tests.utils.operation_designated_operator_timeline',
+                operation=operation,
+                operator=approved_user_operator.operator,
+                end_date=None,
+            )
```
andrea-williams, line 44: this is ugly in the sense that it's adding in more lines of code, but it's the only solution I could find to get around this error:

`django.core.exceptions.ValidationError: {'__all__': ['Constraint “unique_active_designated_operator_per_operation” is violated.']}`

bc_obps/service/tests/test_facility_designated_operation_timeline_service.py
```diff
@@ -47,19 +45,14 @@ def test_get_timeline_by_operation_id_industry_user():
...
-            status=FacilityDesignatedOperationTimeline.Statuses.TRANSFERRED,
-        )
+        # user's timeline
         baker.make_recipe(
             'registration.tests.utils.facility_designated_operation_timeline',
             operation=users_operation,
             _quantity=10,
-            status=FacilityDesignatedOperationTimeline.Statuses.ACTIVE,
         )
+        # mimic transferred facilities
```
BCerki, line 54: Missing code here? It looks like we might be missing the code to handle transferred facilities in the service too

bc_obps/service/tests/test_operation_service_v2.py
```diff
@@ -1443,17 +1441,17 @@ def test_gets_unfiltered_sorted_list_for_industry_user():
...
             operator=approved_user_operator.operator,
-            status=OperationDesignatedOperatorTimeline.Statuses.TRANSFERRED,
+            end_date=None,
         )
         baker.make_recipe(
             'registration.tests.utils.operation_designated_operator_timeline',
             operator=approved_user_operator.operator,
-            status=OperationDesignatedOperatorTimeline.Statuses.ACTIVE,
+            end_date=None,
         )
```
BCerki, line 1455: Interesting that these three don't trigger the unique constraint. Does it only happen if you use _quantity?

BCerki (commented):
I can't get into dev right now to manually test, so I'll do that part later

bc_obps/service/facility_designated_operation_timeline_service.py
```diff
@@ -21,7 +21,6 @@ def get_timeline_by_operation_id(
 
         if user.is_industry_user():
             UserOperatorService.get_current_user_approved_user_operator_or_raise(user)
-            base_queryset = base_queryset.exclude(status=FacilityDesignatedOperationTimeline.Statuses.TRANSFERRED)
```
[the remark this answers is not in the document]
BCerki: This one still applies, I made this transfer:
image
And I can still see it in the old operator's grid:
image

BCerki (commented):
I was able to get and do the manual testing, operation timeline grid looks great, but I think there's some missing code for facilities, more details inline

bc_obps/service/tests/test_operation_service_v2.py
andrea-williams: yes, it was only happening when using `_quantity`. When I tried 

```
baker.make_recipe('operation_designated_operator_timeline', operation=baker.make_recipe('operation'), end_date=None, _quantity=20)
```

it was only creating 1 mock operation, and then trying to assign 20 active timelines for that 1 operation, which isn't allowed because we have a Django constraint that forbids it.

andrea-williams: that's why I instead had to first make 20 different operations, and then loop through them to create one timeline for each unique operation

andrea-williams:
Holding myself accountable for manual regression testing before requesting review:

1. Sign in as CAS user and transfer Facility 38 from Bees LFO to Alpha Enterprises, effective yesterday
Screenshot 2025-03-10 at 4 31 05 PM

2. Sign in as bc-cas-dev and view Bees LFO - Facility 38 is no longer visible because it's been transferred
Screenshot 2025-03-10 at 4 33 08 PM

BCerki (approved):
:cartwheeling:

PR #2874, opened 2025-02-19
Chore/fixed no past facilities bug

ayeshmcg:
A small PR to fix the bug with operations(LFO) with no past facilities
To test:-
1. Select Bark operation
2. go to review facilities
3. You should not see past facilities section and should be able to proceed


acatchpole (approved):
well that's a lot safer now

PR #2883, opened 2025-02-20
refactor: simplify user setup in endpoint permissions tests

Sepehr-Sobhani:


fviduya:
Just an unused code ```auth_headers``` to be removed then it's good! 🥇 Thank you!

PR #2889, opened 2025-02-20
Practice migration: set purpose for reg 1 ops with boro id

BCerki:
card: https://github.com/orgs/bcgov/projects/123/views/16?pane=issue&itemId=97842660&issue=bcgov%7Ccas-registration%7C2841

This PR:
- migration to set purpose for reg1 records
- see https://github.com/bcgov/cas-registration/pull/2868 for testing instructions (steps 1-4 exactly the same, and then run migrations incrementally if desired)
- this doesn't require any update to the mock data or any other parts of the app, so no AC added to the ticket for the real migration


andrea-williams (commented):
Informally approved - not going to actually Approve it just to ensure that this doesn't get merged into develop prematurely.

Have confirmed it works as described in the ticket AC. Had a horrible thought while looking at the prod data - we have the `opt_in` boolean flag and I know at some point we were talking about setting the purpose for those to "Opted-In" by default but I can't remember whether we actually agreed to that or not. I'll check with Patricia.

Also, minor nit: could you please rename the "auto_blah" migration filename to a better descriptor?

BCerki:
> Informally approved - not going to actually Approve it just to ensure that this doesn't get merged into develop prematurely.
> 
> Have confirmed it works as described in the ticket AC. Had a horrible thought while looking at the prod data - we have the `opt_in` boolean flag and I know at some point we were talking about setting the purpose for those to "Opted-In" by default but I can't remember whether we actually agreed to that or not. I'll check with Patricia.
> 
> Also, minor nit: could you please rename the "auto_blah" migration filename to a better descriptor?

Per Teams conversation, if the opt-in flag is True, we'll set the purpose to Opted-In instead of OBPS Regulated

andrea-williams:
re-approved! Thanks Bri!

BCerki:
Closing, work finished in a different PR

PR #2891, opened 2025-02-21
Task/fix lfo enum wording/2882

andrea-williams:


BCerki (commented):
Nice work!
A couple overall things:
- `Linear Facility Operation` still appears in a couple of the model comments and one of the reporting components
- Some of the CI steps are failing
- Re this note: `Ideally we'd create a sub-class on the Operation model so that we can restrict the choices for type, but that would require a model migration and changing data in prod, so I gave that a swerve.` Are you talking about adding `models.TextChoices` for type, like we have in the facility model? That would need a migration but I don't think it requires a change to prod data

BCerki (approved):
Looks great! I manually tested the registration form and everything still works and I can change types

andrea-williams:
> Looks great! I manually tested the registration form and everything still works and I can change types

Thank you for always testing so thoroughly @BCerki !

PR #2892, opened 2025-02-21
2195 set up bciers apps for owasp zap scan

Vikram-Kumar-BCGov:


pbastia:
Closing, was the base for #2921

PR #2901, opened 2025-02-25
chore: fixed registration purpose issue

ayeshmcg:
Card: https://github.com/bcgov/cas-reporting/issues/494
Changes:
1. Modified model ReportOperation: added registration_purpose field.
2. Modifed schema, api, service and tests


bc_obps/reporting/models/report_operation.py
```diff
@@ -49,6 +49,12 @@ class ReportOperation(TimeStampedModel):
         blank=True,
         null=True,
     )
+    registration_purpose = models.CharField(
+        max_length=1000,
+        db_comment="The Registration purpose of the operation",
+        blank=True,
+        null=True,
```
pbastia, line 56: The registration purpose shouldn't be allowed to be null. We won't have any data in the db by the time this is deployed, so you can give django any value when you run `make migrations`

bc_obps/reporting/schema/report_operation.py
```diff
@@ -61,6 +62,7 @@ class ReportOperationIn(Schema):
     operation_name: str
     operation_type: str
     operation_bcghgid: str
+    registration_purpose: str
```
pbastia, line 65: Fields declared in the `Meta` class don't need to be re-specified manually in the schema itself, this is redundant

bc_obps/service/report_version_service.py
```diff
@@ -19,6 +19,8 @@ def create_report_version(
         operation = report.operation
         operator = report.operator
 
+        print('operation', operation)
```
pbastia, line 22: print left behind

bc_obps/reporting/service/report_additional_data.py
```diff
@@ -15,11 +15,7 @@ def get_report_report_additional_data(
 
     @staticmethod
     def get_registration_purpose_by_version_id(version_id: int) -> dict:
-        registration_purpose = (
-            ReportVersion.objects.select_related('report', 'report__operation')
-            .get(id=version_id)
-            .report.operation.registration_purpose
-        )
+        registration_purpose = ReportOperation.objects.get(report_version__id=version_id).registration_purpose
```
pbastia, line 18: Registration purpose should default to the value in registration. I think we could make the `get_report_report_additional_data` return an in-memory object with that value set this way we wouldn't even have to have that `get_registration_purpose_by_version_id` endpoint

pbastia (changes requested):
Very nice! Just a couple of things that we can streamline in how the data is seeded from registration

pbastia:
> Thanks for the feedback @pbastia ! Just to clarify, are you referring to the create_report_version method in report_version_service when mentioning streamlining how data is seeded from registration?

Actually, thanks for catching that! I read your code too fast and thought that the registration purpose was pulled directly from the registration module. If we pull it from the report itself, this makes more sense.

ayeshmcg:
yes now the registration_purpose is populated in reporting on report creation and later on fetched from reporting only

pbastia (approved):
🎉 thank you!

Issue #2747, opened 2025-01-28
Disable login/access Reg 1: Landing Page Update


zoeyli-46:
#### Description:
Decommission/migrate Reg 1

When there is a transition/dark period between transitioning from Reg 1 to Reg 2, the landing page will be updated with new information and the login button is disabled
[Wireframe](https://www.figma.com/design/LsDmLDCdnJqI8UahyqvOD8/BCIERS-Designs?node-id=30626-6549&t=2oKSUvfB9Y2uQ0Xt-4)
#### Acceptance Criteria:

Given that BCIERS is undergoing decommissioning/migrating, 
When industry user is on the landing page
Then the login button is disabled and there is a banner at the top of the screen

When Reg 1 is decommissioned, the current URL registration.industrialemissions.gov.bc.ca will automatically redirect to the new: industrialemissions.gov.bc.ca, so that users accessing the currently URL will automatically be brought to the new site 

#### Development Checklist:

- [x] build the component
- [ ] put it behind a growthbook flag (maybe?)
- [ ] Checklist item
- [ ] Meets the DOD

#### Definition of Ready (Note: If any of these points are not applicable, mark N/A)

- [ ] User story is included
- [ ] User role and type are identified
- [ ] Acceptance criteria are included
- [ ] Wireframes are included (if required)
- [ ] Design / Solution is accepted by Product Owner
- [ ] Dependencies are identified (technical, business, regulatory/policy)
- [ ] Story has been estimated (under 13 pts)

#### Definition of Done (Note: If any of these points are not applicable, mark N/A)

- [ ] Acceptance criteria are tested by the CI pipeline
- [ ] UI meets accessibility requirements
- [ ] Configuration changes are documented, documentation and designs are updated
- [ ] Passes code peer-review
- [ ] Passes QA of Acceptance Criteria with verification in Dev and Test
- [ ] Ticket is ready to be merged to main branch
- [ ] Can be demoed in Sprint Review
- [ ] Bugs or future work cards are identified and created
- [ ] Reviewed and approved by Product Owner

#### Notes:

-

zoeyli-46:
@patriciarussellCAS @patrickisaac @andrea-williams here is the ticket for the landing page update / decommissioning reg 1

patrickisaac:
Refinement Notes:
- Use of Growthbook to toggle button/banner

BCerki:
@patriciarussellCAS , what date do you want to put on the banner?

[...]

PR #2902, opened 2025-02-25
chore: update landing page

BCerki:
card: https://github.com/bcgov/cas-registration/issues/2747

This PR:
- adds a banner to the reg1 landing page
- disables the button 
- no need for a growthbook flag since we're not doing regular releases to reg1 prod and reg1 and 2 have different urls

### DIFF
--- a/bciers/apps/registration1/app/(onboarding)/home/page.tsx
+++ b/bciers/apps/registration1/app/(onboarding)/home/page.tsx
@@ -4,12 +4,11 @@
 import Button from "@mui/material/Button";
 import Grid from "@mui/material/Grid/Grid";
 import events from "@/app/data/home/events.json";
-import { signIn } from "next-auth/react";
-import { getEnvValue } from "@bciers/actions";
 import {
   bcObpsLink,
   bcObpsGuidanceLink,
   carbonTaxExemptionLink,
+  ghgRegulatorEmail,
 } from "@bciers/utils/src/urls";
 import Note from "@bciers/components/layout/Note";
 import Link from "next/link";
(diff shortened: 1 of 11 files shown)


bciers/apps/registration1/app/(onboarding)/home/page.tsx
```diff
@@ -66,174 +60,195 @@ export default function Page() {
...
+    <>
+      <Note variant="important">
+        <div className="mt-4">
+          <b>Service Updates in Progress</b>
+        </div>
+        <div>
+          BCIERS is currently undergoing scheduled updates until mid-March.
+          During this time, user log-in will be temporarily unavailable. If you
+          require urgent support, please reach out to{" "}
+          <Link href={"mailto:GHGRegulator@gov.bc.ca"}>
```
Sepehr-Sobhani, line 72: nit: We have hardcoded `mailto:GHGRegulator@gov.bc.ca` in a few places, but perhaps we could use `TILE_REPORT_PROBLEM` instead. However, this enum is found under the e2e/libs directory, which isn't particularly related.

Sepehr-Sobhani (changes requested):
💯 Great stuff!
We should also disable the other button because during this period no one will be able to access the web app.
Not sure if we should update other parts of the landing page since the wireframe differs from the current version. 🤔

bciers/apps/registration1/app/(onboarding)/home/page.tsx
BCerki: We have an urls file in libs, I'll pop it in there

BCerki:
> 💯 Great stuff! We should also disable the other button because during this period no one will be able to access the web app. Not sure if we should update other parts of the landing page since the wireframe differs from the current version. 🤔

I'll check with Zoey

Sepehr-Sobhani (approved):
🕺

PR #2921, opened 2025-02-26
Chore: OWASP Zap baseline scan for the whole app

pbastia:


.github/workflows/zap-owasp.yaml
```diff
@@ -2,55 +2,127 @@ name: ZAP OWASP Scan Code
...
+
+          if [ -s errors.txt ]; then
+
+            echo "Found errors with severity/risk level higher than $ZAP_FAILURE_RISK_LEVEL,"
+            echo "for more information, consult the logs of the individual zap actions."
+
+            echo "Errors found:"
+            cat errors.txt
+            exit 1
+          fi
```
joshgamache, line 128: This is great!

joshgamache (approved):
Nice work spreading it out to the rest of the app. All the scans ran on CI and passed. I skimmed their logs and all the scans seem to be running as expected, good to see they're already passing!

pbastia:
> Nice work spreading it out to the rest of the app. All the scans ran on CI and passed. I skimmed their logs and all the scans seem to be running as expected, good to see they're already passing!

@joshgamache it's passing by design! We should definitely lower the risk threshold and address some of the medium issues.

PR #2924, opened 2025-02-27
Add mock authentication for bypassing Keycloak

Sepehr-Sobhani:


Sepehr-Sobhani:
Only needed for load testing.

PR #2941, opened 2025-03-02
chore: 595 omit extra data in activity form

shon-button:
Addresses:  [595](https://github.com/bcgov/cas-reporting/issues/595)

### 🚀 Impact:

- Updated `onSumit` to use `e.formData` from the event object and `omitExtraData: true` to remove extra fields not in the schema
- Updated `bciers/libs/components/src/form/NavigationForm.tsx\const onSaveAndContinue` to call form's submit() method
- Added JSON propery `id` to ALL activity schemas so to retain `id` as a valid field in the schema

## 📝 Notes:  
- There was a comment on the reason to use `formState` but, I am not sure it ~still applies; so, using `e.formData` from the event object instead of `formState` to ensure no extra data is sent to the server

### **Tests Set-Up:**  

1. Start the API server:  
   ```sh
   cd bc_obps
   make reset_db
   make run
   ```  

   ```sh
   cd bciers && yarn dev-all
   ```  

3. Navigate to [http://localhost:3000](http://localhost:3000)  
4. Click the **"Log in with Business BCeID"** button  

---

#### **Steps:**  
1. Navigate to activities, ex: http://localhost:3000/reporting/reports/1/facilities/f486f2fb-62ed-438d-bb3e-0819b51e3aff/activities
2. Select an activity
3. Select a methodology that displays conditional fields - note the methodology selected
4. Complete required fields
5. Click `Save`
6. Validate the data was saved
7. Select a different methodology that displays conditional fields
8.  Complete required fields
9. Click `Save`
10. Validate the data was saved
11. Select the methodology from step 3
12. Validate the data:
#### **Expected Result**  

✅  There is no data in the conditional field for the methodology from step 3

13. **Repeat steps using `Save & Continue`**
#### **Expected Result**  

✅  There is no data in the conditional field for the methodology from step 3

---

#### **Steps:**  
1. Repeat the test using different activities and different methodologies
#### **Expected Result**  
✅  There is no extra data sent to the server

### DIFF
--- a/bc_obps/reporting/json_schemas/2024/aluminum_production/activity.json
+++ b/bc_obps/reporting/json_schemas/2024/aluminum_production/activity.json
@@ -2,6 +2,10 @@
   "title": "Alumina production",
   "type": "object",
   "properties": {
+    "id": {
+      "type": "number",
+      "description": "The unique identifier for this activity"
+    },
     "smelterTechType": {
       "title": "Smelter technology type",
       "type": "array",
--- a/bc_obps/reporting/json_schemas/2024/carbonates_use/activity.json
+++ b/bc_obps/reporting/json_schemas/2024/carbonates_use/activity.json
@@ -1,5 +1,10 @@
 {
   "title": "Carbonate use",
   "type": "object",
-  "properties": {}
+  "properties": {
+    "id": {
+      "type": "number",
+      "description": "The unique identifier for this activity"
+    }
+  }
 }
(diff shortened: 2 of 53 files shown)


ayeshmcg (commented):
Nice work @shon-button ! 🎉 The previous methodology-related data is no longer saved when clicking the Save button, which works as expected.

I have a question—if we do the same with the Save and Continue button, should it still clear the previous methodology-related field data, or should it retain it?

ayeshmcg (approved):
Great work @shon-button , Now on save and continue also it works as expcected

PR #2950, opened 2025-03-04
docs: add documentation on the steps to apply a hotfix

dleard:
Hotfix documentation pulled from our CIF app


pbastia:
@dleard  as a review for this one, should we do a practice run while screen sharing with all the devs?

pbastia (approved):

Issue #598, opened 2024-01-16
Revise Operator Form


Dianadec:
#### Describe the task

AC:

1. Move parent company question in Operator Information table
- In the My Operator (Operator Information) table, the Question with radio buttons asking parent company information "Does the operator have a parent company" needs to be moved to the bottom of the Operator Information section.
- See wireframe: https://www.figma.com/file/LsDmLDCdnJqI8UahyqvOD8/OBPS-App-Designs?type=design&node-id=2401%3A331&mode=design&t=brhPkfT4W4RpuEjA-1

2. Remove 'Percentage of ownership of operator' field from the Operator Information Form. 
- Here's the wireframe to show this field. https://www.figma.com/file/LsDmLDCdnJqI8UahyqvOD8/OBPS-App-Designs?type=design&node-id=3422%3A3161&mode=design&t=59ovTuKFm0Whkcg5-1
See below for the wireframe of what the design should look like without the field.
- Under 'Parent Company Information', we no longer need this field.
It should look like this: https://www.figma.com/file/LsDmLDCdnJqI8UahyqvOD8/OBPS-App-Designs?type=design&node-id=2401%3A331&mode=design&t=18BwAF7yEAoBZUZy-1

3. Remove fields in flow within 'If Operator Data does NOT exist', Operator Information form:
- remove field 'percentage of owner as operator' under Parent Company Information
- https://www.figma.com/file/LsDmLDCdnJqI8UahyqvOD8/OBPS-App-Designs?type=design&node-id=2615%3A3283&mode=design&t=vNEAKgcCX2f1IfPT-1

#### Acceptance Criteria

- [ ] first
- [ ] second
- [ ] third

#### Additional context

- Add any other context about the task here.
- Or here

Dianadec:
@nanyangpro  did I capture the info here correctly? cc @hannavovk

Dianadec:
@nanyangpro  I've combined the 'Operator Form' tasks into this one ticket. Please review :). (I closed #601 and #611)

BCerki:
Blocked on broken Add Operator/Select Operator workflow

nanyangpro:
Great call, @BCerki!
I double-checked in DEV. You are right, all the AC in this ticket are already checked off. Feel free to close this ticket or send it for PO approval.

PR #2960, opened 2025-03-06
Feat: Tasklist generator for registration purpose

pbastia:
# Reporting flows and tasklist framework

This PR introduces types to support describing reporting flows, pages, and return types for various factories to compute:
- the tasklist elements
- the back and continue URLs
- the list of headers and which one is active

The `NavigationInformation` type replaces the tasklist and continue/back URLs, as well as the top headers, and can be generated once the reporting flows have been implemented.

## 🔧  Building the navigation information
Preferably in a server component, just:
```ts
const navInfo = await getNavigationInformation(
                         HeaderStep.OperationInformation,
                         ReportingPage.PersonResponsible,
                         reportVersionId,
                         facilityId,
                         optionalContext)
```

## 🔧 Adding a reporting flow
1. in `taskList/reportingFlows/someFlow.ts`, create a reporting flow:
```ts
export const someReportingFlow: ReportingFlowDescription = {
  [HeaderStep.step1]: [ReportingPage.page1, ReportingPage.page2, ...],
  [HeaderStep.step2]: ...
  ...
}
```
2. Add it to the list of flows in `reportingFlows.ts`
3. In `reportingFlows.ts`, amend the `getFlow()` method to allow returning it.

## 🔧 Adding a (tasklist) page to use in a reporting flow:
1. Amend the `ReportingPage` enum to include it
2. In `taskList/taskListPages/`, add a tasklist element factory for that page, either in a new file or in an existing one, they're grouped by header step
3. Amend `pageElementFactory.ts` to include the new file if needed
4. That's it! The page is ready to be understood by an existing or new flow.

## 🔧 Passing context down to the factory page
The `getNavigationInformation()` method allows passing a context object of type `TaskListPageFactoryContext`. That context object is passed down to the individual page factories, so we can make use of it when building the TaskList.

Example:
```ts
const navInfo = await getNavigationInformation(..., { facilityName: "MyOverriddenFacilityName" }
```

When building additional factory pages, more fields to that context object requires adding them to the `TaskListPageFactoryContext`, so that developers are aware of what may or may not be in this object.

## 🔧 Passing context up to the navigation information builder
The factory page return type, `TaskListPageFactoryData` can also be amended to add behaviour when building the TaskList. The `extraOptions` field can be extended to include more data, that `getNavigationInformation` might be able to understand and act on.


bciers/apps/reporting/src/app/components/reportInformation/facilities/FacilitiesPage.tsx
```diff
@@ -17,9 +19,20 @@ export default async function FacilitiesPage({
   if (!facilities) {
     return <div>No facilities available.</div>;
   }
+
+  const navInfo = await getNavigationInformation(
+    HeaderStep.ReportInformation,
+    ReportingPage.FacilitiesTable,
+    version_id,
+    facilities.rows[0].id,
```
shon-button, line 27: @pbastia 

I believe `facilities.rows[0].id` should be `facilities.rows[0].ifacility`; otherwise, error is generated:

image
```
[{'type': 'uuid_parsing', 'loc': ('path', 'facility_id'), 'msg': 'Input should be a valid UUID, invalid length: expected length 32 for simple format, found 2', 'ctx': {'error': 'invalid length: expected length 32 for simple format, found 2'}}]
```

bciers/apps/reporting/src/app/components/taskList/taskListPages/2_facilitiesInformation.ts
```diff
@@ -0,0 +1,139 @@
...
+    facilityId,
+  ) => ({
+    element: {
+      type: "Page",
+      title: "Allocation of emissions",
+      isActive: activePage === ReportingPage.AllocationOfEmissions,
+      link: `/reports/${reportVersionId}/facilities/${facilityId}/allocation-of-emissions`,
+    },
+  }),
+  [ReportingPage.EndOfReport]: (activePage, reportVersionId, facilityId) => ({
```
shon-button, line 130: @pbastia 

Questioning if `ReportingPage.EndOfReport` link is a useful link:
image

bciers/apps/reporting/src/app/components/taskList/types.ts
```diff
@@ -0,0 +1,105 @@
...
+}
+
+export enum ReportingFlow {
+  SFO = "SFO",
+  LFO = "LFO",
+  SimpleReport = "Simple Report",
+  EIO = "EIO",
+  ReportingOnlySFO = "Reporting Only - SFO",
+  ReportingOnlyLFO = "Reporting Only - LFO",
+  NewEntrantSFO = "New Entrant - SFO",
```
shon-button, line 39: @pbastia 

The following `ReportingFlow` do not seem to be implemented in the `HeaderStep` and in the `bciers/apps/reporting/src/app/components/taskList/reportingFlows/`

```
  ReportingFlow.NewEntrantSFO
  ReportingFlow.NewEntrantLFO
```

bciers/apps/reporting/src/app/components/taskList/taskListPages/5_signOffSubmit.ts
```diff
@@ -1,4 +1,52 @@
...
+      type: "Page", // Set the type to "Page"
+      title: "Final review",
+      link: `/reports/${reportVersionId}/final-review`,
+      isActive: activePage === ReportingPage.FinalReview,
+    },
+  }),
+  [ReportingPage.Verification]: (
+    activePage,
+    reportVersionId,
+    facilityId,
```
shon-button, line 18: @pbastia 
Is `facilityId` required here?

shon-button:
@pbastia 

Nice refactor, maintainability much improved!

Found a few issues, listed above, we one suggestion for your consideration...

IMO, the header steps would improve user report navigation experience if the header steps had associated routing links.

bciers/apps/reporting/src/app/components/taskList/taskListPages/2_facilitiesInformation.ts
pbastia: It's not, but it's a page that exists, and our pattern is to have a link in the tasklist for it. Otherwise, this is a page that is not part of a reporting flow, the user doesn't know where they are, and the implementation of the tasklist highlighted that. @zoeyli-46  can chime in maybe?

pbastia:
@shon-button I completely agree that the headers should be a link, I find myself clicking on them often enough. However the original design was to not implement them, because they're a guardrail to make sure that the user completes the report in a particular order. For example, making sure that a user selects products and activities in step 1 before reporting emissions in step 2

bciers/apps/reporting/src/app/components/taskList/taskListPages/2_facilitiesInformation.ts
zoeyli-46: Is the thought that End of Facility report is not a useful page and causes the user to not know where they are in the reporting flow? 

I suppose the alternative is that if the user is on the final page of the facility report, when they click save and continue they are directly brought back to the facility reports grid and a success snackbar appears that says “[facility name] report changes saved” so they know that they reached the end of the facility report!

bciers/apps/reporting/src/app/components/reportInformation/facilities/FacilitiesPage.tsx
pbastia: Good catch!

bciers/apps/reporting/src/app/components/taskList/taskListPages/2_facilitiesInformation.ts
pbastia: I suggest leaving the current implementation, since it's the easiest (1-1 mapping between pages and tasklist elements, with no exceptions) and review the design if it proves confusing

bciers/apps/reporting/src/app/components/taskList/types.ts
pbastia: Thanks! I see we already have all the pages for new entrant, added the flows. We're still missing some pages for Simple report and EIOs, these will nee to be added later

bciers/apps/reporting/src/app/components/taskList/taskListPages/5_signOffSubmit.ts
pbastia: it is, the functions are all of type `TaskListPageFactory`. We take a shortcut and a quick of javascript for a lot of other factories, that javascript allows us to omit the last function parameters if they're undefined.
I'll change it to `_` to highlight this

shon-button (approved):
@pbastia 

🤌 Sublime as per always, just one last item for your consideration...

Can this issue be resolved in this refactor?
[598](https://github.com/bcgov/cas-reporting/issues/598)

pbastia:
@shon-button yes, [#598 ](https://github.com/bcgov/cas-reporting/issues/598) will definitely be fixed by this! But probably introduce other navigation bugs hehe

shon-button:
> @shon-button yes, [#598 ](https://github.com/bcgov/cas-reporting/issues/598) will definitely be fixed by this! But probably introduce other navigation bugs hehe

@pbastia 

Sorry, I should have elaborated, my asking "**Can** this issue be resolved in this refactor" is because my tests using this branch indicate [598](https://github.com/bcgov/cas-reporting/issues/598) is still an issue.

PR #2979, opened 2025-03-10
feat: add outstanding balance cell

nataliepiguliak:


vivanovbc (changes requested):
@nataliepiguliak outstanding balance should be BE-driven and dynamic based on monetary payments and applied compliance units. Since both monetary payments and compliance units are not implemented I think you can use excess emissions as part of outstanding balance function calculations on BE.

nataliepiguliak:
@vivanovbc I believe I addressed your comment, please take a look, many thanks!

vivanovbc (approved):
It looks good. There might be a more "elegant" way to set the summary outstanding_balance, but we can improve that later, as we are going to refactor it anyway.

PR #2982, opened 2025-03-10
422 lfo conditions small medium

gdalcengio:
[522](https://github.com/bcgov/cas-reporting/issues/522)

**Changes in this commit:**

- Adding facility type through multiple props and the getFacilityDetails() function to check for Small or Medium facilities
- Activities for Small/Medium facilities have dynamically added 'Not Applicable'
- Production data page for Small/Medium facilities now have:
	- ~Deselect all button~
	- 'Not Applicable' methodology dynamically added to each product
	- The ability to continue w/o a product ~while Large and Single facilities now NEED a product and disable save buttons without it~
- Allocation of Emissions page for Small/Medium facilities now has a **~new~ refactored table** of ReportEmissionAllocation to handle if there is no product
	- Note: this is since methodology is pulled from products which only Small/Medium facilities may not have now, so we bring the methodology to the parent table


bc_obps/reporting/migrations/0073_not_applicable_methodology.py
```diff
@@ -0,0 +1,41 @@
...
+def init_additional_methodology_data(apps, schema_monitor):
+    """
+    Add additional data to erc.methodology
+    """
+    Methodology = apps.get_model("reporting", "Methodology")
+    Methodology.objects.bulk_create(
+        [
+            Methodology(name="Not Applicable"),
+        ]
+    )
```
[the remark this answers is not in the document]
pbastia: Because this methodology is added to the schema based on facility type, not config/activity/source type etc..., we don't need to explicitly add it to the configuration elements (otherwise they'll show up everywhere, regardless of facility type)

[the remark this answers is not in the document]
pbastia: TLDR: nothing to do here :)

bc_obps/reporting/models/report_emission_allocation_no_product.py
```diff
@@ -0,0 +1,75 @@
...
+from reporting.models import report_version
+from reporting.models.facility_report import FacilityReport
+from registration.models.time_stamped_model import TimeStampedModel
+from reporting.models.triggers import immutable_report_version_trigger
+from reporting.models.rls_configs.report_emission_allocation_no_product import (
+    Rls as ReportEmissionAllocationNoProductRls,
+)
+
+
+class ReportEmissionAllocationNoProduct(TimeStampedModel):
```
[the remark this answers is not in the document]
dleard: @pbastia I've suggested keeping the `report_version_id` on the optional child model here since that is the pattern we have followed with all other child models in reporting that _could_ have a transitive relationship to the version, but we have chosen to explicitly pass that ID down in order to aid in flattening the data out for our end users. Any concerns there?

bc_obps/reporting/service/report_emission_allocation_service.py
```diff
@@ -129,14 +143,12 @@ def save_emission_allocation_data(
                     continue
 
                 report_emission_allocation_record, _ = ReportProductEmissionAllocation.objects.update_or_create(
+                    report_emission_allocation=report_emission_allocation,
```
pbastia, line 146: I created a card for this service: https://github.com/bcgov/cas-reporting/issues/694 but we'll leave it like this for now, the issues are not introduced by your PR

bc_obps/reporting/service/report_supplementary_version_service.py
```diff
@@ -469,31 +473,52 @@ def clone_report_version_facility_product_data(
...
+        old_emission_allocation = ReportEmissionAllocation.objects.filter(
+            report_version_id=old_report_version,
+            facility_report=old_facility_report,
+        ).first()
+        if old_emission_allocation:
+            cloned_allocation = copy.deepcopy(old_emission_allocation)
+            cloned_allocation.pk = None
+            cloned_allocation.report_version = new_report_version
+            cloned_allocation.facility_report = new_facility_report
+            cloned_allocation.save()
```
pbastia, line 497: we don't have to use `copy.deepcopy` here, there is an official documented way to duplicate django models: https://docs.djangoproject.com/en/5.0/topics/db/queries/#copying-model-instances

instead, you can do:
```suggestion
        emission_allocation_to_copy = ReportEmissionAllocation.objects.filter(
            report_version_id=old_report_version,
            facility_report=old_facility_report,
        ).first()
        if emission_allocation_to_copy:
            emission_allocation_to_copy.pk = None
            cloned_allocation.save()
```

and it will save a new object since the pk is None

bc_obps/reporting/service/report_supplementary_version_service.py
```diff
@@ -469,31 +473,52 @@ def clone_report_version_facility_product_data(
...
+
+    @staticmethod
+    def clone_report_version_facility_product_emission_allocations(
```
pbastia, line 500: Since they're part of the same model you can make `clone_report_version_facility_emission_allocation` clone both the ReportEmissionAllocation and the children ReportProductEmissionAllocation records, no need to have 2 separate methods for them.
You'll need to clone the parent record first, then the children

bciers/apps/reporting/src/app/components/facility/FacilityEmissionAllocationForm.tsx
```diff
@@ -125,7 +127,23 @@ export default function FacilityEmissionAllocationForm({
...
+  ) {
+    const object = {
+      type: "string",
+      title: "Methodology",
+      enum: ["Not Applicable", "OBPS Allocation Calculator", "Other"],
+    };
+    modifiedEmissionAllocationSchema.properties.allocation_methodology =
+      object as RJSFSchema;
+  }
+
```
pbastia, line 146: For emission allocation, the "Not Applicable" methodology is available to any type of facility, according to the AC of this card. We can add the value straight into the `emissionAllocationSchema` instead. 

Then `facilityType` is not needed and it simplifies things in `FacilityEmissionAllocationPage`

bciers/apps/reporting/src/app/components/finalReview/reviewDataFactory/productionDataFactoryItem.ts
```diff
@@ -8,10 +9,19 @@ const productionDataFactoryItem: ReviewDataFactoryItem = async (
 ) => {
   const productionData = await getProductionData(versionId, facilityId);
 
+  const facilityType = (await getFacilityReportDetails(versionId, facilityId))
+    .facility_type;
+  const productionMethodology = ["Small Aggregate", "Medium Facility"].includes(
+    facilityType,
+  )
+    ? ["Not Applicable", "OBPS Calculator", "other"]
+    : ["OBPS Calculator", "other"];
```
pbastia, line 18: This logic should be in the `buildProductionDataSchema` function, which has the responsiblity of building the schema.
Instead of passing it the `productionMethodology` array, we can pass the `facilityType` to it directly

bciers/apps/reporting/src/app/components/products/ProductionDataForm.tsx
```diff
@@ -97,6 +97,8 @@ const ProductionDataForm: React.FC<Props> = ({
       onChange={(data) => onChange(data.formData)}
       continueUrl={navigationInformation.continueUrl}
       errors={errors}
+      saveButtonDisabled={errors && errors.length > 0}
+      submitButtonDisabled={errors && errors.length > 0}
```
pbastia, line 101: `onChange` doesn't update the errors, so this means that if the server returns errors once, we won't be able to submit anymore. I don't think we need those 2 lines, they weren't part of the AC for your card anyways

bciers/apps/reporting/src/app/components/products/ProductionDataPage.tsx
```diff
@@ -21,10 +22,19 @@ export default async function ProductionDataPage({
...
   }));
 
+  const facilityType = (await getFacilityReportDetails(version_id, facility_id))
+    .facility_type;
+  const productionMethodology = ["Small Aggregate", "Medium Facility"].includes(
+    facilityType,
+  )
+    ? ["Not Applicable", "OBPS Calculator", "other"]
+    : ["OBPS Calculator", "other"];
+
```
pbastia, line 32: Duplicate code, above comment will fix it

bciers/apps/reporting/src/data/jsonSchema/facility/facilityEmissionAllocation.tsx
```diff
@@ -170,7 +170,7 @@ export const emissionAllocationSchema: RJSFSchema = {
     allocation_methodology: {
       type: "string",
       title: "Methodology",
-      enum: ["Not Applicable", "OBPS Allocation Calculator", "Other"],
+      enum: ["OBPS Allocation Calculator", "Other"],
```
pbastia, line 173: This change can also be reverted

pbastia (changes requested):
Very nice refactor! We can make just a few more adjustments that would remove some code

bc_obps/reporting/service/report_supplementary_version_service.py
dleard: Won't the nesting be off if we create these together?The ReportEmissionAllocation object is the parent, but it is 1:1 with a facility_report. The ReportProductEmissionAllocation child object is per product & needs the new ID of each cloned report_product, which is why they have a separate function called later in the nesting order

gdalcengio: Moved it to clone_report_version_facility_product_data function since it does not need to clone for each product. 

`if emission_allocation_to_copy:
            emission_allocation_to_copy.pk = None
            emission_allocation_to_copy._state.adding = True
            emission_allocation_to_copy.save()`

Returns 'Report emission allocation with this Report version and Facility report already exists.', so assigning report version like
`emission_allocation_to_copy.report_version = new_report_version` is necessary.

bc_obps/reporting/service/report_supplementary_version_service.py
```diff
@@ -456,6 +457,17 @@ def clone_report_version_facility_product_data(
...
     ) -> None:
+        # Clone the parent ReportEmissionAllocation
+        emission_allocation_to_copy = ReportEmissionAllocation.objects.filter(
+            report_version_id=old_report_version,
+            facility_report=old_facility_report_to_clone,
+        ).first()
+        if emission_allocation_to_copy:
+            emission_allocation_to_copy.pk = None
+            emission_allocation_to_copy._state.adding = True
+            emission_allocation_to_copy.report_version = new_report_version
```
dleard, line 468: Doesn't this need a fkey set to the new facility_report object?

pbastia: Ah it definitely should

bc_obps/reporting/service/report_supplementary_version_service.py
```diff
@@ -479,21 +491,25 @@ def clone_report_version_facility_product_data(
...
+            facility_report=new_facility_report,
+        ).first()
         for old_allocation_to_clone in old_allocations_to_clone:
-            cloned_allocation = copy.deepcopy(old_allocation_to_clone)
-            cloned_allocation.pk = None
-            cloned_allocation.report_version = new_report_version
-            cloned_allocation.facility_report = new_facility_report
-            cloned_allocation.report_product = new_report_product
-            cloned_allocation.save()
+            cloned_product_allocation = copy.deepcopy(old_allocation_to_clone)
```
pbastia, line 509: This also shouldn't use `deepcopy` and instead use the django method.

pbastia (approved):
Very nice! There's one instance left where we use `deepcopy`, no need to fix it here since I'm realizing it's everywhere in that service. I made a card for it: https://github.com/bcgov/cas-reporting/issues/699

Issue #2879, opened 2025-02-20
Reg1\Prod Data Migration - remove duplicate contacts


andrea-williams:
**Blocked by** #2878 
**Blocking** #2672 

#### Describe the task

Actual implementation of #2878 on prod data

This will need to be done BEFORE #2672 

#### Acceptance Criteria

- [x] run the script created in 2878
- [x] add unique constraint:
```
constraints = [
            models.UniqueConstraint(
                fields=['email', 'operator'],
                name='unique_email_per_operator',
                condition=models.Q(operator__isnull=False)
            )
        ]
```
- [ ] third

#### Definition of Done

- [ ] first
- [ ] second
- [ ] third

#### Additional context

- Add any other context about the task here.
- Or here

Issue #2339, opened 2024-10-07
Admin\Contacts - create new contact records for existing external users in prod


andrea-williams:
#### Describe the task

A clear and concise description of what the task is.

#### Acceptance Criteria

- [ ] write data migration script that creates a new Contact record for each existing external user in our prod database. 
- [ ] script also includes adding the necessary FK references in the `operator_contacts` table to connect the new Contact records to their relevant Operator
- [ ] script does not create a new Contact record for external users who already appear in the contacts table (query by email address to filter out these)

#### Additional context

- to be applied on production data around the time that Reg2 is launched in prod
- Or here

PR #2987, opened 2025-03-11
Prod migrations from reg1 to reg2 - Part 2

BCerki:
cards:
- https://github.com/bcgov/cas-registration/issues/2339
- https://github.com/bcgov/cas-registration/issues/2879

I had to change a couple of the practice migrations to play nice together. I've commented on the lines I've changed and I left the assertions in one of the migrations so the reviewer can more easily check my work. I'll delete them after review.


bc_obps/registration/migrations/0085_make_contacts_from_industry_users.py
```diff
@@ -0,0 +1,55 @@
...
+        users_user_operator = user.user_operators.filter(status="Approved").first()
+        operator = users_user_operator.operator if users_user_operator else None
+        if operator:
+            _ = Contact.objects.get_or_create(
+                email=user.email,
+                defaults={
+                    "first_name": user.first_name,
+                    "last_name": user.last_name,
+                    "phone_number": user.phone_number,
+                    "position_title": user.position_title,
```
BCerki, line 39: This line is new since the practice migrations; position_title was missing

bc_obps/registration/migrations/0087_remove_duplicate_contacts.py
```diff
@@ -0,0 +1,69 @@
...
+def count_stats(Contact) -> Dict[str, int]:
+    """Collects and returns key statistics about the data before and after migration."""
+    return {
+        'total_contacts': Contact.objects.count(),
+        'total_unique_contacts': Contact.objects.values(
+            'first_name', 'last_name', 'email', 'position_title', 'phone_number', 'business_role', 'operator', 'address'
+        )
+        .distinct()
+        .count(),
+        'total_unique_emails_and_operators': Contact.objects.values('email', 'operator').distinct().count(),
```
BCerki, line 23: This line is new since practice migrations. Once we created contacts from existing users, we ended up with some contacts that have the same email but different operators. This is because sometimes users would create an operator when theirs already existed, or make a new one instead of using the one imported from SWRS, etc. So we can only check for uniqueness of the email/operator combination, not email alone

Sepehr-Sobhani (approved):
Bravo! 🤩

Issue #2842, opened 2025-02-13
Reg\Prod Data Migration: set registration purpose for existing Reg1 operations


andrea-williams:
**Blocked by**: #2841, #2747 

#### Describe the task

Manual data migration - set registration purpose to "OBPS Regulated" by default

#### Acceptance Criteria

- [ ] apply the script created in #2841 against production data
- [ ] second
- [ ] third

#### Definition of Done

- [ ] first
- [ ] second
- [ ] third

#### Additional context

- Add any other context about the task here.
- Or here

Issue #2041, opened 2024-08-02
Reg\Prod Data Migration for Operation statuses


andrea-williams:
**Blocked:** teardown of Reg1, #2770 
**Blocking:** launch of Reg2

#### Describe the task

Shortly before launching Registration Part 2 in production, we will need to apply a data migration to update the status of all operations in our prod database.

#### Acceptance Criteria

- [ ] operations with a status of `NOT_STARTED` should keep that same status
- [ ] operations with a status of `DRAFT` should keep that same status
- [ ] operations with a status of either `PENDING`, `APPROVED`, or `CHANGES_REQUESTED` should have their status updated to `DRAFT`
- [ ] operations with a status of `DECLINED` should ??? -- TBC with business area

#### Additional context

- Add any other context about the task here.
- Or here

patrickisaac:
Refinement Notes:
1. Practice run - Write script to confirm this works (pulling prod data locally) -> can be done whenever
2. Once Reg1 is closed, that's when we actually run the script on PROD data -> ideally closer to PROD launch

andrea-williams:
@patriciarussellCAS I'm starting to write the script/practice run for this ticket, but noticed that we still have an outstanding question. What should we do with the one operation in prod that has a status of DECLINED? Do we set the status to "DRAFT", in which case that operator could submit a registration for it? Do we keep it as DECLINED (might create confusion, since Reg2 no longer has the concept of declining an operation)? Or do we just delete that record entirely from the database?

patriciarussellCAS:
If it's only one operation, I think we could just delete it to avoid confusion. It would be helpful to document the name of the Operation somewhere so we have a record of it? or could we see that in Metabase?

andrea-williams:
> Thanks [@andrea-williams](https://github.com/andrea-williams) If it's only one operation, I think we could just delete it to avoid confusion. It would be helpful to document the name of the Operation somewhere so we have a record of it? or could we see that in Metabase?

@patriciarussellCAS  if we archive it (which is really want I meant by 'delete', my bad) then it would still be visible in Metabase but wouldn't show up in the UI. I think this is our best path forward. Thanks!

PR #2993, opened 2025-03-12
Prod migrations from reg1 to reg 2 - Part 3/3

BCerki:
cards:

- https://github.com/bcgov/cas-registration/issues/2842
- https://github.com/bcgov/cas-registration/issues/2041

This PR:
- applies 2 practice migrations
- new migration to remove the unused statuses from the operation choices
- we don't have a po_user in dev so made that part conditional


bc_obps/registration/migrations/0090_migrate_reg1_operation_statuses.py
```diff
@@ -0,0 +1,50 @@
...
+    Operation = apps.get_model('registration', 'Operation')
+    User = apps.get_model('registration', 'User')
+
+    Operation.objects.filter(status="Pending").update(status="Draft")
+    Operation.objects.filter(status="Approved").update(status="Draft")
+    Operation.objects.filter(status="Changes Requested").update(status="Draft")
+
+    # we don't have a po_user outside of prod, so if we're not in prod, we end the migration here
+    if os.environ.get('ENVIRONMENT') != 'prod':
+        return
```
BCerki, line 33: Does this look okay? When we run `make_reset`, all the migrations run before the mock data, so everything after this fails because we don't have a po_user.user_guid

bc_obps/registration/migrations/0089_set_purpose_for_existing_operations.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def migrate_assign_regulated_purpose(apps, schema_monitor):
+    # import the required Django model
+    Operation = apps.get_model('registration', 'Operation')
+
+    for operation in Operation.objects.all():
+        if operation.bc_obps_regulated_operation is not None:
+            if operation.opt_in:
+                operation.registration_purpose = 'Opted-in Operation'
```
Sepehr-Sobhani, line 18: I think we should also set the `opt_in` to True on the operation instance.

bc_obps/registration/migrations/0089_set_purpose_for_existing_operations.py
```diff
@@ -0,0 +1,30 @@
...
+Purpose: set the `registration_purpose` field of Operations to OBPS Regulated Operation if the registration was assigned a BORO ID in Reg1.
+"""
+
+
+def migrate_assign_regulated_purpose(apps, schema_monitor):
+    # import the required Django model
+    Operation = apps.get_model('registration', 'Operation')
+
+    for operation in Operation.objects.all():
+        if operation.bc_obps_regulated_operation is not None:
```
Sepehr-Sobhani, line 16: We could use 
```python
for operation in Operation.objects.filter(bc_obps_regulated_operation__isnull=False):
```

bc_obps/registration/migrations/0090_migrate_reg1_operation_statuses.py
```diff
@@ -0,0 +1,50 @@
...
+
+    filtered_users = User.objects.filter(first_name="Patricia", last_name="Russell")
+
+    po_user = filtered_users.first()
+
+    Operation.objects.filter(status="Declined").update(
+        archived_by_id=po_user.user_guid, archived_at=datetime.datetime.now(datetime.timezone.utc)
+    )  # declined operations will be archived, per PR's instruction and using her user GUID
```
Sepehr-Sobhani, line 41: We could use `set_archive` function!

bc_obps/registration/migrations/0090_migrate_reg1_operation_statuses.py
```diff
@@ -0,0 +1,50 @@
...
+    Operation.objects.filter(status="Approved").update(status="Draft")
+    Operation.objects.filter(status="Changes Requested").update(status="Draft")
+
+    # we don't have a po_user outside of prod, so if we're not in prod, we end the migration here
+    if os.environ.get('ENVIRONMENT') != 'prod':
+        return
+
+    filtered_users = User.objects.filter(first_name="Patricia", last_name="Russell")
+
+    po_user = filtered_users.first()
```
Sepehr-Sobhani, line 37: We can combine these to use `.get(...)` instead!

bc_obps/registration/migrations/0090_migrate_reg1_operation_statuses.py
Sepehr-Sobhani: I suggest that instead of this condition, we check whether po_user exists, like this:
```python
po_user = User.objects.filter(first_name="Patricia", last_name="Russell").first()

if po_user:
    ...
```

Sepehr-Sobhani (approved):
Looks great! 💯

bc_obps/registration/migrations/0089_set_purpose_for_existing_operations.py
BCerki: Isn't it already true, or else we wouldn't make it past `if operation.opt_in`? It already being set to true is the only way we know to set the purpose to opt-in

bc_obps/registration/migrations/0090_migrate_reg1_operation_statuses.py
BCerki: I think @andrea-williams tried that but it wasn't working for some reason. It was in a call so I can't double-check the conversation

bc_obps/registration/migrations/0089_set_purpose_for_existing_operations.py
Sepehr-Sobhani: Hmmm 🤔 But I think we introduced `opt_in` field in Reg2, right

andrea-williams: opt_in will be an obsolete field soon - it's a leftover from Reg1 and won't be needed soon

bc_obps/registration/migrations/0090_migrate_reg1_operation_statuses.py
andrea-williams: that sounds familiar but I can't remember exactly what the issue was

bc_obps/registration/migrations/0089_set_purpose_for_existing_operations.py
Sepehr-Sobhani: Sounds good! Do we need another ticket for that? Do we need to do something about `opted_in_operation` field?!

BCerki: > Hmmm 🤔 But I think we introduced opt_in field in Reg2, right

We have it set in the current prod data, so it must be reg1 like Andrea mentions

> And what to do with opted_in_operation field?

In reg2, as soon as a user submits page 1 of registration with the opted-in purpose, a blank `opted_in_operation` is created.

So I think we have two options:
1. Create a blank `opted_in_operation` for every existing record with this purpose (the only field we need is created_by, and we can use the same person who created the operation)
2. Do nothing. When users come in to re-register, they'll have to go through page 1 and the `opted_in_operation` will be created as soon as they click "save and continue"--this calls OperationService.update_operation:
```
if (
            operation.registration_purpose == Operation.Purposes.OPTED_IN_OPERATION
            and operation.opted_in_operation is None
        ):
            operation = cls._create_opted_in_operation_detail(user_guid, operation)
```
If we do nothing though, we could end up with records that have a opted-in purpose with no record if a user never registers them. So I'm going to do option 1 unless anyone has feelings

Sepehr-Sobhani: I agree!

bc_obps/registration/migrations/0090_migrate_reg1_operation_statuses.py
```diff
@@ -0,0 +1,44 @@
...
+DRAFT              | DRAFT
+DECLINED           | Archive the operation
+"""
+
+
+def migrate_reg1_operation_statuses(apps, schema_monitor):
+    # import the required Django model
+    Operation = apps.get_model('registration', 'Operation')
+    User = apps.get_model('registration', 'User')
+    breakpoint()
```
Sepehr-Sobhani, line 25: leftover!

PR #2994, opened 2025-03-12
Remove and replace fixtures for reporting_year in prod_data migration

acatchpole:
Card: https://github.com/bcgov/cas-reporting/issues/578

In the prod_data migration, changed how the ReportingYear entries are created:
 - Removed the load from fixtures
 - Used the fixture values to generate entries based on the model

Also updated tests that were using the model bakery to create 2025 reporting_year

**** Notes
 - All the datetime values are the same as the fixtures, except they have been made local time (PST/PDT) instead of UTC.
 - 2025 and 2026 were added
 - 2023 was kept in because (despite the fact that it will never be a reporting_year) a compliance migration creating compliance_periods uses 2023 as a fk, and I didn't want to mess with it.

### DIFF
--- a/bc_obps/reporting/migrations/0008_prod_data.py
+++ b/bc_obps/reporting/migrations/0008_prod_data.py
@@ -634,15 +634,35 @@ def init_reporting_years(apps, schema_editor):
     '''
     Add initial year data to erc.reporting_year
     '''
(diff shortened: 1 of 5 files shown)


pbastia (approved):
👌

Issue #2884, opened 2025-02-20
Opt-in detail records are orphaned when archived


BCerki:
#### Describe the task

When the registration purpose changes from opted-in to something else, we set the FK of the opted-in-detail record to null (on Operation model) and archive the opted-in-detail record (OptedInOperationDetail). This orphans the detail record, because the only thing connecting it to an operation was the FK on Operation.

#### Acceptance Criteria

- [ ] first
- [ ] second
- [ ] third

#### Definition of Done

- [ ] first
- [ ] second
- [ ] third

#### Additional context

- Add any other context about the task here.
- Or here

BCerki:
@patriciarussellCAS

patriciarussellCAS:
Thanks! I'd like to chat more about this at next refinement to ensure I understand what's being deleted and what's still in our records. 
I'm not sure if there are instances where we need to keep a record of this. It's probably fine to proceed as Sep recommends, but want to make sure I'm clear first. 
cc: @patrickisaac

BCerki:
> Thanks! I'd like to chat more about this at next refinement to ensure I understand what's being deleted and what's still in our records. I'm not sure if there are instances where we need to keep a record of this. It's probably fine to proceed as Sep recommends, but want to make sure I'm clear first. cc: [@patrickisaac](https://github.com/patrickisaac)

I'll try to explain this one more clearly:
When a user registers an opted-in operation, we store their data in two places: 1) operation table, 2) opt-in details table. There is a link from 1) to 2). 
When a user changes their registration purpose from opt-in to something else, we update the information in the operation table (delete the link), and archive the info in the opt-in table. This means that in the opt-in table, we have archived records that aren't connected to an operation anymore. Sepehr suggests deleting the opt-in record instead of archiving it since we have no way to connect it back to the operation anyway.
On the other hand, if it IS important to maintain that link, then we'd change our implementation so the archived opt-in record and operation are still connected.

patriciarussellCAS:
Thanks for breaking that down so clearly @BCerki 
Given that, I tend to agree that these opt-in records could be deleted. I wouldn't expect EIOs to change their purpose so consider those low-risk as well. 

[...]

PR #3003, opened 2025-03-13
Chore: 2884 remove orphaned opt-in-detail records

Sepehr-Sobhani:
[ISSUE 2884](https://github.com/bcgov/cas-registration/issues/2884)


bc_obps/service/tests/test_operation_service.py
```diff
@@ -106,62 +106,6 @@ def test_assigns_single_selected_purpose():
         assert operation.updated_at is not None
         assert operation.registration_purpose == Operation.Purposes.REPORTING_OPERATION
 
-    @staticmethod
```
BCerki, line 109: I think we should keep/repurpose one of these to confirm that the detail record is deleted

BCerki (approved):
I didn't check any of the poetry parts as I think that's been covered by another review (but if not, let me know and I'll look). I manually tested changing the purpose in both admin and reg and everything worked.

Sepehr-Sobhani:
> I didn't check any of the poetry parts as I think that's been covered by another review (but if not, let me know and I'll look). I manually tested changing the purpose in both admin and reg and everything worked.

Andrea has already tested that part, but it would be great if you could also verify that Poetry installs properly on your local machine.

bc_obps/service/tests/test_operation_service.py
Sepehr-Sobhani: This test, `test_old_purpose_opted_in`, covers our it. However, I can create an additional test if you have any concerns.

BCerki: Right, all good then!

Issue #2914, opened 2025-02-25
UAT Bug: Error message after clicking save and continue when registering EIO -- addresses


TammieV-BCG:
#### Describe the Bug:
Error message after clicking save and continue on operation registration page when registering EIO

#### Steps to reproduce the behaviour:

1. Go to the BCIER test site 
2. Login using test Business BCeID 
3. navigate to register an operation 
4. select Electricity Import Operation as registration purpose 
5. Fill out form registering SFO Operation 7
6. click save and continue 

#### Screenshots:

![Image](https://github.com/user-attachments/assets/fefa00da-0212-4ddc-8e61-8ad330af4751)

#### Additional information:
Browser: chrome

#### Additional Context

User story: As an External User, given that I'm registering my operation as an Electricity Import Operation, then I only fill out the necessary fields for this registration purpose type

UAT ticket: https://teams.microsoft.com/l/entity/com.microsoft.teamspace.tab.planner/mytasks?tenantId=6fdb5200-3d0d-4a8a-b036-d3685e359adc&webUrl=https%3A%2F%2Ftasks.teams.microsoft.com%2Fteamsui%2FpersonalApp%2Falltasklists&context=%7B%22subEntityId%22%3A%22%2Fv1%2Fplan%2FbyyJUkJggkaPhsw6jAP-oWUAGWDg%2Fview%2Fboard%2Ftask%2FN63WvOJQ-Eusk_R_F21PHGUAGk3z%22%7D
-

patriciarussellCAS:
Tagging @TammieV-BCG for your awareness. Thanks @BCerki

fviduya:
Moving to accepted, this should be tested in the TEST environment. cc: @TammieV-BCG

patriciarussellCAS:
Reviewed in Test. no error message.

PR #3009, opened 2025-03-13
UAT Bug: Error message after clicking save and continue when registering EIO

BCerki:
card: https://github.com/bcgov/cas-registration/issues/2914

This error is thrown because:
- when switching from an SFO to an EIO, we update the facility
- EIOs don't have addresses, so we delete the old SFO facility address
- This throws an error because in the mock test data, we use the same address for everything

If we ever did somehow end up with records having the same address (the same fk, not just the same values), we wouldn't want to delete it without investigating further. So the only change necessary is to the test mock data.

This PR:
- first commit is to set up the local db with the fixtures from test. I'll delete it before merging
- second commit gives each facility a unique address

To test: 
- pull only the first commit and run `make clear_db` to clear any existing data, and then `make load_test_data`
- follow the reproduction instructions in the ticket and hit the bug
- pull the second commit and follow the reproduction instructions
- you should not hit the bug

Note that since we've fixed [this bug](https://github.com/orgs/bcgov/projects/123/views/16?pane=issue&itemId=100904556&issue=bcgov%7Ccas-registration%7C2971), if you select a purpose and then an operation, everything in the form will be replaced by the operation's data. We have a ticket to look at that: https://github.com/orgs/bcgov/projects/123/views/16?filterQuery=repo%3Abcgov%2Fcas-registration+-status%3Adone+operation+form&pane=issue&itemId=80445707&issue=bcgov%7Ccas-registration%7C2224

### DIFF
--- a/bc_obps/registration/fixtures/test/address.json
+++ b/bc_obps/registration/fixtures/test/address.json
@@ -18,5 +18,105 @@
       "province": "BC",
       "postal_code": "X0Y 1Z0"
     }
+  },
+  {
+    "model": "registration.address",
+    "pk": 3,
+    "fields": {
+      "street_address": "123 Main St",
+      "municipality": "City",
+      "province": "ON",
+      "postal_code": "A1B 2C3"
+    }
+  },
+  {
+    "model": "registration.address",
+    "pk": 4,
+    "fields": {
+      "street_address": "789 Oak St",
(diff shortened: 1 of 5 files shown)


Sepehr-Sobhani (approved):
Thank you for all the instructions and clarifications. 🤩

Issue #2664, opened 2025-01-09
Django points at the scanned version of a file after upload


pbastia:
#### Description:

Django automatically uploads a file to a bucket.
ClamAV (#2663) will move it to a 'scanned' bucket
- Django needs to know where to look when a user wants to download the file
- Django needs to know what to do when we delete the file

#### Process

1. File uploaded and flag model as 'Pending'
1. Create something to check buckets (ie. every couple minutes?). Cronjob or just a scheduled separate process on the main pod. Cronjob would be a lot of infrastructure work (create pod, startup container, load django and apps, etc) to run a very quick python command.
   1. Look for file in clean bucket
      1. Mark clean, if found
   1. else, check in quarantined bucket
      1. mark error, if found
1. Once found, update file location in model
  1. Mark error (for reupload) if neither found (maybe a dev one? since we'd have to check why ClamAV didn't handle it?)

#### Acceptance Criteria:

Given
When
Then

#### Development Checklist:

- [ ] Checklist item
- [ ] Checklist item
- [ ] Checklist item
- [ ] Meets the DOD

#### Definition of Ready (Note: If any of these points are not applicable, mark N/A)

- [ ] User story is included
- [ ] User role and type are identified
- [ ] Acceptance criteria are included
- [ ] Wireframes are included (if required)
- [ ] Design / Solution is accepted by Product Owner
- [ ] Dependencies are identified (technical, business, regulatory/policy)
- [ ] Story has been estimated (under 13 pts)

#### Definition of Done (Note: If any of these points are not applicable, mark N/A)

- [ ] Acceptance criteria are tested by the CI pipeline
- [ ] UI meets accessibility requirements
- [ ] Configuration changes are documented, documentation and designs are updated
- [ ] Passes code peer-review
- [ ] Passes QA of Acceptance Criteria with verification in Dev and Test
- [ ] Ticket is ready to be merged to main branch
- [ ] Can be demoed in Sprint Review
- [ ] Bugs or future work cards are identified and created
- [ ] Reviewed and approved by Product Owner

#### Notes:

-

joshgamache:
[...]

PR #3024, opened 2025-03-14
Feat/2664 malware scanned files

joshgamache:
Addresses #2664 . This makes Django and the frontend aware of the Malware scanner buckets. 

## Changes 🚧

- Update Django Storages to use `Clean`, `Unscanned` and `Quarantined` buckets in GCS
- Update Django document model to add scanning status
- Update Django document fetcher to use correct bucket based on status
- Add management command to check buckets and update document status
- Update fronted to be aware of scan status and show status-based notifications


bc_obps/service/data_access_service/document_service.py
```diff
@@ -24,6 +25,25 @@ def create_document(
...
             created_by_id=user_guid,
             operation_id=operation_id,
+            status=Document.FileStatus.UNSCANNED,
         )
 
         return document
+
+    @classmethod
+    def check_document_file_status(cls, document: Document) -> Document.FileStatus:
+        file_name = document.file.name
```
pbastia, line 35: is this one the file name seen by the user? or the underlying name with the appended hash?

helm/cas-bciers/templates/cronJobs/check-file-status.yaml
```diff
@@ -0,0 +1,35 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: check-document-file-status
+  labels: {{- include "cas-bciers.labels" . | nindent 4 }}
+    component: backend
+  namespace: {{ .Release.Namespace }}
+spec:
+  suspend: false
+  schedule: "*/1  * * * *"
```
pbastia, line 10: Every minute might be too often - unless we have a mechanism to check whether there is already another instance running. I'd suggest doing that, this way if a bug happens, we don't create a bunch of pods

pbastia: We could use a tool like `flock` to do that, this will just require a small PVC mounted on this pod? https://man7.org/linux/man-pages/man1/flock.1.html

pbastia (approved):
Looks great! Just left a few questions - mostly about the danger of unthrottled Cronjob running every minute

bc_obps/service/data_access_service/document_service.py
joshgamache: This is the file name seen by the user. The Storage handler takes care of knowing the hashed filename.

joshgamache: So it's both. If a filename is duplicate, it gets hashed, and in Django it's just known as that new hashed filename. This is the existing behaviour (see below, on in the Dev deployment), which we might want to address. I'll bring it up in the the teams Dev channel.

image

helm/cas-bciers/templates/cronJobs/check-file-status.yaml
joshgamache: @pbastia I've updated the helm charts with d0872bd to add a `flock` using a tiny 10mb PVC mount.

PR #3025, opened 2025-03-16
feat: fix compliance status

nataliepiguliak:


vivanovbc (changes requested):
I reviewed existing code and looks like we can just use a human string for value like in other places of TextChoices usage -> https://github.com/bcgov/cas-registration/blob/c253ecbf78da7c789a65ae07065fe0693329bb63/bc_obps/registration/models/operation.py#L31

PR #3026, opened 2025-03-17
feat: add obligation id

nataliepiguliak:


vivanovbc (changes requested):
After further consideration, it would be better to move this to the Obligation model. However, since it already has an ID, we should rename this field to `readableId` and store it in the database. The `readableId` should be generated when the obligation is created. In the UI, we will display `readableId` in the `Obligation ID` column.

pbastia:
Agreed with @vivanovbc , we should generate the ID once and store it. I suggest `obligation_id` for the field name, since it's the one our business area will be using and seeing in Metabase.
The reason for generating once and storing is that the format of that ID is subject to legal, regulatory or just regular program changes, and we should be able to update the generation method without impacting the existing obligation IDs that will be set on invoices, internal tracking sheets, etc.

bc_obps/compliance/models/compliance_summary.py
```diff
@@ -70,6 +70,37 @@ def compliance_status(self) -> str:
         else:
             return self.ComplianceStatus.OBLIGATION_FULLY_MET
 
+    @property
+    def obligation_id(self) -> str:
```
pbastia, line 74: This logic should be in a service of some sort, and an `obligation_id = CharField(...)` on the model

bc_obps/compliance/models/compliance_obligation.py
```diff
@@ -44,6 +44,12 @@ class PenaltyStatus(models.TextChoices):
         default=PenaltyStatus.NONE,
         db_comment="The status of the penalty (e.g., NONE, ACCRUING, PAID)",
     )
+    obligation_id = models.CharField(
+        max_length=30,
+        null=True,
```
vivanovbc, line 49: This field should be required for Compliance Obligation and always has value

bc_obps/service/compliance_obligation_service.py
```diff
@@ -72,6 +75,37 @@ def create_compliance_obligation(
...
+        # Get BORO ID if available (format: YY-OOOO)
+        if report_version.report.operation.bc_obps_regulated_operation:
+            # Use the BORO ID directly (already in YY-OOOO format)
+            operation_part = report_version.report.operation.bc_obps_regulated_operation.id
+        else:
+            # The operation has not been approved the BC OBPS
+            # Fallback: construct a similar format using reporting year and a default operation ID
+            # year_part = str(self.report.reporting_year.reporting_year)[-2:]
+            # operation_part = f"{year_part}-0001"
+            return None
```
vivanovbc, line 100: Are we sure about this logic? I don't think we need to create an obligation for non-regulated operations, they only need to report emissions, can you confirm?

bc_obps/service/compliance_summary_service.py
```diff
@@ -64,7 +64,7 @@ def create_compliance_summary(cls, report_version_id: int, user_guid: UUID) -> C
             # Create compliance obligation if there are excess emissions
             if compliance_data.excess_emissions > Decimal('0'):
                 ComplianceObligationService.create_compliance_obligation(
-                    summary.id, compliance_data.excess_emissions, report_version.report.reporting_year_id
+                    summary.id, compliance_data.excess_emissions, report_version
```
vivanovbc, line 67: Good catch

PR #3035, opened 2025-03-17
Fix inconsistent data display on activity form & emission summary forms

gdalcengio:


ayeshmcg (commented):
Great work on fixing Issue [590](https://github.com/bcgov/cas-reporting/issues/590)! The changes are well-implemented, and the issue with the data not showing immediately after saving is resolved. However, we are still encountering an error when no activity is selected, which is preventing the form from functioning as expected in certain cases.

The error happens in the ActivityForm.tsx file (TypeError: Cannot convert undefined or null to object),  This occurs when the form is submitted without selecting an activity.
image


Issue in Ticket 448: The issue reported in [cas-reporting#448](https://github.com/bcgov/cas-reporting/issues/448) still persists. After the form is saved, navigating away and coming back to the form does not refresh the emission summary correctly.

gdalcengio:
> Great work on fixing Issue [590](https://github.com/bcgov/cas-reporting/issues/590)! The changes are well-implemented, and the issue with the data not showing immediately after saving is resolved. However, we are still encountering an error when no activity is selected, which is preventing the form from functioning as expected in certain cases.
> 
> The error happens in the ActivityForm.tsx file (TypeError: Cannot convert undefined or null to object), This occurs when the form is submitted without selecting an activity. image
> 
> Issue in Ticket 448: The issue reported in [cas-reporting#448](https://github.com/bcgov/cas-reporting/issues/448) still persists. After the form is saved, navigating away and coming back to the form does not refresh the emission summary correctly.

The issue you are referencing in the image attached is part of ticket [577](https://github.com/bcgov/cas-reporting/issues/577). I will double check the emission summary page again

shon-button:
@gdalcengio 
@ayeshmcg 

Hi Gabe, I am curious if you tried to refresh the router cache using `router.refresh()` on the NavigationForm functions.

If that `router.refresh()`  resolves the caching issue I might suggest it to be a better solution than using `revalidatePath` to refresh the cache as the `revalidatePath`  strategy would require potentially adding `revalidatePath` to multiple server actions so to resolve the caching issue on pages other than activities page (ex: when "navigating away and coming back to the form does not refresh the emission summary correctly")

gdalcengio:
@ayeshmcg 

After double checking, I could not replicate the incorrect emission summary refreshing issue, the first solution worked fine for me for both instances, but regardless I've changed it to what @shon-button suggested with the router.refresh. I have tried it, Shon, the router.refresh solution works for me just as well. 

This PR is ready for another review

ayeshmcg (approved):
Nice work, @gdalcengio ! The router.refresh() solution works perfectly, @shon-button . Both issues are resolved, and it looks like this also resolves [#577](https://github.com/bcgov/cas-reporting/issues/577) as well.

Approving ! 🚀

PR #3041, opened 2025-03-18
2649 interal user access management

BCerki:
card: https://github.com/orgs/bcgov/projects/123/views/16?pane=issue&itemId=92498895&issue=bcgov%7Ccas-registration%7C2649

Notes:
- As discussed in standup, we don't have a status field on the user model and don't want to add one (I don't remember everyone's reasoning, but mine is that the status column would only apply to the internal users and would be confusing for external users because their access is managed by the user_operator model). This means that the status column in the grid is inferred from the user's role and archived_at field
- I reused the code from the external access requests page for consistency, so things may not look exactly like the design (e.g. button is declined instead of rejected)
- There's a MUI console warning about an out-of-range value. This happens because we don't want to show the pending role in the dropdown

This PR:
- migration to base the user model off TimeStampedModel
- endpoint to fetch the data for the grid (all internal users including archived ones)
- endpoint to update an individual user
- endpoint to check if user is archived
- updates the links in the dashboard tiles (do we need a migration for the prod one?)
- page/grid for cas_admin to update user's access (roles and archived/not archived)
- updated pending page for cas_pending users who log in and don't have access yet
- declined page and redirect to it in middleware

I struggled with how to handle the archived users here and there; suggestions welcome!

NEW
- added a backend check to make sure cas_admins can't change their own role. This came with a slight refactor of the user services
- non-admin internal roles should have view access to the grid according to the miro, so added this and tests


bc_obps/common/fixtures/dashboard/bciers/prod/internal.json
```diff
@@ -73,10 +73,11 @@
             "href": "mailto:GHGRegulator@gov.bc.ca",
             "icon": "Users",
             "content": "View, approve or decline IDIR user access requests, or to assign access type to users here.",
+            "access_roles": ["cas_admin"],
             "links": [
               {
                 "title": "Access Requests",
-                "href": "/tbd/"
+                "href": "/administration/users/"
```
BCerki, line 80: bri this is prod

bc_obps/registration/api/user/user_id.py
```diff
@@ -0,0 +1,35 @@
...
+    tags=USER_TAGS,
+    description="""Updates application role of the user and optionally archives them.
+""",
+    auth=authorize("authorized_irc_user"),
+)
+def update_user_role(request: HttpRequest, user_id: UUID, payload: UserUpdateRoleIn) -> Tuple[Literal[200], User]:
+    user = UserDataAccessService.update_user(user_guid=user_id, updated_data=payload, include_archived=True)
+    if payload.archive:
+        # brianna this isn't working
+        # user.set_archive(get_current_user_guid(request))
```
BCerki, line 28: thoughts?

bc_obps/registration/tests/endpoints/test_users_endpoint.py
```diff
@@ -0,0 +1,43 @@
...
+            custom_reverse_lazy("get_internal_users"),
+        )
+
+        # Assert: Verify the response status
+        assert response.status_code == 200
+
+        # Assert: Validate the response structure and data
+        response_json = response.json()
+        assert len(response_json) == 4
+        # assert response_json.keys() == {'count', 'items'}
```
BCerki, line 33:
```suggestion

```

bciers/apps/dashboard/middlewares/withAuthorizationDashboard.ts
```diff
@@ -40,7 +41,7 @@ export const withAuthorizationDashboard: MiddlewareFactory = (
 
     // Check if the user is authenticated via the jwt encoded in server side cookie
     const token = await getToken();
-
+    // brianna here
```
BCerki, line 44:
```suggestion

```

bciers/libs/utils/src/getUUIDFromEndpoint.ts
```diff
@@ -1,5 +1,6 @@
 export const endpointAllowList = [
   "registration/user/user-app-role",
+  // "registration/user/user-is-archived",
```
BCerki, line 3:
```suggestion
```

bc_obps/common/fixtures/dashboard/bciers/prod/internal.json
```diff
@@ -73,10 +73,12 @@
             "href": "mailto:GHGRegulator@gov.bc.ca",
             "icon": "Users",
             "content": "View, approve or decline IDIR user access requests, or to assign access type to users here.",
+
             "links": [
               {
                 "title": "Access Requests",
-                "href": "/tbd/"
+                "href": "/administration/users/",
```
BCerki, line 80: Do we need to do a migration for this, since it's for prod?

Sepehr-Sobhani (approved):
Amazing work! 🤩 
Left a couple of comments, but nothing blocking.

bc_obps/registration/tests/endpoints/test_users.py
```diff
@@ -0,0 +1,43 @@
...
+            custom_reverse_lazy("get_internal_users"),
+        )
+
+        # Assert: Verify the response status
+        assert response.status_code == 200
+
+        # Assert: Validate the response structure and data
+        response_json = response.json()
+        assert len(response_json) == 4
+        # assert response_json.keys() == {'count', 'items'}
```
Sepehr-Sobhani, line 33: leftover!

Sepehr-Sobhani (approved):
:shipit:

PR #423, opened 2023-12-13
285 readonly pending

BCerki:
card: https://app.zenhub.com/workspaces/team-moose-sprint-board-60d216175a2689000ec774d0/issues/gh/bcgov/cas-registration/285

PR #3044, opened 2025-03-19
Feat/500 report history page

ayeshmcg:
Card: https://github.com/bcgov/cas-reporting/issues/500
Changes:-
1. Added route for report history page
2. Added component report history to show report history table for a report
3. Added api and service to fetch data
4. Added tests

To test:-
1. Go to http://localhost:3000/reporting/reports
2. Click on Report history in more actions
3. If status is not started it will show page like 
image
4. If status is submitted or draft, it show all the report versions. 
image
5. All the report versions wil be visible when supplementary report wil be created which is done in [#423](https://github.com/bcgov/cas-reporting/issues/423)
6. Clicking on view details(for submitted report) wil take user to final review page which is done in [#257](https://github.com/bcgov/cas-reporting/issues/257)


pbastia (changes requested):
Awesome work! just a few things with the backend query and schema that we can simplify

bciers/apps/reporting/src/app/components/reportHistory/ReportHistoryPage.tsx
```diff
@@ -0,0 +1,31 @@
...
+  report_id,
+  searchParams,
+}: {
+  report_id: number;
+  searchParams: ReportHistorySearchParams;
+}) {
+  // Fetch operation name and reporting year only if report_id is valid
+  const operationName = isNaN(report_id)
+    ? ""
+    : (await getOperationName(report_id)).name;
```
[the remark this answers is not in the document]
ayeshmcg: left this one, because when we try to get report history of an operation with report status not started so when there is not report ID it should show the page with message like no report versions.... and if i remove NaN check it throws error right away for not started ones

ayeshmcg:
@pbastia modified the code for middleware, but do you think it would be better to have a separate endpoint for checking if report version is valid which wil only be called if report version exits

bc_obps/reporting/api/validate_user_report_version.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+class UserStatusEnum(Enum):
+    REGISTERED = "Registered"
+    NOT_REGISTERED = "Not Registered"
+    REGISTERED_AND_VALID = "Registered and Valid Report Version"
+    REGISTERED_AND_INVALID = "Registered and Invalid Report Version"
+
+
+@router.get("/validate_user", response=UserStatusResponse)
```
pbastia, line 21: this is much too generic, I'd do: `/report-version/{version_id}/user_operation_status`

pbastia: this way no need to pass query params

bciers/apps/reporting/src/app/utils/getOperationName.ts
```diff
@@ -0,0 +1,12 @@
...
+export async function getOperationName(reportId: number) {
+  const endpoint = `reporting/report-operation/${reportId}`;
+  const response = await actionHandler(endpoint, "GET");
+  if (response.error) {
+    throw new Error(
+      `Failed to fetch the operation Name for report ${reportId}.`,
+    );
+  }
+  return response;
+}
```
pbastia, line 12: we can delete this too

bciers/apps/reporting/src/app/bceidbusiness/industry_user_admin/reports/report-history/[report_id]/page.tsx
```diff
@@ -0,0 +1,4 @@
+import defaultPageFactory from "@bciers/components/nextPageFactory/defaultPageFactory";
```
Comment (pbastia) [inline bciers/apps/reporting/src/app/bceidbusiness/industry_user_admin/reports/report-history/[report_id]/page.tsx:1]: There should be an identical page in `industry_user`

bc_obps/reporting/api/report_history.py
```diff
@@ -0,0 +1,41 @@
...
+    auth=authorize("approved_industry_user"),
+)
+@paginate(PageNumberPagination, page_size=PAGE_SIZE)
+def get_report_history(request: HttpRequest, report_id: int) -> QuerySet[ReportVersion]:
+    report_versions = ReportingHistoryDashboardService.get_report_versions_for_report_history_dashboard(report_id)
+    return report_versions
+
+
+@router.get(
+    "/report-operation/{report_id}",
```
pbastia, line 33: this should be `/reports/{report_id}/operation` to match our API guidelines

pbastia (changes requested):

bciers/apps/reporting/src/app/components/reportHistory/ReportHistoryPage.tsx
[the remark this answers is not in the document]
pbastia: `report_id` is guaranteed to be correct here, since it comes from the URL. It would never be `NaN`

bc_obps/reporting/api/validate_user_report_version.py
pbastia: API route should match file name for this one

pbastia (approved):
👍

Issue #2685, opened 2025-01-14
Contact's address cannot be deleted if contact is an operation rep


BCerki:
#### Describe the task
Blocked by https://github.com/orgs/bcgov/projects/123/views/16?pane=issue&itemId=88758186&issue=bcgov%7Ccas-registration%7C2524

We need to have address data for operation reps for legislation reasons. Currently, a user can delete contact address information (via the backend only; fields are mandatory on the FE) even if the contact is assigned as an operation rep somewhere.

#### Acceptance Criteria

- [x] address fields are editable
- [x] form can't be submitted if address fields are edited to be blank. Address is optional in the db, so will have to do a check in the services
- [ ] third

#### Definition of Done

- [ ] first
- [ ] second
- [ ] third

#### Additional context

- Add any other context about the task here.
- Or here

Sepehr-Sobhani:
Mostly backend/database changes and can't be tested.

PR #3052, opened 2025-03-19
Chore: 2685 operation representative address can not be deleted

Sepehr-Sobhani:
[ISSUE 2685](https://github.com/bcgov/cas-registration/issues/2685)


bc_obps/registration/tests/endpoints/_contacts/test_contact_id.py
```diff
@@ -141,43 +143,50 @@ def test_industry_user_admin_update_contact_and_address(self):
...
-            custom_reverse_lazy("update_contact", kwargs={"contact_id": contact.id}),
-        )
-        assert response.status_code == 200
-        assert Contact.objects.count() == 1
-        assert Address.objects.count() == 1  # One for the operator
+        # override the get method to return the senior officer role(otherwise it will return the operation representative role and the test will fail)
+        with patch('registration.models.BusinessRole.objects.get', return_value=senior_officer_role) as mock_get:
+            # Assert that we have only one contact(to make sure we are updating the contact and not creating a new one)
+            assert Contact.objects.count() == 1
+            # Assert that we are not creating a new address
```
BCerki, line 156: Suggest adding to the comment where the two addresses come from. I always forget which bakers add addresses

bc_obps/registration/tests/models/test_contact.py
```diff
@@ -36,3 +37,70 @@ def setUpTestData(cls):
...
+    def test_no_trigger_when_address_already_null(self):
+        """Test that updating a contact with NULL address doesn’t trigger if address stays NULL."""
+        # Update another field
+        self.null_address_contact.email = "updated_null_op_rep@example.com"
+        self.null_address_contact.save()
+
+        # Verify the update succeeded and address is still NULL
+        self.null_address_contact.refresh_from_db()
+        self.assertEqual(self.null_address_contact.email, "updated_null_op_rep@example.com")
+        self.assertIsNone(self.null_address_contact.address)
```
BCerki, line 106: Similar to above question, do we need to allow this case? If someone somehow got away with giving us a null address, we'd probably want to them fix it

bc_obps/service/contact_service.py
```diff
@@ -136,3 +142,12 @@ def raise_exception_if_contact_missing_address_information(cls, contact_id: int)
             raise Exception(
                 f'The contact {contact.first_name} {contact.last_name} is missing address information. Please return to Contacts and fill in their address information before assigning them as an Operation Representative here.'
             )
+
+    @classmethod
+    def validate_operation_representative_address(cls, contact_id: int, address_data: DictStrAny) -> None:
```
BCerki, line 147: start function name with _?

Sepehr-Sobhani:
> * I don't think your PR introduced this, but when editing an existing contact there's a problem with the error clearing: https://www.loom.com/share/aed330b2cd9342b0b242438d0cb0e4fb

That contact is an Operation Representative, so we must prevent the user from clearing address fields.

Sepehr-Sobhani:
> * I was able to create an new contact with role operation rep without address info:
>   image.

You can create an OP REP without an address, but you cannot delete the address or remove address fields. This is because the contact instance is created first, and the address is assigned afterward (FK relationship). Preventing users from creating the contact isn't an option, as many other contact types don't require an address.

Sepehr-Sobhani:
> * If I try to assign a contact without an address as a op rep (with rjsf required turned off), I get this:
>   image (a user would have a really hard time getting here, probably not worth addressing

This is correct because schema validation prevents the user from doing that. However, we currently don't have a way to display backend schema validation errors since we make fields required in RJSF.  
I'm going to make a small tweak to show the first error when a validation error occurs. I don't expect this to happen often, but we can improve it later if we need to display multiple errors at once.

bc_obps/registration/models/address.py
```diff
@@ -30,5 +31,30 @@ class Address(BaseModel):
...
+                # Prevent setting address fields to empty or NULL for Operation Representative
+                name="protect_fields_empty_update",
+                level=pgtrigger.Row,  # Row-level trigger
+                when=pgtrigger.Before,  # Trigger before the update
+                operation=pgtrigger.Update,  # Only for UPDATE operations
+                func="""
+                    if exists (
+                        select 1
+                        from erc.contact
+                        where address_id = new.id
```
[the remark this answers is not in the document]
Sepehr-Sobhani: Yes, new is from PostgreSQL’s trigger system, used by pgtrigger. In a `BEFORE UPDATE` trigger, new is the proposed row data.
https://www.postgresql.org/docs/current/plpgsql-trigger.html#PLPGSQL-DML-TRIGGER-NEW

bc_obps/registration/models/contact.py
```diff
@@ -44,5 +45,26 @@ class Meta(TimeStampedModel.Meta):
...
+                name="protect_address_null_update",
+                level=pgtrigger.Row,  # Row-level trigger
+                when=pgtrigger.Before,  # Trigger before the update
+                operation=pgtrigger.Update,  # Only for UPDATE operations
+                condition=pgtrigger.Q(
+                    old__address__isnull=False,  # Old address was not null
+                    new__address__isnull=True,  # New address is null
+                ),
+                func="""
+                    if new.business_role_id = 'Operation Representative'
```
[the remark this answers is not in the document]
Sepehr-Sobhani: if new.business_role_id is 'Operation Representative' (either unchanged or explicitly set), the func will raise the exception.

bc_obps/registration/models/contact.py
```diff
@@ -44,5 +45,26 @@ class Meta(TimeStampedModel.Meta):
...
+        triggers = [
+            *TimeStampedModel.Meta.triggers,
+            pgtrigger.Trigger(
+                # Prevent setting address to NULL for Operation Representative
+                name="protect_address_null_update",
+                level=pgtrigger.Row,  # Row-level trigger
+                when=pgtrigger.Before,  # Trigger before the update
+                operation=pgtrigger.Update,  # Only for UPDATE operations
+                condition=pgtrigger.Q(
+                    old__address__isnull=False,  # Old address was not null
```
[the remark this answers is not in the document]
Sepehr-Sobhani: Good catch!

bc_obps/registration/tests/models/test_contact.py
Sepehr-Sobhani: I have changed that part of the trigger so this test is no longer valid.

bc_obps/registration/models/contact.py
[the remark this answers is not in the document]
Sepehr-Sobhani: After implementing this, I noticed that removing the `old__address__isnull=False` condition causes an issue. If a contact already exists in the database and the user selects it in the OP REP step of registration, but the contact has no address, the user is prevented from filling out the address fields. This happens because we update the contact fields first, and by the time we process the address, an error occurs.

bc_obps/registration/tests/models/test_contact.py
Sepehr-Sobhani: Test reverted 😃 
https://github.com/bcgov/cas-registration/pull/3052#discussion_r2019423057

BCerki:
> > * I don't think your PR introduced this, but when editing an existing contact there's a problem with the error clearing: https://www.loom.com/share/aed330b2cd9342b0b242438d0cb0e4fb
> 
> That contact is an Operation Representative, so we must prevent the user from clearing address fields.

Yes! I mean that usually a user triggers an error, fixes it, and then resubmits, the error disappears on successful submission. On the contact page, the error doesn't disappear until you refresh.

Sepehr-Sobhani:
> > > * I don't think your PR introduced this, but when editing an existing contact there's a problem with the error clearing: https://www.loom.com/share/aed330b2cd9342b0b242438d0cb0e4fb
> > 
> > 
> > That contact is an Operation Representative, so we must prevent the user from clearing address fields.
> 
> Yes! I mean that usually a user triggers an error, fixes it, and then resubmits, the error disappears on successful submission. On the contact page, the error doesn't disappear until you refresh.

😮

BCerki (approved):
If you want to address the error message refresh in this PR, I can look again, otherwise it can be a separate bug ticket and this is good to go

Sepehr-Sobhani:
> If you want to address the error message refresh in this PR, I can look again, otherwise it can be a separate bug ticket and this is good to go

I'll try to fix it if it's a quick fix; otherwise, I'll create a bug ticket.

Sepehr-Sobhani:
> If you want to address the error message refresh in this PR, I can look again, otherwise it can be a separate bug ticket and this is good to go

Fixed!

PR #3056, opened 2025-03-20
chore: fixed source type bug

ayeshmcg:
Card: https://github.com/bcgov/cas-reporting/issues/621
This PR fixes the issue on Petroluem refining activity for source type Equipment leaks. The source type should be Equipment leaks at refineries.


dleard (approved):
Confirmed the bug has been fixed, nice work!

PR #3059, opened 2025-03-20
Chore/512 pulp paper compliance exception

dleard:
This PR implements handling an exception for Pulp & Paper. [Issue 512](https://github.com/bcgov/cas-reporting/issues/512)

On the Frontend it should:
- throw a validation error on the Production page if: 
  - the activity Pulp & paper is reported,
  - there are emissions reported that are categorized as both industrial process & excluded 
  - and the chemical pulp product is not reported
- throw a validation error on the Emission allocation page if:
  - the activity Pulp & paper is reported,
  - there are emissions reported that are categorized as both industrial process & excluded
  - and chemical pulp product is not reported OR if the amount allocated to chemical pulp - sum of emissions categorized as both industrial process & excluded is less than 0.

On the Backend it should:
- Subtract the total of emissions categorized as both industrial process & excluded from the total industrial process emissions allocated to the chemical pulp product

### DIFF
--- a/bciers/apps/reporting/src/app/components/products/ProductionDataForm.tsx
+++ b/bciers/apps/reporting/src/app/components/products/ProductionDataForm.tsx
@@ -14,6 +14,8 @@ interface Props {
   initialData: ProductData[];
   schema: RJSFSchema;
   navigationInformation: NavigationInformation;
+  isPulpAndPaper: boolean;
+  overlappingIndustrialProcessEmissions: number;
 }
 
 const ProductionDataForm: React.FC<Props> = ({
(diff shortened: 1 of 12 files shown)


bciers/apps/reporting/src/app/components/products/ProductionDataForm.tsx
```diff
@@ -50,6 +54,17 @@ const ProductionDataForm: React.FC<Props> = ({
...
 
   const onSubmit = async (data: any) => {
+    /*
+      Handle pulp & paper overlapping industrial process exception:
+      If pulp & paper is reported and there are industrial process emissions that are also categorized as excluded (ie: woody biomass)
+      Then the 'Pulp and paper: chemical pulp' product must be reported
+    */
+    if (isPulpAndPaper && overlappingIndustrialProcessEmissions > 0) {
+      if (!data.product_selection.includes("Pulp and paper: chemical pulp")) {
+        setErrors(["Missing Product: 'Pulp and paper: chemical pulp'"]);
```
pbastia, line 64: Should we add to the message: "Please add the product on the operation or facility information page" ?

PR #3071, opened 2025-03-21
fix: malware scanning file status cronjob errors

joshgamache:
The cronjob function to check the status of documents as they move through malware scanning is failing in OpenShift for two reasons: 
- Missing GCP application credentials.
- No 'user' being identified when updating the file status (and therefore the location to target) in the DB, so no update record could be added. 

This PR addresses both of those issues.


dleard (approved):
:shipit:

pbastia (approved):
🥇

PR #3072, opened 2025-03-21
fix: only add source types from the formData to the query string

dleard:
Details on the bug & how to reproduce the path are in the [issue ticket](https://github.com/bcgov/cas-reporting/issues/591)


shon-button:
@dleard 

image

shon-button (approved):

PR #3073, opened 2025-03-21
Feat/583 validate report access

pbastia:
Work done:
- add a "composite_auth(...)" method to compose multiple authorization methods
- A generic method to check whether the request's `current_user` has access to the report version referenced by the URL 
- A generic test method to test whether an API endpoint has that validation in its `authorize` call
- Added the validation to all required reporting endpoints

### DIFF
--- a/bc_obps/reporting/tests/api/test_activity_data_api.py
+++ b/bc_obps/reporting/tests/api/test_activity_data_api.py
@@ -4,6 +4,7 @@ from registration.tests.utils.bakers import operator_baker
 from model_bakery import baker
 
 from registration.utils import custom_reverse_lazy
+from reporting.tests.utils.report_access_validation import assert_report_version_ownership_is_validated
 
 
 class TestActivityData(CommonTestSetup):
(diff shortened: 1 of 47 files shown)


bc_obps/reporting/tests/api/test_activity_data_api.py
Sepehr-Sobhani, line 1: using `_api` suffix to prevent file naming conflict?

bc_obps/reporting/tests/api/test_fuel.py
```diff
@@ -1,22 +1,23 @@
 from django.test import Client
 import pytest
+from registration.tests.utils.helpers import CommonTestSetup, TestUtils
 
 pytestmark = pytest.mark.django_db
```
Sepehr-Sobhani, line 5: We can remove this line, `CommonTestSetup` already has that.

Sepehr-Sobhani (approved):
Neat! 🤩

PR #3078, opened 2025-03-24
chore: add endpoint is_supplementary_report_version

shon-button:
Addresses [646](https://github.com/bcgov/cas-reporting/issues/646)

### 🚀 Impact:

- Added service `bc_obps/service/report_version_service.py\is_initial_report_version` that checks if a report version is the initial report version
- Added endpoint `bc_obps/reporting/api/report_supplementary_version.py\is_supplementary_report_version` used to determine if a given report version is supplementary report (i.e., not the initial report version)


bc_obps/reporting/service/report_supplementary_version_service.py
```diff
@@ -27,6 +27,36 @@
...
+        # Check if there exists any other submitted ReportVersion for the same report.
+        # Exclude the current report_version from the query
+        is_supplementary_report: bool = (
+            ReportVersion.objects.filter(
+                report=report_version.report,
+                status=ReportVersion.ReportVersionStatus.Submitted,
+            )
+            .exclude(id=report_version_id)
+            .exists()
+        )
```
pbastia, line 55: This logic doesn't quite work for the case where we pass the id of the very first submission - not a supplementary report -, but another submitted report version exists.

"Supplementary report" being an official term in the regs and not the only reason one might have re-submission (corrective reports at the director's orders is another one), maybe we could rename this endpoint `is_initial_submission` and return whether this is the lowest id within that report?

```py
is_initial_submission: bool = (
            ReportVersion.objects.filter(
                id=report_version_id,
                status=ReportVersion.ReportVersionStatus.Submitted,
            )
            .annotate(min_report_version=Min("report__report_versions__id")
            .filter(id=F('min_report_version'))
            .exists()
        )
```

pbastia (changes requested):
very nice! Just missing an edge case not covered by the logic

bc_obps/reporting/service/report_supplementary_version_service.py
shon-button: @pbastia 

Thank-you very much for the clarification on the possible scenarios of reports with "Submitted" status - noted!

My goal here was to provide a way to determine if a report was created using service `create_report_supplementary_version` and use that information in the `Supplementary Report - ` users stories with conditions such as `In supplementary report’s sign off page, remove...and replace by...`

Perhaps, there is a better way to determine `is_supplementary_report` then with an endpoint?
Would adding a field to a model be a better choice?

Appreciate your guidance on this! :)

pbastia: I think `is_supplementary_report` is just `not is_initial_submission` ?

shon-button: **NOTE:** 
created service `is_initial_report_version` instead of `is_initial_submission` which returned `false` on initial versions in `Draft` and would not be useful in determining if supplementary report

pbastia (approved):
👌

PR #3095, opened 2025-03-26
Hotfix/populate operation/facility timeline with prod data

Sepehr-Sobhani:


andrea-williams (approved):
I haven't tested this at all, but it all looks good to me and I trust ya!

Thanks for fixing this Sep!

Sepehr-Sobhani:
> I haven't tested this at all, but it all looks good to me and I trust ya!
> 
> Thanks for fixing this Sep!

I’ve tested this locally, which is why it took me a while to create the PR. Hopefully, everything goes well. :)

Sepehr-Sobhani:
This one is on top of Develop so replaced by: https://github.com/bcgov/cas-registration/pull/3099

PR #3096, opened 2025-03-26
refactor: uncouple report sign-off data type from api

acatchpole:
Card: https://github.com/bcgov/cas-reporting/issues/663
Related Card: https://github.com/bcgov/cas-reporting/issues/451

#### Changes
Refactored report_sign_off_service to not use ninja schema data type that was also used by API. service data type has been uncoupled and is now defined within the service. 

Other changes were refactoring the data to and in the API to lower transformations away from the base formData.

### DIFF
--- a/bc_obps/reporting/service/report_sign_off_service.py
+++ b/bc_obps/reporting/service/report_sign_off_service.py
@@ -1,14 +1,50 @@
 import datetime
 from django.db import transaction
 from reporting.models.report_sign_off import ReportSignOff
-from reporting.schema.report_sign_off import ReportSignOffAcknowledgements, ReportSignOffIn
 from django.core.exceptions import ValidationError
+
+
+class ReportSignOffAcknowledgements:
+    def __init__(
+        self,
+        acknowledgement_of_review: bool,
+        acknowledgement_of_records: bool,
+        acknowledgement_of_information: bool | None,
+        acknowledgement_of_possible_costs: bool,
+        acknowledgement_of_new_version: bool | None,
+    ):
+        self.acknowledgement_of_review = acknowledgement_of_review
+        self.acknowledgement_of_records = acknowledgement_of_records
+        self.acknowledgement_of_information = acknowledgement_of_information
+        self.acknowledgement_of_possible_costs = acknowledgement_of_possible_costs
+        self.acknowledgement_of_new_version = acknowledgement_of_new_version
+
+
+class ReportSignOffData:
+    def __init__(
+        self,
+        acknowledgement_of_review: bool,
+        acknowledgement_of_records: bool,
+        acknowledgement_of_information: bool | None,
+        acknowledgement_of_possible_costs: bool,
+        acknowledgement_of_new_version: bool | None,
+        signature: str,
+    ):
+        self.acknowledgements = ReportSignOffAcknowledgements(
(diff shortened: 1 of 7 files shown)


bc_obps/reporting/service/report_sign_off_service.py
```diff
@@ -1,14 +1,50 @@
...
+        acknowledgement_of_records: bool,
+        acknowledgement_of_information: bool | None,
+        acknowledgement_of_possible_costs: bool,
+        acknowledgement_of_new_version: bool | None,
+    ):
+        self.acknowledgement_of_review = acknowledgement_of_review
+        self.acknowledgement_of_records = acknowledgement_of_records
+        self.acknowledgement_of_information = acknowledgement_of_information
+        self.acknowledgement_of_possible_costs = acknowledgement_of_possible_costs
+        self.acknowledgement_of_new_version = acknowledgement_of_new_version
```
pbastia, line 20: Looks like this would do the exact same thing! I'd be up for experimenting this as a pattern for our service communication across our app

```suggestion
@dataclass
class ReportSignOffAcknowledgements:
        acknowledgement_of_review: bool,
        acknowledgement_of_records: bool,
        acknowledgement_of_information: bool | None,
        acknowledgement_of_possible_costs: bool,
        acknowledgement_of_new_version: bool | None,
```

pbastia (commented):
Awesome! This looks pretty good, except for the fact that it's now a long rebase. Given the scope of the changes it'll probably be way quicker to just start a new PR

bc_obps/reporting/service/report_submission_service.py
```diff
@@ -10,15 +12,46 @@
...
+                    raise Exception("missing_report_verification")
+
+                # Check for the attachment only if mandatory
+                ReportAttachment.objects.get(
+                    report_version_id=version_id,
+                    attachment_type=ReportAttachment.ReportAttachmentType.VERIFICATION_STATEMENT,
+                )
+        except ReportAttachment.DoesNotExist:
+            raise Exception("verification_statement")
+
```
pbastia, line 51: I don't think this method exists anymore, it's been replaced by the `ReportValidationService` called below

bciers/apps/reporting/src/app/components/signOff/SignOffForm.tsx
```diff
@@ -66,10 +66,8 @@ export default function SignOffForm({
...
 
       const payload = {
-        acknowledgements: {
-          ...rest,
-          ...supplementary,
-        },
+        ...rest,
+        ...supplementary,
         signature,
         date,
```
pbastia, line 72: with this change, we now have `payload == formState` so we could submit directly?

pbastia (approved):
Just a couple of things from the rebase! Looks good to me otherwise

bciers/apps/reporting/src/app/components/signOff/SignOffForm.tsx
acatchpole: implemented

bc_obps/reporting/service/report_submission_service.py
acatchpole: made sure using the latest version of this file and updated sign off data type

PR #3097, opened 2025-03-26
Add django-querycount package

Sepehr-Sobhani:


andrea-williams (commented):
is there a way to set default settings for querycount that all devs could share? Otherwise I could see there might be some problems if the high/medium/etc. thresholds are configured differently for each dev

Sepehr-Sobhani:
> is there a way to set default settings for querycount that all devs could share? Otherwise I could see there might be some problems if the high/medium/etc. thresholds are configured differently for each dev

Since we are not defining those thresholds, the package uses the default setting in: https://github.com/bradmontgomery/django-querycount/blob/main/querycount/qc_settings.py

andrea-williams (approved):

Sepehr-Sobhani:
@andrea-williams made a small tweak to show the 2 most duplicated SQL queries in terminal:
Screenshot 2025-03-27 at 10 11 35 AM

PR #3114, opened 2025-03-27
Fix/639 incorrectly persisted report data

dleard:
Should fix cas-reporting issues [639](https://github.com/bcgov/cas-reporting/issues/639), [643](https://github.com/bcgov/cas-reporting/issues/643), [652](https://github.com/bcgov/cas-reporting/issues/652), [651](https://github.com/bcgov/cas-reporting/issues/651).

The root problem for all of these issues was that on removing activities from the facility review or operation review pages, the many to many tables were being updated, but the actual report data was persisted. This fix delete/cascades all report_activity data for the removed activity IDs & all allocation records as well (as they will be incorrect now that we've removed an activity worth of emission data).

Also found & fixed during this work: LFOs were using the `FacilityReportService` when updating facility_report data, but SFOs were not. SFOs were only updating the set of activity IDs when changes were made on the OperationReview page directly in the `ReportService`.

Refactored some issues with the services being dependent on ninja schemas.

### DIFF
--- a/bc_obps/service/report_service.py
+++ b/bc_obps/service/report_service.py
@@ -1,19 +1,49 @@
 from uuid import UUID
+from django.db.models import Case, When, Value, BooleanField
 from django.db import transaction
 from django.db.models import QuerySet
-from django.db.models import Case, When, Value, BooleanField
-from registration.models import Activity, RegulatedProduct
+from registration.models import Activity
+from registration.models import RegulatedProduct
 from registration.models.operation import Operation
 from reporting.models import ReportOperationRepresentative
 from reporting.models.report import Report
 from reporting.models.facility_report import FacilityReport
 from reporting.models.report_operation import ReportOperation
 from reporting.models.report_version import ReportVersion
-from reporting.schema.report_operation import ReportOperationIn
 from service.data_access_service.report_service import ReportDataAccessService
 from service.data_access_service.reporting_year import ReportingYearDataAccessService
 from service.report_version_service import ReportVersionService
 from django.forms.models import model_to_dict
+from service.facility_report_service import FacilityReportService, SaveFacilityReportData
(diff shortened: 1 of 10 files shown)


bc_obps/service/report_service.py
```diff
@@ -1,19 +1,49 @@
...
+        self.operation_name = operation_name
+        self.operation_type = operation_type
+        self.registration_purpose = registration_purpose
+        self.bc_obps_regulated_operation_id = bc_obps_regulated_operation_id
+        self.activities = activities
+        self.regulated_products = regulated_products
+        self.operation_report_type = operation_report_type
+        self.operation_representative_name = operation_representative_name
+        self.operator_trade_name = operator_trade_name
+        self.operation_bcghgid = operation_bcghgid
```
pbastia, line 46: Dataclasses are friends now! They build constructors, equality, and a bunch of other things like frozen/read-only, ... automatically, we just need to list which fields.

https://docs.python.org/3/library/dataclasses.html

```suggestion
@dataclass
class SaveReportOperationData:
        operator_legal_name: str,
        operation_name: str,
        operation_type: str,
        registration_purpose: str,
        bc_obps_regulated_operation_id: str,
        activities: List[int],
        regulated_products: List[int],
        operation_report_type: str,
        operation_representative_name: List[int],
        operator_trade_name: Optional[str] = None,
        operation_bcghgid: Optional[str] = None,
    
```

pbastia (approved):
Very nice! I found out on Friday about dataclasses, I think we should use them everywhere now, it just cuts down on code

bc_obps/service/report_service.py
dleard: Nice! I'll update this in the follow up PR for https://github.com/bcgov/cas-reporting/issues/650

PR #3053, opened 2025-03-19
feat: add validation for report activity data before saving

acatchpole:

PR #3116, opened 2025-03-27
Feat/452 flexible validation service

pbastia:
## Work done:
Refactoring the way reports are validated, to let developers add and test independent validation plugins.

- Added a `report_validation_service`
- Refactored the existing validation from the report submission service into the new pattern.

<br>
<br>
<br>

## Pattern

The report validation pattern follows a plugin architecture.
Plugins are then automatically discovered and collected by the validation service, run in a sequence, and errors are then aggregated and returned.

## Adding a new validation plugin

- Report validators must be python modules in the `reporting/service/report_validation/validators` folders.
- They must contain a `validate` function returning a dictionary of errors
- They must be registered in the **init**.py file to be discovered by the validation service

#### To add a new validator:

- Create a `name_of_validator.py` file in `reporting/service/report_validation/validators`
- In `name_of_validator.py`, create a `validate` function such as:

```py
def validate(report_version: ReportVersion): dict[str, ReportValidationError]:
    ...do_stuff()
    return {...}
```

- In `reporting/service/report_validation/validators/__init__.py`, import the validator and add it to the **all** array

## Validate a report version

The `ReportValidationService` class collects the validation plugins at startup time.
To validate a report version, call `validate_report_version(version_id)`, and the service will iterate through the registered validation plugins.

A dictionary of `ReportValidationError` objects is returned, with keys determined by the validation plugins.

Error severity (`Severity.WARNING`, `Severity.ERROR`) is also determined by the implementation of the plugins, and can be used on the backend or frontend to throw exceptions or display appropriate warning or error messages.


dleard (approved):
:100: This looks great! A couple questions about follow-up work:
- Does this collide with the work in #3053 (should the in-progress work in that PR become a registered validator that follows this pattern?)
- Now that we have a well defined pattern for creating validators, should we create a ticket or brainstorming space somewhere for things that we could/should validate? (ie: On the final review form, could this allow us to implement a set of errors to display like we did in CIIP). The discussion we had in stand-up around empty activity forms comes to mind here.
- I see there is a cas-reporting [issue (446)](https://github.com/bcgov/cas-reporting/issues/446) for validation exploration. It's kinda old & some things have already been handled, but we could either repurpose it or close it and create another one to capture validator creation

pbastia:
@dleard to answer:
>Does this collide with the work in https://github.com/bcgov/cas-registration/pull/3053 (should the in-progress work in that PR become a registered validator that follows this pattern?)

Yes, and @acatchpole handed his PR over to me before his vacation :D 

> Now that we have a well defined pattern for creating validators, should we create a ticket or brainstorming space somewhere for things that we could/should validate? (ie: On the final review form, could this allow us to implement a set of errors to display like we did in CIIP). The discussion we had in stand-up around empty activity forms comes to mind here.

> I see there is a cas-reporting https://github.com/bcgov/cas-reporting/issues/446 for validation exploration. It's kinda old & some things have already been handled, but we could either repurpose it or close it and create another one to capture validator creation

Definitely. Maybe re-use [#446](https://github.com/bcgov/cas-reporting/issues/446) for that is a great idea, and we can create a shared space with the BA? I've tagged you and patricia in there

PR #3144, opened 2025-04-02
chore:fixed activity form issue

ayeshmcg:
Card: https://github.com/bcgov/cas-reporting/issues/593
There was an issue related to rendering forms in React where duplicate fields were being identified as the same, causing unexpected behavior. This was due to React's internal key handling during re-renders.

To fix this, I added unique key props to the form fields to ensure React correctly distinguishes between them during updates.


shon-button (approved):
Brilliant
🔥
🎉
🤸‍♀️

PR #3148, opened 2025-04-02
Feat/521 activity form builder caching

pbastia:
- Adding a caching strategy for the source type pieces when building activity forms.
- Fixed an N+1 query where we weren't using the prefetched data


dleard (approved):
:heart:

PR #3149, opened 2025-04-03
chore: fixed report history page version number

ayeshmcg:
Card: https://github.com/bcgov/cas-reporting/issues/691
A small PR to fix how the version is generated

To test 
Create multiple report versions using supplementart reports and go to 
http://localhost:3000/reporting/reports/report-history/{report_id}
You should see like 
image

### DIFF
--- a/bc_obps/reporting/service/report_history_dashboard_service.py
+++ b/bc_obps/reporting/service/report_history_dashboard_service.py
@@ -1,19 +1,28 @@
-from django.db.models import QuerySet, Window
-from django.db.models.functions import RowNumber
+from django.db.models import Window, QuerySet, Case, When, Value, CharField
+from django.db.models.functions import RowNumber, Cast, Concat
 from django.db.models import F
 from reporting.models import ReportVersion
 
 
 class ReportingHistoryDashboardService:
     """
-    Service providing data operations for the reporting dashboard
+    Service providing data operations for the report history dashboard
     """
 
     @classmethod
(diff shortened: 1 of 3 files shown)


bc_obps/reporting/service/report_history_dashboard_service.py
```diff
@@ -1,19 +1,28 @@
...
             .annotate(version_number=Window(expression=RowNumber(), order_by=F("id").desc()))
-            .order_by("-id")
+            .annotate(
+                version=Case(
+                    When(version_number=1, then=Value("Current Version")),
+                    default=Concat(Value("Version "), Cast(total_versions - F("version_number") + 1, CharField())),
+                    output_field=CharField(),
+                )
+            )
+            .order_by("version_number")
```
pbastia, line 25: Very very nice! We should be able to just reuse the same base query to make sure we only hit the database once, and annotate with the count (I haven't tested it):
```py
filtered_versions = ReportVersion.objects.filter(report_id=report_id)
report_versions = (
    filtered_versions
    .annotate(total_count=Cast(filtered_versions.count(), IntegerField())
    .annotate(version_number=Window(expression=RowNumber(), order_by=F("id").desc()))
    .annotate(
        version=Case(
                    When(version_number=1, then=Value("Current Version")),
                    default=Concat(Value("Version "), Cast(F('total_count') - F("version_number") + 1, CharField())),
                    output_field=CharField()
               )
           )
          .order_by('version_number')
)
```

pbastia (approved):
Very nice! I have a suggestion to reduce the query count, but haven't tested it if you want to give it a whirl.

Approved regardless

PR #3164, opened 2025-04-04
add compliance invoice

nataliepiguliak:


vivanovbc (approved):
I know that we are still deciding on an approach and this change uses mock data but looks good. I would get another approval from a team but it's a good base.

bc_obps/compliance/services/invoice_service.py
```diff
@@ -0,0 +1,93 @@
...
+
+        Args:
+            compliance_summary_id: ID of the compliance summary (not used in mock version)
+
+        Returns:
+            Tuple of (PDF bytes, filename)
+        """
+        context = InvoiceService._prepare_invoice_context()
+
+        logo_path = Path(__file__).parent.parent / 'static' / 'logo.png'
```
pbastia, line 28: This may not be working in our production environment anymore.
If this is meant to be added to static files, we can use the `settings.STATIC_ROOT` to access it, this is where django will put all the static files it collects when running the `collectstatic` command

bciers/apps/compliance/src/app/actions/downloadInvoice.ts
```diff
@@ -0,0 +1,93 @@
...
+        const contentDisposition = response.headers.get("Content-Disposition");
+        let filename = "invoice.pdf";
+        if (contentDisposition) {
+          const filenameMatch = contentDisposition.match(/filename="(.+)"/i);
+          if (filenameMatch && filenameMatch[1]) {
+            filename = filenameMatch[1];
+          }
+        }
+
+        const base64Data = Buffer.from(pdfData).toString("base64");
```
pbastia, line 71: Just a flag that this has us load the whole PDF file on the server-side of the frontend pod. This shouldn't be an issue, the invoices pdf are pretty small, but would you mind writing a tech debt card about streaming from the backend to the client?

pbastia: Another question: why do we encode to base64? Shouldn't we be able to pass the buffer back to the client directly, so that it doesn't have to be decoded on the other side?

pbastia (approved):
Very nice work! I've approved it, but left a few questions

bc_obps/compliance/api/invoice.py
```diff
@@ -0,0 +1,13 @@
...
+from .router import router
+
+
+@router.get("/invoice/generate/{summary_id}")
+def generate_invoice(request: HttpRequest, summary_id: int) -> HttpResponse:
+
+    pdf_bytes, filename = InvoiceService.generate_invoice_pdf(summary_id)
+
+    response = HttpResponse(pdf_bytes, content_type='application/pdf')
+    response['Content-Disposition'] = f'attachment; filename="{filename}"'
```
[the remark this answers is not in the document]
nataliepiguliak: done

bc_obps/compliance/services/invoice_service.py
nataliepiguliak: fixed for local and prod, thank you!

bciers/apps/compliance/src/app/actions/downloadInvoice.ts
nataliepiguliak: addressed, thank you Pierre!

PR #3165, opened 2025-04-04
Feat: 641 validate boro bcghg ids upon report submission

pbastia:
Work done:
- report validation plugin checking if the `ReportOperation` is missing BORO ID or BCGHGID
- report validation plugin checking if any of the `FacilityReport` is missing BCGHGID


bc_obps/reporting/tests/utils/baker_recipes.py
```diff
@@ -84,7 +84,7 @@ def json_seq(
 facility_report = Recipe(
     FacilityReport,
     report_version=foreign_key(report_version),
-    facility=foreign_key(facility),
+    facility=foreign_key(facility, one_to_one=True),
```
pbastia, line 87: This is to tell the model bakery not to associate multiple facility reports to the same facility when building it with `_quantity`

bc_obps/reporting/service/report_validation/validators/operation_boroid_presence.py
```diff
@@ -0,0 +1,20 @@
+from reporting.models.report_operation import ReportOperation
+from reporting.models.report_version import ReportVersion
+from reporting.service.report_validation.report_validation_error import (
+    ReportValidationError,
+    Severity,
+)
+
+
+def validate(report_version: ReportVersion) -> dict[str, ReportValidationError]:
```
shon-button, line 9: @pbastia 

Wondering if we need to exclude operations with registration purpose = Reporting Operation since the Review Operation Info page won't have a BORO ID, as detailed [here](https://github.com/bcgov/cas-reporting/issues/209)

pbastia: Very good call, just making the fix

ayeshmcg (approved):
Looks great @pbastia .

PR #3169, opened 2025-04-07
fix: source subtypes in other fugitive sources schema for o&g other t…

pbastia:
…han non compression


bc_obps/reporting/migrations/0081_update_fugitive_sources_schema.py
```diff
@@ -0,0 +1,65 @@
...
+        valid_to_id=Configuration.objects.get(valid_to="2099-12-31").id,
+        has_unit=True,
+        has_fuel=False,
+    ).update(json_schema=og_other_fugitive_sources_schema)
+
+
+def revert_other_fugitive_sources_schema(apps, schema_monitor):
+    cwd = os.getcwd()
+    with open(
+        f"{cwd}/reporting/json_schemas/2024/og_extraction_other_than_ncnp/other_fugitive_sources.json"
```
shon-button, line 39: I believe this is loading the "old" file with the incorrect drop-down values:

image


Personally, I would reconsider "added a new *_001.json schema" while keeping the "old" .json files for possible revert, seems like is a YAGNI case, and I would just make changes in the original `reporting/json_schemas/2024/og_extraction_other_than_ncnp/other_fugitive_sources.json` schema

pbastia: Yes, this is the `revert` function of the migration
We need to be able to revert somehow, because this data has already been deployed to prod

shon-button: All good now:

image

shon-button (approved):

PR #3173, opened 2025-04-08
chore: replace cron job with python loop for status checker

joshgamache:
Replaces cronjob-within-cronjobpod with a loop in the Python command.

### DIFF
--- a/bc_obps/registration/management/commands/check_document_file_status.py
+++ b/bc_obps/registration/management/commands/check_document_file_status.py
@@ -3,29 +3,46 @@ from registration.models import Document
 from service.data_access_service.document_service import DocumentDataAccessService
 import time
 
-# Repeat the check every 30 seconds for 59 times, equalling a total of 29:30 minutes with 30 seconds for pod leeway
-REPETITIONS = 59  # Number of times to repeat
-REPEAT_DELAY = 30  # Delay between scans in seconds
-
 
 class Command(BaseCommand):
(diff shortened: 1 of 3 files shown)


bc_obps/registration/management/commands/check_document_file_status.py
```diff
@@ -1,20 +1,36 @@
 from django.core.management.base import BaseCommand
 from registration.models import Document
 from service.data_access_service.document_service import DocumentDataAccessService
+import time
+
+# Repeat the check every 45 seconds for 37 times, equalling a total of 27:45 minutes with 2:75 minutes for pod leeway
+REPETITIONS = 2  # Number of times to repeat
+REPEAT_DELAY = 45  # Delay between repeats in seconds
```
pbastia, line 8: instead of building leeway into our countdown to account for processing time, we can compute the appropriate amount of time to wait after our code runs:

```py
start_time = int(time.time())
iteration = 0

while iteration <= REPETITIONS:
   # compute sleep duration (first iteration happens at current_time)
   next_attempt_time = current_time + iteration * REPEAT_DELAY
   sleep_duration = int(time.time()) - next_attempt_time

   # move to next iteration
   iteration = iteration + 1

   # skip next iteration if we spent too much time on the previous iteration
   if sleep_duration < 0:
      continue
   
   # wait the appropriate amount of time
   time.sleep(sleep_duration)

    ... doStuff()
```

joshgamache: Thanks for that, I integrated this change with a996b25!

dleard (approved):
Awesome! Looks great. We'll make a metabase alert to tell us if anything gets flagged as quarantined as well

PR #3174, opened 2025-04-08
E2E setup and registration workflow test

BCerki:
card: https://github.com/orgs/bcgov/projects/123/views/16?pane=issue&itemId=106101096&issue=bcgov%7Ccas-registration%7C3196
This PR:
- updates env variables and imports (some were changed since we last ran e2e tests)
- puts the registration e2e job back in CI
- e2e test for registration (see the docs for details on approach and what's covered)
- fix inconsistency in opt-in capitalization
- skips the malware scan in local and CI environments
- takes happo shots

I suggest making the e2e mandatory but not the happo one, as we don't know how flaky it's going to be yet and I don't want to block reporting


bciers/apps/registration-e2e/poms/registration.ts
```diff
@@ -0,0 +1,233 @@
...
+
+  async fillLfoFacilityInformation() {
+    await clickButton(this.page, /add facility/i);
+    await this.page.getByLabel(/facility name\*/i).fill("a");
+    await fillComboxboxWidget(this.page, /facility type+/i, "Small Aggregate");
+  }
+
+  async fillOptInInformation() {
+    await this.page
+      .locator("#root_meets_section_3_emissions_requirements-0")
```
BCerki, line 144: I couldn't get a generic checkAllCheckboxes helper to work--playwright didn't seem to pick them up without a more specific locator

fviduya:
Nice work 🥇!  A few changes I'd like to suggest:

**For the submission page**
- Check for the exact match since this page's wording does not really change + I imagine we'd want to be stricter and catch whatever may slip there 

**For Opt-In**
- Do we also wanna check for exact word matches for that?
- And if not, I think I found a solution to checkAllCheckboxes:
```
export async function checkAllRadioButtons(page: Page) {
  const radioButtons = page.getByRole("radio", { name: "Yes" });
  const count = await radioButtons.count();
  
  for (let i = 0; i < count; i++) {
    const radio = radioButtons.nth(i);
    if (await radio.isEnabled()) {
      await radio.check();
    }
  }
}
```

andrea-williams (commented):
Looks great to me! (You're at risk of building a reputation for yourself as a devops expert with all the yaml files you wrote!)

Haven't been able to run the E2E tests locally because poetry in my local is having a breakdown at the moment.

bciers/apps/registration-e2e/poms/registration.ts
fviduya: *_Moving my comment here for better visual_

Think I found a way to get around this! 
```
export async function checkAllRadioButtons(page: Page) {
  const radioButtons = page.getByRole("radio", { name: "Yes" });
  const count = await radioButtons.count();
  
  for (let i = 0; i < count; i++) {
    const radio = radioButtons.nth(i);
    if (await radio.isEnabled()) {
      await radio.check();
    }
  }
}
```

BCerki:
> Looks great to me! (You're at risk of building a reputation for yourself as a devops expert with all the yaml files you wrote!)
> 
> Haven't been able to run the E2E tests locally because poetry in my local is having a breakdown at the moment.

I didn't write the yamls! They were from reg1

BCerki:
> Nice work 🥇! A few changes I'd like to suggest:
> 
> **For the submission page**
> 
> * Check for the exact match since this page's wording does not really change + I imagine we'd want to be stricter and catch whatever may slip there
> 
> **For Opt-In**
> 
> * Do we also wanna check for exact word matches for that?

Happo will catch wording changes--the new screenshots won't match the baseline ones.

andrea-williams:
when I pull the code and run it locally, all my pythontests pass and I'm able to make reset_db just fine too, so I don't understand why those checks are failing.

However, I'm having an issue when I try to run the E2E on my local:

Screenshot 2025-04-15 at 3 05 04 PM

any ideas what I'm doing wrong?

BCerki:
> when I pull the code and run it locally, all my pythontests pass and I'm able to make reset_db just fine too, so I don't understand why those checks are failing.
> 
> However, I'm having an issue when I try to run the E2E on my local:
> 
> Screenshot 2025-04-15 at 3 05 04 PM
> 
> any ideas what I'm doing wrong?

This looks like a problem with the frontend server. Do you see logs about it crashing in the FE terminal when you try to run the test?

fviduya:
Ran it on my end and got no issues, looks good to me! 🌟

fviduya (approved):

PR #3175, opened 2025-04-08
Fix/ignore unregulated products in production data

dleard:
The _initial_ issue that caused the bug in https://github.com/bcgov/cas-reporting/issues/543 is that the unregulated product FOG had no units & our pydantic schema required it. This has been fixed by making units required & setting N/A for products with no units (as in the product dropdown list provided by IRC).

Some further changes were needed here to ensure we are not capturinig production data for unregulated products, AND that the FOG unregulated product could allocate emissions for its special case.

- The production_data API now filters out unregulated products. This means unregulated products will not show up on the production data page.
- The report_product service auto creates a report_product record for FOG if it is in an operation's list of products with 0 for all numeric values. We don't want any report_product data for FOG, but it still needs a report_product record created so that the child ReportProductEmissionAllocation records can be created. This is a special case for FOG, we don't need to allocate emissions to other unregulated products.
- The compliance service ignores unregulated products when iterating through an operation's report_product records. This ensures we ignore the FOG product. It could potentially have side effects since it does not have a PWAEI value associated to it. The allocated emissions are instead used when calculating reporting_only emissions.

This PR also adds some missing products from the dropdown list provided by IRC and the associated PWAEI values.

Some things to check that this is working as intended:
Setup:
- Add the unregulated product FOG to a report & 1 other unregulated product (maybe Oil and gas non-processing, non-compression)
- Add at least 1 regulated product as well

Checks:
- The unregulated products do not show up on the production data page
- A report_product record is not created for the non-FOG product
- A report_product record is auto-created for the FOG product
- The FOG product shows up on the allocation of emissions page
- The other unregulated product does not show up on the allocation of emission page
- None of the unregulated products show up on the compliance summary page
- The reporting_only emissions should include the emissions allocated to FOG


pbastia (approved):
Very nice! Just a couple of comments, but they're not critical at all.

dleard:
Thanks, I'll take a look at resolving those comments in my follow up PR

PR #3185, opened 2025-04-09
fix: disallow 'e' in number fields

gdalcengio:


pbastia (changes requested):
I think this is complex logic that exposes us to edge cases. What if the user doesn't copy paste but drags a chunk of text? what if they enter a capital e?
The documentation warns us about the `type="number"` and encourages us to render a `NumberField` instead:
https://mui.com/material-ui/react-text-field/#type-quot-number-quot

gdalcengio:
Redid in this PR https://github.com/bcgov/cas-registration/pull/3236

PR #3193, opened 2025-04-10
fix: report_product data should be deleted if product removed from report_operation

dleard:
Follow up to #3114. Pretty much the same fix, just for report_products.

To test this work:
- Create a report with a couple regulated_products
- Go through the report and report data on the products
- Go back to the review operation page & remove one of the products from the set
- Confirm that the associated report_product record has been deleted


ayeshmcg (approved):
🥇

PR #3199, opened 2025-04-11
Chore/fix product prod migrations

dleard:
This PR fixes a migration that was edited from behind a release tag, which meant production was not aware of the change.
It undoes the change in the old migration & applies the change in a migration prod can see.
Also in this PR:
- Remove reliance on hardcoded ID to fetch FOG product
- adds unregulated product for line tracing


Sepehr-Sobhani (approved):
🤩

PR #3205, opened 2025-04-11
Contact and user bug

BCerki:
card: https://github.com/orgs/bcgov/projects/123/views/16?pane=issue&itemId=102535049&issue=bcgov%7Ccas-registration%7C3038

This PR:
- there are two services to do with contacts when user_operators are  1) created or the 2) status changes. I think this bug only affects the case of status changes, but since the code was the same in both places, I updated it in both.
- updates the services to reset the contact info to the user's if it's been changed before the contact was approved (I ~think the only way this happens is mock data, or a user is approved, edited, and then unapproved)
- updates tests. I had to get rid of magic mock in one place because it doesn't create a real UUID
- sonar duplication is on tests so I'm not worried about it

### DIFF
--- a/bc_obps/service/tests/test_user_operator_service.py
+++ b/bc_obps/service/tests/test_user_operator_service.py
@@ -1,5 +1,5 @@
 from itertools import cycle
-from unittest.mock import patch, MagicMock
+from unittest.mock import patch
 
 import pytest
 from model_bakery import baker
@@ -135,7 +135,7 @@ class TestUserOperatorService:
         "service.data_access_service.user_operator_service.UserOperatorDataAccessService.get_or_create_user_operator"
     )
     @patch("service.operator_service.OperatorService.update_operator")
-    def test_create_operator_and_user_operator(
+    def test_create_operator_and_user_operator_with_new_contact(
         mock_update_operator, mock_get_or_create_user_operator, mock_save_operator
     ):
         user = baker.make_recipe('registration.tests.utils.industry_operator_user')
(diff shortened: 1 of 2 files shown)


andrea-williams (approved):
Manual Testing steps:

1. logged in with bc-cas-dev to confirm I had access to operator "Bravo"
2. log out as bc-cas-dev, log in as bc-cas-dev-secondary
3. request access to Bravo operator
4. log out as bc-cas-dev-secondary, log in as bc-cas-dev
5. approve bc-cas-dev-secondary's access request --> confirmed that Bravo operator now also has a Contact record for bc-cas-dev-secondary
6. edited the Contact record for bc-cas-dev-secondary (changed their last name)
7. reset bc-cas-dev-secondary's access request status back to "Pending"
8. refreshed the page
9. changed bc-cas-dev-secondary's access request status back to "Approved" - confirmed that the Contact record for bc-cas-dev-secondary kept the changed last name, was not overwritten by the last name stored in the User record

Another manual test case:
1. log in as bc-cas-dev for "Bravo" operator
2. created a new operator contact with the same email address that's associated with my business bceid (cas-dev-andrea), but with different first & last name
3. log out as bc-cas-dev, sign in as cas-dev-andrea
4. request access to Bravo operator as cas-dev-andrea (had to change the bceid_business_name and business_guid in the User table manually, so that it matched the one for bc-cas-dev)
5. log out as cas-dev-andrea, log in as bc-cas-dev
6. approve Andrea's access request - was able to approve it without overwriting the existing Contact record matching the email address used for my business bceid (and without creating a duplicate Contact).

While playing around with this a bit more, I tried to edit an existing contact to have the same email as another contact - I was prevented from doing so, but only got an "Internal Server Error" message in the UI, which is not very helpful to users. However, I think the fix for this is outside the scope of this ticket. Tagging @fviduya so it's on her radar!

Issue #3196, opened 2025-04-11
Reg/e2e/register an operation workflow


BCerki:
#### Describe the task

Write e2e tests for the register an operation workflow

#### Acceptance Criteria

- [x] LFO test
- [x] SFO tests
- [x] any updates necessary since we last ran e2e (env variables, etc.)

#### Definition of Done

- [ ] first
- [ ] second
- [ ] third

#### Additional context

- Add any other context about the task here.
- Or here

BCerki:
Nothing to test in dev. If you'd like to see the automated test results and screenshots, you can go to any PR and look for the e2e and happo registration jobs, e.g.

![Image](https://github.com/user-attachments/assets/a1cf3ac5-2795-4e2a-9020-cbdaf1c49e8e)

![Image](https://github.com/user-attachments/assets/c3f4f141-74a5-48c9-ae6d-89ddbe02d97f)

PR #3213, opened 2025-04-14
3206 e2e select operator

BCerki:
card: https://github.com/bcgov/cas-registration/issues/3196

This PR:
- e2e tests for the three select operator cases: new, request admin access, request non-admin access
- a lot of this^ was pulled over from reg1. Reg1 tested a lot more than ^ because we didn't have vitests at the time, but I've pared it down now (e.g., no longer checking declined and all the confirmation pages)


BCerki:
closing, will finish in https://github.com/bcgov/cas-registration/pull/3259

PR #3229, opened 2025-04-16
Chore/705 unregulated product handling

dleard:
This PR adds handling for all unregulated products in the same way as the FOG product. There are 3 unregulated products in our system. An example of how the handling should work is outlined in the [ticket](https://github.com/bcgov/cas-reporting/issues/705).

To test this:
- Add one or more unregulated products to a report in the operation review page and at least one regulated product
- Add emissions for an activity. The emissions should have a combination of excluded & non-excluded fuels (suggest using Fuel Combustion Mobile activity for this)
- On the product data page, it should not show the unregulated products
- Upon saving on the product data page, report_product records should be created in the db by the backend for the unregulated products with production values of 0
- On the emission allocation page, the unregulated products should show up & allow for allocation of emissions to those products
- Allocate some emissions to all products in all categories (including the excluded categories)
- On the compliance summary page, the reporting-only emissions should be a sum of: Emissions from basic categories allocated to unregulated products + Emissions from excluded categories allocated to regulated_products.
- The reporting-only emissions should not include emissions allocated from excluded categories to unregulated products (double counting)

### DIFF
--- a/bc_obps/reporting/service/report_product_service.py
+++ b/bc_obps/reporting/service/report_product_service.py
@@ -58,18 +58,18 @@ class ReportProductService:
             )
 
         # Add a report_product record for unregulated products reported by this operation (for emission allocation only)
-        for unregulated_product_id in unregulated_product_ids:
-            if unregulated_product_id in ReportOperation.objects.get(
-                report_version_id=report_version_id
-            ).regulated_products.values_list("id", flat=True):
+        for r_product_id in ReportOperation.objects.get(
+            report_version_id=report_version_id
+        ).regulated_products.values_list("id", flat=True):
+            if r_product_id in unregulated_product_ids:
                 ReportProduct.objects.update_or_create(
                     report_version_id=report_version_id,
                     facility_report=facility_report,
-                    product_id=unregulated_product_id,
+                    product_id=r_product_id,
                     defaults={
                         "report_version_id": report_version_id,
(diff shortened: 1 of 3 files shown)


ayeshmcg (changes requested):
sorry added review in this one

bc_obps/reporting/service/report_product_service.py
```diff
@@ -59,24 +57,25 @@ def save_production_data(
...
-                    "production_methodology_description": "auto-generated report_product record for unregulated product",
-                },
-            )
+        # Add a report_product record for unregulated products reported by this operation (for emission allocation only)
+        for unregulated_product_id in unregulated_product_ids:
+            if unregulated_product_id in ReportOperation.objects.get(
+                report_version_id=report_version_id
+            ).regulated_products.values_list("id", flat=True):
+                ReportProduct.objects.update_or_create(
+                    report_version_id=report_version_id,
```
pbastia, line 66: - we might want to query the list of product ids outside of the for loop to avoid multiple database hits
- this means that if an unregulated product is removed from the list on the report operation page, we'll still have a dangling record, we should remove it earlier with the other ones on line 39.

I think the easy thing is to just use the list of reported unregulated products (Line 20) every time, instead of querying the system for all unregulated products and filtering later whether they're to be reported

pbastia (changes requested):

PR #3233, opened 2025-04-16
172 Create Compliance Fixtures

fviduya:
Card: https://github.com/bcgov/cas-compliance/issues/172

### DIFF
--- a/bc_obps/registration/fixtures/mock/address.json
+++ b/bc_obps/registration/fixtures/mock/address.json
@@ -211,7 +211,7 @@
   },
   {
     "model": "registration.address",
-    "pk": 21,
+    "pk": 22,
     "fields": {
       "street_address": "Incomplete address",
       "municipality": "Grove"
--- a/bc_obps/registration/fixtures/mock/operation.json
+++ b/bc_obps/registration/fixtures/mock/operation.json
@@ -10,7 +10,7 @@
       "swrs_facility_id": 1001,
       "status": "Registered",
       "bc_obps_regulated_operation": "24-0014",
-      "activities": [1, 5],
+      "activities": [1],
       "registration_purpose": "Reporting Operation",
       "contacts": [1, 2]
     }
@@ -42,7 +42,7 @@
       "naics_code": 20,
       "regulated_products": [1],
       "status": "Draft",
-      "activities": [1, 5],
+      "activities": [1],
       "registration_purpose": "OBPS Regulated Operation"
     }
   },
@@ -77,7 +77,7 @@
       "regulated_products": [1],
       "status": "Registered",
       "bc_obps_regulated_operation": "23-0002",
-      "activities": [1, 5],
+      "activities": [1],
       "registration_purpose": "OBPS Regulated Operation",
       "contacts": [1]
     }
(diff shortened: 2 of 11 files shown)


bc_obps/reporting/management/commands/load_reporting_fixtures.py
```diff
@@ -14,13 +14,20 @@ def handle(self, *args, **options):
...
+        additional_fixtures = [
+            f'{self.fixture_base_dir}/report_product.json',
+            # f'{self.fixture_base_dir}/report_emission.json',
+            f'{self.fixture_base_dir}/report_emission_allocation.json',
+            f'{self.fixture_base_dir}/report_product_emission_allocation.json',
+        ]
+
+        self.load_fixtures(additional_fixtures)
+
     def load_reports(self):
```
fviduya, line 31: The records for report_version and facility_report gets created here

bc_obps/registration/fixtures/mock/address.json
```diff
@@ -211,7 +211,7 @@
   },
   {
     "model": "registration.address",
-    "pk": 21,
+    "pk": 22,
```
fviduya, line 214: Updated the pk for this as it I noticed it had the same key as the one above it

bc_obps/reporting/management/commands/load_reporting_fixtures.py
Sepehr-Sobhani, line 1: Could we simplify this to something like:
```python
from django.core.management.base import BaseCommand
from django.core.management import call_command


class Command(BaseCommand):
    help = 'Load fixtures for the reporting application'
    FIXTURE_BASE_DIR = 'reporting/fixtures/mock'

    FIXTURES = [
        'reporting_year',
        'report',
        'report_version',
        'facility_report',
        'report_activity',
        'raw_report_activity_data',
        'report_source_type',
        'report_unit',
        'report_fuel',
        'report_emission',
        'report_product',
        'report_emission_allocation',
        'report_product_emission_allocation'
    ]

    def handle(self, *args, **options):
        for fixture in self.FIXTURES:
            fixture_path = f'{self.FIXTURE_BASE_DIR}/{fixture}.json'
            self.stdout.write(self.style.SUCCESS(f"Loading: {fixture_path}"))
            call_command('loaddata', fixture_path)
```

fviduya: `load_reports` in the file creates the data for report_version and facility_report. I tried the approach you provided but it returns an error when I click View Details on the reports though, I'm not sure what I am missing though. I'll keep looking into it but I didn't add this yet to the latest update I pushed!

Sepehr-Sobhani: I see now! we are calling the service `create_report` in this function.
In that case we can still use this approach, but just keep the `load_reports` function as is.

bc_obps/registration/fixtures/mock/operation.json
```diff
@@ -286,14 +286,14 @@
     "fields": {
       "operator": "4242ea9d-b917-4129-93c2-db00b7451051",
       "submission_date": "2024-01-25T15:27:00.000Z",
-      "name": "Bees LFO - Registered",
+      "name": "Registered Bees LFO - Excess Emission",
```
vivanovbc, line 289: 👍

Issue #3228, opened 2025-04-16
Bug: Internal Server Error when trying to Save and Continue operation information.


patrickisaac:
#### Describe the Bug:
Five different operators have reporting Internal Server Error when trying to Save and Continue operation information.

#### Probability:

(How likely the bug is to happen, scored from 1-5. For example, probability of 5 is something like "it happens to all users every time they log in." whereas probability of 1 "only happens to certain users when a really specific and unlikely path is followed.")

#### Impact:

(How bad the bug is when it does happen, scored from 1-5. For example, effect of 5 is "the entire app crashes and makes it unusable for all users" or "the bug causes the wrong data to be saved, with critical information (e.g. payment) being affected." whereas effect of 1 is "It makes some styling look a little bit weird.")

<img src="https://www.getzephyr.com/sites/default/files/riksk%20matrix.jpg" alt="Bug Matrix" width="250" height="250"/>

#### Steps to reproduce the behaviour:

1. Log into BCIERS
2. Confirm Operator Information
3. Attempt to register existing operation
4. Press 'Save and Continue'
5. Observe Internal Server Error message at the bottom of the screen.

#### Screenshots:
![Image](https://github.com/user-attachments/assets/3732676e-773b-4611-ac50-be36943aa073)

#### Additional information:

Browser: [e.g. chrome, safari]
Version: [e.g. 22]

#### Additional Context

-

patriciarussellCAS:
I've marked this one as high priority because it's occurring to a number of users.

PR #3234, opened 2025-04-16
Fix/boro id for draft operations

andrea-williams:
Card #3228 

This PR:
- modifies the existing pgtriggers on `Operation` model regarding "Registered" status and BORO and BCGHG IDs so that if an operation already has a BORO or BCGHG ID when the operation.status is updated to "Registered", a database error won't be triggered (mimics scenario we have in prod, where operations that received a BORO ID last year, or who have a BCGHG ID imported from SWRS, but who are trying to register now)
- modifies the existing pgtrigger on `Facility` model regarding the facility's operation's status, so that if a facility already has a BCGHG ID (i.e., it was imported from SWRS), then the facility data can be updated as part of the registration workflow without raising errors
- added 2 new unit tests in test_operation_service to assert that the modified pgtriggers are behaving as expected


bc_obps/registration/tests/models/test_operation.py
```diff
@@ -252,17 +252,6 @@ def test_change_status_then_assign_id_succeeds(self):
...
-        # Ensure changing status from 'REGISTERED' prevents updates while retaining IDs
-        self.registered_operation.bcghg_id = baker.make_recipe("registration.tests.utils.bcghg_id")
-        self.registered_operation.bc_obps_regulated_operation = baker.make_recipe("registration.tests.utils.boro_id")
-        self.registered_operation.save()
-
-        self.registered_operation.status = Operation.Statuses.DRAFT
-        with self.assertRaises(ProgrammingError):
-            self.registered_operation.save()
-        self.assertIsNotNone(self.registered_operation.bcghg_id)
-
```
andrea-williams, line 265: deleted this test because it was failing and as far as we're aware at the moment, there won't ever be a scenario where an operation's status reverts from "Registered" back to "Draft"

pbastia (approved):

PR #3235, opened 2025-04-17
chore: fix fetching facility report with facility id and report version

ayeshmcg:
Card: https://github.com/bcgov/cas-reporting/issues/711
When a supplementary report is created, an additional entry is added to the FacilityReport table for the same facility_id, but with a different report_version_id.
Previously, we were using .get(facility_id=...), which caused a MultipleObjectsReturned error due to the presence of multiple entries for the same facility.
This PR updates the query to include report_version_id as a filter to ensure that a single, specific FacilityReport is retrieved.


pbastia (approved):
🥇

PR #3236, opened 2025-04-17
fix: disallow 'e' in number fields by separating text and number fields

gdalcengio:


pbastia (changes requested):
Very nice!! The requested changes is for the max value, which could be hit pretty easily for Gt of emissions or kW/h values. The rest is nitpicking

gdalcengio:
> Very nice!! The requested changes is for the max value, which could be hit pretty easily for Gt of emissions or kW/h values. The rest is nitpicking

@pbastia 

Changes have been made and tests related have been fixed accordingly

pbastia (approved):
🎉

PR #3238, opened 2025-04-17
Chore: 483 EIO Reporting Workflow

shon-button:
Addresses: 
[483](https://github.com/bcgov/cas-reporting/issues/483)


ayeshmcg:
Changes:
1. Added Schema and UISchema for EIO
2. Create Electricity Import Data page
3. Created eio flow
4. updated headerElementFactory 
5. Added model, api, service and schema for Electricitty Import Data
6. Updated Sign Off model : added acknowledgement_of_errors
7. Updated report submit signal to skip create compliance if not regulated
8. Added tests for frontend and backend


To test:-
1. Register an Operation(with registration purpose= Electricity Import Operation): http://localhost:3000/registration/register-an-operation
image
 
   
2 Go to http://localhost:3000/reporting/reports and start EIO report
image
You should see only three headers: Operation Information, Report Information and Sign-off & submit
3. Go to Report Information 
image

You should be able to see Electricity Import Data page
4. Go to /sign-off
      - **If Supplementary Report:** 
    
image

     - **Not supplementary**
     
image

bciers/apps/reporting/src/app/components/taskList/headerElementFactory.ts
```diff
@@ -16,7 +16,9 @@ export const headerElementFactories: {
   }),
   [HeaderStep.ReportInformation]: (context) => ({
     type: "Section",
-    title: `${context?.facilityName ?? "Facility"} information`,
+    title: context?.isElectricityImport
+      ? "Report Information"
+      : `${context?.facilityName ?? "Facility"} information`,
```
pbastia, line 21: Should we just create another `HeaderStep` entry for EIOs instead of adding conditionals to conditionals? It would simplify everything

bciers/apps/reporting/src/app/utils/getReportNeedsVerification.ts
```diff
@@ -1,6 +1,15 @@
...
+import { OperationTypes } from "@bciers/utils/src/enums";
 
 export async function getReportNeedsVerification(reportVersionId: number) {
+  // 🚀 Fetch the operation associated with the specified version ID
+  const reportOperation = await getReportingOperation(reportVersionId);
+  // Check for EIO bypass
+  const operationType = reportOperation?.operation_type;
+  if (operationType === OperationTypes.EIO) {
+    return true;
+  }
```
pbastia, line 12: This should happen on the server, the `ReportVerificationService.get_report_needs_verification(...)` method called through the API below is there for exactly that case

bciers/apps/reporting/src/middlewares/withRuleHasReportRouteAccess.ts
```diff
@@ -220,11 +221,7 @@ export const permissionRules: PermissionRule[] = [
       return Boolean(reportVersionId && pathname.match(pathRegex));
     },
     validate: async (reportVersionId, token) => {
-      // Directly call fetchResponse here since it may not be reused elsewhere.
-      const needsVerification = await fetchResponse(
-        `${REPORT_VERSION_API_BASE}${reportVersionId}${ApiEndpoints.NEEDS_VERIFICATION}`,
-        token.user_guid,
-      );
+      const needsVerification = getReportNeedsVerification(reportVersionId);
```
pbastia, line 224: I think it was by design that we weren't using the `actionHandler` to fetch that, since it's in the middleware and the actionhandler is a pretty heavy piece of machinery. @shon-button  might be able to confirm?
Were you running into issues with the previous implementation?

bciers/apps/reporting/src/data/jsonSchema/signOff/signOff.ts
```diff
@@ -4,54 +4,107 @@ import {
...
+  const properties: RJSFSchema["properties"] = {
+    ...baseFields(flow),
+  };
+
+  if (isSupplementary) {
+    properties.supplementary = {
+      type: "object",
+      properties: getSupplementaryFields(isRegulated),
+    };
+  } else if (flow !== ReportingFlow.EIO) {
```
pbastia, line 93: This is a lot of if/then/else that don't follow the same patterns, nor are tested.
Mixing conditions about supplementary, flows etc. is a recipe for multiplying cases having to handle and test a lot of possible options, or introducing bugs when we forget about certain cases.

We can talk about it if you want, we propably want to expose mutliple json schemas/pages through our factories instead

bciers/apps/reporting/src/tests/components/eio/ElectricityInformationForm.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+      <ElectricityInformationForm
+        versionId={mockVersionId}
+        initialFormData={{}}
+        navigationInformation={dummyNavigationInformation}
+      />,
+    );
+
+    expect(
+      await screen.findByText("Electricity Import Data"),
+    ).toBeInTheDocument();
```
pbastia, line 43: Maybe an assertion counting the fields on the page

bciers/apps/reporting/src/tests/components/signOff/SignOffForm.test.tsx
```diff
@@ -24,65 +64,54 @@ describe("SignOffForm Component", () => {
...
+    const texts = [
+      "I certify that I have reviewed the annual report, and that I have exercised due diligence to ensure that the information included in this report is true and complete.",
+      "I understand that the Ministry responsible for the administration and enforcement of the Greenhouse Gas Industrial Reporting and Control Act may require records from the Operator evidencing the truth of this report.",
+      "I understand that this information is being collected for the purpose of emission reporting under the Greenhouse Gas Industrial Reporting and Control Act and may be disclosed to the Ministry responsible for the administration and enforcement of the Carbon Tax Act.",
+      "I understand that the information provided in this report will impact the compliance obligation of this operation and that any errors, omissions, or misstatements can lead to an additional compliance obligation or administrative penalties.",
+    ];
+
+    texts.forEach((text) => {
+      expect(screen.getByText(text)).toBeInTheDocument();
+    });
```
pbastia, line 80: This test is now tautological - we mock something and test the thing we mocked. 

It's now not testing the actual schema contents anymore. The schema generation method introduced in this PR needs to be tested at the very least, or refactored to make use of multiple schemas (see previous comment), so we can trust that it gets rendered on screen

pbastia (changes requested):
Very nice work, this is a lot!
For next time, we should really split the work between PRs where it makes sense, especially if the features are disconnected like it's the case here:
- EIO flow
- Electricity Import Data Page
- Changes to sign-off page
- Update to signal to only take in account regulated operation

This will let us review and merge more efficiently, and reduce the risk of introducing bugs

bciers/apps/reporting/src/middlewares/withRuleHasReportRouteAccess.ts
shon-button: @pbastia 
@ayeshmcg 

PB, Good eyes and good recall...the background was that we had had `actionHandler` issues in the past which I believe was based on user guid token issues.  So, as a resolution we fetched directly in the middleware.  

In this case, for the change in `needsVerification` fn I added the server action to the middleware as a "just to see" (because I didn't want to have to change in 2 places), and it worked- cool!  
Perhaps a change in the auth.js libraries or some other magic bugaboo has happened since the original middleware actionHanlder issues.

In short, I think if a sever action functions as expected in the middleware then we should use the shared fn(s).  

I believe the other "mimicked" functions in the middleware could be replaced with the shared fns - tech debt?

ayeshmcg: @pbastia , @shon-button 
Yes for changes related to needsVerification we can create a techdebt

bc_obps/reporting/schema/electricity_import_data.py
```diff
@@ -0,0 +1,47 @@
+from ninja import ModelSchema
+from pydantic import alias_generators
+from reporting.models import ElectricityImportData
+
+
+def to_snake(string: str) -> str:
+    return alias_generators.to_snake(string)
```
pbastia, line 7: this code is not useful, it just passes on the call. We can just do `from pydantic.alias_generators import to_snake` and use that in the schema below

bc_obps/reporting/schema/electricity_import_data.py
```diff
@@ -0,0 +1,47 @@
...
+            'import_specified_emissions',
+            'import_unspecified_electricity',
+            'import_unspecified_emissions',
+            'export_specified_electricity',
+            'export_specified_emissions',
+            'export_unspecified_electricity',
+            'export_unspecified_emissions',
+            'canadian_entitlement_electricity',
+            'canadian_entitlement_emissions',
+        ]
```
pbastia, line 47: Those two schemas are identical, we should make them the same class

bc_obps/reporting/models/electricity_import_data.py
```diff
@@ -0,0 +1,95 @@
...
+        null=True,
+        decimal_places=4,
+        max_digits=20,
+        db_comment="Emissions from Canadian Entitlement Power",
+    )
+
+    class Meta(TimeStampedModel.Meta):
+        db_table = 'erc"."electricity_import_data'
+        app_label = "reporting"
+        db_table_comment = "Table storing Electricity Import Data for the reporting system"
```
pbastia, line 89: We're missing a unique constraint: There can't be more than one ElectricityImportData model per ReportVersion object

bc_obps/reporting/tests/api/test_electricity_import_endpoint.py
```diff
@@ -0,0 +1,92 @@
...
+            import_specified_electricity='0.05',
+            import_specified_emissions='0.05',
+            import_unspecified_electricity='0.05',
+            import_unspecified_emissions='0.05',
+            export_specified_electricity='0.05',
+            export_specified_emissions='0.05',
+            export_unspecified_electricity='0.05',
+            export_unspecified_emissions='0.05',
+            canadian_entitlement_electricity='0.05',
+            canadian_entitlement_emissions='0.05',
```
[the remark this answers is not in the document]
pbastia: I was just suggesting setting some of these values to None, int, 0 or large numbers, just to cover a few more things

bc_obps/reporting/tests/service/test_electricity_import_data_service.py
```diff
@@ -0,0 +1,97 @@
...
+    def test_save_with_integers(self):
+        """Should save correctly with integer inputs."""
+        self._save_and_assert(self.INTEGER_PAYLOAD)
+
+    def test_save_with_zeros(self):
+        """Should save correctly when all fields are zero."""
+        self._save_and_assert(self.ZERO_PAYLOAD)
+
+    def test_save_with_none_for_optional_fields(self):
+        self._save_and_assert(self.NONE_PAYLOAD)
```
pbastia, line 72: This is a really good case for `@pytest.parametrize` , the docs are well written https://docs.pytest.org/en/stable/example/parametrize.html, and we also have some examples like in `test_changing_registration_purpose.py`

bciers/apps/reporting/src/app/components/attachments/AttachmentsPage.tsx
```diff
@@ -32,7 +36,7 @@ const AttachmentsPage: React.FC<HasReportVersion> = async ({ version_id }) => {
       version_id={version_id}
       initialUploadedAttachments={uploadedAttachmentsDict}
       navigationInformation={navInfo}
-      isVerificationStatementMandatory={needsVerification}
+      isVerificationStatementMandatory={isEIO ? false : needsVerification}
```
pbastia, line 39: We should add that logic in the backend in the `getReportNeedsVerification(version_id)` call, not here
Otherwise:
- we have an extra place (here) where we implement additional logic that needs to be tested
- the getReportNeedsVerification method returns something inaccurate and doesn't take in account EIOs, for other consumers

bciers/apps/reporting/src/app/components/verification/VerificationForm.tsx
```diff
@@ -50,6 +52,7 @@ export default function VerificationForm({
 
   // 🛠️ Function to handle form submit
   const handleSubmit = async () => {
+    if (isEIO) return true;
```
pbastia, line 55: This prevents the form from being submitted if that flag is set, I don't think it's the behaviour we want

bciers/apps/reporting/src/app/components/verification/VerificationPage.tsx
```diff
@@ -56,6 +60,7 @@ export default async function VerificationPage({
         isSupplementaryReport={
           isSupplementaryReport.is_supplementary_report_version
         }
+        isEIO={flow === ReportingFlow.EIO}
```
pbastia, line 63:
```suggestion
        isEIO={isEIO}
```

pbastia (changes requested):
🎉 
Just a couple of things left, including a unique constraint that I missed the last time around, and a few nits.
Great work!

bciers/apps/reporting/src/app/components/verification/VerificationForm.tsx
ayeshmcg: Regarding this, we want this page to be optional for EIO workflow, so do you think we should make the fields to have null value (model) ?

ayeshmcg:
@pbastia  created card https://github.com/bcgov/cas-reporting/issues/743
to handle additional work required for verification and attachment page

bc_obps/reporting/models/electricity_import_data.py
```diff
@@ -0,0 +1,101 @@
+from django.db import models
+from registration.models.time_stamped_model import TimeStampedModel
+from reporting.models.report_version import ReportVersion
+from reporting.models.triggers import immutable_report_version_trigger
+from reporting.models.rls_configs.electricity_import_data import Rls as ElectricityImportDataRls
+
+
+class ElectricityImportData(TimeStampedModel):
```
pbastia, line 8: This breaks the convention of naming all our report data `ReportSomethingSomething`, I think we should name it `ReportElectricityImport`

bc_obps/reporting/schema/electricity_import_data.py
```diff
@@ -0,0 +1,25 @@
+from ninja import ModelSchema
+from pydantic import alias_generators
+from reporting.models import ElectricityImportData
+
+
+def to_snake(string: str) -> str:
+    return alias_generators.to_snake(string)
```
pbastia, line 7: This code is still not needed, just import it directly:

```py
from pydantic.alias_generators import to_snake

class ElectricityImportDataSchema(ModelSchema):
  class Meta
    alias_generator = to_snake

```

pbastia (changes requested):
Just a couple of changes left! This will be good to go after

PR #3249, opened 2025-04-17
fix: error loading existing data into emission allocation form

pbastia:


ayeshmcg (approved):
Awesome, It works perfectly

ayeshmcg (commented):
@pbastia 
It complaining about circular import
`ImportError: cannot import name 'ReportEmissionAllocationsSchemaIn' from partially initialized module 'reporting.schema.report_emission_allocation' (most likely due to a circular import) (/Users/ayeshaayesha/Ayesha/cas-registration/bc_obps/reporting/schema/report_emission_allocation.py)`

PR #3250, opened 2025-04-17
Fix/572 cas number non responsive

dleard:
Fixes the issue reported in https://github.com/bcgov/cas-reporting/issues/572 where the cas-number was not being updated when a gasType selection was changed.

This PR moves the responsibility for displaying that CAS Number field from the form_builder_service (which wasn't working) into a custom field template for the gasType field. The gasType records are fetched by activityInit.tsx & passed down through activityForm.tsx via the formContext. The selected gasType is matched against those records & the CAS Number field is populated there.

To test:
- This change means **all** emission forms will now show a cas-number (previously it was only shown for a specific source type)
- Create a report with the Electricity Generation activity
- Select the Cooling Units source type
- When changing the gasType selection, the cas_number field below should now auto-update

**Gas Type Selection:**
Screenshot from 2025-04-17 16-06-41

**On Selecting a new gas type from the dropdown:**
Screenshot from 2025-04-17 16-06-58


pbastia (approved):
:shipit:

PR #3258, opened 2025-04-22
chore: add compliance summary models to persist compliance data

dleard:
This PR adds a couple models, a new service to the compliance service & modifies the report submission service.
On submission of a report_version, the compliance data is now saved to the new models.
To test this manually:
- Create a report & follow it through all the way to submission
- Confirm that the compliance data has been saved to the new models (report_compliance_summary & report_compliance_summary_product)


bc_obps/reporting/models/report_compliance_summary.py
```diff
@@ -0,0 +1,73 @@
...
+    reduction_factor = models.DecimalField(
+        db_comment="The Province developed distinct reduction factors for products in the B.C. OBPS with disproportionately higher industrial process emissions than those produced in other sectors. https://www2.gov.bc.ca/assets/gov/environment/climate-change/action/carbon-tax/obps-technical-backgrounder.pdf",
+        decimal_places=4,
+        max_digits=10,
+    )
+    tightening_rate = models.DecimalField(
+        db_comment="Tightening rates are planned, yearly, gradual increases to BC OBPS stringency. https://www2.gov.bc.ca/assets/gov/environment/climate-change/action/carbon-tax/obps-technical-backgrounder.pdf",
+        decimal_places=4,
+        max_digits=10,
+    )
```
pbastia, line 56: Any reason why `max_digits` is different for those 2?

bc_obps/reporting/service/compliance_service.py
```diff
@@ -5,9 +5,12 @@
...
 from reporting.service.emission_category_service import EmissionCategoryService
+from reporting.models import ReportComplianceSummary, ReportComplianceSummaryProduct
+from registration.models import RegulatedProduct
 from decimal import Decimal
 from django.db.models import Sum
 from typing import Dict, List
+from django.db import transaction
 
 
 class RegulatoryValues:
```
pbastia, line 16: Those classes can be `@dataclass`, it'll just be less code to write

bc_obps/reporting/service/compliance_service.py
```diff
@@ -288,3 +294,45 @@ def get_calculated_compliance_data(cls, report_version_id: int) -> ComplianceDat
...
+            reporting_only_emissions=compliance_data_to_save.reporting_only_emissions,
+            emissions_attributable_for_compliance=compliance_data_to_save.emissions_attributable_for_compliance,
+            emissions_limit=compliance_data_to_save.emissions_limit,
+            excess_emissions=compliance_data_to_save.excess_emissions,
+            credited_emissions=compliance_data_to_save.credited_emissions,
+            reduction_factor=compliance_data_to_save.regulatory_values.reduction_factor,
+            tightening_rate=compliance_data_to_save.regulatory_values.tightening_rate,
+            initial_compliance_period=compliance_data_to_save.regulatory_values.initial_compliance_period,
+            compliance_period=compliance_data_to_save.regulatory_values.compliance_period,
+        )
```
pbastia: `update_or_create` works a bit differently and will try to match every named field we pass, to apply `defaults` or `create_defaults` to the matching records, depending on the scenario: https://docs.djangoproject.com/en/5.2/ref/models/querysets/#update-or-create

This should be (with the dataclass usage it becomes very easy):
```py
compliance_summary_record, _ = ReportComplianceSummary.objects.update_or_create(
    id=compliance_summary_record_id,
    report_version=report_version_record,
    defaults=dataclasses.asdict(compliance_data_to_save)
)
```

bc_obps/reporting/service/compliance_service.py
```diff
@@ -288,3 +294,45 @@ def get_calculated_compliance_data(cls, report_version_id: int) -> ComplianceDat
...
+                id=product_data_id,
+                report_version=report_version_record,
+                report_compliance_summary=compliance_summary_record,
+                product=RegulatedProduct.objects.get(id=product_data_to_save.product_id),
+                annual_production=product_data_to_save.annual_production,
+                apr_dec_production=product_data_to_save.apr_dec_production,
+                emission_intensity=product_data_to_save.emission_intensity,
+                allocated_industrial_process_emissions=product_data_to_save.allocated_industrial_process_emissions,
+                allocated_compliance_emissions=product_data_to_save.allocated_compliance_emissions,
+            )
```
pbastia: same update_or_create comment

bc_obps/reporting/service/report_submission_service.py
```diff
@@ -37,12 +40,17 @@ def submit_report(version_id: int, user_guid: UUID, sign_off_data: ReportSignOff
...
+            ComplianceService.save_compliance_data(version_id)
+
         # Set the new version as lastest submitted
         report_version.is_latest_submitted = True
         report_version.status = ReportVersion.ReportVersionStatus.Submitted
         report_version.save()
 
-        # Send a signal that the report has been submitted
-        report_submitted.send(sender=ReportSubmissionService, version_id=version_id, user_guid=user_guid)
+        if is_regulated_operation:
```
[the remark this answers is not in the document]
pbastia: I would argue against restricting this here - the contract of the signal is just that a report was submitted, and this makes assumptions about who is listening to this signal or how they're using it

@ayeshmcg [has a PR](https://github.com/bcgov/cas-registration/pull/3238/files#diff-5776b99d151575a9964455017dd976ee8006b82470b896770b7eba6e37ae00ba) that implements this check in the signal receiver, which is in my opinion a better place to check
An alternative would be to make a specific "regulated_operation_report_submitted" signal

bc_obps/reporting/tests/service/test_compliance_service/test_compliance_service_class.py
```diff
@@ -63,3 +64,21 @@ def test_new_entrant(self):
...
+        report_compliance_product_records = ReportComplianceSummaryProduct.objects.filter(
+            report_compliance_summary_id=report_compliance_summary_record.id
+        )
+
+        assert (
+            report_compliance_summary_record.emissions_attributable_for_compliance
+            == result.emissions_attributable_for_compliance
+        )
+        for idx, p in enumerate(report_compliance_product_records):
+            assert p.annual_production == result.products[idx].annual_production
```
pbastia, line 84: This is only testing the `create` case, which is why it worked with the `update_and_create` bug above, we might want to also test the update case

bc_obps/reporting/tests/utils/immutable_report_version.py
```diff
@@ -28,7 +30,10 @@ def assert_immutable_report_version(
         **{path_to_report_version: report_version},
     )
 
-    setattr(model_under_test, str_field_to_update, "{'test': 'allow change'}")
+    if decimal_value_to_update != Decimal('0'):
+        setattr(model_under_test, str_field_to_update, Decimal('1.0'))
```
pbastia, line 34:
```suggestion
        setattr(model_under_test, str_field_to_update, decimal_value_to_update)
```

pbastia (changes requested):
Very nice! Just a bug with the `update_or_create` usage that will be an easy fix

bc_obps/reporting/models/report_compliance_summary.py
dleard: They're smaller values. It could honestly be even less max digits I think

pbastia (approved):
Looks great! Once the ignore trigger is added to the migration it'll be good to go

PR #3259, opened 2025-04-23
3206 2 e2e select operator

BCerki:
card: https://github.com/bcgov/cas-registration/issues/3196

This PR:

- e2e tests for the three select operator cases: new, request admin access, request non-admin access
- a lot of this^ was pulled over from reg1. Reg1 tested a lot more than ^ because we didn't have vitests at the time, but I've pared it down now (e.g., no longer checking declined and all the confirmation pages)
- the job skips the e2e test if no files in the `administration` directory are affected, so I added a temporary commit with a comment that I'll delete before merging. I looked into running the tests if only the corresponding -e2e project is affected, but it looked somewhat complicated and it's going to be rare that we only touch an e2e repo and not any code


bciers/apps/administration/app/data/jsonSchema/operationInformation/administrationOperationInformation.ts
```diff
@@ -16,6 +16,7 @@ import { Apps, OperationStatus } from "@bciers/utils/src/enums";
 
 import { optedInOperationDetailsUiSchema } from "./optedInOperation";
 import { RegistrationPurposes } from "@/registration/app/components/operations/registration/enums";
+// a comment to trigger e2e
```
andrea-williams, line 19: leftover

andrea-williams (approved):

PR #3261, opened 2025-04-23
3178 missing validation

BCerki:
card:https://github.com/orgs/bcgov/projects/123/views/16?pane=issue&itemId=105741444&issue=bcgov%7Ccas-registration%7C3178

Notes:
- In a `oneOf` schema, it appears when the docs say exactly one schema must match, that means required fields too. So we will always get a `must be const` error on registration purpose if activities or products are missing. To deal with this, I moved to if/then. This makes activities and products appear on the page on initial load. Zoey is okay with this.
- There's some interaction between the form (state-related?) and the MultiStepWidget that's producing weird behaviour. Only the last widget on the page that uses the widget gets a value of []. The other keys don't even show in the form data. E.g. {activities: []}. This means that activities correctly gets the minItems error, but the other fields get no error (or required, if we turn it on). e.g. image. I tried to untangle this and it has something to do with the `onChange` in MultiSelectWidget, but I didn't  figure it out and this is low probability bug, so I fixed it with customTransformErrors instead.

### DIFF
--- a/bciers/apps/registration/app/data/jsonSchema/operationInformation/registrationPurpose.ts
+++ b/bciers/apps/registration/app/data/jsonSchema/operationInformation/registrationPurpose.ts
@@ -85,62 +85,65 @@ export const createRegistrationPurposeSchema = async () => {
       },
     },
 
-    dependencies: {
-      registration_purpose: {
-        oneOf: registrationPurposes.map((purpose: RegistrationPurposes) => {
-          const isRegulatedProducts =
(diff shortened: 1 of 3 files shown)


andrea-williams (approved):
Manually tested in my local. Tried registering an existing operation as OBPS Regulated without activities or products, and got the expected error message. Tried again by creating a new operation instead and still got the expected error message.

Nice work Bri!

PR #3264, opened 2025-04-25
Add sentry to all frontend apps

Sepehr-Sobhani:
**NOTES:**
- We’re seeing a warning about not setting up a `global-error.js` file. However, each app already has its own `error.ts` file, and according to the `Next.js` documentation, having separate error files per app is the preferred approach.
We can suppress this warning by setting `SENTRY_SUPPRESS_GLOBAL_ERROR_HANDLER_FILE_WARNING` in the `.env` file, but doing so requires upgrading the Sentry package. That upgrade introduces breaking changes with the current Next.js version we’re using (v14).


bciers/apps/administration/next.config.js
```diff
@@ -1,15 +1,13 @@
...
 // Next.js doesn't use TS's paths, so we need to use the relative path
-const nextConfigBase = require("../../next.config.base");
+const { nextConfigBase, withSentry } = require("../../next.config.base");
 
 /**
  * @type {import('@nx/next/plugins/with-nx').WithNxOptions}
  **/
 const nextConfig = {
-  ...nextConfigBase,
+  ...nextConfigBase, // Spread shared base config
```
pbastia, line 10: extra comment with no real purpose, did copilot add this while we weren't looking? :D

bciers/apps/dashboard/next.config.js
```diff
@@ -8,13 +8,13 @@ const {
...
 // Next.js doesn't use TS's paths, so we need to use the relative path
-const nextConfigBase = require("../../next.config.base");
+const { nextConfigBase, withSentry } = require("../../next.config.base");
 
 /**
  * @type {import('@nx/next/plugins/with-nx').WithNxOptions}
  **/
 const nextConfig = {
-  ...nextConfigBase,
+  ...nextConfigBase, // Spread shared base config
```
pbastia, line 17: same

pbastia (approved):
Looks awesome! :shipit:

bciers/apps/administration/next.config.js
Sepehr-Sobhani: It did! I really need to push the tab button less :D

PR #3277, opened 2025-04-29
added warning note for certain methodologies

acatchpole:
Card: https://github.com/bcgov/cas-reporting/issues/571

### Changes
 - Created MethodologyFieldTemplate that sets the "description" value to the desired warning note if the selected methodology should display a warning : ```* By selecting {methodology}, you may be contacted by Ministry staff to
        verify compliance``` if the ```methodology``` is "Alternative Parameter Measurement Methodology" or
    "Replacement Methodology"
 - Replaced InlineFieldTemplate with MethodologyFieldTemplate in all activity uischemas
 
### The Why
```description``` is a field that is already checked by the InlineFieldTemplate for displaying notes on a field. That is why all the MethodologyFieldTemplate does is set that ```description``` value.

I tried some other methods around trying to inject the ```description``` property into the formSchema. I was unsuccessful in trying to mutate the formSchema directly in the front end. And in the backend, we could add the ```description``` field to the methodology object in the oneOf dependency array, but then it wouldn't render for some reason. And if we added it to the parent ```methodology``` object, it *would* render, but unconditionally for every selected methodology.

### To Test

 1. With the front-end and back-end running, progress through the Report Submission process to any activity form that includes emissions.
 2. Select a gas type to show the methodology field.
 3. Select either Alternative Parameter Measurement Methodology" or "Replacement Methodology" to show the warning message displayed below the field.
 4. Select any other methodology to see the warning note disappear.


gdalcengio (approved):
works perfectly, no notes

Issue #2641, opened 2024-12-27
E2E Test CI Flag Logic


Sepehr-Sobhani:
#### Describe the task

Currently, we use the `if: ${{ inputs.is_nx_affected }}` flag in our e2e tests to bypass test runs when no changes are detected in a particular frontend app. However, this approach does not account for changes made in the backend code, particularly in the e2e test setup and fixtures. As a result, when backend-related changes occur, the flag fails to recognize that e2e tests might still be impacted, leading to missed issues that are only noticed in the next PR. This behavior introduces the risk of undetected problems.

Sepehr-Sobhani:
cc @pbastia / @dleard for visibility.

I'm not sure if this might become an issue again, so it could be worth considering proactive measures.

BCerki:
It did become an issue for https://github.com/bcgov/cas-registration/pull/3259. I ended up doing a temporary commit with a comment to trigger the condition

PR #3291, opened 2025-05-02
add/edit contact

fviduya:
These scripts:
1. Tests functionality of manually adding a contact
2. Tests functionality of editing a contact
3. Tests the functionality of adding a contact through the registration flow
4. Adds a test setup file to be used globally 

The test for adding a contact through user access approval is missing here, will be adding that when we create the test for user access and requests.

~*I also couldn't seem to get sonarcloud to not tag the block as duplicate?~


bciers/apps/administration-e2e/tests/workflows/contact.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+    const viewLink = await viewDetails.getAttribute("href");
+
+    await takeStabilizedScreenshot(happoPlaywright, page, {
+      component: "Contacts grid",
+      variant: "filled",
+    });
+    // TODO: investigate accessibility failing
+    // await analyzeAccessibility(page);
+    // Will use goto since navigation to this link is done manually in the frontend
+    await page.goto(viewLink);
```
fviduya, line 90: Decided to use the link instead, when I try to click the View Details button, it doesn't seem to be redirecting to the page.

bciers/libs/e2e/src/setupBeforeAll.ts
```diff
@@ -0,0 +1,25 @@
+import { test } from "@playwright/test";
+import { AppName } from "@/administration-e2e/utils/constants";
+import { setupTestEnvironment } from "@bciers/e2e/utils/helpers";
+import { getStorageStateForRole } from "@bciers/e2e/utils/helpers";
+
+const happoPlaywright = require("happo-playwright");
+
+export function setupBeforeAllTest(role: string) {
```
BCerki, line 8: Thanks for refactoring this!

bciers/apps/administration-e2e/tests/workflows/contact.spec.ts
BCerki: Odd!

BCerki (commented):
Because of [this issue](https://github.com/bcgov/cas-registration/issues/2641), it looks like the e2e tests and happo aren't running in CI on this PR. My workaround has been to make a temporary commit that affects a FE file, and then delete that commit just before merge.
Things look great, but I'd like to see the happo shots before approving. Thanks for your work on this!

bciers/apps/administration-e2e/poms/contacts.ts
```diff
@@ -0,0 +1,154 @@
...
+    this.fieldMunicipality = page.getByLabel(ContactFormField.MUNICIPALITY);
+
+    this.fieldPostalCode = page.getByLabel(ContactFormField.POSTAL_CODE);
+  }
+
+  // ###  Actions ###
+
+  async route() {
+    await this.page.goto(this.url);
+    await this.page.waitForTimeout(500);
```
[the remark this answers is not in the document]
fviduya: This was an odd behavior as well? The main reason why I put timeout here was because when I click `Add Contact` it doesn't seem to be routing to the `add-contact` page without this or without using networkidle. 

These are what I have tried so far:
1. waitForLoadState('networkidle') works, I think it essentially does the same thing and more - which is to wait for at least 500ms until there are no more network requests. ESLint doesn't like this and flags it as an error, I think because we use playwright/recommended.
2. I also tried Promise.all specifically click the Add button then wait for URL
3. I tried to assert visibility of elements: Add Contact button, grid visibility using stabilizeGrid but still runs into an error right after clicking `Add Contact`

The other way I was thinking of was to go to the URL directly, but that would not mimic real user journey. But here is how I structured it before so it doesn't just wait:

```
  async clickAddButton(){
    try {
      await this.page.waitForLoadState('networkidle', { timeout: 10000 });
      await this.page.getByRole('button', {name: /add contact/i}).click();
    } catch (error) {
      console.warn('UI Navigation failed. Falling back to direct URL');
      await this.page.goto(`${this.url}/add-contact`);
    }
  }
```

If clicking the button fails, then just go directly to the URL. I can get past eslint by allowing networkidle, only if team allows it 😄

bciers/apps/registration-e2e/tests/register-an-operation.spec.ts
```diff
@@ -125,7 +106,18 @@ test.describe("Test register operations", () => {
       variant: "default",
     });
     await analyzeAccessibility(page);
+
+    // Check operation representative in contacts grid
```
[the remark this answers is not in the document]
fviduya: Thanks! I'll remove the check here then

fviduya:
The rebase was messy, I created another PR for this. Closing this one now.

Issue #3296, opened 2025-05-05
Add reporting-only NAICS codes


BCerki:
#### Describe the task
List of NAICS codes to add:
https://bcgov.sharepoint.com/:x:/r/teams/00608-CPRInternal/Shared%20Documents/Carbon%20Pricing%20Transition%20SMEs/Registration%20App/Registration%20Pt.%202%20Fields.xlsx?d=w8cf871009f8d43c19d1d79c69fa259e8&csf=1&web=1&e=rvmUoQ&nav=MTJfQjQ0X3tGQjgwQTNGNi1BMjYwLTQ1MDgtOEJBRC05NzIzREVCNzcxMzJ9

Note from Meeting with Kirk May 6: All Reporting Operation NAICS Codes (i.e. not regulated) will have the standard reduction factor (0.65) and tightening rate (0.01). 
This will ensure that if a Regulated Operation selected one of these NAICS codes in error, it will NOT cause a crash or error when they get to the compliance summary. 

#### Acceptance Criteria

- [x] migration to add new code
- [x] Add an `is_regulated` column to the NaicsCode model
- [x] ComplianceService should only compute values for regulated NaicsCodes, throw an error otherwise
- [x] Tests updated - NaicsRegulatoryValues should only have entries for regulated naics codes

#### Definition of Done

- [ ] first
- [ ] second
- [ ] third

#### Additional context

- Add any other context about the task here.
- Or here

fviduya:
I have tested this in the DEV environment, [here is the test document for reference](https://buttoninc-my.sharepoint.com/:w:/g/personal/francesca_button_is/EeV0VajWzPtBr_t7dPadOgQBas0SiyyOJqsoH6UjDaRlIA?e=0L0dts).

I can verify that:
- [x] The following NAICS codes have been added to the database:
   - [x] 111412	Cannabis grown under cover 
   - [x] 111419	Other food crops grown under cover 
   - [x] 221111	Hydro-Electric Power Generation  
   - [x] 221112	Fossil-Fuel Electric Power Generation  
   - [x] 221119	Other Electric Power Generation  
   - [x] 221121	Electric Bulk Power Transmission and Control  
   - [x] 221210	Natural Gas Distribution  
   - [x] 221320	Sewage Treatment Facilities  
   - [x] 221330	Steam and Air-Conditioning Supply  
   - [x] 486110	Pipeline Transportation of Crude Oil  
   - [x] 493110	General Warehousing and Storage  
[...]

PR #3297, opened 2025-05-05
Add unregulated NAICS codes

BCerki:
https://github.com/bcgov/cas-registration/issues/3296

This PR:
- migration to add the codes and regulatory data
- added is_regulated to the NaicsCode model (we're not doing anything with it at the moment, but it may be helpful later if we restrict the code choices based on registration purpose)
- update pytests


bc_obps/registration/migrations/0109_historicalnaicscode_is_regulated_and_more.py
```diff
@@ -0,0 +1,29 @@
...
+            model_name='historicalnaicscode',
+            name='is_regulated',
+            field=models.BooleanField(
+                db_comment='Whether the NAICS code applies to regulated operations', default=True
+            ),
+            preserve_default=False,
+        ),
+        migrations.AddField(
+            model_name='naicscode',
+            name='is_regulated',
```
BCerki, line 23: I left this in because it sounds like we might later have restrictions on this. However, if the reviewer prefers, I can pull it out and we can add it at that later point if we need it

andrea-williams (commented):
Is the plan that the FE work re: filtering out the NAICS codes that don't apply to regulated operations will go into a separate PR?

BCerki:
> Is the plan that the FE work re: filtering out the NAICS codes that don't apply to regulated operations will go into a separate PR?

From the meeting, it sounded like we were going to do any filtering now; it could be an improvement for next year. Since I'd already implemented (I started on the card before we figured out everything with the biz area), I left it in, but if it's confusing I can remove it. It's not hard to redo later if we need it later.

bc_obps/registration/migrations/0110_add_non_regulated_naics_codes.py
```diff
@@ -0,0 +1,213 @@
...
+                    valid_to='9999-12-31',
+                ),
+            ]
+        )
+
+    def reverse_unregulated_regulatory_value_data(apps, schema_monitor):
+        '''
+        Remove data for non-regulated NAICS codes to erc.naics_regulatory_value
+        '''
+        NaicsRegulatoryValue = apps.get_model('registration', 'NaicsRegulatoryValue')
```
andrea-williams, line 190:
```suggestion
        NaicsRegulatoryValue = apps.get_model('reporting', 'NaicsRegulatoryValue')
```

BCerki:
closing, finished in https://github.com/bcgov/cas-registration/pull/3312

PR #3312, opened 2025-05-08
3296 2 unregulated naics codes

BCerki:
card: https://github.com/bcgov/cas-registration/issues/3296

This PR:

- migration to add the codes and regulatory data
- added is_regulated to the NaicsCode model (we're not doing anything with it at the moment, but it may be helpful later if we restrict the code choices based on registration purpose)
- update pytests


andrea-williams (commented):
I don't think the reversal methods in migration 110 are working. Would you mind trying these steps out and seeing if you get the same results as me?

1. make reset_db to get fully up-to-date db
2. `poetry run python manage.py migrate registration 0108` to reverse the 2 new migrations
3. `poetry run python manage.py migrate registration 0110` to re-apply the 2 newest migrations.

I think everything's ok with 109, but in 110 I then get an error because it's trying to add in Naics codes that already exist. When I did a count on the table rows after step 2 it was 47, which is also what it was after step 1.

andrea-williams (approved):

PR #3316, opened 2025-05-08
chore:added migration to fix gwp for C2F6

ayeshmcg:
Card: https://github.com/bcgov/cas-reporting/issues/752
Added migration to update gwp for C2F6


dleard (approved):
:shipit:

PR #3319, opened 2025-05-08
wip: rls reporting policy

acatchpole:
Current state: 
Helper functions:
```generate_report_policy_mapping_from_grants()```: creates the map object that is used to generate the policies
```generate_rls_policies()```: actually creates the RlsPolicy objects. should be project-agnostic.

With an appropriately formatted ```using_statement``` these functions work with the already-existing RLS code to implement RLS and RLS policies on tables. 

HAVE NOT tested that these policies work as expected.

### CODE REVIEW

File: bc_obps/reporting/models/rls_configs/report.py [hunk 1]
```diff
@@ -16,4 +19,28 @@ class Rls:
...
+        exists (
+            with approved_operator as (
+                select operator_id from erc.user_operator where
+                user_id = current_setting('my.guid', true)::uuid
+                and status='Approved'
+            )
+            select 1 from erc.operation_designated_operator_timeline tline
+            join approved_operator ao on ao.operator_id = tline.operator_id
+            and tline.operation_id = report.operation_id
+            and (start_date <= concat(report.reporting_year_id::text, '-12-31')::date and (end_date is null or end_date >= concat(report.reporting_year_id::text, '-12-31')::date))
```
pbastia, line 35: We can use the [built in `make_date`](https://www.postgresql.org/docs/current/functions-datetime.html#FUNCTIONS-DATETIME-TABLE) to make this more readable.

in case a future model has a start_date / end_date we can also make sure those fields come from the `tline` table

```suggestion
            and (tline.start_date <= make_date(report.reporting_year_id, 12, 31) and (tline.end_date is null or tline.end_date >= make_date(report.reporting_year_id, 12, 31)))
```

bc_obps/rls/utils/policy.py
```diff
@@ -79,23 +79,25 @@ def add_draft_check_to_report_using_statement(report_using_statement: str) -> st
...
+                select operator_id from erc.user_operator where
+                user_id = current_setting('my.guid', true)::uuid
+                and status='Approved'
+            )
+            select 1 from erc.report_version rv
+            join erc.report r on rv.report_id = r.id
+            join approved_operator ao on r.operator_id = ao.operator_id
+            join erc.operation_designated_operator_timeline tline on tline.operator_id = r.operator_id
+            and rv.id = report_version_id
+            and (start_date <= concat(r.reporting_year_id::text, '-12-31')::date and (end_date is null or end_date >= concat(r.reporting_year_id::text, '-12-31')::date))
```
pbastia: similarly 

```suggestion
            and (tline.start_date <= make_date(r.reporting_year_id, 12, 31) and (tline.end_date is null or tline.end_date >= make_date(r.reporting_year_id, 12, 31)))
```

bc_obps/reporting/models/rls_configs/facility_report.py
```diff
@@ -31,4 +41,37 @@ class Rls:
...
+            JOIN erc.report_version rv ON fr.report_version_id = rv.id
+            JOIN erc.report r ON rv.report_id = r.id
+            WHERE r.operator_id IN (
+                SELECT uo.operator_id
+                FROM erc.user_operator uo
+                WHERE uo.user_id = current_setting('my.guid', true)::uuid
+                AND uo.status = 'Approved'
+            )
+        )
+    """
```
pbastia, line 58: I wonder if we're missing the timeline clause in here?

pbastia: We might not, since we're joining with accessible report and report_version records and RLS **should** apply there

bc_obps/reporting/models/rls_configs/report_emission.py
```diff
@@ -16,8 +25,26 @@ class Rls:
...
+                JOIN erc.report_version rv ON re.report_version_id = rv.id
+                JOIN erc.report r ON rv.report_id = r.id
+                WHERE r.operator_id IN (
+                    SELECT uo.operator_id
+                    FROM erc.user_operator uo
+                    WHERE uo.user_id = current_setting('my.guid', true)::uuid
+                    AND uo.status = 'Approved'
+                )
+            )
+        """
```
pbastia: same comment as facility_report_report_activity

bc_obps/rls/utils/policy.py
```diff
@@ -79,23 +79,25 @@ def add_draft_check_to_report_using_statement(report_using_statement: str) -> st
...
+        )"""
+
```
pbastia, line 92: Wild idea, and tell me if I'm out to lunch.
Since we already have that timeline check on the report_version table, joining on the report_version should be enough for a RLS check

So we could do just verify we have access to the report version
```
REPORT_USING_STATEMENT = """
  exists(
    select 1 from erc.report_version rv where report_version_id = rv.id
  )
"""
```

bc_obps/reporting/models/rls_configs/facility_report.py
pbastia: Maybe checking the facility_report access is enough actually, and RLS on that will delegate it up the chain

```py
facility_report_activities_using_statement = """
  exists (
    select 1 from erc.facility_report fr where fr.id = facilityreport_id
  )
"""
```

pbastia (commented):
First batch of comments, I think in general we only need to apply the timeline check on the Report model, and any report down the line could just check access to the parent with a simple `exists(select 1 from ...)`

Issue #3055, opened 2025-03-19
Add e2e tests for Registration part 2


Sepehr-Sobhani:
#### Describe the task

A clear and concise description of what the task is.

#### Acceptance Criteria

Note: some of these have separate tickets, added as subissues below.

- [x] Refactor the conditionality in operation_service per https://teams.microsoft.com/l/message/19:81544b60877c4a5eb4fc7946ecbcb525@thread.tacv2/1745443723831?tenantId=6fdb5200-3d0d-4a8a-b036-d3685e359adc&groupId=29821293-3244-455d-beae-e01d27762cfe&parentMessageId=1745441837467&teamName=External%3A%20CAS%20-%20Clean%20Growth%20Digital%20Service%20Team&channelName=Developers&createdTime=1745443723831&ngc=true

Industry user
- [ ] login and profile: https://github.com/bcgov/cas-registration/issues/3450
- [ ] dashboard (may hit this bug: https://github.com/bcgov/cas-registration/issues/3230) ; #2311 
- [x] register a new operation
- [x] register an existing operation
- [x] select operator workflow
- [ ]  user and access requests (we may have a bug https://github.com/bcgov/cas-registration/issues/3227): https://github.com/bcgov/cas-registration/issues/3441
- [ ] edit operation - Change registration purpose (specifically useful maybe for reporting operation to regulated operation OR potential reporting to reporting?) (may hit bug: https://github.com/bcgov/cas-registration/issues/3269): https://github.com/bcgov/cas-registration/issues/3442
- [ ] add/edit facility: https://github.com/bcgov/cas-registration/issues/3443
- [x] add/edit contact
- [ ] edit operator? It's the same form as add, which is covered in select operator

Internal user
- [ ] login and dashboards: https://github.com/bcgov/cas-registration/issues/3444, https://github.com/orgs/bcgov/projects/123/views/16?pane=issue&itemId=81783909&issue=bcgov%7Ccas-registration%7C2311
- [ ]  Issue BORO ID and BCGHG ID: https://github.com/bcgov/cas-registration/issues/3445
- [ ] operator admin and access requests: https://github.com/bcgov/cas-registration/issues/3446
- [ ] transfers: https://github.com/bcgov/cas-registration/issues/3447
- [ ] internal access requests: https://github.com/bcgov/cas-registration/issues/3448
- [ ] one of the grids (probably the one with the most columns), mainly for accessibility: https://github.com/bcgov/cas-registration/issues/3449

[...]

PR #3321, opened 2025-05-08
chore: util for checking document scan

BCerki:
Part of https://github.com/bcgov/cas-registration/issues/3055

This PR:
- moves the conditionality about ignoring document scan in CI/local to a helper
- uses the helper in operation_service
- pytests (operation_service already had tests for this (search monkeypatch), so the new ones are only for the util)

### DIFF
--- a/bc_obps/registration/utils.py
+++ b/bc_obps/registration/utils.py
@@ -156,10 +156,9 @@ class CustomPagination(PageNumberPagination):
 
 
 def is_document_scan_complete(operation: Operation) -> bool:
-    # If we're in the CI or local environment, we don't need to check for document scanning
+    # If we're in the local environment, we don't hit GCS; we use local file storage (see settings.py)
     ENVIRONMENT = settings.ENVIRONMENT
-    CI = settings.CI
-    if CI == 'true' or ENVIRONMENT == 'local':
+    if ENVIRONMENT == 'local':
         return True
 
     return not operation.documents.filter(status=Document.FileStatus.UNSCANNED).exists()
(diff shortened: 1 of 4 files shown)


bc_obps/registration/utils.py
```diff
@@ -151,3 +152,17 @@ def paginate_queryset(
             "items": queryset[offset : offset + page_size],
             "count": self._items_count(queryset),
         }  # noqa: E203
+
+
+def is_document_scan_complete(operation: Operation) -> bool:
+    # If we're in the CI or local environment, we don't need to check for document scanning
+    ENVIRONMENT = os.environ.get("ENVIRONMENT")
+    CI = os.environ.get("CI")
```
Sepehr-Sobhani, line 160: Suggest importing these variables from `from django.conf import settings` and there is a good reason for that: https://docs.djangoproject.com/en/5.2/topics/settings/#using-settings-in-python-code

bc_obps/registration/utils.py
```diff
@@ -151,3 +152,17 @@ def paginate_queryset(
...
+    if CI == 'true' or ENVIRONMENT == 'local':
+        return True
+
+    if not operation.documents.filter(
+        status=Document.FileStatus.UNSCANNED,
+    ).exists():
+        return True
+    return False
```
Sepehr-Sobhani, line 168: Can be simplified to:
```python
    return not operation.documents.filter(
        status=Document.FileStatus.UNSCANNED
    ).exists()
```

Sepehr-Sobhani (approved):
LGTM! I’m going to approve this, but just a heads-up—I recall @pbastia had some thoughts or suggestions around this, so it might be worth double-checking with him.

BCerki:
@fviduya ,  testing is making sure people can still upload attachments in the registration workflow.

PR #3329, opened 2025-05-09
v2: add/edit contact 

fviduya:
Created a new PR to address code review suggestions from https://github.com/bcgov/cas-registration/pull/3291

1. Tests functionality of manually adding a contact
2. Tests functionality of editing a contact
3. ~Tests the functionality of adding a contact through the registration flow~ - removed since registration will fail if there are no operation representative set anyway
4. Adds a test setup file to be used globally 
5. Refactored fillRequiredFormFields - not being used by any E2E tests, I think it still needs more refactoring but figured it can be refactored when needed? 
6. Added successful snackbar helper

The test for adding a contact through user access approval is missing here, will be adding that when we create the test for user access and requests.

### DIFF
--- a/bciers/apps/administration-e2e/poms/contacts.ts
+++ b/bciers/apps/administration-e2e/poms/contacts.ts
@@ -122,7 +122,10 @@ export class ContactsPOM {
     const addButton = this.page.getByRole("button", {
       name: ContactButtonText.ADD_CONTACT,
     });
-    await this.page.waitForTimeout(500);
+    await this.page.waitForTimeout(1000);
+    // Wait for the button to be visible and enabled
+    await expect(addButton).toBeVisible();
+    await expect(addButton).toBeEnabled();
     await addButton.click();
     await this.assertFootnoteIsVisible(true);
   }
--- a/bciers/apps/administration-e2e/poms/contacts.ts
+++ b/bciers/apps/administration-e2e/poms/contacts.ts
@@ -123,14 +123,8 @@ export class ContactsPOM {
       name: ContactButtonText.ADD_CONTACT,
     });
     await this.page.waitForTimeout(500);
-    try {
-      await addButton.click();
-      await this.assertFootnoteIsVisible(true);
-    } catch (error) {
-      console.warn("UI Navigation failed, falling back to direct URL");
-      await this.page.goto(`${this.url}/add-contact`);
-      await this.assertFootnoteIsVisible(true);
-    }
+    await addButton.click();
+    await this.assertFootnoteIsVisible(true);
   }
 
   // ###  Assertions ###
--- a/bciers/apps/administration-e2e/poms/contacts.ts
+++ b/bciers/apps/administration-e2e/poms/contacts.ts
@@ -12,7 +12,11 @@ import {
   ContactButtonText,
   ContactFootnote,
 } from "@/administration-e2e/utils/enums";
-import { fillRequiredFormFields } from "@bciers/e2e/utils/helpers";
+import {
+  fillRequiredFormFields,
+  clickWithRetry,
+  clickButton,
+} from "@bciers/e2e/utils/helpers";
 // ℹ️ Environment variables
 import * as dotenv from "dotenv";
 dotenv.config({ path: "./e2e/.env.local" });
@@ -119,15 +123,22 @@ export class ContactsPOM {
   }
 
   async clickAddButton() {
-    const addButton = this.page.getByRole("button", {
-      name: ContactButtonText.ADD_CONTACT,
-    });
-    await this.page.waitForTimeout(1000);
-    // Wait for the button to be visible and enabled
-    await expect(addButton).toBeVisible();
-    await expect(addButton).toBeEnabled();
-    await addButton.click();
-    await this.assertFootnoteIsVisible(true);
+    const addButton = ContactButtonText.ADD_CONTACT;
+    await this.page.waitForTimeout(5000);
+
+    try {
+      await expect(
+        this.page.getByRole("button", { name: addButton }),
+      ).toBeVisible();
+      // Click the button
+      await clickButton(this.page, addButton);
+      await this.assertFootnoteIsVisible(true);
+    } catch (error) {
+      // Re-clicking the button here if it fails the first time, putting max retries to 3
+      console.warn("Button click failed, retrying");
+      await clickWithRetry(this.page, ContactButtonText.ADD_CONTACT, 2);
+      await this.assertFootnoteIsVisible(true);
+    }
   }
 
   // ###  Assertions ###
--- a/bciers/apps/administration-e2e/poms/contacts.ts
+++ b/bciers/apps/administration-e2e/poms/contacts.ts
@@ -1,0 +1,139 @@
+/**
+ * 📖 https://playwright.dev/docs/pom
+ * Page objects model (POM) simplify test authoring by creating a higher-level API
+ * POM simplify maintenance by capturing element selectors in one place and create reusable code to avoid repetition. *
+ */
+import { Locator, Page, expect } from "@playwright/test";
+// ☰ Enums
+import {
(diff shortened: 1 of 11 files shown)


bciers/apps/administration-e2e/poms/contacts.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  async clickAddButton() {
+    await this.page.waitForTimeout(500);
+    try {
+      await this.page
+        .getByRole("button", { name: ContactButtonText.ADD_CONTACT })
+        .click();
+    } catch (error) {
+      console.warn("UI Navigation failed, falling back to direct URL");
+      await this.page.goto(`${this.url}/add-contact`);
```
fviduya, line 128: I originally had waitForTimeout in route(), but moved it here since it's not needed anymore for searching in grid after using stabilizeGrid.

Also tried using different assertions but none seems to work expect for this or waitForLoadState('networkidle') which is not recommended by playwright. Opted to use this and add a try-catch method so in case clicking the button still results to an error, it will just go directly to the URL and the test _shouldn't_ fail.

bciers/apps/administration-e2e/tests/workflows/contacts.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+      "bill.blue@example.com",
+    );
+
+    // Click View Details
+    const viewDetails = searchResult
+      .getByRole("link", {
+        name: /view details/i,
+      })
+      .first();
+    await stabilizeGrid(page, 1);
```
fviduya, line 65: This worked and allowed me to click the View Details link, thanks!! 

ref: https://github.com/bcgov/cas-registration/pull/3291#discussion_r2071905537

bciers/libs/e2e/src/utils/helpers.ts
```diff
@@ -269,24 +270,35 @@ export async function fieldsClear(page: Page, formFields: string | any[]) {
...
-      // Click the field to focus it
-      await inputField.click();
-      switch (labelText) {
-        case FormField.PHONE:
-          await inputField.fill(E2EValue.INPUT_PHONE); //Format should be ### ### ####
-          break;
-        default:
-          await inputField.fill(`${E2EValue.PREFIX} ${labelText}`);
-          break;
+export async function fillRequiredFormFields(
```
fviduya, line 273: This isn't being used elsewhere, in the meantime I wrote it this way, I think it still needs some refactoring to make it generic but maybe we can do that as we move forward. I can write up a ticket to go over what we already have in the helpers and avoid duplication.

bciers/apps/registration/app/data/jsonSchema/operationInformation/registrationPurpose.ts
```diff
@@ -62,7 +62,7 @@ export const createRegistrationPurposeSchema = async () => {
       },
       registration_purpose: {
         type: "string",
-        title: "The purpose of this registration is to register as a:",
+        title: "the purpose of this registration is to register as a:",
```
BCerki, line 65: Typo?

bciers/apps/administration/app/components/contacts/ContactPage.tsx
```diff
@@ -27,7 +27,7 @@ export default async function ContactPage({
   }
 
   const noteMsg = isCreating
-    ? "Once added, this new contact can be selected wherever needed or applicable."
+    ? "Test Once added, this new contact can be selected wherever needed or applicable."
```
BCerki, line 30: Typo

bciers/apps/administration-e2e/poms/contacts.ts
BCerki: Going directly to the URL isn't how a user would do it, so I think we should avoid that. Timeout or networkidle aren't ideal, but better than a direct navigation, IMO.

Also, in the happo shot it looks like the grid hasn't finished loading (action column looks funky), so I wonder if using stabilize grid before trying to click "Add Contact" might help? 
image

BCerki (commented):
Nice work, thanks for the refactors and cleanup!
There are a couple of things to look at in the happo diffs:
- There's an internal server error on the Contact form 
- The grid looks like it hasn't finished loading, see the inline comment for more details

bciers/apps/administration-e2e/poms/contacts.ts
fviduya: Currently, the timeout set in line 125 and in the function in line 128 works as how a user would interact with the page. But yeah, I agree that using the URL is not intuitive, there are timeouts in place already so I can omit that since it's working without it as well.

Also, the screenshot is taken after searching in the grid then clicking View Details (so maybe this is where I'm wrong if I wanted to see the grid?)

bciers/apps/registration/app/data/jsonSchema/operationInformation/registrationPurpose.ts
fviduya: I tried removing this, I thought the happo screenshots would show without this temp change but it skipped it. So I'm putting this back to check happo again 🙂

bciers/apps/administration-e2e/poms/contacts.ts
```diff
@@ -0,0 +1,160 @@
...
+      // Click the button
+      await clickButton(this.page, addButton);
+      await this.assertFootnoteIsVisible(true);
+    } catch (error) {
+      // Re-clicking the button here if it fails the first time, putting max retries to 3
+      console.warn("Button click failed, retrying");
+      await clickWithRetry(this.page, ContactButtonText.ADD_CONTACT, 2);
+      await this.assertFootnoteIsVisible(true);
+    }
+  }
```
fviduya, line 142: I tried more solutions on this one. The issue I was getting here was that this fails the first run, but passes on the second run. The original try-catch I had was to navigate directly to the URL, it's not how a user would normally interact. I omitted that and created another function to keep on clicking the button for a maximum number of 3 tries (it usually resolves after the 2nd try, but still I feel it's a messy fix). 

I've additionally tried `stabilizeGrid` before clicking the button to make sure all the expected records are there before clicking the button but it was still failing. The other thing I tried, which worked but I'm not sure why, is this:

```
  async clickAddButton() {
    const addButton = ContactButtonText.ADD_CONTACT;
    await this.page.waitForTimeout(5000);
    await expect(
      this.page.getByRole("button", { name: addButton }),
    ).toBeVisible();
    await clickButton(this.page, addButton);
    await this.page.reload({waitUntil:'load'});
    await this.assertFootnoteIsVisible(true);
  }
```

I'm not sure what else I can try, or which solution is more suitable.

fviduya:
I've resolved the Internal Server Error that was showing in Happo, but the grid alignment hasn't been resolved yet. I'm not sure what the issue there is, I tried for instance to wait for about 10 seconds just to check if it's because the page is still rendering, but the result was the same. I checked the others that we have right now, and I saw that the same thing was happening for the LFO Facility Information screenshot. 

Also, I couldn't seem to get the happo screenshots to show without the mock changes I made for the frontend files in reg and admin, so I'll remove that after the review just before merging.

bciers/apps/administration-e2e/poms/contacts.ts
BCerki: Thanks for doing so much exploration on this one! I think the above solution is better than navigating directly, and having a test with a timeout is way better than not having a test at all, so let's use this. We can always iterate on this later if we get another idea.

BCerki (approved):
:cartwheeling:

bciers/apps/administration-e2e/poms/contacts.ts
fviduya: Thanks for the review, @BCerki. I just got back to this. I saw you approved it already so I wanted to confirm first, is above solution the snippet I provided in my comment or the current one implemented right now? (The one with 3 tries) 😄

Issue #3332, opened 2025-05-09
Old emails going out


BCerki:
#### Describe the Bug:

A user received an old email after registering instead of the reg2 version.

#### Probability:

(How likely the bug is to happen, scored from 1-5. For example, probability of 5 is something like "it happens to all users every time they log in." whereas probability of 1 "only happens to certain users when a really specific and unlikely path is followed.")

#### Impact:

(How bad the bug is when it does happen, scored from 1-5. For example, effect of 5 is "the entire app crashes and makes it unusable for all users" or "the bug causes the wrong data to be saved, with critical information (e.g. payment) being affected." whereas effect of 1 is "It makes some styling look a little bit weird.")

<img src="https://www.getzephyr.com/sites/default/files/riksk%20matrix.jpg" alt="Bug Matrix" width="250" height="250"/>

#### Steps to reproduce the behaviour:

1. Step #1
2. Step #2
3. Step #3

#### Additional information:

Browser: [e.g. chrome, safari]
Version: [e.g. 22]

#### Additional Context

-

fviduya:
I have tested this in the DEV environment, [here is the test document for reference](https://buttoninc-my.sharepoint.com/:w:/g/personal/francesca_button_is/ETZtIKktFShCpgnAfAVwwWIBfgnrOq4Ex8Ujse0mmUq-wA?e=5MnNml).

I can verify that:
- [x] After user completes registration of operation, an updated email with subject "BCIERS Receipt Acknowledgement – Registration Submitted for <operation> of <operator>" is received
- [x] After the BORO ID is issued, the user receives another email with subject "BCIERS Notification - BORO ID issued for <operation> of <operator>" is received
- [ ] The contents of the emails received matches the contents specified in [Reg2 email template](https://bcgov.sharepoint.com/:w:/r/teams/00608-CPRInternal/Shared%20Documents/Carbon%20Pricing%20Transition%20SMEs/Registration%20App/Automatic%20Emails%20-%20Reg%202.0.docx?d=w96a3eadf60224795852fd84510d99756&csf=1&web=1&e=h1KOAA).

Moving to Pending PO for review of test. Thank you!

patriciarussellCAS:
Correct email received in Test! Hooray

PR #3335, opened 2025-05-09
3332 end correct email template to confirm registration

BCerki:
card: https://github.com/bcgov/cas-registration/issues/3332

This PR:
- removes the `send_email` service (we weren't using)
- renames the email template enums for hopefully more clarity
- uses the new email template in operation_service
- pytests for ^
- tweak docs to explain how to test emails in local

I've also made a tech debt ticket with some nice-to-haves for emails: https://github.com/bcgov/cas-registration/issues/3336


andrea-williams (approved):
Thanks for fixing this Bri!

Issue #3269, opened 2025-04-25
USER BUG: Editing operation information


zoeyli-46:
[Thread](https://teams.microsoft.com/l/message/19:3b0107d27a7b452a89e6ffd7e16ed4a8@thread.tacv2/1745535412867?tenantId=6fdb5200-3d0d-4a8a-b036-d3685e359adc&groupId=29821293-3244-455d-beae-e01d27762cfe&parentMessageId=1745535412867&teamName=External%3A%20CAS%20-%20Clean%20Growth%20Digital%20Service%20Team&channelName=BCIERS%20Internal%20Users&createdTime=1745535412867&ngc=true)

#### Describe the Bug:
Two operators reported the following error code when they've attempted to make changes to their operation information in the Admin module. 
One made changes to their registration purpose and the other attempted to update their operation name. 
One of the same operators reported the same error when they've attempted to view their draft report.

#### Steps to reproduce the behaviour:

1. Go to Operation Information in Admin module
2. Edit information
3. See error message 

#### Screenshots:
![Image](https://github.com/user-attachments/assets/338e7cd3-0044-4919-80d8-63c7fdf13652)

#### Additional information:

Browser: [e.g. chrome, safari]
Version: [e.g. 22]

#### Additional Context

-

patriciarussellCAS:
I've added ticket 747 from reporting as a sub-issue, as it's mostly the same Operation experiencing multiple issues (Tidewater HDRD) - the second operation is NGTL. I think these and #3340 are all connected or even the same issue so we may not need all these separate tickets. I'll leave as is for now.

fviduya:
I noticed that these are migration files. Should this be transferred to the Accepted lane? @andrea-williams

andrea-williams:
Yes the migration file is only to decouple contacts from operations when the operation has been transferred to a new operator and the operation's contact is from the old operator. It's been tested locally by myself and Bri using prod data, but there's no easy way for anyone to test it in DEV or TEST.

PR #3344, opened 2025-05-12
fix: added forward-only prod data migration to remove old contacts as…

andrea-williams:
Card: #3269 

Forward-only prod data migration to remove the contacts associated with an operation that has been transferred, if the contact is associated with the from_operator (not the to_operator) of the transfer.

#### Suggested steps to test locally:
1. Use check_migrations_with_prod_data script to pull a copy of the existing prod data into your local (proceed with caution, remember to delete data afterwards)
2. Once the prod data is in your local db, run `make migrate` to apply the migration in this PR.
3. Check output of command - I've included print statements in the migration to assist with testing - we can (and probably should) delete this before merging this PR in. 

Expected output:
- there are 3 operations that have already been transferred
- 1 of those operations has 2 associated contacts (both need to be removed), and the other 2 operations have 1 contact each. So in total 4 operation_contact records should be removed
- ZERO contacts should actually be deleted from the database


bc_obps/registration/migrations/0112_prod_fix_contacts_post_transfer.py
```diff
@@ -0,0 +1,62 @@
+# Generated by Django 5.0.14 on 2025-05-12 20:10
+
+from django.db import migrations
+
+"""
+One-time forward-only migration to be applied to prod data.
+Purpose: delete operation representatives from operation_contacts table for operations that have been transferred to a new operator.
```
BCerki, line 7: I think the PR description is clearer, we could resuse: `Forward-only prod data migration to remove the contacts associated with an operation that has been transferred, if the contact is associated with the from_operator (not the to_operator) of the transfer.`

bc_obps/registration/migrations/0112_prod_fix_contacts_post_transfer.py
```diff
@@ -0,0 +1,62 @@
...
+        to_operator = transfer_event.to_operator
+        operation_contacts = operation.contacts.all()
+
+        operation_contacts_count_before = operation_contacts.count()
+
+        # for each contact in operation_contacts, check if the contact is associated with the from_operator. If so, delete it.
+        for contact in operation_contacts:
+            if contact.operator == from_operator:
+                # log the removal
+                print(
```
BCerki, line 33: Yes, I agree, let's pull this out before merging. It was useful for reviewing, though!

BCerki (approved):
This fixes the internal server error in admin.

One of the users was also reporting an error with viewing their draft report, and that's still a problem :( But it can be handled in https://github.com/bcgov/cas-reporting/issues/757

PR #3348, opened 2025-05-13
758 numberfield fix

gdalcengio:
- **fix: Allow zeroes for numberfields**
- **chore: make style of numberfield widget consistent**

User can type and submit 0s again if the field allows and the border color is fixed

### DIFF
--- a/bciers/libs/components/src/form/widgets/TextWidget.tsx
+++ b/bciers/libs/components/src/form/widgets/TextWidget.tsx
@@ -34,7 +34,6 @@ const TextWidget: React.FC<WidgetProps> = ({
   };
 
   const handleNumberChange = (val: number | null) => {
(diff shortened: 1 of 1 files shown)


bciers/libs/components/src/form/widgets/TextWidget.tsx
```diff
@@ -34,7 +34,7 @@ const TextWidget: React.FC<WidgetProps> = ({
   };
 
   const handleNumberChange = (val: number | null) => {
-    if (!val) return;
+    if (val === null || val === undefined) return;
```
pbastia, line 37: This prevents the user from emptying a number box. I suggest removing it

pbastia (changes requested):

pbastia (approved):

Issue #3180, opened 2025-04-09
Admin/Transfers should show ALL operations, not just registered ones


andrea-williams:
#### Describe the task

Current behaviour: when an internal user creates a new transfer, the dropdown menu for Operations only shows operations with a status of "Registered".

This is not desired - if an operation or facility has already transferred ownership in the real world, the old operator should not have to register the operation just so they can get it transferred to the actual (new) owner in BCIERS.

See `bciers/apps/registration/app/data/jsonSchema/transfer/transferDetail.ts` file for how dropdown menus are populated.

#### Dev Checklist
- [ ] populate drop-down menu of Operations in the Transfers page based solely on the `Operations` table (using `operator_id` field to track current ownership), DON'T filter by operation status
- [ ] populate dropdown menu of Facilities in the Transfers page based solely on the `Facilities` table (using `operation_id` field to track current ownership)

#### Acceptance Criteria
- [x] once an operator has been selected, ALL operations associated with that operator should appear in the operation dropdown menu, regardless of the operation's status (i.e., Registered or not)
- [x] once an operator has been selected, ALL facilities associated with any operation belonging to the selected operator should appear in the facility dropdown menu, regardless of the operation's status
- [ ] third

#### Definition of Done

- [ ] first
- [ ] second
- [ ] third

#### Additional context

- Add any other context about the task here.
- Or here

BCerki:
I went rogue on this one and did it differently than the dev checklist suggsted. To test:
- confirm you can see draft records in the transfers form dropdowns
- confirm you cannot see draft records in the admin operations grid

fviduya:
I have tested this in DEV, [here is the test document for reference](https://buttoninc-my.sharepoint.com/:w:/g/personal/francesca_button_is/EXGmGQUQzN5MhsmbdH_ftkUBEH0pNRRYccj1FZNSK_xPfQ?e=v6EuTc).

I can verify that:
- [x] Internal users can see Draft operations in the dropdown
- [x] Internal users can transfer Draft operations
- [x] Internal users cannot see Draft operations in the operations grid

Moving to Pending PO for review of test. Thank you!

patriciarussellCAS:
Will need to double check this in TEST. Moving to accepted.

patriciarussellCAS:
Was able to transfer an operation that was not fully registered.

PR #3349, opened 2025-05-13
3180 transfers show all operations

BCerki:
card: https://github.com/bcgov/cas-registration/issues/3180

This PR:
- changes the service that pulls operation data for IRC users to not filter automatically on status
- use the status filter on the endpoint instead (sent via the front end grid component)
- I chose to do this^ to avoid making a new endpoint/service that was specific to filtering on the operation status. I don't ~think there's any problem with having the service fetch all operation statuses by default. If IRC users are allowed to see everything in transfers, then we aren't restricting it in the grid for security, just for convenience
- update pytests

### DIFF
--- a/bc_obps/service/data_access_service/operation_designated_operator_timeline_service.py
+++ b/bc_obps/service/data_access_service/operation_designated_operator_timeline_service.py
@@ -77,8 +77,8 @@ class OperationDesignatedOperatorTimelineDataAccessService:
         )
 
         if user.is_irc_user():
-            # IRC users only see registered operations
-            return queryset.filter(operation__status=Operation.Statuses.REGISTERED, end_date__isnull=True)
+            # IRC users see all operations (subject to filtering that's done on the endpoint)
+            return queryset.filter(end_date__isnull=True)
         else:
             # Industry users can only see operations associated with their own operator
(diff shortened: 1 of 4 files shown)


andrea-williams:
While logged in as an IDIR user, when I view the Operations grid in Admin module, I can see all operations, not just registered ones (e.g., Bojangles EIO, Blue LFO, etc.)

BCerki:
> While logged in as an IDIR user, when I view the Operations grid in Admin module, I can see all operations, not just registered ones (e.g., Bojangles EIO, Blue LFO, etc.)

Thanks for catching this! I'm assuming you saw the draft etc. operations when using the filters or going to page 2 of the grid? I've fixed that now; I forgot to account for pagination.

Any suggestions on how to test for that besides an e2e test? In pytest I'm testing that the service applies the filters correctly, and in vitest I'm testing that mocked data is displayed correctly, but my mistake was the interaction between the two.

andrea-williams (approved):
Hmm I haven't been able to come up with any bright ideas for how we could test that without having to create an E2E test. I think it's probably safe to leave it as-is for now - we are already aware that there are multiple concerns with the Transfers feature, and the business area is too.

Issue #3268, opened 2025-04-25
USER BUG: Logout warning modal inconsistent


zoeyli-46:
Reported from 1 user - Enbridge: the system will log me out without any warning after some time , even when I am still working on them. I found it when I tried to save but cannot save and had to re-enter the data

*NOTE: This appears to be happening to a number of users. Enbridge was the first to report it. 

Logout modal might be working correctly after being implemented in #2925 BUT it might be dissapearing after the user gets signed out. So the user likely never even saw the count down modal, even though it appeared. If we shorten the timeout, (instead of after 25 mins) will there be a higher chance of the user seeing this modal?  
- [ ] "you've been signed out page" not appearing consistently, it sometimes shows and sometimes doesn't. when it doesn't show, the tab remains on the last page that the user was on, for example, the report

zoeyli-46:
hey @Sepehr-Sobhani @BCerki @fviduya do you guys know why the logout warning modal did not appear? Or perhaps it appeared but disappeared after the time out happened, without telling the user that they got logged out.

BCerki:
There's some more info about this in this ticket: https://github.com/orgs/bcgov/projects/123/views/16?filterQuery=repo%3Abcgov%2Fcas-registration++logout&pane=issue&itemId=99603844&issue=bcgov%7Ccas-registration%7C2925

patriciarussellCAS:
Tested this in TEST today (May 6) by logging in to the main dashboard and leaving the page. 
Here is the screenshot after I have been logged out. No pop-up warning (though I was working in a different tab so maybe it popped up and then disappeared before I saw it??)

![Image](https://github.com/user-attachments/assets/c50fe284-eb81-4b58-a5dc-19d56903b3a8)

Sepehr-Sobhani:
> Tested this in TEST today (May 6) by logging in to the main dashboard and leaving the page. Here is the screenshot after I have been logged out. No pop-up warning (though I was working in a different tab so maybe it popped up and then disappeared before I saw it??)
> 
> ![Image](https://github.com/user-attachments/assets/c50fe284-eb81-4b58-a5dc-19d56903b3a8)

Yes, the modal will appear only in the specific tab you're using, not in other tabs. It showed up and logged you out after 5 minutes.

zoeyli-46:
The "you've been signed out" page doesn't always appear. I left the tab open yesterday on the final review page, and came back today to the tab and it was still on the final review page instead of on the "you've been signed out"

[...]

PR #3352, opened 2025-05-13
3268 logout redirect

BCerki:
card: https://github.com/bcgov/cas-registration/issues/3268

This card talks about two things:
- logout modal doesn't show
- user is not redirected after logout (Zoey's comment `The "you've been signed out" page doesn't always appear. I left the tab open yesterday on the final review page, and came back today to the tab and it was still on the final review page instead of on the "you've been signed out"`)

These problems are caused by having the app open in more than one tab. The non-active tab logs the user out of all tabs after 30 minutes but only redirects them from the tab the triggered the logout. This makes it seem like the modal never showed up (it did, but it was in the inactive tab). It also makes it seem like the user is still logged in when they're not because they're not redirected to the logout page (only the inactive tab redirected).

We considered polling the token expiration and using that as a proxy for user activity. However, the token's expiry is only updated when `useSession` (client components) or `getSession` (server components) is called. This doesn't happen often enough to be a good indicator of user activity, e.g., you can get through the whole registration workflow without updating expiry. (Sidenote, calling `getToken` in the middleware does not refresh the session--can't refresh from middleware because it runs before cookies are set, and `getToken` is just a getter anyway). 

In the end, we used Broadcast Channel to send updates across the tabs. ([Nextauth already uses broadcast channel](https://leapcell.medium.com/deep-dive-into-nextauth-js-a-powerful-and-flexible-authentication-solution-c9b4ebd37b42) to sign users out of all tabs if they sign out of one.) 

This PR:
- set the throttle value lower
- fetch the `logoutUrl differrently
- set a default `logoutUrl`
- tab handling as part of user activity monitoring
- set up Sentry error capturing specifically for log out
- updated deprecated `callbackUrl` to `redirectTo`
- channel for redirecting extra tabs to the logout page, as Nextauth doesn't seem to handle that
- channel to refresh session in the extra tabs anytime (throttled) there's user activity

To test:
- change maxAge, MODAL_DISPLAY_SECONDS, and ACTIVITY_THROTTLE_SECONDS so you don't have to wait half an hour
- open the app in two tabs. The logout and extend session buttons should affect both apps
- let the timer rundown and you should be logged out and redirected to the keycloak logout page
- to test the fallback, temporarily delete `SITEMINDER_KEYCLOAK_LOGOUT_URL` from your env and you'll be redirected to the onboarding page instead of logout page

Notes:
Every once in awhile I'll get "logout failed" on one of the tabs: 
image. I ~think this is because I've already been logged out by one of the other tabs

### DIFF
--- a/bciers/libs/components/src/auth/SessionTimeoutHandler.test.tsx
+++ b/bciers/libs/components/src/auth/SessionTimeoutHandler.test.tsx
@@ -44,8 +44,6 @@ describe("SessionTimeoutHandler", () => {
   const mockUpdate = vi.fn();
   const mockSignOut = signOut as ReturnType<typeof vi.fn>;
   const mockGetEnvValue = getEnvValue as ReturnType<typeof vi.fn>;
-  const mockCreateThrottledEventHandler =
-    createThrottledEventHandler as ReturnType<typeof vi.fn>;
 
   const defaultSession = {
     data: { expires: new Date(Date.now() + 180 * 1000).toISOString() },
@@ -64,7 +62,6 @@ describe("SessionTimeoutHandler", () => {
   beforeEach(() => {
     vi.clearAllMocks();
     mockGetEnvValue.mockResolvedValue("http://logout.url"); // NOSONAR
-    mockCreateThrottledEventHandler.mockReturnValue(vi.fn());
   });
 
   it("does not render anything when not authenticated", () => {
@@ -72,13 +69,9 @@ describe("SessionTimeoutHandler", () => {
     expect(screen.queryByTestId("logout-modal")).not.toBeInTheDocument();
   });
 
-  it("sets up throttled event handler on mount when authenticated and modal is not shown", () => {
+  it("starts polling on mount when authenticated", () => {
     renderWithSession();
-    expect(mockCreateThrottledEventHandler).toHaveBeenCalledWith(
-      expect.any(Function),
-      ["mousemove", "keydown", "mousedown", "scroll", "visibilitychange"],
-      ACTIVITY_THROTTLE_SECONDS,
-    );
+    expect(startTokenExpirationPolling).toHaveBeenCalled();
   });
 
   it("shows modal when session is about to expire", async () => {
@@ -104,12 +97,12 @@ describe("SessionTimeoutHandler", () => {
     );
   });
 
-  it("logs out when session timeout reaches zero", async () => {
+  it("logs out and broadcasts when session timeout reaches zero", async () => {
     renderWithSession({
       data: { expires: new Date(Date.now() - 1 * 1000).toISOString() },
     });
 
-    expect(mockGetEnvValue).toHaveBeenCalledOnce();
+    expect(mockGetEnvValue).toHaveBeenCalledOnce(); // maybe twice would catch broadcast
     await waitFor(() => expect(mockSignOut).toHaveBeenCalledOnce());
   });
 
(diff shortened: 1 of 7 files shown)


bciers/libs/components/src/auth/SessionTimeoutHandler.tsx
```diff
@@ -30,11 +30,19 @@ const SessionTimeoutHandler: React.FC = () => {
       .catch((error) => console.error("Failed to fetch logout URL:", error));
   }, []);
 
-  const handleLogout = () => signOut({ callbackUrl: logoutUrl });
```
BCerki, line 33: If there's no `logoutUrl`, then the fallback is the current page (see the `signOut` function in `bciers/node_modules/next-auth/src/react.tsx`). So when a user isn't being redirected, possibly what's happening is that something went wrong with `getEnvValue` in the `useEffect`, and we don't actually have a `logoutUrl`. I couldn't trigger this locally to check, unfortunately.

If we don't get the `logoutUrl`, we could:
- redirect the user to a different page (maybe make a fallback one that just says "You've been logged out" or something)
- throw an error

bciers/libs/components/src/auth/SessionTimeoutHandler.tsx
```diff
@@ -30,11 +30,19 @@ const SessionTimeoutHandler: React.FC = () => {
...
+    }
+    signOut({ callbackUrl: logoutUrl });
+  };
 
   // Refreshes the session and updates the timeout based on new expiration
   const refreshSession = async (): Promise<void> => {
-    if (status !== "authenticated") return;
+    if (status !== "authenticated") {
+      await handleLogout();
+      return;
```
BCerki, line 44: Returning nothing could have also been why a user doesn't know they've been logged out?

bciers/libs/components/src/auth/SessionTimeoutHandler.test.tsx
```diff
@@ -208,7 +209,34 @@ describe("SessionTimeoutHandler", () => {
...
+    });
+
+    renderWithSession();
+
+    // Manually invoke the refreshSession to simulate activity
+    await capturedRefreshSession();
+
+    await waitFor(() => {
+      expect(mockSignOut).toHaveBeenCalledWith({
+        redirectTo: "http://logout.url", // NOSONAR
```
BCerki, line 229: This assertion fails because the redirectTo is '/', the fallback. When I log it, I do get the correct url

shon-button:
@BCerki 

I can consistently reproduce the "the system will log me out without any warning after some time" in the following scenarios (also documented [here](https://github.com/bcgov/cas-reporting/issues/741#issuecomment-2877595956):

1. Have multiple tabs open where on tab is "abandoned" which logs itself out and results in my active tab being unauthenticated without my knowledge
2. Have my frontend service stopped, wait for the throttle activity listener to check my session which then fails and  makes tab being unauthenticated without my knowledge...start the FE service and try to save data returns error.

For scenario 1, it seems we should be using the server side token to get the actual expiration which wil be updated by the other browsers.

For scenario 2, perhaps extending the throttle time is a better strategy to prevent catching our pods being down.

bciers/libs/components/src/auth/SessionTimeoutHandler.tsx
```diff
@@ -21,20 +22,42 @@ const SessionTimeoutHandler: React.FC = () => {
...
+    channel.onmessage = (event) => {
+      if (event === "logout") {
+        window.location.href = "/your-logout-page"; // or `logoutUrl`
+      }
+    };
+
+    return () => {
+      channel.close();
+    };
   }, []);
```
BCerki, line 41: This section is weirding me out because channel.onmessage doesn't seem to be doing anything, but I am properly being logged out and redirected to the logout page in both tabs. Nextauth uses broadcasting already (https://leapcell.medium.com/deep-dive-into-nextauth-js-a-powerful-and-flexible-authentication-solution-c9b4ebd37b42) so maybe it's some weird interaction between broadcasts?

Issue #3186, opened 2025-04-09
WAVA SCAN Medium priority #1


patrickisaac:
#### Description of the Tech Debt
Description has been provided in the doc saved in https://bcgov.sharepoint.com/:w:/r/teams/00608-Developers/Shared%20Documents/Developers/WAVA%20Scan%20Ticket%20Information/Ticket%20%233186.docx?d=wbf8496c00bc7400aa59af12094b83351&csf=1&web=1&e=ygiY1g

#### Tech Debt Triage

The purpose of our technical debt triage process is to analyze technical debt to determine risk level of the technical debt and the value in tackling that technical debt.

#### Risk Value Scoring:

| Level  | Value                 |
| ------ | --------------------- |
| High   | <div align="center">3 |
| Medium | <div align="center">2 |
| Low    | <div align="center">1 |

| Technical Debt - Risk Types                                                                                                   | Level | Value |
| ----------------------------------------------------------------------------------------------------------------------------- | ----- | ----- |
| Business Area Risk - Risk of business area visibility / damage to user experience                                             | 0     | 0     |
| Developer Fault Risk - How likely will this tech debt cause a future error related to coding on top of it                     | 0     | 0     |
| System Fault Risk - Risk of system errors or application downtime                                                             | 0     | 0     |
| Time Scale Risk - Compound risk effect if left alone. How much more difficult to fix or dangerous will this become over time? | 0     | 0     |
| Time Sink Risk - How much will this tech debt slow the development process down                                               | 0     | 0     |
| <div align="right">**TOTAL SCORE:**                                                                                           | 0     | 0     |

#### Development Checklist:

- [ ] Checklist item
- [ ] Checklist item
- [ ] Checklist item

patriciarussellCAS:
@andrea-williams I think that makes sense, so long as we can properly document somewhere why we decided it's no longer relevant. Perhaps in scrum team?

patriciarussellCAS:
Thanks @patrickisaac - I don't have access to Developers channel so will leave that up to Andrea. 
In the meantime, I can confirm that developers have evaluated and reviewed the medium-risk issue identified and determined that all required work for this issue has been completed. Closing this ticket.

andrea-williams:
I've added the rationale Pierre provided to the doc. @patriciarussellCAS would it be useful to you if we moved those docs somewhere else so that you can access them?

PR #3354, opened 2025-05-14
3186: add sameSite= "strict" to auth cookies

andrea-williams:
Card: #3186 

### 🚀 Impact:

- In `auth.config` using prefix ` __Secure` in production
- In  `/onboarding`  forces a true top‑level GET to `/dashboard`
- added routing logic from withAuthorizationDashboard (SSR) also in /onboarding page (CSR)

### 🧪 Testing:
- if testing in your local env: I'd recommend trying the steps below twice, with `reactStrictMode` set to both `true` (which we normally use in local envs) and `false` (matches our DEV, TEST, and PROD environments). This setting can be changed in `bciers/next.config.base.js` within the `baseConfig` dict. When reactStrictMode is set to true, all React components are rendered twice. The behaviour should be the same now for both (but it wasn't always this way, and gave me trouble).
- in a fresh clean browser session (so either clear your cookies & cache, or start a new Incognito browser window), sign in with your IDIR. Once you've authenticated and been redirected to either the profile page or the main dashboard, inspect your cookies for the Application. You should cookies for `session-token`, `csrf-token`, and `callback-url`, and for each of them, the SameSite attribute should be set to "Strict" (where before it was set to "Lax" by default)
- - repeat the same when logging in with Business BCeID

### ❗ Help
- E2E tests pass in local, consistently fail in CI. I don't know why. Please let me know if you have any ideas/suggestions.

### DIFF
--- a/bciers/libs/e2e/src/utils/helpers.ts
+++ b/bciers/libs/e2e/src/utils/helpers.ts
@@ -383,6 +383,11 @@ export async function setupTestEnvironment(
 ) {
   let browser: Browser | null = null;
 
+  const configs = {
+    headless: true,
+    args: ["--disable-animations"],
+  };
+
   // Attempt launching browsers in order of preference
   for (const browserType of ["chromium", "firefox", "webkit"]) {
     //    console.log(browserType);  // ? always Chromium in e2e:ui
@@ -390,11 +395,11 @@ export async function setupTestEnvironment(
       browser = await (async () => {
         switch (browserType) {
           case "chromium":
-            return chromium.launch();
+            return chromium.launch(configs);
           case "firefox":
-            return firefox.launch();
+            return firefox.launch(configs);
           case "webkit":
-            return webkit.launch();
+            return webkit.launch(configs);
           default:
             throw new Error(`Unsupported browser: ${browserType}`);
         }
--- a/bciers/apps/dashboard/app/onboarding/page.tsx
+++ b/bciers/apps/dashboard/app/onboarding/page.tsx
@@ -1,6 +1,7 @@
 "use client";
 
 // 🏷 import {named} can be significantly slower than import default
+import { useEffect } from "react";
 import Button from "@mui/material/Button";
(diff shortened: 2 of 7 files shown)


shon-button:
@andrea-williams 

As you saw, the custom configurations of cookies session and callback worked as expected with `SameSite: strict` but, the CSRF cookie remained as `SameSite: lax` . 
**It ~seems** 
Auth.js's internal CSRF handling logic explicitly sets the `SameSite` attribute of the csrfToken cookie to `lax` and does not allow you to override this setting. This ensures the core CSRF protection mechanism functions reliably.
Since `SameSite=Strict` behaviour forces the browser to only send the cookie in two scenarios...
```
- When the user navigates directly in the site (e.g., by clicking a link within the domain)
- When a top-level GET request originates from within your site
```
...then fetching the CSRF token from the getCsrfToken() helper (`/api/auth/csrf`) might not qualify as "same-site navigations" or "top-level GETs"  in the way `SameSite=strict` defines them.
So, trying to enforce `SameSite: "strict"` would likely break the token fetching process and render the app vulnerable to CSRF attacks.


Why we are stuck on `onboarding` after login is again related to the new restrictions of when cookies are sent (detailed above).  Since we need a `top-level GET request` any attempts to redirect the app to dashboard using server-side functionality of `middleware` or `auth.cong\callback\redirect` or `signIn("keycloak", { callbackUrl: "/dashboard" }` do not include the cookie so the app stays on `onboarding`.

To resolve this I have added client side router redirect from onboarding for authenticated users.
image

bciers/libs/e2e/src/utils/helpers.ts
```diff
@@ -383,18 +383,23 @@ export async function setupTestEnvironment(
 ) {
   let browser: Browser | null = null;
 
+  const configs = {
```
andrea-williams, line 386: speeds up E2E tests by approx. 2 seconds.

PR #3356, opened 2025-05-14
chore: fixed supplementary report creation

ayeshmcg:
Card: https://github.com/bcgov/cas-reporting/issues/760
Changes: 
Added check on create supplementary report, if registration purpose different in operation and report operation call create report version otherwise call create supplementary report version

### DIFF
--- a/bc_obps/reporting/tests/api/test_report_supplementary_version_api.py
+++ b/bc_obps/reporting/tests/api/test_report_supplementary_version_api.py
@@ -4,6 +4,7 @@ from model_bakery import baker
 from registration.models import Operation
 from registration.tests.utils.helpers import CommonTestSetup, TestUtils
 from registration.utils import custom_reverse_lazy
+from reporting.models import ReportEmission, ReportProduct
 from reporting.tests.utils.report_access_validation import assert_report_version_ownership_is_validated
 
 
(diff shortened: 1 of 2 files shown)


bc_obps/reporting/tests/api/test_report_supplementary_version_api.py
```diff
@@ -1,74 +1,95 @@
...
         assert response.json() == self.new_report_version.id
 
-        # Assert: Ensure the service method was called with the correct version_id
-        mock_create_report_supplementary_version.assert_called_once_with(self.old_report_version.id)
+        if expected_called_method == "supplementary":
+            mock_create_supplementary.assert_called_once_with(report_version.id)
+            mock_create_report_version.assert_not_called()
+        else:
+            mock_create_report_version.assert_called_once_with(report_version.report)
+            mock_create_supplementary.assert_not_called()
```
pbastia, line 74: Could we get an extra test that uses "real" (model bakery) database instances for report versions, and verify that the new version is created with or without the emission/production records?
It would be a small integration test, and make sure that the behaviour we want happens all the way down

pbastia (approved):
Very nice! Approving, conditional to having that extra test

shon-button (approved):
Super fix! ⚡

bc_obps/reporting/tests/api/test_report_supplementary_version_api.py
```diff
@@ -1,74 +1,134 @@
...
+            custom_reverse_lazy(
+                "create_report_supplementary_version",
+                kwargs={"version_id": self.old_report_version.id},
+            ),
+        )
+        assert response.status_code == 201
+        new_report_version_id = response.json()
+        # Verify that the new report version has no related entries
+        assert not ReportEmission.objects.filter(report_version_id=new_report_version_id).exists()
+        assert not ReportProduct.objects.filter(report_version_id=new_report_version_id).exists()
```
pbastia, line 113: 🥇

PR #3369, opened 2025-05-16
Feat: 750 internal can download attachments

pbastia:


bc_obps/reporting/api/permissions.py
```diff
@@ -65,6 +65,10 @@ def check_version_ownership_in_url(
     version_id_param: str,
 ) -> Callable[[HttpRequest], bool]:
     def validate_func(request: HttpRequest) -> bool:
+        # Internal users can access all report versions
+        if request.current_user.is_irc_user():  # type: ignore
+            return True
+
```
pbastia, line 71: @Sepehr-Sobhani  I ended up doing this instead of duplicating API calls and next functions and more frontend switches depending on the user type and the page

ayeshmcg (approved):
Looks good @pbastia

PR #3373, opened 2025-05-20
Fix: 774 wrong reporting year info

pbastia:


ayeshmcg (approved):
Looks great @pbastia , just need to fix operationReviewFactoryItem

PR #3375, opened 2025-05-20
3268 2 logout redirect

BCerki:
card: https://github.com/bcgov/cas-registration/issues/3268

This card talks about two things:
- logout modal doesn't show
- user is not redirected after logout (Zoey's comment `The "you've been signed out" page doesn't always appear. I left the tab open yesterday on the final review page, and came back today to the tab and it was still on the final review page instead of on the "you've been signed out"`)

These problems are caused by having the app open in more than one tab. The non-active tab logs the user out of all tabs after 30 minutes but only redirects them from the tab the triggered the logout. This makes it seem like the modal never showed up (it did, but it was in the inactive tab). It also makes it seem like the user is still logged in when they're not because they're not redirected to the logout page (only the inactive tab redirected).

We considered polling the token expiration and using that as a proxy for user activity. However, the token's expiry is only updated when `useSession` (client components) or `getSession` (server components) is called. This doesn't happen often enough to be a good indicator of user activity, e.g., you can get through the whole registration workflow without updating expiry. (Sidenote, calling `getToken` in the middleware does not refresh the session--can't refresh from middleware because it runs before cookies are set, and `getToken` is just a getter anyway). 

In the end, we used Broadcast Channel to send updates across the tabs. ([Nextauth already uses broadcast channel](https://leapcell.medium.com/deep-dive-into-nextauth-js-a-powerful-and-flexible-authentication-solution-c9b4ebd37b42) to sign users out of all tabs if they sign out of one.) 

This PR:
- set the throttle value lower
- fetch the `logoutUrl differrently
- set a default `logoutUrl`
- tab handling as part of user activity monitoring
- set up Sentry error capturing specifically for log out
- updated deprecated `callbackUrl` to `redirectTo`
- channel for redirecting extra tabs to the logout page, as Nextauth doesn't seem to handle that
- channel to refresh session in the extra tabs anytime (throttled) there's user activity

To test:
- change maxAge, MODAL_DISPLAY_SECONDS, and ACTIVITY_THROTTLE_SECONDS so you don't have to wait half an hour
- open the app in two tabs. The logout and extend session buttons should affect both apps
- let the timer rundown and you should be logged out and redirected to the keycloak logout page
- to test the fallback, temporarily delete `SITEMINDER_KEYCLOAK_LOGOUT_URL` from your env and you'll be redirected to the onboarding page instead of logout page

Notes:
1. Every once in awhile I'll get "logout failed" on one of the tabs. I ~think this is because I've already been logged out by one of the other tabs
image. 
2. If a user opens the app in one tab, and does nothing, and then opens the app in another tab and does nothing, I've set it up so the countdowns are the same.

### DIFF
--- a/bciers/apps/dashboard/auth/index.ts
+++ b/bciers/apps/dashboard/auth/index.ts
@@ -7,7 +7,7 @@ import authConfig from "./auth.config";
 export const { handlers, auth, signIn, signOut } = NextAuth({
   session: {
     strategy: "jwt",
-    maxAge: 60, // 30 minutes matching Keycloak token expiration time
+    maxAge: 60 * 30, // 30 minutes matching Keycloak token expiration time
   },
   ...authConfig,
 });
--- a/bciers/apps/dashboard/auth/index.ts
+++ b/bciers/apps/dashboard/auth/index.ts
@@ -7,7 +7,7 @@ import authConfig from "./auth.config";
 export const { handlers, auth, signIn, signOut } = NextAuth({
   session: {
     strategy: "jwt",
-    maxAge: 30 * 60, // 30 minutes matching Keycloak token expiration time
+    maxAge: 60, // 30 minutes matching Keycloak token expiration time
   },
   ...authConfig,
 });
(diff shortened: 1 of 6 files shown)


andrea-williams (approved):
Awesome work Brianna!

I tested it in my local (decreasing the timeout to 2 minutes) with 2 open tabs. I noticed there was a slight discrepancy in the countdown timer in the modal between the 2 tabs (approx. 8 seconds difference), but I don't think that needs to be addressed now. This is a huge improvement!

Issue #3286, opened 2025-05-01
Admin/Facilities: write migration that copies SFO BCGHG IDs over to their associated facilities


andrea-williams:
Blocked by: https://github.com/orgs/bcgov/projects/123/views/16?pane=issue&itemId=108763205&issue=bcgov%7Ccas-registration%7C3287 (we'd have to run this migration multiple times if someone adds a new facility before ^ goes in)

#### Describe the task

Based on a CAS user request, we need a Django migration that applies the BCGHG ID (where it exists) for any and all SFOs to their respective facility.

#### Acceptance Criteria

- [x] all SFOs in our prod database that already have a BCGHG ID should have the same BCGHG ID applied to their associated facility, too
- [x] all LFOs in our prod database should **NOT** have their BCGHG ID copied over to their facilities
- [x] *NEW*: Add the BCGHGID field to the Facility Information page so that it is visible to users (*will need a wireframe update)

#### Definition of Done

- [ ] first
- [ ] second
- [ ] third

#### Additional context

- Add any other context about the task here.
- Or here

patriciarussellCAS:
Note: I added an AC to include the BCGHGID field on the facility info page. Even though it's the same number as what we see on the operation info page, it will be helpful for external and internal users to see it on this page as well. BUT if this blocks or slows down this work we could consider adding this field in later! Devs, let me know. @andrea-williams @BCerki

BCerki:
@patriciarussellCAS , now that we've added the BCGHG ID to the facility page, cas_directors will be able to assign BCGHG IDs to LFOs by clicking the button, same as they can for operations. For SFOs, there will be no button because the ID should be assigned from the operation form. Is that correct?

patriciarussellCAS:
I think so - if I'm understanding you correctly. 
LFOs - can be assigned a BCGHGID on the operation form and now on each facility form
SFOs - can be assigned a BCGHGID on the operation form only. 
yes?

fviduya:
I have tested this in the DEV environment, [here is the test document for reference](https://buttoninc-my.sharepoint.com/:w:/g/personal/francesca_button_is/EaMfCYXwHFhJs6LI6WO0AF8B5ZqS6vlqCbi4Vz75AmZK9A?e=prGYeM).

I can verify that:
- [x] The BCGHG ID is visible in the Facilities grid
- [x] `Issue BCGHG ID` button is available only for the following Facility Types: Small Aggregate, Medium Facility, Large Facility
[...]

PR #3376, opened 2025-05-20
chore: assign SFO operation BCGHGID to facility too

BCerki:
card: https://github.com/orgs/bcgov/projects/123/views/16?pane=issue&itemId=108763205&issue=bcgov%7Ccas-registration%7C3287

This PR:
- Introduces tests in TestGenerateBCGHGId to validate user access and proper BCGHG ID assignment for SFO and non-assignment for LFO.
- Wraps the generate_bcghg_id method in a transactional block and updates the facility’s BCGHG ID for SFO operations.
- Updates mock data so SFO facilities have the same BCGHG IDs as their operations. Also made Alien a draft because it can't be registered if it doesn't have a facility

Note that we don't show BCGHG ID on the SFO form (will be added in https://github.com/bcgov/cas-registration/issues/3286), so to test this, you'll have to temporarily add it, or check in the db.

### DIFF
--- a/bc_obps/registration/fixtures/mock/bc_greenhouse_gas_id.json
+++ b/bc_obps/registration/fixtures/mock/bc_greenhouse_gas_id.json
@@ -26,14 +26,6 @@
   {
     "model": "registration.bcgreenhousegasid",
     "fields": {
-      "id": "23219990002",
-      "issued_at": "2023-10-13T15:27:00.000Z",
-      "comments": "For Operation 4"
-    }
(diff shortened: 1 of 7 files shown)


andrea-williams (approved):
Nothing was working locally, then I took a break, came back and it all worked fine even though I changed nothing. 🤷‍♀️ 

BCGHG ID appears for SFOs whether they are brand new or modified from existing LFO in mock fixtures.

PR #3383, opened 2025-05-22
chore: Added default page for activity not implemented

ayeshmcg:
Card: https://github.com/bcgov/cas-reporting/issues/754
Changes: 
1. Implemented logic to return a fallback schema when no valid ActivityJsonSchema is found for the specified activity_id.
2. Updated Activity form to check for fallbackSchema, if exist then update the form according and do not  mak a post request.
3. Updated tests

To test:-
1. Go to http://localhost:3000/reporting/reports
2. Start a report and select activities like Electronics manufacturing, Phosphorous acid Production or other activities which havent been implemented yet on http://localhost:3000/reporting/reports/1/review-operation-information. 
3. On activities page select this activities and you will see page like 
image

### DIFF
--- a/bciers/apps/reporting/src/app/components/activities/ActivityForm.tsx
+++ b/bciers/apps/reporting/src/app/components/activities/ActivityForm.tsx
@@ -126,8 +126,8 @@ export default function ActivityForm({
   // 🛠️ Function to submit user form data to API
   const submitHandler = async (e: IChangeEvent) => {
     if (isFallbackSchema) {
-      console.log("Fallback schema, skipping submit");
-      return true; // Exit early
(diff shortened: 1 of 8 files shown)


bciers/apps/reporting/src/app/components/activities/ActivityForm.tsx
```diff
@@ -182,6 +187,8 @@ export default function ActivityForm({
 
     return false;
   };
+  console.log("initialJsonSchema", initialJsonSchema);
+  console.log("uischema", getUiSchema(currentActivity.slug));
```
acatchpole, line 191: leftover log statements?

PR #3388, opened 2025-05-23
3268 2 logout redirect

BCerki:
card: #3268 
In the local environment, the previous fix for 3268 was working fine, but in the dev environment, we saw problems with nextauth's `update()` failing (maybe because of race conditions when it tried to update across multiple tabs?). In this PR, instead of using `update()` to refresh on user activity, we're using `getSession()`, which also refreshes the session. I also put in a sentry log so that if update fails in the other place it's used, we get more info.

If this doesn't work either, some other things to consider:
- https://next-auth.js.org/getting-started/client - session provider polling (unrelated to activity)
- if there's trouble with the stay logged in button, use getSession in the other tabs
- https://teams.microsoft.com/l/message/19:81544b60877c4a5eb4fc7946ecbcb525@thread.tacv2/1747948485010?tenantId=6fdb5200-3d0d-4a8a-b036-d3685e359adc&groupId=29821293-3244-455d-beae-e01d27762cfe&parentMessageId=1747948485010&teamName=External%3A%20CAS%20-%20Clean%20Growth%20Digital%20Service%20Team&channelName=Developers&createdTime=1747948485010&ngc=true


shon-button:
@BCerki 

Just to re-iterate here:

1. Fantastic efforts on a difficult and multi-faceted issue
2. I strongly suggest we re-think the customization code used for logout management and leverage auth.js internals as much as possible...the auth.js related code in this "poc", "draft" SessionProvider is worth a look, I think - found here: https://github.com/bcgov/cas-registration/pull/3389/files#diff-05ad09897140efba00671a4c92fe17b500ef9877ded80999c99b936917de8cd3

BCerki:
Completed in #3390

PR #3392, opened 2025-05-26
chore: fix regulated product typo

ayeshmcg:
Card: https://github.com/bcgov/cas-reporting/issues/783
Fix typo

### DIFF
--- a/bc_obps/registration/migrations/0115_fix_typo_in_regulated_product_name.py
+++ b/bc_obps/registration/migrations/0115_fix_typo_in_regulated_product_name.py
@@ -16,5 +16,9 @@ class Migration(migrations.Migration):
     ]
 
     operations = [
-        migrations.RunPython(fix_typo),
+        migrations.RunPython(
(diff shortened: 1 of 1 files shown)


bc_obps/registration/migrations/0115_fix_typo_in_regulated_product_name.py
```diff
@@ -0,0 +1,20 @@
...
+    )
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ('registration', '0114_V3_2_1'),
+    ]
+
+    operations = [
+        migrations.RunPython(fix_typo),
```
BCerki, line 19: In reg we often include reverse_code=migrations.RunPython.noop and elidable=True with these kinds of forward-only migrations, e.g. /home/briannacerkiewicz/cas-registration/bc_obps/registration/migrations/0095_populate_operation_timeline_with_prod_data.py

PR #3402, opened 2025-05-27
chore: fixed compliance data bug for zero production

ayeshmcg:
Card: https://github.com/bcgov/cas-reporting/issues/786
Changes: added check if annual production is zero then set allocated_for_compliance_2024=0


joshgamache (approved):
Tests passing and code's clean. Poked around the frontend where this is called and no errors being thrown. Nice work!

Issue #782, opened 2024-01-30
Dashboards - MUI Datagrid serverside pagination


marcellmueller:
#### Describe the task

Currently our Dashboards fetch all of the available data we have in the database which can be a very large and slow request when there is a lot of records in the database. We need to enable server side pagination and just request the data as we change pages. Not sure if this should be split up per dashboard but hopefully it's not too much work. There is some info in the MUI docs as it is supported by the library:

https://mui.com/x/react-data-grid/pagination/

#### Acceptance Criteria

- [ ] Data is fetched in between pages
- [ ] Operator dashboard 
- [ ] Operation dashboard

#### Additional context

- Add any other context about the task here.
- Or here

marcellmueller:
Moving this to spring backlog and putting high priority label on it since it will help mitigate performance issues found during load testing.

PR #3403, opened 2025-05-27
Fix/reporting routes/780 782

andrea-williams:
Cards: #780, #782

The problem: `pages` in `navigationInformation.ts` was a deep copy, so if a page was removed from navigation, it would then not get added back in. 
The fix: make a shallow copy of `pages` instead.

To test https://github.com/bcgov/cas-reporting/issues/780:
- create an operation with registration purpose "Reporting Operation", and register it
- create a new report for your operation, and first set the operation's total emissions to below 25k. Navigate all the way through to the Sign Off & Submit section, but don't submit yet. The "Verification" page should not appear because emissions < 25k.
- navigate back to the emissions reporting section, and edit the emissions to be above 25k. Then click through all the way to Sign Off & Submit
- you should now see the "Verification" page because total emissions exceed 25k.

To test https://github.com/bcgov/cas-reporting/issues/782:
(this was only found to be reproducible in DEV, not local)
- create a supplementary report from a submitted annual report
- make arbitrary change to the report's data
- navigate through to Sign Off & Submit section, confirm that "Change Review" page appears

### DIFF
--- a/bciers/apps/reporting/src/app/components/changeReview/ChangeReviewPage.tsx
+++ b/bciers/apps/reporting/src/app/components/changeReview/ChangeReviewPage.tsx
@@ -32,8 +32,6 @@ export default async function ChangeReviewPage({
       skipChangeReview: !isSupplementaryReport,
     },
   );
-  console.log("skipVerificationPage:", !showVerificationPage);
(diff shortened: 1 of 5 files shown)


acatchpole (approved):
working as advertised

Issue #3379, opened 2025-05-21
CRA business number field has commas


BCerki:
#### Describe the Bug:

The CRA business number field has commas in it in edit mode: 

![Image](https://github.com/user-attachments/assets/a43450ef-3654-4a4c-99da-e29e1f2f758d)

#### Probability:

(How likely the bug is to happen, scored from 1-5. For example, probability of 5 is something like "it happens to all users every time they log in." whereas probability of 1 "only happens to certain users when a really specific and unlikely path is followed.")

#### Impact:

(How bad the bug is when it does happen, scored from 1-5. For example, effect of 5 is "the entire app crashes and makes it unusable for all users" or "the bug causes the wrong data to be saved, with critical information (e.g. payment) being affected." whereas effect of 1 is "It makes some styling look a little bit weird.")

<img src="https://www.getzephyr.com/sites/default/files/riksk%20matrix.jpg" alt="Bug Matrix" width="250" height="250"/>

#### Steps to reproduce the behaviour:

1. Step #1
2. Step #2
3. Step #3

#### Additional information:

Browser: [e.g. chrome, safari]
Version: [e.g. 22]

#### Additional Context

-

patriciarussellCAS:
@BCerki added a link to the BCIERS internal channel as we now have some screenshots from the user.

BCerki:
The commas don't preventing submission, so I've skipped that part of the bug for now. To fix the commas properly (and the other number-styling related bugs, hover doesn't work) we could use https://s-yadav.github.io/react-number-format/docs/numeric_format like we did in CIF.

pbastia:
@BCerki there are a couple of other options:
- configure the `<NumberField.Root format={...}` in our TextWidget with the proper parameters to display the number as we want it
- fix it properly: make the database model a string, with a regex validation, since the CRA business number is not an integer in that sense, but a string of digits

patriciarussellCAS:
Working in TEST

PR #3405, opened 2025-05-28
3379 CRA business number bug

BCerki:
card: https://github.com/bcgov/cas-registration/issues/3379

This PR:
- removes mouse events from the TextWidget, which were causing validation errors that blocked submission of the operator form (not entirely sure why, copilot said it could have something to do with the mouse events interfering with the onChange)
- my changes break the hover styling. The NumberField doesn't accept an sx prop and styles doesn't accept hover, which is why the mouse events were in the widget in the first place. I think the proper way to fix this (and the other number-styling related bugs) is to use https://s-yadav.github.io/react-number-format/docs/numeric_format like we did in CIF. You can see that the field outline doesn't change on hover on the Operator Form in admin and the activity forms in reporting.
- I think it's worth breaking the hover styling to make the form submittable again. @patriciarussellCAS , is that OK with you? The hover styling is fixable, but it'll take a little while, and I have other high priority things to do.


BCerki:
This was fixed by https://github.com/bcgov/cas-registration/pull/3408, closing

BCerki:
No longer needed since #3408

PR #3408, opened 2025-05-28
Hotfix: add SessionProvider

BCerki:
card: https://github.com/orgs/bcgov/projects/123/views/16?pane=issue&itemId=112797707&issue=bcgov%7Ccas-reporting%7C792

This PR:
- `useSession` causes a re-render every time it's called, so if someone is typing in a form and it gets called (as it often does, every 30s due to the `handleActivity` function in `SessionTimeoutHandler`), their form data gets wiped. We made a new hook to replace useSession because most of the time we don't care about the whole sesssion, we just need to know the role so that we can conditionally display things to internal vs. external users
- replaced useSession with new hook where possible
- replaces useSession with getSession where the new hook wasn't appropriate
- left useSession on pages that are fine/desired to re-render
- updated vitests
- commented out the sessionUtils vitests for later fixing in https://github.com/bcgov/cas-reporting/issues/795


bciers/libs/utils/src/sessionRoleContext.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+export const SessionRoleContext = createContext(undefined);
+
+const SessionRoleContextProvider = (props: any) => {
+  return (
+    <SessionRoleContext.Provider value={props.value}>
+      {props.children}
+    </SessionRoleContext.Provider>
+  );
+};
```
pbastia, line 19: we can probably type this a bit better:
```suggestion

export const SessionRoleContext = createContext<string | undefined>(undefined);

interface SessionRoleContextProviderProps {
  value: string | undefined;
  children: React.ReactNode;
}

const SessionRoleContextProvider: React.FC<SessionRoleContextProviderProps> = (props) => {
  return (
    <SessionRoleContext.Provider value={props.value}>
      {props.children}
    </SessionRoleContext.Provider>
  );
};
```

pbastia (approved):
Nice!
- one console log
- some typing that can be improved

PR #3409, opened 2025-05-29
fix: disable sticky sessions in k8s routes to even out load among pods

dleard:
Our routes were configured to send all requests to one pod in the service's endpoints list, making our HPA pointless. 
Disabling sticky sessions will allow our balancing method to come into effect and even out the load across all pods in the service's endpoints.

### DIFF
--- a/helm/cas-bciers/templates/administration/route.yaml
+++ b/helm/cas-bciers/templates/administration/route.yaml
@@ -20,7 +20,7 @@ metadata:
     haproxy.router.openshift.io/rewrite-target: {{ .Values.administrationFrontend.route.path }}
     haproxy.router.openshift.io/disable_cookies: 'true'
     "helm.sh/hook": "pre-install,pre-upgrade"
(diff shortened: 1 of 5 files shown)


helm/cas-bciers/templates/administration/route.yaml
```diff
@@ -18,6 +18,9 @@ metadata:
   annotations:
     haproxy.router.openshift.io/balance: roundrobin
     haproxy.router.openshift.io/rewrite-target: {{ .Values.administrationFrontend.route.path }}
+    haproxy.router.openshift.io/disable_cookies: 'true'
+    "helm.sh/hook": "pre-install,pre-upgrade"
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
```
pbastia, line 23:
```suggestion
    "helm.sh/hook-delete-policy": before-hook-creation
```

pbastia (approved):

PR #3412, opened 2025-05-29
sessionUtils vitests

BCerki:
card: https://github.com/bcgov/cas-reporting/issues/795

This PR:
- updates the vitests for sessionUtils, tech debt from #3408

### DIFF
--- a/bciers/libs/testConfig/src/global.tsx
+++ b/bciers/libs/testConfig/src/global.tsx
@@ -28,6 +28,7 @@ import {
   useSessionRole,
   getSessionRole,
   getSession,
+  useContext,
 } from "./mocks";
 import createFetchMock from "vitest-fetch-mock";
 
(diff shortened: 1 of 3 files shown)


bciers/libs/utils/src/sessionUtils.test.ts
```diff
@@ -1,40 +1,37 @@
-/* eslint-disable */
-import { auth, useSession } from "@bciers/testConfig/mocks";
+import { auth, useContext } from "@bciers/testConfig/mocks";
+// we need both unmocks or else the sessionUtils will be spies instead of the actual functions
```
BCerki, line 2: I don't know why; we're already importing the real function, not the mock

bciers/libs/testConfig/src/global.tsx
```diff
@@ -40,6 +41,14 @@ declare module "vitest" {
...
 expect.extend(matchers);
 
+vi.mock("react", async () => {
+  const actualReact = await vi.importActual<typeof import("react")>("react");
+  return {
+    ...actualReact,
+    useContext: useContext,
+  };
+});
+
```
andrea-williams, line 51: I would never have known that this would be necessary in order to use the `useContext` mock. Great solve!

andrea-williams (approved):

Issue #2311, opened 2024-10-01
Dashboard\e2e - BCIERS (zone) dashboards


shon-button:
Blocked by:
https://github.com/bcgov/cas-registration/issues/2341
https://github.com/bcgov/cas-registration/issues/2048

#### Description:

Tests BCIERS Dashboards by zone:
- Dashboard
- Administration
- COM
- Registration
- Reporting

#### Development Steps

- [ ] Suggesting, create `bciers/apps/dashboard-e2e/tests/dashboards.spec.ts` and add code to loop zones and take selfie (see example code below).
- [ ] Perhaps leverage  `setupTestEnvironment` for setting up the fixtures to reduce SQL DB CRUD dependencies
- [ ] Loop all available roles
- [ ] Take happios
- [ ] Assert accessibly 
- [ ] Add tests for conditional tiles, see:
https://github.com/bcgov/cas-registration/issues/2341
https://github.com/bcgov/cas-registration/issues/2048

Possible starting point  `bciers/apps/dashboard-e2e/tests/dashboards.spec.ts` 

```
import { test } from "@playwright/test";
// import { analyzeAccessibility } from "@/e2e/utils/helpers";
// ☰ Enums
import { AppRole, UserRole, UserOperatorStatus } from "@/e2e/utils/enums";
// 🥞 DB CRUD
import {
  deleteUserOperatorRecord,
  upsertUserRecord,
  upsertOperatorRecord,
  upsertUserOperatorRecord,
} from "@/e2e/utils/queries";
// ℹ️ Environment variables
import * as dotenv from "dotenv";
dotenv.config({ path: "./e2e/.env.local" });

const happoPlaywright = require("happo-playwright");
test.beforeAll(async () => {
  try {
    // Scenario FrontEndRoles.INDUSTRY_USER_ADMIN where UserOperatorStatus.APPROVED && OperatorStatus.APPROVED;
    // Upsert a User record: bc-cas-dev
    await upsertUserRecord(UserRole.INDUSTRY_USER_ADMIN);
    await upsertUserRecord(UserRole.CAS_ADMIN);
    // Upsert an Operator record, using default values
    await upsertOperatorRecord();
    // Upsert an User Operator record: industry_user_admin, operator id 2
    await upsertUserOperatorRecord(
      process.env.E2E_INDUSTRY_USER_ADMIN_GUID as string,
      AppRole.ADMIN,
      UserOperatorStatus.APPROVED,
    );
[...]

PR #3432, opened 2025-06-03
chore: dashboard e2e test

fviduya:
For #2311 

Dashboard E2E tests for both internal and external user.
- Specifically chose cas_analyst and cas_admin for internal since the only difference in the dashboard is the Transfer operation or facility link
- For external, updated the reporter role through DB instead. Open to suggestions on how this should be handled!


BCerki (commented):
`For external, updated the reporter role through DB instead. Open to suggestions on how this should be handled!`--updating through the db is fine for this test, I think;we've got role-setting via the app on the to-do list for its own test. However, I got errors on the reporter test locally: image 

Re the CI fail, we used to have a dashboard test in reg1 (https://github.com/bcgov/cas-registration/pull/2864), so maybe there's something in there that might give us a hint? I thought we might need to add the rest of the apps to the `needs` array because we go into each app's dashboard, and I've pushed a commit that does that, we'll see if it helps.

.github/workflows/main.yaml
```diff
@@ -86,6 +86,24 @@ jobs:
...
+        build-reporting,
+        check-nx-affected,
+      ]
+    uses: ./.github/workflows/test-nx-project-e2e.yaml
+    with:
+      docker_compose_file: "docker-compose-bciers-apps.yaml"
+      image_url: "ghcr.io/bcgov/cas-dash-frontend"
+      is_nx_affected: ${{contains(needs.check-nx-affected.outputs.NX_AFFECTED_E2E, 'dashboard') || github.ref_name == 'develop' || github.ref_name == 'main' || github.head_ref == 'develop' || github.head_ref == 'main'}}
+      nx_project: dashboard
+    secrets: inherit
```
fviduya, line 106: Looks like what's missing was the `docker_compose_file`. The code was from here: https://github.com/bcgov/cas-registration/pull/2852/files

BCerki: Nice work!

bciers/apps/dashboard-e2e/tests/external/dashboard.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+      // 🛸 Navigate to dashboard page
+      const dashboardPage = new DashboardPOM(page);
+      await dashboardPage.route();
+      let component = "";
+
+      // Say cheese!
+      component = `Main Dashboard for ${role}`;
+      await takeStabilizedScreenshot(happoPlaywright, page, {
+        component: component,
+        variant: "filled",
```
BCerki, line 54: I don't know that variant filled makes sense for a dashboard (I think it was originally for forms because in reg1 we took shots of the forms when they were filled vs. empty vs. had validation errors). I think variant is optional, so we could just drop it

BCerki (commented):
I'm still having trouble running a couple of the tests locally: 
image

bciers/apps/dashboard-e2e/tests/external/dashboard.spec.ts
fviduya: I set the value to `default` instead since the variant filled is required, or I can change the variant field to be optional if that's better.

bciers/apps/dashboard-e2e/tests/external/dashboard.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+  "INDUSTRY_USER",
+  "INDUSTRY_USER_ADMIN",
+  "INDUSTRY_USER_REPORTER",
+];
+
+test.describe.configure({ mode: "serial" });
+userRoles.forEach((role) => {
+  test.describe("External user dashboard", () => {
+    const roleTest = setupBeforeAllTest(UserRole.INDUSTRY_USER_ADMIN);
+    roleTest(`Dashboard for role: ${role}`, async ({ page }) => {
```
andrea-williams, line 31: I think some comments here explaining what's going on would be useful. I find it confusing to determine why `roleTest` is needed when we're also testing against each entry in `userRoles`.

andrea-williams (approved):
There's one piece of code that I think would benefit from some in-line documentation, but otherwise LGTM! And everything runs and passes in my local.

bciers/apps/dashboard-e2e/tests/external/dashboard.spec.ts
fviduya: Ooh thanks for the feedback @andrea-williams! Looking into this again I realized we don't need roleTest anymore and can simply just call `setupBeforeAllTest` outside and loop through each value of userRoles. The main reason why I had roleTest was for setting up the specific role, but since on this one I'm actually just using industry_user_admin and changing the role in the backend, the code can be simplified.

PR #3440, opened 2025-06-05
chore:fix final review page

ayeshmcg:
Card: https://github.com/bcgov/cas-reporting/issues/535

### DIFF
--- a/bc_obps/reporting/api/report_final_review.py
+++ b/bc_obps/reporting/api/report_final_review.py
@@ -1,0 +1,40 @@
+from django.http import HttpRequest
+from ninja.responses import Response
+from django.db.models import Prefetch
+from reporting.constants import EMISSIONS_REPORT_TAGS
+from reporting.schema.generic import Message
+from ..models import ReportVersion, FacilityReport, ReportRawActivityData
+from ..schema.report_final_review import ReportVersionSchema
+from .router import router
+
+
+@router.get(
+    "/report-version/{version_id}/final-review",
+    response={200: ReportVersionSchema, 404: Message},
+    tags=EMISSIONS_REPORT_TAGS,
+    description="Fetch final review data for a given report version ID.",
+)
+def get_report_final_review_data(request: HttpRequest, version_id: int):
+    try:
+        report = (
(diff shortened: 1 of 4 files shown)


bc_obps/reporting/api/report_final_review.py
```diff
@@ -0,0 +1,84 @@
...
+@router.get(
+    "/report-version/{version_id}/final-review",
+    response={200: ReportVersionSchema, custom_codes_4xx: Message},
+    tags=EMISSIONS_REPORT_TAGS,
+    description="Fetch final review data for a given report version ID.",
+    auth=approved_industry_user_report_version_composite_auth,
+)
+def get_report_final_review_data(request: HttpRequest, version_id: int) -> Response:
+    try:
+        report = (
```
pbastia, line 32: should be `report_version`

bc_obps/reporting/api/report_final_review.py
```diff
@@ -0,0 +1,84 @@
...
+        return Response({"message": "ReportVersion not found"}, status=404)
+
+    for facility in report.facility_reports.all():
+        facility.emission_summary = EmissionCategoryService.get_facility_emission_summary_form_data(facility.id)  # type: ignore
+
+    compliance_summary = ComplianceService.get_calculated_compliance_data(version_id)
+
+    serialized_report = ReportVersionSchema.from_orm(report)
+    data = serialized_report.dict()
+    data["report_compliance_summary"] = compliance_summary
```
pbastia, line 82: For consistency I would put all these in schemas in resolvers too.
Django-ninja is a middleware between the HTTP request and the API layer, and here by using this `ReportVersionSchema.from_orm` we're using it as a service instead - making it essentially run twice in different ways.

- the `facility_report.emission_summary` could be added with a resolver on the FacilityReport Schema
- the `report_compliance_summary` field could be also be added with a resolver on the `ReportVersionSchema` directly

This way no need to do extra work manipulating the existing records, adding extra fields, or adding to the resulting json data

ayeshmcg:
@pbastia  This has been updated

pbastia (approved):
🥇

PR #3453, opened 2025-06-06
feat: middleware to allow keycloak token rotation

pbastia:


bciers/apps/dashboard/middlewares/withTokenRefresh.ts
```diff
@@ -0,0 +1,95 @@
...
+  return async (request: NextRequest) => {
+    // Casting to our augmented JWT type
+    const jwt = (await getToken({
+      req: request,
+      secret: process.env.NEXTAUTH_SECRET,
+    })) as JWT | null;
+
+    const response = NextResponse.next();
+
+    if (!jwt) return response;
```
pbastia, line 60: @shon-button  maybe here we could redirect to the `/onboarding` landing page instead

bciers/apps/dashboard/middlewares/withTokenRefresh.ts
shon-button, line 1: @pbastia 
Perhaps this logic should be within the `auth.config` instead of in a routing middleware so that the logic would also apply on non-routing events such as calls to the API?

bciers/apps/dashboard/middlewares/withTokenRefresh.ts
shon-button: Perhaps more defensive if we do something like:


``` if (!jwt) {
  const response = NextResponse.redirect(new URL('/onboarding', request.url));

  // Get all cookies from the incoming request
  const cookies = request.cookies.getAll();

  // Identify and clear all session-related cookies
  cookies.forEach(({ name }) => {
    if (name.startsWith(SESSION_COOKIE)) {
      response.cookies.set(name, "", {
        maxAge: 0,
        path: "/",
      });
    }
  });

  return response;
}
```

bciers/apps/dashboard/middlewares/withTokenRefresh.ts
```diff
@@ -0,0 +1,109 @@
...
+      client_secret: `${process.env.KEYCLOAK_CLIENT_SECRET}`,
+      grant_type: "refresh_token",
+      refresh_token: `${refreshToken}`,
+    }),
+  });
+
+  if (!newTokensResponse.ok)
+    throw new Error("Error refreshing KeyCloak access token");
+
+  const newTokens = await newTokensResponse.json();
```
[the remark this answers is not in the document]
pbastia: The response contains both new refresh and access tokens, the plural makes more sense to me!

bciers/apps/dashboard/middlewares/withTokenRefresh.ts
```diff
@@ -0,0 +1,118 @@
...
+      // split the string into an array of strings
+      const tokenChunks = newSessionToken.match(regex);
+
+      if (tokenChunks) {
+        tokenChunks.forEach((tokenChunk, index) => {
+          response.cookies.set(`${SESSION_COOKIE}.${index}`, tokenChunk, {
+            httpOnly: true,
+            maxAge: 30 * 60, // 30 min to match KC
+            secure: SESSION_SECURE,
+            sameSite: "strict",
```
pbastia, line 110: @andrea-williams  am I going to run into trouble here?

andrea-williams: hmmm possibly? the test will be whether you're able to refresh your token again after your `tokenChunks` expire. And if you haven't already, I would recommend checking that the cookies are actually being set with sameSite=strict - next-auth likes to override what you specify most of the time in favour of their own (Lax) default.

bciers/apps/dashboard/middlewares/withTokenRefresh.ts
```diff
@@ -0,0 +1,118 @@
...
+
+  return {
+    access_token: newTokens.access_token,
+    refresh_token: newTokens.refresh_token,
+    expires_at:
+      Math.round(Date.now() / 1000) + OAUTH_TOKEN_ROTATION_INTERVAL_SECONDS,
+  };
+}
+
+function signOut(request: NextRequest) {
```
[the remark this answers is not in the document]
pbastia: Oooh great call! I guess we don't sign out of Keycloak, and we'd want to try to sync Next auth and Keycloak at that point.

bciers/apps/dashboard/middleware.ts
```diff
@@ -24,8 +25,12 @@ Conditional statements
 */
 
 export const config = {
-  matcher: ["/((?!static|.*\\..*|api|_next|sw.js|favicon.ico|bciers/libs/).*)"],
+  matcher: ["/((?!static|.*\\..*|_next|sw.js|favicon.ico|bciers/libs/).*)"],
```
shon-button, line 28: @pbastia 

I think the dashboard middleware matcher can stay the same so to allow unauthenticated calls to our authentication api routes (or add the api route to `withAuthorizationDashboard\unauthAllowedPaths` 

For all other apps, I think the middleware matcher should remove the exception match on api routes so that other app api routes have the authentication logic applied from `bciers/libs/middlewares/src/withAuthorization.ts`

bciers/apps/dashboard/middlewares/withTokenRefresh.ts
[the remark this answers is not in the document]
shon-button: @pbastia 

Perhaps useful: `bciers/node_modules/next-auth/src/react.tsx\export async function signOut<R extends boolean = true>(`

bciers/apps/dashboard/middlewares/withTokenRefresh.ts
```diff
@@ -0,0 +1,111 @@
...
+}
+
+export const withTokenRefreshMiddleware: MiddlewareFactory = () => {
+  /**
+   * Middleware that supplements next-auth to add access token rotation.
+   * Huge thanks to the community: https://github.com/nextauthjs/next-auth/discussions/9715#discussioncomment-12818495
+   */
+  return async (request: NextRequest) => {
+    // Casting to our augmented JWT type
+    const jwt = (await getToken({
```
shon-button, line 65: @pbastia 

Is there a reason not to use the `bciers/apps/dashboard/app/api/auth/token/route.ts` for `getToken()`?

bciers/apps/dashboard/middlewares/withTokenRefresh.ts
```diff
@@ -0,0 +1,111 @@
...
+      req: request,
+      secret: process.env.NEXTAUTH_SECRET,
+    })) as JWT | null;
+
+    const response = NextResponse.next();
+    if (request.nextUrl.pathname === "/auth/logout" || !jwt) return response;
```
shon-button, line 71: @pbastia 

If jwt is null (i.e. the token fetch fails), then NextResponse.next() is returned, which lets the request go through unblocked to the requested route — even if the user is not authenticated. 

Although the other middlewares should also check the token using getToken(); perhaps, we might ensure sign out if jwt is null?

Something like?

```

import { getToken } from "@bciers/actions";
import isInAllowedPath from "@bciers/utils/src/isInAllowedList";

export const withTokenRefreshMiddleware: MiddlewareFactory = () => {
  return async (request: NextRequest) => {
    const response = NextResponse.next();
    const { pathname } = request.nextUrl;
    const unauthAllowedPaths = ["auth", "onboarding"];
    if (isInAllowedPath(pathname, unauthAllowedPaths)) return response;

    // Casting to our augmented JWT type
    const jwt = await getToken() as JWT | null;
 
    if (!jwt) return signOut(request);

```

shon-button (approved):
Thanks for all your effort on this tricky implementation.
App security is now much improved! 🔐 💫

Added two small suggestions for your consideration 🙂

PR #3458, opened 2025-06-09
chore: update selectors to avoid backend collision

joshgamache:
Fixes configuration of new file check deployment to be selectable independent from the backend. Adds additional network selectors so that the pod can properly access postgres.


dleard (approved):
awesome, thanks for the quick fix! :shipit:

PR #3459, opened 2025-06-09
3455 e2e test: issue boro and bcghg id

fviduya:
Card: #3445 

To test locally: 
- Update your .env.local file to get the storage state for cas_director (OBPS FE E2E STORAGESTATES AUTH v5)
- Comment line 20 in the test file and uncomment line 21

Few changes:
- I've updated the naming for dashboard happo screenshots so it doesn't get overwritten and each one gets a screenshot
- I've added my IDIR in user.json, the test fails without it since we're using my IDIR in the storage state
- I've registered Brine LFO for this test - please let me know if I should just create a new one!

### To do before merging:
- [x] Remove the updates made in 
   - [x] test-nx-project-e2e.yaml, 
   - [x] enums.ts, and 
   - [x] test file regarding use of CAS_DIRECTOR_1
- [x] Remove changes made in operationColumns.ts

### DIFF
--- a/bciers/apps/administration/app/components/datagrid/models/operations/operationColumns.ts
+++ b/bciers/apps/administration/app/components/datagrid/models/operations/operationColumns.ts
@@ -32,7 +32,7 @@ const operationColumns = (
           case !!boroId:
             return boroId; // Return 'boroId' if it exists
           case regulatedOperationPurposes.includes(purpose):
-            return "Pendings"; // Return "Pending" if 'purpose' is in the list
+            return "Pending"; // Return "Pending" if 'purpose' is in the list
           default:
             return "N/A"; // Default case
         }


bciers/apps/administration-e2e/tests/workflows/issue-boro-bcghg-id.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+  analyzeAccessibility,
+  takeStabilizedScreenshot,
+  assertSuccessfulSnackbar,
+  clickButton,
+  urlIsCorrect,
+} from "@bciers/e2e/utils/helpers";
+
+const happoPlaywright = require("happo-playwright");
+
+const test = setupBeforeAllTest(UserRole.CAS_DIRECTORS);
```
fviduya, line 20: The CAS Director secret is being used in the dashboard test file, so in order to not disrupt other PRs and merges to develop, I've opted to create a mock secret and use that instead. To test this locally, comment line 20 and uncomment line 21.

BCerki (approved):
Looking good!

I didn't entirely understand what was going on with the mock secret. We'll be using CAS_DIRECTOR (not DIRECTORS) for all the e2e tests, right?

Re `I've added my IDIR in user.json, the test fails without it since we're using my IDIR in the storage state`, I think we've tried to avoid putting our real guids in the mock data (though there may be some in there currently). I'd suggest asking about it in the developers channel before merging.

Nice work!

Issue #3262, opened 2025-01-02
RLS Policies - Registration app


dleard:
work started in: https://github.com/bcgov/cas-registration/pull/3330
blocked by bcgov/cas-reporting#486 
#### Description:

Add all the necessary RLS policies for the models in the registration app. ~Use the example in the POC PR https://github.com/bcgov/cas-registration/pull/2558 as a guide on how to apply policies.~ There's a PR in-progress here: https://github.com/bcgov/cas-registration/pull/3330

#### Development Checklist:

- [ ] All models that require RLS policies have policies applied using the RLS app classes
- [ ] All models with RLS policies have tests that confirm the policy logic

#### Definition of Ready (Note: If any of these points are not applicable, mark N/A)

- [ ] User story is included
- [ ] User role and type are identified
- [ ] Acceptance criteria are included
- [ ] Wireframes are included (if required)
- [ ] Design / Solution is accepted by Product Owner
- [ ] Dependencies are identified (technical, business, regulatory/policy)
- [ ] Story has been estimated (under 13 pts)

#### Definition of Done (Note: If any of these points are not applicable, mark N/A)

- [ ] Acceptance criteria are tested by the CI pipeline
- [ ] UI meets accessibility requirements
- [ ] Configuration changes are documented, documentation and designs are updated
- [ ] Passes code peer-review
- [ ] Passes QA of Acceptance Criteria with verification in Dev and Test
- [ ] Ticket is ready to be merged to main branch
- [ ] Can be demoed in Sprint Review
- [ ] Bugs or future work cards are identified and created
- [ ] Reviewed and approved by Product Owner

#### Notes:

-

TammieV-BCG:
Refinement note: 

- Exploration aim: audit this to understand what is still missing and update ticket

PR #3463, opened 2025-06-11
Rls policy setup

BCerki:
Setup for #3262 and #3318 

This PR:

- comments RLS setup work back in
- policy generation functions
- drop policy functions (unlike grants, you can't make policies over and over, you have to drop and re-create)
- apply policies to Operation and Report Attachment
- tests for all of the above

I think the sonar stuff is OK (in tests, or code that was previously written), so if the reviewer agrees, I'll dismiss it before merge


dleard (approved):
Fantastic! Clean solution, thorough and very well written! We'll need to handle the "timeline" aspect that gets introduced by transferring operations/facilities before enabling rls. Approving this PR & we'll split out a couple follow up cards to handle the timeline stuff.
Great work!

PR #3465, opened 2025-06-12
Chore/193 refresh on request services

dleard:
Adds a service to refresh data from elicensing into our own models before returning the data from our models.
This is now used when creating an invoice/fee on emission report submission.
The API endpoint/service that gets the list of payments for an obligation has also been refactored to use this service.

A way to manually test this is:

1. In reporting, submit a report for the obligation not met operation
2. Confirm in the database that there are matching records in the elicensing_invoice and elicensing_line_item tables for the invoice/fee
3. In elicensing, manually create a payment for that invoice
4. From the python shell, call `get_compliance_obligation_payments_by_compliance_report_version_id` (ComplianceDashboardService)
5. Confirm in the database that there is a matching record in the elicensing_payment table for the payment created in step 3


bc_obps/compliance/models/elicensing_line_item.py
shon-button, line 1: @dleard 

Do we not need an amount field to reflect the data found in an InvoiceFee?

Something like:
```

    amount = models.DecimalField(
        decimal_places=2,
        max_digits=20,
        db_comment="The amount of this fee in dollars from elicensing",
    )
```

bc_obps/compliance/service/elicensing/obligation_elicensing_service.py
shon-button, line 1: nit: perhaps rename to `elicensing_obligation_service` to align with folder structure?

bc_obps/compliance/service/elicensing/operator_elicensing_service.py
shon-button, line 1: nit: perhaps rename to `elicensing_operator_service` to align with folder structure?

bc_obps/compliance/service/elicensing/elicensing_data_refresh_service.py
```diff
@@ -0,0 +1,102 @@
...
+        Returns:
+            None
+        """
+
+        client_operator = ElicensingClientOperator.objects.get(id=client_operator_id)
+        invoice_response = elicensing_api_client.query_invoice(
+            client_id=client_operator.client_object_id, invoice_number=invoice_number
+        )
+
+        invoice_record, _ = ElicensingInvoice.objects.update_or_create(
```
Sepehr-Sobhani, line 65: Question: We have quite a few `update_or_create` calls — are we expecting a high volume of object creation? If so, is there a risk of orphaned records in our data models?

bc_obps/compliance/service/elicensing/elicensing_data_refresh_service.py
```diff
@@ -0,0 +1,102 @@
...
+        cls, compliance_report_version_id: int
+    ) -> RefreshWrapperReturn:
+        data_is_fresh = True
+        invoice = ComplianceObligation.objects.get(
+            compliance_report_version_id=compliance_report_version_id
+        ).elicensing_invoice
+        try:
+            ElicensingDataRefreshService.refresh_data_by_invoice(
+                client_operator_id=invoice.elicensing_client_operator_id, invoice_number=invoice.invoice_number  # type: ignore[union-attr]
+            )
```
Sepehr-Sobhani, line 40: Here, we should either return the updated invoice from this service or call `invoice.refresh_from_db()` after that to ensure we have the latest data. Otherwise, even if `refresh_data_by_invoice` succeeds, we would still be returning the old reference to the invoice object defined on line 34.

bc_obps/compliance/service/elicensing/elicensing_data_refresh_service.py
```diff
@@ -0,0 +1,102 @@
...
+        except:  # noqa: E722
+            data_is_fresh = False
```
Sepehr-Sobhani, line 42: Suggestion:
```python
except Exception as e:
    logger.error(f"Failed to refresh data by invoice: {str(e)}")
data_is_fresh = False
```

bc_obps/compliance/tests/api/_compliance_report_versions/_compliance_report_version_id/test_payments.py
Sepehr-Sobhani, line 1: I suggest keeping this test but adjusting its structure to focus on verifying that the shape of the data returned from this endpoint matches our expectations.

bc_obps/compliance/tests/service/elicensing/test_elicensing_data_refresh_service.py
```diff
@@ -0,0 +1,114 @@
...
+from decimal import Decimal
+
+import pytest
+from model_bakery.baker import make_recipe
+
+
+class TestOperatorELicensingService:
+    """Tests for the ElicensingDataRefreshService class"""
+
+    @pytest.mark.django_db
```
Sepehr-Sobhani, line 16: nit: We can add `pytestmark = pytest.mark.django_db` before the test class to avoid repeating it on every test function.

Sepehr-Sobhani (approved):
Great work! 🤩

bc_obps/compliance/service/elicensing/elicensing_data_refresh_service.py
Sepehr-Sobhani: Yes!

bc_obps/compliance/models/elicensing_line_item.py
dleard: these fields have been added to the model / refresh service

Issue #3125, opened 2025-03-28
Reword the backend errors


zoeyli-46:
### description

<img width="497" alt="Image" src="https://github.com/user-attachments/assets/f9055a0c-4318-4f9d-b844-6fde91822605" />

[Spreadsheet of where the errors occur](https://bcgov.sharepoint.com/:x:/r/teams/00608-DesignersInternal/_layouts/15/Doc.aspx?sourcedoc=%7BF9A939B5-E785-4CF0-B3D0-B015D9F54BE3%7D&file=Error%20Messages.xlsx&nav=MTVfezAwMDAwMDAwLTAwMDEtMDAwMC0wMDAwLTAwMDAwMDAwMDAwMH0&action=default&mobileredirect=true)

Users get an ambiguous internal server error message but do not have instructions on what to do next. Error message should inform the user of how to proceed if they run into an error

### Task
Replace the following internal server error messages: 
- [ ] An error occurred in the Server Components render. The specific message is omitted in production builds to avoid leaking sensitive details. A digest property is included on this error instance which may provide additional detials about the nature of the error. BC: not sure how much we can do about this one: https://github.com/vercel/next.js/issues/71101 and https://github.com/vercel/next.js/discussions/78135. My best guess for what triggered this was an error in one of the create schema functions, which didn't all have error handling, so I've added that, and hopefully that will hit our error boundary instead of this
- [x] An internal server error has occured. Please try refreshing the page or contact support if the problem persists. If reporting this issue, mention that the error has been logged. Reference Code: 
- [x] Cannot delete some instance of model '...' beacuse they are referenced through protected foreign keys: '...'  BC: this is one is no longer reproducible via the path in the spreadsheet, but it could pop up again, so I implemented a message
- [x] ~Error comparing files: file does not exist: documents/....~ BC: we removed the document comparison function as part of endpoint optimization and I can't reproduce this error anymore

For errors where dropdowns fail to get populated: for example:
- [x] Operation Representative Page - "Failed to Retrieve Contact Information"
- [x] Contact Page - "Failed to retrieve contact information"

Replace with the error message
- [x] An internal server error has occurred. Please refresh the page. If the problem persists, contact [ghgregulator@gov.bc.ca](mailto:ghgregulator@gov.bc.ca) for help and include the reference code: [code]
- ~[ ] Operation/facility information - "notFound"~ BC: this is struck out in the spreadsheet, and in admin Operation and Facility info pages have error handling, so I think it's been fixed

[...]

PR #3468, opened 2025-06-13
3125 backend error wording

BCerki:
card: https://github.com/bcgov/cas-registration/issues/3125

The card asks for the backend errors to be handled by the same message as the ErrorBoundary, so this PR:
- creates a new component with the bulk of the ErrorBoundary component that's reusable across client/server components (ErrorBoundary only catches render errors in client component, not fetch ones in server components)

### DIFF
--- a/bciers/libs/components/src/error/ErrorBox.tsx
+++ b/bciers/libs/components/src/error/ErrorBox.tsx
@@ -1,0 +1,51 @@
+"use client";
+import { useEffect, useState } from "react";
+import * as Sentry from "@sentry/nextjs";
+import { ghgRegulatorEmail } from "@bciers/utils/src/urls";
+
+import { Alert, AlertTitle } from "@mui/material";
+
+interface Props {
+  error?: Error & { digest?: string };
+}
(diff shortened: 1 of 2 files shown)


bciers/libs/components/src/error/ErrorBox.tsx
```diff
@@ -0,0 +1,51 @@
...
+      >
+        <AlertTitle>
+          <strong>Error</strong>
+        </AlertTitle>
+        <p>An internal server error has occured.</p>
+        <p>
+          Please refresh the page. If the problem persists, contact{" "}
+          <a href={ghgRegulatorEmail}>ghgregulator@gov.bc.ca</a> for help and
+          include the reference code:{" "}
+          <strong>{eventId ? eventId : "pending"}</strong>
```
BCerki, line 46: I put 'pending' because it takes a moment for the code to show up, and if there's a problem logging to sentry, we'll never get a code

andrea-williams (approved):
Looks great! And thank you x1000 for adding in suggestions on how to test - saved me a lot of time!

PR #3477, opened 2025-06-17
chore: external access request tests

fviduya:


bc_obps/registration/fixtures/mock/user_operator.json
```diff
@@ -228,5 +228,17 @@
       "verified_by": null,
       "user_friendly_id": 20
     }
+  },
+  {
+    "model": "registration.useroperator",
```
fviduya, line 233: This is bc-cas-dev-secondary. Changed this to pending access request approval state. Let me know if this shouldn't be changed!

fviduya: Removed this change as I realized that we are using this user in select-operator test. Opted to upsert the record in the test file instead of making it permanent in the fixtures so we won't have an issue in the select-operator test.

bciers/apps/administration-e2e/tests/workflows/users-and-access-requests.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+    await assertSuccessfulSnackbar(page, /is now approved/i);
+
+    await takeStabilizedScreenshot(happoPlaywright, page, {
+      component: "EXTERNAL: Approve an administrator",
+      variant: "default",
+    });
+
+    const newPage = await openNewBrowserContextAs(UserRole.INDUSTRY_USER);
+    // const context = await openNewBrowserContextAs(UserRole.INDUSTRY_USER);
+    // const newPage = await context.newPage();
```
fviduya, line 101: This is leftover, I'll clean it once I do the cleanup at the end

bciers/apps/administration/app/components/datagrid/models/operations/operationColumns.ts
```diff
@@ -32,7 +32,8 @@ const operationColumns = (
           case !!boroId:
             return boroId; // Return 'boroId' if it exists
           case regulatedOperationPurposes.includes(purpose):
-            return "Pending"; // Return "Pending" if 'purpose' is in the list
+            return "Pendings"; // Return "Pending" if 'purpose' is in the list
+          // chesca again
```
fviduya, line 36: This is to force e2e test

bciers/libs/e2e/src/utils/helpers.ts
```diff
@@ -375,11 +400,7 @@ export async function tableHasExpectedRowCount(
   await expect(rows).toHaveCount(expectedRowCount);
 }
 
-// 🛠️ Function: calls api to seed database with data for workflow tests
-export async function setupTestEnvironment(
-  workFlow?: string,
-  truncateOnly?: boolean,
-) {
+export async function getBrowser() {
```
fviduya, line 403: This was originally in setupTestEnvironment, I created a separate function for this since this is also needed in the new function `openNewBrowserContextAs`

bciers/libs/e2e/src/utils/helpers.ts
```diff
@@ -409,7 +430,40 @@ export async function setupTestEnvironment(
...
+export async function openNewBrowserContextAs(
+  role: string,
+  happoPlaywright?: any,
+) {
+  const browser = await getBrowser();
+  const storageState = await getStorageStateForRole(role);
+  const context = await browser.newContext({ storageState });
+
+  // To take a screenshot in the new browserContext, we first need to close the current context for the screenshot to be saved properly, then re-initialize using the new browserContext
+  if (happoPlaywright) {
```
fviduya, line 453: This function _might_ be used for other tests, opted to put this here. I learned that the happoPlaywright needs to be reinitialized based on the new context, otherwise it won't take a screenshot for the new page and will result in an error. 

I also learned that it's best to call `happoPlaywright.finish()` first before reinitializing, so screenshots taken before reinitializing can be saved properly.

bciers/libs/e2e/src/utils/helpers.ts
```diff
@@ -579,7 +629,45 @@ export async function clickWithRetry(
...
+}
+
+export async function assertSpinnerIsDone(row: Locator) {
+  const spinner = row.locator('svg[aria-label="loading"]').first();
+  await expect(spinner).toBeVisible();
+  await expect(spinner).toBeHidden();
+}
+
+// This is specifically for a combobox that does not allow manual entries
+export async function selectOptionFromCombobox(
```
fviduya, line 649: The comboBox that we have in the user access request page does not have a search capability, unlike the other ones we have in administration and registration.

bciers/apps/administration-e2e/poms/users-access-request.ts
```diff
@@ -0,0 +1,168 @@
...
+  readonly url: string = process.env.E2E_BASEURL;
+
+  readonly userAccessRequestURL = this.url + AppRoute.USER_ACCESS_REQUEST;
+  // Button Locators
+
+  // Field Locators
+
+  readonly fieldUsername: Locator;
+
+  readonly fieldPassword: Locator;
```
fviduya, line 36: These are also going to be deleted since it's not used anymore. The original plan was to log out and log in manually, now we are opening a new browser context and using the current storage state that we have.

bciers/libs/e2e/src/utils/helpers.ts
BCerki: We have a few different widgets in the code that look almost the same in the UI: ComboBoxWidget, SelectWidget, and MultiSelectWidget. In case you haven't come across this already, you can tell which one a field is using by looking at the RJSF ui:schema, e.g. 
image

BCerki (commented):
Nice work on this!

To answer this question: `Add a test for verifying that contact is created upon approval of request?` I don't think we need to check for this. It's already handled in the pytests.

Happo shots look great.

I hit errors when running the decline test locally, even with a longer timeout: 
image
That could definitely be a local problem though, so if you're not seeing anything like that, we can pair on my setup.

fviduya:
> I hit errors when running the decline test locally, even with a longer timeout: image That could definitely be a local problem though, so if you're not seeing anything like that, we can pair on my setup.

Did you run just the declined test? If yes, I think it is expected to give an error. The test uses beforeAll so you'd have to run the test as a whole since approval of admin and declined tests are dependent on the approval of reporter. I made it this way so that the test is not going to be too slow, but I can also change it to beforeEach if that's going to be more convenient.

BCerki:
> > I hit errors when running the decline test locally, even with a longer timeout: image That could definitely be a local problem though, so if you're not seeing anything like that, we can pair on my setup.
> 
> Did you run just the declined test? If yes, I think it is expected to give an error. The test uses beforeAll so you'd have to run the test as a whole since approval of admin and declined tests are dependent on the approval of reporter. I made it this way so that the test is not going to be too slow, but I can also change it to beforeEach if that's going to be more convenient.

Tests shouldn't be depedent on each other (if we introduce a change that makes them fail, it's much harder to find the regression if we have to detangle the tests first), so even if they're going to run slower, I think it's better to have them isolated.

bciers/apps/administration-e2e/poms/users-access-request.ts
```diff
@@ -0,0 +1,173 @@
...
+
+  async declineRequest(row: Locator) {
+    await row.getByRole("button", { name: "Decline" }).click();
+  }
+
+  async editRequest(row: Locator) {
+    let action = "Edit";
+    await row.getByRole("button", { name: action }).click();
+    await assertSpinnerIsDone(row);
+  }
```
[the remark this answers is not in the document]
fviduya: Thanks for this, I thought to combine the approve and decline actions together, and leave editRequest as its own since that's not going to assign any role to the user. I added the repetitive assertions in another function as well to make this cleaner.

bciers/apps/administration-e2e/poms/users-access-request.ts
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    const remainingActions = actions.filter(
+      (action) => !actionList.includes(action),
+    );
+
+    return { actionList, remainingActions };
+  }
+
+  async approveRequest(row: Locator, originalRole: string) {
```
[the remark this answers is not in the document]
fviduya: This was a leftover from the original approach I had, thanks for pointing this out!

bciers/apps/administration-e2e/poms/users-access-request.ts
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    for (let x = 0; x < hiddenActions.length; x++) {
+      await expect(
+        row.getByRole("button", { name: hiddenActions[x] }),
+      ).toBeHidden();
+    }
+  }
+
+  async assertCorrectRole(row: Locator, originalRole: string) {
```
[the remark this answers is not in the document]
fviduya: Thanks for pointing this out, it was also a leftover from the previous approach.

fviduya:
> > > I hit errors when running the decline test locally, even with a longer timeout: image That could definitely be a local problem though, so if you're not seeing anything like that, we can pair on my setup.
> > 
> > 
> > Did you run just the declined test? If yes, I think it is expected to give an error. The test uses beforeAll so you'd have to run the test as a whole since approval of admin and declined tests are dependent on the approval of reporter. I made it this way so that the test is not going to be too slow, but I can also change it to beforeEach if that's going to be more convenient.
> 
> Tests shouldn't be depedent on each other (if we introduce a change that makes them fail, it's much harder to find the regression if we have to detangle the tests first), so even if they're going to run slower, I think it's better to have them isolated.

That makes sense, the user_operator record will now be added before each test. Because of that, the checking for edit functionality has been removed, so I added another test instead specifically just for checking that when user clicks Edit, they still see the actions and correct status since we had an issue before where if they click the said button, the actions and status becomes blank.

BCerki (approved):
:cartwheeling:

PR #3483, opened 2025-06-17
3379 CRA Business Number to string

acatchpole:
Card: https://github.com/bcgov/cas-registration/issues/3379
### Problem
The CRA Business number was appearing with commas in it as though it was a large number and not an identifier.

### Solution
Refactor cra_business_number from an int to a string. This 'number' is an identifier that is unchangeable and no math is done it. Therefore, it should properly be a string rather than an int. It being a string means that it will display in whatever format we define.

### Changes
 - Added validation to models - regex for 9 digits
 - Created migration to change operator related tables. This changes the field expected data type and casts the existing values from an int to a string
 - Added a custom format for front end validation
 - updated all fixtures and test values
 - updated expected types in functions where needed
 
### Questions for reviewer
 - Did not update the ETF script for importing from SWRS. Is this necessary?
 - As this migration will change PROD data when deployed, is there anything that should be done regarding that *inside* this PR?

### to test
1. login to application as a user already associated with an operator (eg. bc-cas-dev)
2. From the Dashboard, click "My Operator" link under Administration
3. On the my-operator page, click the "Edit" button
4. Scroll up to CRA Business Number field
   - Expect to see a 9 digit number with no commas


BCerki (commented):
Nice work! I manually tested with both mock and real data, and everything looks good in the db.

A couple of things that came up in the FE (if these pre-dated your PR, we can make new tickets for them):
We're missing the RJSF validation on multiple operators: 
image
(We should have gotten something like the corporate registry number. We do get the proper CRA validation for operator, partner, and parent; it's only missing on multiple operator)

Overlap on the select operator error: 
image

bc_obps/registration/tests/models/test_operator.py
```diff
@@ -72,12 +72,12 @@ def setUpTestData(cls):
         ]
 
     def test_check_cra_business_number_format(self):
-        valid_numbers = [123454321]
+        valid_numbers = ['123454321', 123454321]
```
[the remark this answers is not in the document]
acatchpole: this was left in to show that it DOES work. The model has the value defined as a string, but apparently it also handles an otherwise valid int value seamlessly.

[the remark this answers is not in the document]
BCerki: I guess it gets coerced to a string at some point. Could we add another assertion to confirm that the type is correct after saving? Maybe something like assert isinstance(self.test_object.cra_business_number, str)

bc_obps/registration/migrations/0124_cra_business_number_to_string_in_operator_models.py
```diff
@@ -90,7 +162,7 @@ class Migration(migrations.Migration):
             name='cra_business_number',
             field=models.CharField(
                 blank=True,
-                db_comment='The CRA business number of an operator',
+                db_comment='The CRA business number of an _operator',
```
BCerki, line 165: typo?

BCerki (approved):
I re-checked the overlapping error and select operator field and it's happening even when I've got my accessibility settings turned off, so I'll make a card for it.

Other code changes look good! I had one more suggestion in a test, and some of the CI steps are failing, but that might just need a rebase. I'm approving, but I'd suggest waiting until you get an answer on whether to include the tests in the migration before merging.

bc_obps/registration/migrations/0124_cra_business_number_to_string_in_operator_models.py
```diff
@@ -0,0 +1,187 @@
...
+            field=models.CharField(
+                db_comment='The CRA business number of an operator',
+                validators=[
+                    django.core.validators.RegexValidator(
+                        message='CRA Business Number should be 9 digits.', regex='^\\d{9}$'
+                    )
+                ],
+            ),
+        ),
+        migrations.RunPython(assert_cra_business_numbers_casted, reverse_code=migrations.RunPython.noop),
```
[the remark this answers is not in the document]
acatchpole: per discussion with other devs:
Migrations in production are only run once ever (we don't rollback). Therefore, after it is confirmed that the migration works locally, the tests should be removed to keep the migration code clean and readable.

bc_obps/registration/migrations/0124_cra_business_number_to_string_in_operator_models.py
acatchpole: aye

bc_obps/registration/schema/partner_operator.py
```diff
@@ -13,7 +13,7 @@ class PartnerOperatorIn(ModelSchema):
 
     legal_name: str = Field(..., alias="partner_legal_name")
     trade_name: Optional[str] = Field("", alias="partner_trade_name")
-    cra_business_number: int = Field(..., alias="partner_cra_business_number")
+    cra_business_number: str = Field(..., alias="partner_cra_business_number")
```
pbastia, line 16: Should we add the pattern here too?

bc_obps/registration/schema/operator.py
```diff
@@ -95,7 +96,7 @@ class OperatorIn(ModelSchema):
 
     trade_name: Optional[str] = ''
     business_structure: str
-    cra_business_number: int
+    cra_business_number: str = Field(pattern=CRA_BUSINESS_NUMBER_REGEX)
```
pbastia, line 99:
```suggestion
    cra_business_number: str = Field(...,pattern=CRA_BUSINESS_NUMBER_REGEX)
```

bc_obps/registration/schema/validators.py
```diff
@@ -1,10 +1,10 @@
 from typing import Optional
 
 
-def validate_cra_business_number(value: int) -> Optional[int]:
+def validate_cra_business_number(value: str) -> Optional[str]:
     """
     Validate that the cra_business_number has 9 digits.
     """
-    if not (isinstance(value, int) and 100000000 <= value <= 999999999):
+    if not (value.isdigit() and len(value) == 9):
```
pbastia, line 8: I would use a regex here, the `isdigit()` method can return true with funny characters https://stackoverflow.com/a/54912545

pbastia (approved):
Approved conditionally! Nice work! Just a comment about using `isdigit()` that could match some not-really-numeric characters and have our system behave weirdly

PR #3484, opened 2025-06-17
chore: updated equivalent emission in report raw activity data json

ayeshmcg:
Card: https://github.com/bcgov/cas-reporting/issues/820
Changes: Updated Report Save activity data to save to raw activity json after all the emissions are calculated
Added migration to update all the existing raw activity data json


bc_obps/reporting/migrations/0114_update_report_raw_activity_data_equivalent_emission.py
```diff
@@ -0,0 +1,53 @@
...
+        record.save(update_fields=["json_data"])
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("reporting", "0113_V3_7_0"),
+    ]
+
+    operations = [
+        migrations.RunPython(migrate_json_data),
```
pbastia, line 52: Given that we're modifying production data, we might want to write tests for these methods!

bc_obps/reporting/service/report_activity_save_service.py
```diff
@@ -81,13 +80,18 @@ def save(self, data: dict) -> ReportActivity:
...
+                id__in=retrieve_ids(data["sourceTypes"])
+            ).delete()
 
-        for source_type_key in data["sourceTypes"]:
-            self.save_source_type(report_activity, source_type_key, data["sourceTypes"][source_type_key])
+            for source_type_key in data["sourceTypes"]:
+                self.save_source_type(report_activity, source_type_key, data["sourceTypes"][source_type_key])
+
+        # Save raw data after processing source types
+        self.save_raw_data(data)
```
pbastia, line 94: I think you're right, the save method modifies in-place the json data. This is not best practice, but works well for this particular case.
We also need a test to validate this new behaviour

pbastia (changes requested):
Excellent work! We just need a few tests to validate the new behaviour and verify that we're modifying production data properly

pbastia (approved):
🎉

PR #3486, opened 2025-06-18
chore: remove resolve method for excess emissions

vivanovbc:
### Changes
- removed the resolve method and rely on the alias for compliance report excess emissions
- since the database field already has 4 decimal places, we could just remove the resolve method entirely
- that should fix an issue with the value of Excess Emission in the Compliance Report Details page is rounded off to the nearest whole number. This affects the value of Equivalent Value and Outstanding Balance (currently hardcoded)


Sepehr-Sobhani (approved):
🤩

PR #3490, opened 2025-06-18
478 thousand separator and number/string types

BCerki:
card: https://github.com/bcgov/cas-reporting/issues/478

This PR:
- changes type: string to type: number in rjsf schemas for fields that should be numbers. This fixes the thousand separator display
- migration to update all the activity json schemas
- updates text widget to better handle number values
- updates final review string widget to better handle numbers
- new and updated vitests
- a bit of documentation about the TextWidget works with string vs. number type

Notes: 
- For the migration, I pulled the activity name and source type out of the `ActivitySourceTypeJsonSchema` model, so those are unlikely to have typos. I manually matched up the json schema path with the activity and source type. I tried to do some something more programmatic, but the filenames and json `title` didn't exactly match the db in all cases. I put in some testing code (ie, there are 139 json files and 139 ActivitySourceTypeJsonSchema records were updated), and also some print statements for manual verification
- I pulled prod data and looked at some of the reports in the UI. The equivalent emission field looks ok.
- Pre-commit is failing because of the test code in the migrations. I'll pull it out before merging


bciers/apps/reporting/src/app/components/facility/FacilityEmissionAllocationForm.tsx
```diff
@@ -160,13 +159,21 @@ export default function FacilityEmissionAllocationForm({
...
+        ? Number(initialData.facility_total_emissions)
+        : initialData.facility_total_emissions,
+      products:
+        initialData.report_product_emission_allocation_totals?.map(
+          (product: { [key: string]: any }) => ({
+            ...product,
+            allocated_quantity: product.allocated_quantity
+              ? Number(product.allocated_quantity)
+              : product.allocated_quantity,
+          }),
```
pbastia, line 172: I think the logic here could be put behind a function? is the idea to "parse number or undefined" ?
We might be able to handle things a bit better by having the code be explicit about what is expected. Otherwise empty string, empty array, null or undefined would all stay unchanged

bciers/apps/reporting/src/app/components/finalReview/formCustomization/FinalReviewStringField.tsx
```diff
@@ -1,5 +1,40 @@
...
+      <NumberField.Root
+        id={props.id}
+        name={props.name}
+        disabled
+        value={props.formData ? Number(props.formData) : props.formData}
+        format={{
+          // set the fraction digits based on how many decimal places that value returned from the API has. Lat/long can have up to 8, and some calculated values can have up to 4.
+          maximumFractionDigits: props.formData
+            ? props.formData.toString().split(".")[1]?.length
+            : 4,
```
pbastia, line 25: I think this overrides the 4 digits as soon as we have a decimal point in the number, so we might start seeing numbers like 123.00000000000000456 if they ever happen.
For lat/lon we should probably have a separate widget

bciers/libs/components/src/form/widgets/TextWidget.tsx
```diff
@@ -63,18 +63,25 @@ const TextWidget: React.FC<WidgetProps> = ({
...
         style={widthStyle}
-        format={{ maximumFractionDigits: 4 }}
+        format={{
+          // set the fraction digits based on how many decimal places that value returned from the API has. Lat/long can have up to 8, and some calculated values can have up to 4.
+          maximumFractionDigits: value
+            ? value.toString().split(".")[1]?.length
+            : 4,
+          minimumFractionDigits: value
+            ? value.toString().split(".")[1]?.length
+            : 0,
```
pbastia, line 83: Same comment as for the FinalReviewStringField! I suggest
- Keep the default to 4 decimal points
- Have a special case for lat/lon, probably as a different widget, or a [ui:decimalPoints] field in the ui schemas to accomodate that

pbastia (changes requested):
Very nice work, that was a lot of updates! There is just one breaking change in the widget about the number of decimal places

bciers/apps/reporting/src/app/components/finalReview/formCustomization/FinalReviewStringField.tsx
BCerki: Actually, max decimals should be purely a formatting thing, so that error must be about something else, will investigate

bciers/apps/reporting/src/app/components/additionalInformation/operationEmissionSummary/OperationEmissionSummaryForm.test.tsx
```diff
@@ -35,22 +35,22 @@ const mockNavigationInformation = {
 };
 
 const mockSummaryFormData = {
-  attributable_for_reporting: "1866666648.0000",
```
BCerki: Now that values are numbers instead of strings, we would never get trailing 0s

pbastia (changes requested):
Just a couple of things left! Very nice!

bciers/apps/reporting/src/app/components/activities/ActivityForm.tsx
```diff
@@ -52,9 +52,39 @@ export default function ActivityForm({
...
+          newObj[key] = typeof value === "number" ? value : Number(value);
+        } else {
+          newObj[key] = transformEquivlantEmission(value);
+        }
+      }
+      return newObj;
+    }
+
+    return obj;
+  }
```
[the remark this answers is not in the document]
BCerki: It turned out that it wasn't pydantic changing the type; it was one of our services, so I all I had to do was change the service.

bc_obps/reporting/tests/service/test_report_activity_save_service/test_save_report_emission.py
```diff
@@ -281,4 +281,4 @@ def test_save_equivalent_emission(self):
         )
 
         # Rounds to 4 decimal places and returns correct equivalentEmission value
-        assert return_value.json_data["equivalentEmission"] == "100.1234"
+        assert return_value.json_data["equivalentEmission"] == 100.1234
```
BCerki, line 284: This is the sonarcloud fail about not using == with floats. Since it's a test, I'll dismiss before merging if the reviewer is ok

Issue #102, opened 2023-09-12
IO updates/confirms information for an operator that already exists in our database


andrea-williams:
#### Description:

Brief description of the work involved, including any user research context or key decisions.

#### Acceptance Criteria:

Given I am an Industrial Operator (person) who has logged into my Business BCeID account
When
Then

#### Development Checklist:

- [ ] Checklist item
- [ ] Checklist item
- [ ] Checklist item
- [ ] Meets the DOD

**Definition of Ready** (Note: If any of these points are not applicable, mark N/A)

- [ ] User story is included
- [ ] User role and type are identified
- [ ] Acceptance criteria are included
- [ ] Wireframes are included (if required)
- [ ] Design / Solution is accepted by Product Owner
- [ ] Dependencies are identified (technical, business, regulatory/policy)
- [ ] Story has been estimated (under 13 pts)

·**Definition of Done** (Note: If any of these points are not applicable, mark N/A)

- [ ] Acceptance criteria are tested by the CI pipeline
- [ ] UI meets accessibility requirements
- [ ] Configuration changes are documented, documentation and designs are updated
- [ ] Passes code peer-review
- [ ] Passes QA of Acceptance Criteria with verification in Dev and Test
- [ ] Ticket is ready to be merged to main branch
- [ ] Can be demoed in Sprint Review
- [ ] Bugs or future work cards are identified and created
- [ ] Reviewed and approved by Product Owner

#### Notes:

-

andrea-williams:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Climate-Action-Secretariat-60ca4121764d710011481ca2/issues/bcgov/cas-registration/102?planning-poker) with Zenhub @andrea-williams @BCerki @dleard @JoshLarouche @Sepehr-Sobhani

hannavovk:
I think this is actually covered by #26. Or are there additional aspects here that are yet to be worked on?  @andrea-williams, @BCerki - can you confirm?

hannavovk:
Update: #26 is for operation, this is for operator.

suhafa:
Marked this as 'Needs Wireframe', there needs to be a design for the purpose of just reviewing operator information and updating it

hannavovk:
@suhafa and @andrea-williams  Let's talk about this one and figure out what exactly is needed so that we can get that wireframe going.

suhafa:
[...]

PR #3498, opened 2025-06-20
feat: add declined and changes required status for issuance requests

nataliepiguliak:


Sepehr-Sobhani (approved):
Great stuff 🤩 , Approved! However, as I mentioned, if implementing `TrackStatusOfIssuancePage` for internal users is within the scope of this ticket, we need to include it—since there’s currently no such page available for internal users.


Screenshot 2025-06-20 at 4 23 19 PM
Screenshot 2025-06-20 at 4 23 50 PM

nataliepiguliak:
Many thanks Sepehr! The above functionality for internal users is coming in #102

PR #3501, opened 2025-06-20
1893 ci protecting migration code behind tags

pbastia:


dleard (approved):
Nice! Definitely a good check to have in place

PR #3509, opened 2025-06-23
Add friendly error messages for contact duplicate emails

acatchpole:


bc_obps/service/contact_service.py
```diff
@@ -91,12 +92,15 @@ def update_contact(
             raise Exception(UNAUTHORIZED_MESSAGE)
 
         address_data = payload.dict(include={'street_address', 'municipality', 'province', 'postal_code'})
-
+        contact_data: Dict = payload.dict(include={*ContactIn.Meta.fields})
```
pbastia, line 95: We shouldn't make the service depend on the API layer (the `ContactIn` schema). The service should declare exactly which fields it needs here instead:
```py
payload.dict(include=['first_name', 'last_name', 'email', 'phone_number', 'position_title'])
```
(this solution is already iffy since the type of the payload is an API type instead of a Service layer type)

bc_obps/service/contact_service.py
```diff
@@ -152,3 +156,23 @@ def _validate_operation_representative_address(cls, contact_id: int, address_dat
...
+        existing_contact = contacts_with_email.first()
+        if existing_contact:
+            if contact_id is None:
+                raise UserError(
+                    f"A contact with the email '{email}' already exists. Please add a different contact or edit the existing contact."
+                )
+            else:
+                raise UserError(
+                    f"The email '{email}' is in use by another contact. Please use a different email address."
+                )
```
pbastia, line 178: This could be simplified, we can safely exclude the case `id=None` without making a special case for it:

```py
if Contacts.objects.filter(email=email).exclude(id=contact_id).exists():
  message = f'A contact...' if contact_id is None else f'The email ...'
  raise UserError(message)
```

pbastia (approved):
Looks good to me! Just a couple of nitpicks about style

bc_obps/service/contact_service.py
acatchpole: very pythonic. much nicer. implemented

PR #3521, opened 2025-06-26
Supplementary report status 751

BCerki:
card: https://github.com/bcgov/cas-reporting/issues/751

This PR:
- shows 'Draft Supplementary Report' as the status for reports with status draft and a report_version_id that doesn't match the first_report_version_id
- previously, filtering and sorting for Not Started wasn't working, fixed
- add filtering and sorting for 'Draft Supplementary Report'
- comment to clarify how the transfer grid filter works (I used it as a model for the status filter and it took me awhile to figure out what was going on)
- pytests for filtering and sorting the report_status column


acatchpole (approved):
looks good to me

PR #3547, opened 2025-07-03
bug: added operator to contact email validation

acatchpole:
Card: [no card]

### The Issue
A validation function was added to the ```contact_service``` to prevent duplicate emails and provide a user friendly error. There was an error in the logic of this function that assumed a Contact email had to be universally unique. The actual constraint is that an email must be unique ***for a given operator***.

### The fix
The validation now filters by email AND operator to check for uniqueness

### To Test


bc_obps/service/contact_service.py
```diff
@@ -99,8 +99,9 @@ def update_contact(
         cls._validate_operation_representative_address(contact_id, address_data)
 
         # Prevent updating contact if the email is already in use by another contact
```
BCerki, line 101: Suggest adding to this comment to mention the operator

BCerki (approved):
Looks great, thanks for the detailed testing instructions!

Issue #3517, opened 2025-06-25
Add Operator submit does not augment token app_role to "industry_user_admin"


shon-button:
#### Describe the Bug:

**Issue:**
`Administration\Add Operator Form` submit does not augment auth token with app_role to "industry_user_admin"

**Reproduction Steps:**
Complete workflow `Add Operator  → Register Operation` to display link for `My Compliance`
Click link `My Compliance`  
App routes to `Could not find requested resource` indicating the authenticated role is NOT `industry_user_admin`

**Note:** On subsequent log out and log in, `My Compliance` links does route to expected route indicating the authenticated role is now set to `industry_user_admin`

![Image](https://github.com/user-attachments/assets/fe81affa-601d-454a-ae8c-68a071ff7d87)
![Image](https://github.com/user-attachments/assets/aadede50-e9fb-4573-8c7c-bef21e4b5e1a)

![Image](https://github.com/user-attachments/assets/3d920ec6-95c3-49c5-919c-71b8a26e1f1d)

**Suspected Root Cause:** 
`bciers/apps/administration/app/components/operators/OperatorForm.tsx\onSubmit` does NOT augment token app_role to "industry_user_admin" since changing the method from `await update()` to ` await getSession();`

![Image](https://github.com/user-attachments/assets/eac2fdfb-74f5-4bb8-bafa-1d054202ecd7)

#### Probability:

(How likely the bug is to happen, scored from 1-5. For example, probability of 5 is something like "it happens to all users every time they log in." whereas probability of 1 "only happens to certain users when a really specific and unlikely path is followed.")

#### Impact:

(How bad the bug is when it does happen, scored from 1-5. For example, effect of 5 is "the entire app crashes and makes it unusable for all users" or "the bug causes the wrong data to be saved, with critical information (e.g. payment) being affected." whereas effect of 1 is "It makes some styling look a little bit weird.")

<img src="https://www.getzephyr.com/sites/default/files/riksk%20matrix.jpg" alt="Bug Matrix" width="250" height="250"/>

#### Steps to reproduce the behaviour:

1. Step #1
2. Step #2
3. Step #3

#### Additional information:

Browser: [e.g. chrome, safari]
Version: [e.g. 22]

#### Additional Context

-

patriciarussellCAS:
@shon-button or @vesselak would you like this transferred to compliance repo?

shon-button:
[...]

PR #3553, opened 2025-07-07
Augment industry_user role

BCerki:
card: https://github.com/bcgov/cas-registration/issues/3517


andrea-williams (approved):
Tested the bug before and after this fix - can confirm it works. Well done!

Issue #3482, opened 2025-06-17
Internal Users can update/consolidate/merge duplicate Operation records in Admin


patriciarussellCAS:
#### Describe the task

There are a number of Operations in the system with duplicate records. This is due to a number of factors/situations which need to be reviewed and properly cleaned up. 
The intent of this ticket is to confirm the issues and come up with functionality that can be built in the Internal user view to allow Internal users (roles to be confirmed) to make the necessary updates. 
Tiegan from IRC has created a list of known issues in our teams channel [here](https://bcgov.sharepoint.com/:w:/r/teams/00608-RegistrationBCIERSInternalUsers/Shared%20Documents/BCIERS%20Internal%20Users/Duplicate%20Operation%20Tracker.docx?d=wbf114a20e6d44a1f99a460dc2cd3c445&csf=1&web=1&e=0BfA9t)

This includes: 
- moving/consolidating BCGHGIDs, BOROIDs and/or SWRS facility ID's from one operation to another
- ~~deleting, or removing a duplicated record from the system~~  <- create sub-issue for deleting records! https://github.com/bcgov/cas-registration/issues/3568

We have one ticket already created for one of the operations: CertainTeed. See #3271 

* [Wireframe](https://www.figma.com/design/LsDmLDCdnJqI8UahyqvOD8/BCIERS-Designs?node-id=43352-10423&t=wnfBIGPn15gnsVdZ-4)
loom recording
https://www.loom.com/share/4fb94c84e4da4521bffe137335ec83d3

#### Acceptance Criteria

Given that I am on the operation information page or facility information page
The BCGHGID field has 1 button Issue BCGHGID and an editable text field 

Given that I enter something in the text field
Then the button Issue BCGHGID is disabled 

Given I have issued a BCGHGID or entered one,
Then I can still edit the BCGHGID

~~-[ ] Give CAS staff the ability to fix Operation records and delete unnecessary duplicates in the system~~
- [ ] Ensure that no data is lost: only operations without reports associated can be deleted
- [ ] A django command to be run manually
- [ ] Tests have been written for this command

(UX Design considerations: we might want to rename the 'issue bcghgid/boro id' buttons to 'generate bcghgid/boro id' now that we have an editable bcghgid field, to make it clear)

#### Definition of Done

- [ ] first
- [ ] second
- [ ] third

#### Additional context

- Add any other context about the task here.
- Or here

BCerki:
[...]

Issue #1606, opened 2024-05-05
Delete contacts


zoeyli-46:

BCerki:
A couple questions (probably for @zoeyli-46):

- Do we have a wireframe for this part: "Then I will first be prompted to confirm"? Does the prompt mean a modal?
- In the roles miro board, both industry_user and industry_user_admin have "manage contacts." This ticket is only allowing industry_user_admin to delete. Is that correct?

andrea-williams:
- users should not be allowed to delete themselves or change their role
- only industry_user_admins should be allowed to delete contacts

patrickisaac:
Refinement Notes:
- This particular ticket is about deleting a contact that is not a user

patrickisaac:
Refinement Notes:
- Moving this to the `Backlog`

BCerki:
Re this note `This particular ticket is about deleting a contact that is not a user`. Does this mean contacts that are also users cannot ever be deleted?

BCerki:
FYI, this has some effects on reporting:
- If a contact is assigned as a person responsible in a _draft_ report and the contact is deleted, they're automatically removed from the report (I didn't build this; it's just how this feature interacts with the rest of what's been built)
- If a contact is assigned in a _submitted report_ and then the contact is deleted, the contact stays in the submitted report (can't change anything about a report after submission)
Also, "Person Responsible" doesn't show up in the places assigned list. Not sure if this is a bug or desired.

PR #3561, opened 2025-07-09
Delete contacts 1606

BCerki:
card: https://github.com/bcgov/cas-registration/issues/1606

This PR:
- adds a delete button and conditional props to the SingleStepTasklist (currently the contact form is the only one with a delete button, but it looks like it's needed for https://github.com/bcgov/cas-registration/issues/3482 too, and it was clearer to add to that component vs. trying to wrestle it into ContactForm)
- modal when the delete button is clicked; text and buttons in the modal conditional on whether or not the contact is assigned places
- endpoint to archive contact
- snackbar on grid
- added bc-cas-dev and bc-cas-dev-secondary as contacts (more representative of real app; every user should have a contact record too unless someone later deletes it)
- pytests
- vitests
- added an env variable so we can make the playwright timer longer in local (PB: the default timeout on await somestuff.click() is 5s and is too little for nextjs to compile the page)
- some of the happo diffs are real because we now have a delete button

### DIFF
--- a/bc_obps/common/tests/endpoints/auth/constants.py
+++ b/bc_obps/common/tests/endpoints/auth/constants.py
@@ -433,6 +433,7 @@ ENDPOINTS = {
             "method": "get",
             "endpoint_name": "get_current_user_operator_access_requests",
         },
+        {"method": "patch", "endpoint_name": "archive_contact", "kwargs": {"contact_id": MOCK_UUID}},
     ],
     "authorized_irc_user": [
         {"method": "get", "endpoint_name": "list_operators"},
--- a/bc_obps/registration/api/_contacts/contact_id.py
+++ b/bc_obps/registration/api/_contacts/contact_id.py
@@ -9,6 +9,7 @@ from registration.api.router import router
 from service.contact_service import ContactService
 from service.error_service.custom_codes_4xx import custom_codes_4xx
 from registration.schema import ContactWithPlacesAssigned
+from ninja.types import DictStrAny
 
 
 @router.get(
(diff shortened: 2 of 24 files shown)


andrea-williams (commented):
When I ran the code locally, I logged in as bc-cas-dev and was able to successfully delete the 2 mock contacts who aren't assigned anywhere and everything looked fine in the UI, but in the logs in my FE server I got this error both times:

`app/components/contacts/ContactPage.tsx (21:13) @ ContactPage
 ⨯ Error: Failed to retrieve contact information
    at ContactPage (./app/components/contacts/ContactPage.tsx:29:66)
digest: "2909405295"
  19 |     contactFormData = await getContact(contactId, `/contacts/${contactId}`);
  20 |     if (contactFormData && "error" in contactFormData)
> 21 |       throw new Error("Failed to retrieve contact information");
     |             ^
  22 |   } else {
  23 |     // Retrieves the list of users associated with the operator of the current user
  24 |     userOperatorUsers = await getUserOperatorUsers("/contacts/add-contact");`
  
  Also - a really pedantic note that I even resent myself for raising, but I think there could potentially be confusion in situations where multiple operation reps are assigned per operation. For example, using the default mock data, I removed Bill Blue from all the operations where he was the only operation rep, so that the only remaining places assigned for him were "Banana LFO" & "Bugle SFO", where Bob Brown is also assigned as an operation rep for both of these operations. I still am blocked from deleting Bill Blue's contact, getting the modal with the error message "Before you can delete this contact, please replace them in the places they are assigned with another contact first." This _could_ be confusing for users as is there already another contact assigned, so they might think that they need to have 2 operation reps assigned (which isn't true). 
  
Flagging this as a potential issue, but I also think it could be punted as a future UI enhancement, since the scenario I described isn't captured in the AC for the current ticket.

PR #3573, opened 2025-07-16
feat: past reports dashboard tab

acatchpole:
Card: https://github.com/bcgov/cas-reporting/issues/504 and https://github.com/bcgov/cas-reporting/issues/832
Some changes adjacent to: https://github.com/bcgov/cas-reporting/issues/838
### Changes:
 - Added dashboard tiles and links to past-reports
 - Updated dashboard tile link names to be more accurate (per 832)
 - Updated URL for /reports to reports/current-reports (NOTE: /reports still works to show the current-page, but all of the report actions are broken when using this URL)
 - updated name of function ```get_all_operations_for_user```-> ```get_all_current_operations_for_user``` for clarity 
 - changed name of ```schema/operations.py``` to dashboard.py as this is reporting grid schemas and not just operations
 - added API, service, and related tests to get past reports
 - updated reports page to have tabs accessing /current-reports and /past-reports
 - refactored some columns components of current-reports so they could be shared with past-reports
### Testing:
**NOTE**: I only added one "past report" to the data fixtures, so I recommend commenting out the current year filter (ln.128 of reporting_dashboard_service.py) so that you can see more reports and test their actions.

#### Remaining issue:
i left the page at reporting/reports in. It shows the exact same grid as reporting/reports/current-reports, but none of the actions work from this page as the last domain in the url gets stripped as part of defining the actions' routes. It was from the desire to share all those actions between current-reports and past-reports that I updated the URL in the first place. 
example of problem: 
reporting/reports/current-reports. Click on ```Report history``` and get routed to reporting/reports/report-history. GOOD
reporting/reports. Click on ```Report history``` and get routed to reporting/report-history. BAD. Doesn't exist. Error.

#### Be brutal
This PR includes refactoring of code and changing of some structures. I did not seek advice when i was making these changes. Please let me know if you think I was mucking around with things that I didn't need to or shouldn't have.


bc_obps/reporting/api/reporting_dashboard.py
```diff
@@ -23,7 +28,7 @@
     "/operations",
     response={200: List[ReportingDashboardOperationOut], custom_codes_4xx: Message},
     tags=DASHBOARD_TAGS,
-    description="""Returns a list of operators for the current user, current reporting year, current report version. Populates
+    description="""Returns a list of operations for the current user, current reporting year, current report version. Populates
```
acatchpole, line 31: just a typo fix

bciers/apps/reporting/src/tests/components/dataGrid/model/pastReports/pastReportsColumns.test.tsx
```diff
@@ -0,0 +1,186 @@
...
+    expect(columns[6].headerName).toBe("More Actions");
+    expect(columns[6].width).toBe(120);
+  });
+
+  it("renders an empty string in UpdatedAtCell if report_status is DRAFT", () => {
+    const columns = pastReportsColumns();
+    const params = {
+      row: { report_status: ReportOperationStatus.DRAFT },
+    };
+    const cell = columns[2].renderCell!(params);
```
andrea-williams, line 63: TIL that `renderCell!()` is a thing. Thank you!

andrea-williams (commented):
Saving progress - have reviewed all code, now to manually test in local.

andrea-williams:
honestly, now that I think about it more, it might not even be a logical issue, but just a matter of not having modified the logic for the ReportHistory page, which should be a separate ticket anyway. Sorry if that's the case!

acatchpole:
RE: Wrong year on the report history page -- 

That year (2024) is hard coded right now. I also noticed this issue and created this follow up ticket https://github.com/bcgov/cas-reporting/issues/854

bciers/apps/reporting/src/app/components/datagrid/models/dashboard/dashboardColumns.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+const UpdatedAtCell = ({ row, value }: GridRenderCellParams) =>
+  row.report_status && row.report_status !== ReportOperationStatus.DRAFT
+    ? value
+      ? formatTimestamp(value)
+      : ""
+    : "";
+
+const SubmittedByCell = ({ row }: GridRenderCellParams) =>
+  row.report_status && row.report_status !== ReportOperationStatus.DRAFT
```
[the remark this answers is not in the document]
acatchpole: the report_submitted_by value is actually just the updated_by for the latest report version. If the report version HAS been submitted, then this last updated_by IS who submitted the report. If it is a draft, there is still an updated_by (probably) but the report has never been submitted so we make this value an empty string.

PR #3598, opened 2025-07-22
220 pending invoice

gdalcengio:
- **feat: set action cell to pending invoice creation if no invoice number**
- **test: update action cells with pending invoice creation**


bc_obps/compliance/schema/compliance_report_version.py
```diff
@@ -35,11 +35,18 @@ class ComplianceReportVersionListOut(ModelSchema):
...
     class Meta:
         model = ComplianceReportVersion
         fields = ['id', 'status']
 
+    @staticmethod
+    def resolve_invoice_number(obj: ComplianceReportVersion) -> Optional[str]:
+        if not obj.obligation.elicensing_invoice:
+            return None
+        return obj.obligation.elicensing_invoice.invoice_number
+
```
Sepehr-Sobhani, line 49: You don't need this resolver if you use:
```python
invoice_number: Optional[str] = Field(None, alias="obligation.elicensing_invoice.invoice_number")
```

PR #3603, opened 2025-07-24
225: add HasComplianceReportVersion type

BCerki:
card: https://github.com/bcgov/cas-compliance/issues/225

This PR:
- new HasComplianceReportVersion FE type
- refactor props to use ^
- rename args in the actionHandlers
- rename args in a service (pretty sure this is correct, but flagged with a comment for reviewer)
- update typing
- update vitests
- the `getComplianceSummary` function actually gets a report version, not a summary, but I left the name as-is because it's being used on the `ComplianceSummaryReviewPage`. However, if the reviewer thinks it would be clearer to change it to `getComplianceReportVersion`, happy to.

### DIFF
--- a/bciers/apps/compliance/src/app/components/compliance-summary/manage-obligation/review-compliance-summary/ComplianceSummaryReviewComponent.tsx
+++ b/bciers/apps/compliance/src/app/components/compliance-summary/manage-obligation/review-compliance-summary/ComplianceSummaryReviewComponent.tsx
@@ -20,7 +20,6 @@ export function ComplianceSummaryReviewComponent({
   complianceReportVersionId,
 }: Readonly<Props>) {
   const [errors, setErrors] = useState<string[]>([]);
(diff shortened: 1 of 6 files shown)


bc_obps/compliance/service/payment_instructions_service.py
```diff
@@ -16,19 +16,19 @@ class PaymentInstructionsService:
...
 
         Args:
-            compliance_summary_id: ID of the compliance summary
+            compliance_report_version_id: ID of the compliance summary
 
         Returns:
             Tuple of (PDF data generator, filename, total_size_in_bytes)
         """
         try:
-            context = PaymentInstructionsService._prepare_payment_instructions_context(compliance_summary_id)
```
BCerki, line 31: I think this must be the report version id because that's the name of the arg in `_prepare_payment_instructions_context`

Sepehr-Sobhani (approved):
Great work! 🥇

PR #3610, opened 2025-07-25
Chore: replace datetime.now with timezone.now

Sepehr-Sobhani:


BCerki (changes requested):
Nice work! Thanks for always keeping an eye on things like this.

I see datetime.now in a couple files still:
- bc_obps/service/reporting_year_service.py
- bc_obps/reporting/tests/service/test_reporting_year_service.py

Do we need to do any sort of migration for the records that had timestamps entered with code that was just datetime.now() vs. datetime.now(ZoneInfo("UTC"))? If it was just datetime.now(), then are we missing timezone in our existing records?

Also, it would be nice to have some documentation about how we store all times in UTC (often when people look at timestamps in the db or metabase to check something they've just done, they wonder why the time looks wrong).

Sepehr-Sobhani:
@BCerki Thanks for the thoughtful questions and for always thinking outside the box!

> * bc_obps/service/reporting_year_service.py

- [ ] I believe the use of `datetime.now(ZoneInfo("America/Vancouver"))` here is intentional. That said, I’ll double-check with the team to confirm whether this should be updated.

> bc_obps/reporting/tests/service/test_reporting_year_service.py

This is mocking `.now()` and is already using `tzinfo=timezone.utc`, so we're good there.

> Do we need to do any sort of migration for the records that had timestamps entered with code that was just datetime.now() vs. datetime.now(ZoneInfo("UTC"))? If it was just datetime.now(), then are we missing timezone in our existing records?

I reviewed all the changes and found that the only place where time was saved incorrectly (without timezone awareness) is in the `save_report_sign_off` method. I’ll reach out to the Regi-porting team to confirm whether a migration is needed for those existing records.


> Also, it would be nice to have some documentation about how we store all times in UTC (often when people look at timestamps in the db or metabase to check something they've just done, they wonder why the time looks wrong).

- [ ] Sure

docs/backend/timezone-handling.md
```diff
@@ -0,0 +1,85 @@
...
+
+  return `${date}\n${timeWithTimeZone}`;
+};
+```
+
+## Querying Database
+
+When querying the database directly (e.g., in Metabase, SQL queries, debugging):
+
+1. **Remember you're seeing UTC times**: A timestamp showing `2024-01-15 22:00:00` is 10:00 PM UTC, which is 2:00 PM Pacific Time
```
BCerki, line 67: Is UTC 22:00 always 2pm pacific? Ie, does UTC have daylight savings on the same schedule as us? If not, maybe we could update this to `2:00 PM Pacific Standard (or Daylight, whichever it is) Time`

docs/backend/timezone-handling.md
```diff
@@ -0,0 +1,85 @@
...
+## Common Scenario
+
+### User Submits Form at 2:00 PM Pacific Time
+
+**What happens:**
+
+- Frontend captures: `2024-01-15T14:00:00-08:00` (Pacific Time)
+- Backend receives: `2024-01-15T22:00:00Z` (UTC)
+- Database stores: `2024-01-15 22:00:00+00:00`
+- Frontend displays: "Jan 15, 2024 2:00 PM PST"
```
BCerki, line 80: Thanks for this! And for the docs on timestamps in general

BCerki (approved):
:cartwheeling:

docs/backend/timezone-handling.md
Sepehr-Sobhani: Good call, Thanks! 👍

PR #3614, opened 2025-07-26
Feat 245 handle supplementary report decreased excess emissions

Sepehr-Sobhani:


dleard (approved):
Nice work! I do have two suggestions:

1. For the status of the compliance_report_version, I don't think "obligation not met" makes sense here since this report version has no obligation. I think it makes more sense to use "no obligation or earned credits" here.

2. For the description of the adjustment pushed to elicensing "Compliance Units and/or Payments Applied" doesn't make sense. I think we should come up with something that describes what this adjustment is for. Something like "supplementary report adjustment"?

Everything works great in this PR. I'd be fine with merging this and creating a small tech debt card to handle the status/description semantics.

PR #3619, opened 2025-07-29
265 fix outstanding balance calculation

BCerki:
https://github.com/bcgov/cas-compliance/issues/265

This PR:
- put a negative in front of the adjustment value in the apply compliance unit service (if we're applying units, the balance is always going to go down)
- pytest to check that the apply service is called with a negative number


Sepehr-Sobhani (approved):
🤩

PR #3621, opened 2025-07-29
251 fully met obligation

gdalcengio:
- **feat: Update compliance status to fully met on adjustment creation if obligtion reaches 0**
- **test: test updating of compliance status to fully obligated**
- **chore: update action cell logic and where status gets updated for fully met compliance status**


dleard:
This will be handled in https://github.com/bcgov/cas-compliance/issues/275

PR #3632, opened 2025-07-31
chore: added delete activity fron report raw activity data and also migration

ayeshmcg:
**Card:** https://github.com/bcgov/cas-reporting/issues/850
**Issue:** When reviewing facilities, updates to activities were properly reflected in the report raw activity data, but deletions were not. When an activity was removed from a facility report during the review process, the corresponding raw activity data remained in the database, creating orphaned records that could cause data inconsistencies on the final review page.

**Root Cause**
The set_activities_for_facility_report method in FacilityReportService was only updating the many-to-many relationship between facility reports and activities, but not cleaning up the related ReportRawActivityData records when activities were removed.

**Changes**

1. Added to delete from report raw activity data if activity was deleetd from review facilities
2. Added test to verify that
3. Added migration to fix the existing report raw activity data table
4. Added test for to test that migration

**To Test:**
1. Start an LFO report.
2. Navigate to the Review Facilities page.
- Select any two activities.
- Complete the rest of the report.
- On the Final Review page, you should see both selected activities listed.
3. Return to the Review Facilities page.
- Unselect one of the previously selected activities.
- Complete the report again.
- Navigate to the Final Review page.

4. You should now see data for only the remaining selected activity.


bc_obps/reporting/migrations/0123_cleanup_orphaned_raw_activity_data.py
```diff
@@ -0,0 +1,55 @@
...
+            print(f"Deleted {len(records_to_delete)} orphaned ReportRawActivityData records")
+    else:
+        print("No orphaned ReportRawActivityData records found")
+
+
+def reverse_cleanup_orphaned_raw_activity_data(apps, schema_editor):
+    """
+    This migration is irreversible as we cannot restore deleted orphaned data.
+    """
+    pass
```
andrea-williams, line 41: the preferred method for making a migration non-reversible is to use 

```
    operations = [
        migrations.RunPython(
            code=<function_name>,
            reverse_code=migrations.RunPython.noop,
            elidable=True,
        ),
    ]
```
   
see registration/migrations/0095 for an example. 
    
With the recommended approach, Django knows that the migration hasn't been reversed, whereas with the current script, it thinks it has applied the reversal (even though it did nothing). We should also add `elidable=True` in case we ever squash our migrations.

andrea-williams (commented):
One piece of feedback re: code. Will now try running the code on my local

andrea-williams (approved):
Ran the code locally and followed the suggested testing steps - behaviour observed is as expected.

Please make the suggested changes to the 0123 migration before merging!

ayeshmcg:
Thanks a lot @andrea-williams . Added the requested changes

PR #3636, opened 2025-08-01
chore: fix excess emissions calculation

BCerki:
#3594 

This PR:
- service method to get the excess emission value we want to display on the frontend (which is not the same as the db value for excess emissions for supplementary reports)
- use ^ in both dashboard services so we see the correct display value in both the grid and the individual page


bc_obps/compliance/tests/service/test_compliance_dashboard_service.py
```diff
@@ -305,135 +305,67 @@ def test_get_compliance_report_version_by_id_sets_calculated_fields(
...
         assert result.id == version.id
-        assert result.report_compliance_summary.excess_emissions == Decimal(4.0)
 
+        assert result.excess_emissions == Decimal(4.0)
         assert result.compliance_charge_rate == Decimal("25.00")
         assert result.equivalent_value == Decimal("100.00")  # 4.0 * 25.0
         # assert result.outstanding_balance_tco2e == Decimal("2.0")
         # assert result.outstanding_balance_equivalent_value == Decimal("50.0")  # 2.0 * 25.0
 
-    @patch(
```
BCerki, line 315: I deleted these because they're covered in the new service tests instead now

vivanovbc (approved):
LGTM!

PR #3647, opened 2025-08-06
fix: let operation name in compliance summary grid take from reporting rather than admin

gdalcengio:
https://github.com/bcgov/cas-compliance/issues/173

### DIFF
--- a/bc_obps/compliance/schema/compliance_report_version.py
+++ b/bc_obps/compliance/schema/compliance_report_version.py
@@ -6,7 +6,7 @@ from registration.models import Operation
 
 # report aliases
 OPERATOR_NAME_ALIAS = "compliance_report.report.operator.legal_name"
-OPERATION_NAME_ALIAS = "compliance_report.report.operation.name"
+OPERATION_NAME_ALIAS = "report_compliance_summary.report_version.report_operation.operation_name"
 OPERATION_BCGHG_ID_ALIAS = "compliance_report.report.operation.bcghg_id.id"
 
 # reporting_year aliases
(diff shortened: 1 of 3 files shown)


shon-button (approved):
Perfect fix!

I wonder though if the other "report aliases" should also derive from `report_compliance_summary.report_version.report_`

@dleard ? 👀

dleard:
> Perfect fix!
> 
> I wonder though if the other "report aliases" should also derive from `report_compliance_summary.report_version.report_`
> 
> @dleard ? 👀
> 
> ```
> 
> # report aliases
> OPERATOR_NAME_ALIAS = "compliance_report.report.operator.legal_name"
> OPERATION_NAME_ALIAS = "report_compliance_summary.report_version.report_operation.operation_name"
> OPERATION_BCGHG_ID_ALIAS = "compliance_report.report.operation.bcghg_id.id"
> ```
> 
> If so, suggest we make a new ticket to scope out any changes to `OPERATOR_NAME_ALIAS` that **may** also affect ` def get_compliance_report_versions_for_dashboard`; such as, perhaps:
> 
> ```
>   # Get all compliance report versions for the filtered operations
>             compliance_report_versions = compliance_report_version_queryset.filter(
>                 compliance_report__report__operation_id__in=operations,
>                 compliance_report__report__operator=UserOperatorService.get_current_user_approved_user_operator_or_raise(
>                     user
>                 ).operator,
>             )
> ```

Yes, this makes sense to me. We should be consistent in where we retrieve data from & the compliance report data should be a result of the emissions report data

PR #3648, opened 2025-08-07
Feat: 3482 internal can merge operations

pbastia:


bc_obps/service/tests/operation_service/test_operation_service_bcghgid.py
```diff
@@ -0,0 +1,146 @@
...
+        manual_bcghg_id = "11234567890"
+
+        result = OperationService.generate_bcghg_id(cas_director.user_guid, operation.id, manual_bcghg_id)
+
+        facility.refresh_from_db()
+        operation.refresh_from_db()
+
+        assert operation.bcghg_id.id == manual_bcghg_id
+        assert facility.bcghg_id.id == manual_bcghg_id
+        assert result.id == manual_bcghg_id
```
andrea-williams, line 138: I'm confused by this test. The name of the test suggests that when an LFO receives a BCGHG ID (applied manually), it should only be applied to the operation and not the facilities within it. This is the behaviour that I would expect. But then the actual logic in the test asserts that the facility receives the same BCGHG ID as the operation?

andrea-williams (commented):
Finished reviewing code - will run locally and manually test next.

PR #3650, opened 2025-08-07
Chore/penalty calc tests

dleard:
Tests for #3637 
The tests were split out into a separate branch so the original branch could be merged more quickly to unblock other dependent work.


Sepehr-Sobhani (approved):
🥇

PR #3660, opened 2025-08-11
214 compliance api permissions

BCerki:
card: https://github.com/bcgov/cas-compliance/issues/214


bc_obps/compliance/api/permissions.py
```diff
@@ -0,0 +1,97 @@
...
+        OperationDesignatedOperatorTimeline.objects.filter(
+            operation=operation,
+            start_date__lte=created_at.date(),  # ownership started on or before the report was created
+        )
+        .filter(
+            Q(end_date__gte=created_at.date())
+            | Q(end_date__isnull=True)  # ownership is ongoing or didn't end until after the report was created
+        )
+        .first()
+    )
```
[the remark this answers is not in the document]
Sepehr-Sobhani: If my foggy brain is working correctly, the intention here was to give access to users who used to own that operation, not just those who own it currently. Am I correct, @dleard ?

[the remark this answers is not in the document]
dleard: I had the same thought as Brianna after I wrote[this section](https://github.com/bcgov/cas-registration/blob/develop/bc_obps/compliance/service/compliance_dashboard_service.py#L39-L56), and then forgot about it.
We save both the operation_id and operator_id on the emissions report, so it may not be necessary for compliance to use the timeline, since we've saved who the report belonged to at the time on the report model.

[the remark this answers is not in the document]
Sepehr-Sobhani: @dleard Should we follow the same approach for implementing RLS policies?

bc_obps/compliance/tests/api/test_permissions.py
```diff
@@ -0,0 +1,130 @@
...
+    @pytest.mark.parametrize("role,status,expected_validity", role_status_access_tuples)
+    def test_compliance_version_ownership_from_url(self, role, status, expected_validity):
+        # Setup
+        user_operator = make_recipe("registration.tests.utils.user_operator", role=role, status=status)
+
+        compliance_report_version = make_recipe(
+            "compliance.tests.utils.compliance_report_version",
+            compliance_report__report__operator=user_operator.operator,
+            compliance_report__created_at='2024-06-15',
+            # compliance_report__report__operation=operation,
```
Sepehr-Sobhani, line 34: leftover!

bc_obps/compliance/tests/api/test_permissions.py
```diff
@@ -0,0 +1,130 @@
...
+
+        mock_request = MagicMock()
+        mock_request.resolver_match.kwargs = {"test_id": compliance_report_version.id}
+        mock_request.current_user = user_operator.user
+
+        is_valid = validator_under_test(mock_request)
+
+        assert is_valid == expected_validity
+
+    def test_version_ownership_from_url_denies_access_if_no_timeline_record(self):
```
Sepehr-Sobhani, line 47: We are not checking timeline!

bc_obps/compliance/tests/api/test_permissions.py
```diff
@@ -0,0 +1,130 @@
...
+
+        validator_under_test = check_compliance_version_ownership_in_url("test_id")
+
+        mock_request = MagicMock()
+        mock_request.resolver_match.kwargs = {"test_id": compliance_report_version.id}
+        mock_request.current_user = user
+
+        is_valid = validator_under_test(mock_request)
+
+        assert not is_valid
```
Sepehr-Sobhani, line 59: I suggest using `assert is_valid is False` to be more explicit about the expected value of `is_valid`.

bc_obps/compliance/tests/api/test_permissions.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def test_version_ownership_from_url_denies_access_if_operation_transferred(self):
```
Sepehr-Sobhani, line 61: Is this test still necessary? We only check direct relationships, not those through the timeline table.

bc_obps/compliance/tests/api/test_permissions.py
Sepehr-Sobhani: Same for other assertions too.

Sepehr-Sobhani (approved):
🤩

bc_obps/compliance/tests/api/test_permissions.py
Sepehr-Sobhani: 😮 How about `is_valid is False`?

BCerki: One day I'll remember how write python

PR #3661, opened 2025-08-11
Fix: 844 LFO Facility-level activity data is overwritten by operation-level activity updates

pbastia:


acatchpole (approved):
Code looks good. Works as expected. This is certainly better behaviour than before as activity data that has been input is protected. 

The fact that activities that have been removed from a facility report still get added back when you go through and save again on the Review Operation Information page is behaviour that could still be improved upon in later work.

PR #3664, opened 2025-08-12
chore: correct calculation for earned credit cap limit

dleard:
Fixes an issue with how the earned credit limit/cap was being calculated.

The issue:
When an adjustment is made via a supplementary report. It should be considered when calculating the 50% limit, but not count against the cap remaining (adjustments made when applying credits should lower the cap remaining, not supplementary adjustments).

The fix:
Updated the get adjustments function to differentiate between those two adjustment types and fetch them in the appropriate places in the function logic.


Sepehr-Sobhani (approved):
🤩

PR #3670, opened 2025-08-13
Penalty fee date

BCerki:
follow up to card: https://github.com/bcgov/cas-compliance/issues/73

This PR:
- adds fee_date to the compliance_penalty model
- uses fee_date instead of invoice.created_at date when creating the invoice (it's going to be the same date, but this is more consistent with how the obligation invoice is done)
- pytests for ^


Sepehr-Sobhani (approved):
:shipit:

PR #3672, opened 2025-08-14
287 - emissions attributable for compliance

BCerki:
card: https://github.com/bcgov/cas-compliance/issues/287


BCerki:
Finished in a different PR

PR #3642, opened 2025-08-06
chore: 219 compliance routing rules

shon-button:
Addresses: 
[219](https://github.com/bcgov/cas-compliance/issues/219)

### 🚀 Impact:
- ⚠️ Added `user_compliance_access_service` get user compliance access status for registered operations; compliance report ownership, and compliance report status
- Added middleware `withRuleHasComplianceRouteAccessBCeID` for industry user rules- 
- Used user compliance access status in `withRuleHasComplianceRouteAccessBCeID` rules
- Added middleware `withRuleHasComplianceRouteAccessIDIR` for cas user rules
- Relocated component based routing rules using `redirect()` to associated middleware rules

**BCeID**
   -   `Compliance` app restricted to user-operator with **registered operation**
   - `compliance-summaries/[compliance_report_version_id]/*` routes restricted to user's compliance report's operator- **ownership**
   -    `compliance-summaries/[compliance_report_version_id]/(no-obligation)` routes restricted to compliance report version with **status** **"No obligation or earned credits"**
   -    `compliance-summaries/[compliance_report_version_id]/(manage-obligation)` routes restricted compliance report version with **status** **"Obligation not met"**
   - **redirect suite** for the status **"Obligation not met"** rule:
       - `apply-compliance-units` (**ApplyComplianceUnitsPage**) redirect to→ `Compliance Summaries` when getComplianceAppliedUnits.can_apply_compliance_units !== true
   -    `compliance-summaries/[compliance_report_version_id]/(request-issuance)` routes restricted to compliance report version with **status** **"Earned credits"**
   - **redirect suite** for the status **"Earned credits"** rule:
       - `request-issuance-of-earned-credits` (**RequestIssuanceOfEarnedCreditsPage**) redirect to→ `track-status-of-issuance` when issuance is **ISSUANCE_REQUESTED | APPROVED | DECLINED**
       - `request-issuance-review-summary` (compliance-summary/request-issuance/review-compliance-summary/**ComplianceSummaryReviewPage**) redirect to→ `track-status-of-issuance` when issuance is **ISSUANCE_REQUESTED | APPROVED | DECLINED**
       - `track-status-of-issuance` (**TrackStatusOfIssuancePage**) redirect to→ `request-issuance-of-earned-credits` when issuance is **CREDITS_NOT_ISSUED | CHANGES_REQUIRED**
 

**IDIR**
   - **redirect suite** for route `review-by-director`:
[...]

PR #3684, opened 2025-08-18
143 compliance tile visibility

BCerki:
card: https://github.com/bcgov/cas-compliance/issues/143

This PR:
- new endpoint and service to check if an operator has any registered and regulated operations (I couldn't just replace the one we already have that checks for registered operations because it's used to show the 'action required' message)
- compliance tile shown conditionally based on results of ^
- migration to update tiles
- use the new endpoint in the middleware too since users who shouldn't see the tile also shouldn't be able to navigate directly to compliance
- pytests


bciers/apps/compliance/src/middlewares/withRuleHasComplianceRouteAccess.ts
shon-button, line 1: @BCerki 
Since PR #3642 is deleting this page; perhaps, we can revert your midleware changes and create a tech debt similar to the other middleware to dos- something like "Compliance/Middleware - Tile Visibility"?

PR #3686, opened 2025-08-19
chore: redirect from /reports to current-reports

acatchpole:
Card: https://github.com/bcgov/cas-reporting/issues/857

### Changes
added a redirect from /reporting/reports -> /reporting/reports/current-reports


acatchpole:
implemented a different way in https://github.com/bcgov/cas-registration/pull/3698

PR #3689, opened 2025-08-19
Chore: 3482 missing edit bcghgid function

pbastia:


bc_obps/registration/api/_facilities/_facility_id/bcghg_id.py
```diff
@@ -26,3 +26,18 @@ def facility_bcghg_id(
...
+    tags=FACILITY_TAGS,
+    description="""Clears the BCGHG ID for the facility if it is already set.""",
+    auth=authorize('cas_director'),
+)
+def delete_facility_bcghg_id(request: HttpRequest, facility_id: UUID) -> Literal[200]:
+    FacilityService.clear_bcghg_id(
+        get_current_user_guid(request),
+        facility_id,
+    )
+    return 200
```
pbastia, line 43: this *could* be a "204 No Content" but we've standardized this app to return a "200 OK" on success, which is also completely acceptable

bciers/libs/components/src/form/widgets/BcghgIdWidget.tsx
```diff
@@ -77,65 +98,97 @@ const BcghgIdWidget: React.FC<WidgetProps> = ({
...
+      >
+        edit
+      </Link>{" "}
+      to enter a BCGHGID
+    </div>
+  );
+
+  return (
+    <div className="flex flex-col w-full" style={{ paddingLeft: "14px" }}>
+      <div className="flex items-center w-full">
```
pbastia, line 149: This is the interesting stuff to align content horizontally then vertically

andrea-williams:
Manual testing:

on a Facility page, I (as a cas_director) tried manually entering a BCGHG ID in the wrong format (too many digits), and the error message appeared, as expected. Then when I corrected it and clicked "Save", the BCGHG ID was applied to the facility successfully, but the error message remained on the page.

Screenshot 2025-08-21 at 2 35 20 PM

bc_obps/service/tests/test_facility_service.py
```diff
@@ -555,6 +555,24 @@ def test_generate_bcghg_id_with_existing_manual_id():
...
+            operator=approved_user_operator.operator,
+            status=Operation.Statuses.REGISTERED,
+        )
+        facility = baker.make_recipe(
+            'registration.tests.utils.facility',
+            operation=operation,
+            bcghg_id=baker.make_recipe('registration.tests.utils.bcghg_id'),
+        )
+
+        FacilityService.clear_bcghg_id(approved_user_operator.user.user_guid, facility.id)
```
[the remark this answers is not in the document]
pbastia: Interesting! Not a Copilot error, more of a pilot error.
I copy-pasted the logic, trusting the `get_if_authorized` method implemented by the same service, from the generate BCGHGID endpoint.

This is actually an existing bug, I'll fix it right away. Fantastic eyes and sharp review!!

PR #3700, opened 2025-08-20
fix: selected activities without emissions reports get deselected when creating a supplementary report

pbastia:
Refactored some tests to mirror database structure a bit better


ayeshmcg (approved):
🎉

PR #3709, opened 2025-08-25
268 request issuance button

BCerki:
card: https://github.com/bcgov/cas-compliance/issues/268

The button already correctly stays disabled during form submission (`const isDisabled = disabled || isSubmitting;` in the button component, and we have a vitest and manual testing to confirm). The reason it looks broken is because it takes a second for the redirect to take effect, so there's a moment where the button is renabled because the submission is complete but the redirect isn't. To fix it, I moved `setIsSubmitting(false) into the error condition so that the button doesn't get renabled if the submission was successful. This pattern appears in some of the other compliance components.

Also fixed the typo in the button text.


Sepehr-Sobhani (approved):
🤩

PR #3711, opened 2025-08-25
chore: fix bandit dependencies in CI

BCerki:
This PR:
- adds additional dependencies to bandit in pre-commit to get past pbr error, e.g. https://github.com/bcgov/cas-registration/actions/runs/17213663266/job/48837263110?pr=3709


acatchpole (approved):
i'm going to judge this by the fact that CI worked here

PR #3713, opened 2025-08-26
224 - use operator mailing instead of physical address on invoices

BCerki:
card: https://github.com/bcgov/cas-compliance/issues/224

This PR:
- uses operator.mailing_address instead of physical address in the invoice service
- there was a duplicate address in the fixtures which made it harder to see which address was being used, so I updated the fixtures to be unique


Sepehr-Sobhani (approved):
🥇

Issue #3574, opened 2025-07-16
Add help text on user profile page


zoeyli-46:
#### Description:
Users go to their user profile and expect the ability to change their email but the email field is not editable. To address this, add instructional text to the email field redirecting users to their contact page. 

[Wireframe](https://www.figma.com/design/LsDmLDCdnJqI8UahyqvOD8/BCIERS-Designs?node-id=43436-10129&t=SusUXkVXWNXUa6tw-4)

<img width="1392" height="538" alt="Image" src="https://github.com/user-attachments/assets/e5930f61-7e7b-4558-876b-e3829ebd825a" />

#### Acceptance Criteria:

Given I am an external user and I am on my user profile 
Then I see help text underneath the email field 
"This email is used to log in. To change your login email, contact ghgregulator@gov.bc.ca. To change the email you are contacted with, edit the email in your contact details page. "

ghgregulator@gov.bc.ca is hyperlinked mail to link
contact details page is hyperlinked to the users contact details page

Given I click on the contact link
Then I am redirected to MY contact information page. (not the contacts grid but the users contact information page where they can edit their contact email). I.e., If I am Mia then clicking the link takes me to Mia's contact information page (not the contacts grid with all the other contacts)

Given I am requesting access to an operator and filling in the user profile page in the request access workflow,
I do not see this help text

#### Development Checklist:

- [x] vitests for help text
- [x] help text
- [ ] Meets the DOD

#### Recommended testing steps:
- [ ] sign in with your IDIR and confirm that your profile page loads ok (there won't be any help text)
- [ ] sign in as `bc-cas-dev` (an industry user that already exists in the DEV database, or your local db) and confirm that the full help text message appears on the profile page

<img width="747" height="609" alt="Screenshot 2025-09-11 at 8 36 42 AM" src="https://github.com/user-attachments/assets/7e71b779-a8f0-4d1c-95ea-adf30af009fd" />

- [ ] sign in with another Business BCeID account that doesn't already exist in the db (can create your own if needed) and confirm that the profile page loads fine, but without any help text yet

<img width="950" height="548" alt="Screenshot 2025-09-11 at 8 37 58 AM" src="https://github.com/user-attachments/assets/9598e3a3-2244-4977-8cdc-da46b3cee353" />

#### Definition of Ready (Note: If any of these points are not applicable, mark N/A)

- [ ] User story is included
- [ ] User role and type are identified
- [ ] Acceptance criteria are included
[...]

PR #3714, opened 2025-08-26
Help text on User Profile page

andrea-williams:
ticket #3574

### DIFF
--- a/bciers/apps/administration/app/components/profile/ProfileForm.tsx
+++ b/bciers/apps/administration/app/components/profile/ProfileForm.tsx
@@ -1,5 +1,5 @@
 "use client";
-import { useEffect, useState } from "react";
+import { useState } from "react";
 import { Alert } from "@mui/material";
 import { actionHandler } from "@bciers/actions";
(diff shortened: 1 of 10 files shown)


bciers/apps/administration/app/components/profile/ProfileForm.tsx
```diff
@@ -55,6 +48,75 @@ export default function ProfileForm({ formData, isCreate }: Props) {
...
+            // fallback if no contact found
+            setEmailHelpText(emailHelpTextFirstClause);
+          }
+        } catch (e) {
+          setEmailHelpText(emailHelpTextFirstClause);
+        }
+      }
+    };
+    fetchHelpText();
+  }, [isCreate]);
```
pbastia, line 118: I think we could avoid this whole 'useEffect', 'getSession' and logic by running this in a server side component. We can then pass the contact id as a parameter for the client side page to render the help text or not.
The server side component might also be able to make use of the `getUserRole` utility method that already exists

pbastia (approved):
Fantastic!

PR #3720, opened 2025-08-27
290 datetime fields

BCerki:
card: https://github.com/bcgov/cas-compliance/issues/290

This PR changes date/datetime fields from DateTimeField to DateField in compliance models and updates all related code to use date objects instead of datetime objects. The change affects three key models: ElicensingInvoice.due_date, ElicensingLineItem.fee_date, and ElicensingAdjustment.adjustment_date.

- Changes model fields from DateTimeField to DateField for date-only data
- Updates service layer to parse ISO date strings using date.fromisoformat() instead of datetime.fromisoformat()
- Modifies test fixtures and comparison logic to use date objects consistently
- There are still some datetime types in the BCCR workflows. These seem unrelated to our db, so I left them as-is


Sepehr-Sobhani (approved):
Neat! 🤩 
Just to share something I learned, you could also use `timezone.localdate()` instead of `timezone.now().date()`.

PR #3725, opened 2025-08-28
chore: remove fixed height and margin from compliance units component…

nataliepiguliak:
… layout


Sepehr-Sobhani (changes requested):
I’d argue against this change. We’re `using min-h-[62vh]` across other components in Compliance, and removing it would bring the buttons much closer to the input field instead of keeping them near the bottom of the page. Also, removing `className="mt-8"` would push the error component too far away from the buttons.

Image

Sepehr-Sobhani (approved):

PR #3728, opened 2025-08-29
chore: lfo final review

ayeshmcg:
Card: https://github.com/bcgov/cas-reporting/issues/800
**Changes:**
- Added endpoints to fetch Final Review data specifically for the LFO workflow.
- Added an endpoint to retrieve FacilityReport data for an individual facility.
- Updated the Final Review page to show a facility grid when the operation type is LFO.
- Created a reusable component, FinalReviewFacilityGrid, to display all facilities in a grid with pagination and search functionality.
- Added a page to display Final Review data for an individual facility, including an anchor so that navigating back to the Final Review page scrolls to the grid instead of the top of the page.

**To Test:**
1. Start a report with operation type LFO, fill in the required details, and navigate to the Final Review page. You should see a facility grid under “Report Information” instead of accordions.
image

2. Click View Details on a facility row to navigate to the Final Review section for that individual facility report.
image


bciers/apps/reporting/src/app/utils/getLfoFinalReviewData.ts
```diff
@@ -0,0 +1,15 @@
+import { actionHandler } from "@bciers/actions";
+
+export async function getLfoFinalReviewData(
```
pbastia, line 3: this file can also go

pbastia (changes requested):
Fantastic work! I think we can simplify the LFO/non-LFO case by handling everything on the server, this will remove frontend logic and API endpoints

pbastia (approved):
Looks good! I'll let you be the judge of whether you want to put some of these comments in a tech debt card or fix them before merging. Approved!

ayeshmcg:
@pbastia  updated and added context for back url

PR #3730, opened 2025-08-29
314 - remove reporting fields from compliance rjsf schemas

BCerki:
card: https://github.com/bcgov/cas-compliance/issues/314

This PR:
- removes reporting data from rjsf schemas
- removes the met-obligation rjsf schema (not being used anywhere)
- remove data we're no longer showing from the API return
- renamed `calculate_computed_value_excess_emissions` from `display` because we're no longer displaying excess_emissions (but we are still using the function; the excess_emissions/their delta is used to calculate the equivalent value, which we do still  show)


shon-button (approved):
💥

PR #3739, opened 2025-09-02
304 - internal declined note is conditional

BCerki:
card: https://github.com/bcgov/cas-compliance/issues/304

This PR:
- the wording of internal declined note on the track status of issuance page is conditional based on how the declined happened (analyst requested supp report vs. director declined). This PR adds that conditionality
- update vitests


Sepehr-Sobhani (approved):
🤩

PR #3740, opened 2025-09-03
chore: fix review changes page

ayeshmcg:
Card: https://github.com/bcgov/cas-reporting/issues/479
**Changes:** 
1. Fixed Production Data section in Review changes page and FInal review
2. Fixed New Entrant data in Review changes page
3. Fixed: Changing operation Name if SFO, or changing Facility name now doesnt show facilities as added/deleted, instead shows a field Facility name with edited value
4. Changed Change Review to Review Changes
5. Added Save button 
6. Disabled save and continue button if no data
7. Moved Compliance alert on the top and added condition to render it
8. If Registration purpose changed then just show, Review operation information and Reason for change
9. Added logic for Facility name in the backend.


bc_obps/reporting/service/report_review_changes_service.py
```diff
@@ -137,9 +138,84 @@ def process_change(
...
+
+        # --- Special handling for facility_reports ---
+        if "facility_reports" in previous_data and "facility_reports" in current_data:
+            old_facilities = previous_data["facility_reports"]
+            new_facilities = current_data["facility_reports"]
+
+            removed = set(old_facilities.keys()) - set(new_facilities.keys())
+            added = set(new_facilities.keys()) - set(old_facilities.keys())
+
+            if len(removed) == 1 and len(added) == 1:
```
andrea-williams, line 167: is this the part that handles the name change of a facility?

I'm suspicious about the hard-coded value of 1. What if multiple facilities get renamed?

bciers/apps/reporting/src/app/components/changeReview/ChangeReviewForm.tsx
```diff
@@ -7,19 +7,22 @@ import { ReviewChanges } from "@reporting/src/app/components/changeReview/templa
...
+  isReportingOnly?: boolean;
 }
 
 export default function ChangeReviewForm({
   versionId,
   initialFormData,
   navigationInformation,
   changes,
+  showChanges = true,
+  isReportingOnly,
```
andrea-williams, line 25: did you mean to set the default value of `true` on the optional `isReportingOnly` param instead of `showChanges` ? (which is required, so the default value will always be overwritten)

andrea-williams (approved):
Approving based on review of the code. I didn't run the code locally because I'm not familiar with what changes were requested in QA testing.

PR #3742, opened 2025-09-03
feat: add automatic overdue penalty flow for industry users

nataliepiguliak:


BCerki (commented):
Nice work! There was one part of the AC that's outdated and needs a change, but the bulk of this looks great!

shon-button (approved):

PR #3770, opened 2025-09-09
330 - support for handling multiple penalties

BCerki:
card: https://github.com/bcgov/cas-compliance/issues/330

This PR:
- adds penalty type to model
- filters by this new type in endpoints
- renames `get_penalty_data` for clarity
- previously we had an enum for penalty type; replaced its usage with the choices
- pytest updates
- pytest for `get_automatic_overdue_penalty_data`


shon-button (approved):
🥇

PR #3784, opened 2025-09-10
feat: add cc field for ghg regulator in email service

Sepehr-Sobhani:


BCerki (approved):
:cartwheeling: Looks great! 
image

PR #3795, opened 2025-09-11
chore: change default Alert icon color on Director Review page

nataliepiguliak:


nataliepiguliak:
@shon-button sorry, I didn't add the note, with the new AlertNote changes, using the alertType prop also changes the background. But I discussed with design, we only needed to change the icon color to yellow, and that’s what I implemented

shon-button:
> @shon-button sorry, I didn't add the note, with the new AlertNote changes, using the alertType prop also changes the background. But I discussed with design, we only needed to change the icon color to yellow, and that’s what I implemented

@nataliepiguliak 

Yes, agreed, it is the AlertIcon we are updating; perhaps, we can refactor AlertNote to also be the central place to associate a fill color with each type?  
I'll leave it to you to decide :)

Sepehr-Sobhani (approved):
🌟

PR #3805, opened 2025-09-12
chore: add environment detection utilities and integrate in middleware and token APIs

Sepehr-Sobhani:
This is to prevent middleware vitest tests from failing locally.


shon-button (approved):
✨ 
Nice and tidy! :)

PR #3816, opened 2025-09-16
chore/817 adding alert note for missing product on emissions allocation page

pbastia:
PR that does a couple of things:

- adds a `has_missing_products` flag to the facility emissions allocation API response
- uses this flag to display an alert


bc_obps/reporting/tests/service/test_report_product_service.py
```diff
@@ -1,14 +1,13 @@
...
 from reporting.models.report_product import ReportProduct
 from reporting.service.report_product_service import ReportProductService
 from model_bakery.baker import make_recipe
 
-pytestmark = pytest.mark.django_db
 
-
-class TestReportProductService:
-    def setup_method(self):
+class TestReportProductService(TestCase):
```
pbastia, line 9: I had to change the test from pytest to django to be able to use "assertQuerySetEquals"

acatchpole (approved):
works great. a needed imporvement with the response definition

PR #3819, opened 2025-09-16
chore: used get flow in final review

ayeshmcg:
**Card:** https://github.com/bcgov/cas-reporting/issues/841
**Changes:**
1. Added flowHelpers utility to fetch the reporting flow based on the flow type.
2. Updated the Final Review page to use getFlow from flowHelpers.
3. Refactored the Final Review form to remove duplicated code.

### DIFF
--- a/bciers/apps/reporting/src/app/components/finalReview/FinalReviewForm.tsx
+++ b/bciers/apps/reporting/src/app/components/finalReview/FinalReviewForm.tsx
@@ -58,7 +58,7 @@ export const FinalReviewForm: React.FC<Props> = ({
               key={facilityKey}
               facilityName={facilityReport.facility_name}
               facilityData={facilityReport}
-              showReportingOnlyConditions={!isReportingOnly}
+              showWhenNotReportingOnly={!isReportingOnly}
             />
           ),
         )}
--- a/bciers/apps/reporting/src/app/components/changeReview/templates/FacilityReportChanges.tsx
+++ b/bciers/apps/reporting/src/app/components/changeReview/templates/FacilityReportChanges.tsx
@@ -135,7 +135,7 @@ export const FacilityReportChanges: React.FC<FacilityReportChangesProps> = ({
         facilityName={facilityName}
         facilityData={facilityData.facilityData}
         isAdded={true}
-        showReportingOnlyConditions={true}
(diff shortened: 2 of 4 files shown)


bciers/apps/reporting/src/app/components/finalReview/FinalReviewForm.tsx
```diff
@@ -66,13 +58,87 @@ export const FinalReviewForm: React.FC<Props> = ({
               key={facilityKey}
               facilityName={facilityReport.facility_name}
               facilityData={facilityReport}
-              showReportingOnlyConditions={!isReportingOnly()}
+              showReportingOnlyConditions={!isReportingOnly}
```
andrea-williams, line 61: I'm confused by this part (and maybe it's just the naming that's confusing). But if the operation is reporting-only, I would expect `isReportingOnly` to be true, and therefore I would expect `showReportingOnlyConditions` to be sent a value of `true`, not the inverse. Should the `!` be there?

bciers/apps/reporting/src/app/components/taskList/reportingFlows.ts
```diff
@@ -26,17 +28,32 @@ export const reportingFlows: {
...
+/**
+ * Resolves the appropriate ReportingFlow based on operation type and registration purpose.
+ * Optionally includes new cases like POTENTIAL_REPORTING_OPERATION and OPTED_IN_OPERATION.
+ *
+ * @param operationType - The type of operation ("Single Facility Operation" or "Linear Facilities Operation").
+ * @param registrationPurpose - The registration purpose of the report.
+ * @param includeNewCases - Whether to include the new flows for POTENTIAL_REPORTING_OPERATION and OPTED_IN_OPERATION.
+ * @returns The corresponding ReportingFlow enum value.
+ * @throws Error if the operation type or registration purpose cannot be resolved.
+ */
```
andrea-williams, line 40: I love that you added documentation that we didn't previously have. Thank you!

andrea-williams:
Ran locally, created a report for a regulated SFO and the final review page worked as expected. Also tested with a Reporting Only LFO, which also worked :)

andrea-williams (approved):

PR #3821, opened 2025-09-16
837 allocation validator

acatchpole:
Card: https://github.com/bcgov/cas-reporting/issues/873

## Changes
Validates that emissions allocated add up to emissions reported. This is done per facility report, and checks each emission category, basic and fuel-excluded.
If allocation methodology is "Not Applicable", then skips validation for that facility.

## Existing issue
Having a link in the error that directs them to the appropriate allocation of emissions page where the user can fix the cause of the error would be very friendly. But to do this requires an update to the error component. That work will be done in [this ticket](https://github.com/bcgov/cas-reporting/issues/910).

There might also be a need for further validation of allocations: 
 - Validating that emissions allocated to products in different category types (basic, excluded) make sense (can't have 50000tonnes allocated to product1 in an excluded category when only 10000tonnes total were allocated in the basic categories)
 - Some special case for Pulp&Paper emissions and ensuring that required products are present and allocated to
But if these cases are found to be required, they can be added on to this validator later, or a new validator can be created to handle them.

## Testing
Proceed through the report work flow, and input valid values. Input emission data for an activity. You can create invalid allocations two different ways.
 A):
   1. proceed through to the allocation-of-emissions page and allocate all emissions to 100% and save. 
   2. navigate back to an activity and change an emission amount value and save.
   3. Go directly to the reporting/reports/[version]/final-review page by inputting the URL, thereby skipping the allocation of emissions page.
   4. Proceed through the rest of the report
   5. Hit "Submit" button
    - Expect to see error "Facility {UUID} Category {categoryID} Allocation Mismatch: **Emissions allocated for {facility_name} in '{emission_category_name}' category do not match reported emissions.**"
 
 B):
 
   - Same as above but instead of doing ANY allocations and then going back to make changes, you just skip the allocation page entirely.

### DIFF
--- a/bc_obps/reporting/service/report_validation/validators/report_emission_allocation_validator.py
+++ b/bc_obps/reporting/service/report_validation/validators/report_emission_allocation_validator.py
@@ -33,9 +33,9 @@ def validate(report_version: ReportVersion) -> dict[str, ReportValidationError]:
                 total_allocated = sum(product.allocated_quantity for product in category.products)
                 if total_allocated != category.emission_total:
                     errors[
-                        f"facility_{fr.facility_id}_category_{category.emission_category_id}_allocation_mismatch"
+                        f"allocation_mismatch_facility_{fr.facility_id}_category_{category.emission_category_id}"
                     ] = ReportValidationError(
                         Severity.ERROR,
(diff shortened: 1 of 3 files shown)


ayeshmcg (approved):
🎉

PR #3840, opened 2025-09-18
Chore 308 prepare backend for compliance prod

Sepehr-Sobhani:


bc_obps/bc_obps/settings.py
```diff
@@ -80,22 +80,13 @@
 BCCR_CLIENT_ID = os.getenv("BCCR_CLIENT_ID")
```
BCerki, line 80: I encountered something in my email PR that I'm also seeing here. If I change my env ENVIRONMENT to 'prod', as bc-cas-dev I can login but I keep getting redirected to the profile page. I see a 401 in my logs:
```
[17:54:05] WARNING | basehttp | "GET /api/common/dashboard-data?dashboard=bciers HTTP/1.1" 401 26
```
Is that expected?

Sepehr-Sobhani: @BCerki Did you switch the environment to prod before running the migrations, or afterward?

PR #3842, opened 2025-09-18
chore: fixed review changes bug for lfo

ayeshmcg:
Fixes a bug in the review changes logic for LFO reports.
**To Test**
1. Submit an initial LFO report.
2. Create a supplementary report.
3. Add a new facility or edit an existing one.
4. Verify that all facility changes are correctly reflected in the review changes view.


bc_obps/reporting/schema/report_final_review.py
```diff
@@ -388,22 +386,29 @@ def resolve_operation_emission_summary(obj: ReportVersion) -> Optional[dict]:
...
+        fields = [
+            "report_type",
+            "is_latest_submitted",
+            "reason_for_change",
+            "status",
+        ]
+
+
+class FinalReviewVersionSchema(BaseReportVersionSchema):
+    facility_reports: Union[Dict[str, FacilityReportSchema], List[FacilityReportLFOSchema]] = {}
```
pbastia, line 400: I'm not a big fan of having surprise (non-deterministic) types for our variables. Maybe something we can refactor at a later point

PR #3850, opened 2025-09-19
fix: update declined issuance note to include appeal information and …

nataliepiguliak:
…registry links


shon-button:
@nataliepiguliak 

Regression test (?) for `external - request credits - Change of ID`, implemented in (?), fails as the message displayed after analyst request change of id is the initial instructions:

image

image

Note: Figma design details for `external - request credits - Change of ID`
image

shon-button:
@nataliepiguliak 

Test for `external - request credits - Declined`  fails to display decline message:

image

image

PR #3857, opened 2025-09-22
Fix: review changes with lots of facilities

pbastia:


ayeshmcg:
@pbastia  rebased this

PR #3858, opened 2025-09-22
fix: update alert icon color to use BC_GOV_YELLOW constant

nataliepiguliak:


BCerki (approved):
Colour is correct when viewing as analyst: 
image
and as director: 
image

PR #3860, opened 2025-09-22
chore: fix zero comparison for earned_credits to avoid records where amount is zero

dleard:
One line fix: Cast the decmial credited_emissions to integer before comparing with zero


BCerki (approved):
:cartwheeling: 
Created a report with a tiny earned credit: 
image
And it was correctly treated as no earned credits: 
image

Suggest a tech debt card to add a test for this later
Sidenote I had to pull in a couple changes from https://github.com/bcgov/cas-registration/pull/3857/files to access the compliance summary page and submit the report

PR #3889, opened 2025-10-01
chore: remove sub dashboards

shon-button:


bc_obps/common/migrations/0010_dashboarddata.py
```diff
@@ -7,9 +7,9 @@ def load_dashboard_fixtures(apps, schema_editor):
     from django.core.management import call_command
 
     fixture_files = [
-        'common/fixtures/dashboard/registration/external.json',
-        'common/fixtures/dashboard/reporting/external.json',
-        'common/fixtures/dashboard/reporting/internal.json',
+        "common/fixtures/dashboard/bciers/external.json",
```
BCerki, line 10: We should avoid changing migrations that have already been applied. I think we can make all the changes we need in a new migration using `reset_dashboard_data`. But if that isn't possible, we should check the prod migrations locally before merging this so we find out about any problems before release time.

bciers/apps/compliance/src/middlewares/withResponseCompliance.ts
```diff
@@ -6,16 +6,20 @@ import { appName } from "../middleware";
 export const withResponseCompliance: MiddlewareFactory = () => {
   return async (request: NextRequest) => {
     const { pathname } = request.nextUrl;
+    console.log(pathname);
```
BCerki, line 9:
```suggestion

```

BCerki (commented):
Nice work! Some notes:
1. I can still click on the clean growth user access and report a problem tiles (but there's no subdashboard, so maybe that's fine?)
2. This testing instruction didn't behave as expected:
Navigate to http://localhost:3000/compliance/my-compliance/compliance-summaries
I get not found (but I can get there via http://localhost:3000/compliance/compliance-administration/compliance-summaries)
Image

3. Some of the CI checks are failing

4. This is unrelated to subdashboards so not suggesting we do anything in this PR, but for later: 
Transfers crumbs are inconsistent with the other dashboards:

Image

Image

bc_obps/common/migrations/0010_dashboarddata.py
shon-button: @BCerki 

Due to the deleted .json files, the migrations do not complete without the changes made.

shon-button:
> Nice work! Some notes:
> 
> 1. I can still click on the clean growth user access and report a problem tiles (but there's no subdashboard, so maybe that's fine?)
> 2. This testing instruction didn't behave as expected:
>    Navigate to http://localhost:3000/compliance/my-compliance/compliance-summaries
>    I get not found (but I can get there via http://localhost:3000/compliance/compliance-administration/compliance-summaries)
> 
> Image
> 3. Some of the CI checks are failing
> 4. This is unrelated to subdashboards so not suggesting we do anything in this PR, but for later:
>    Transfers crumbs are inconsistent with the other dashboards:
> 
> Image Image

@BCerki 
Valid point, I was just focused on sub-dashboards, but agree with you that it makes sense to remove all tile href - done.

bc_obps/common/migrations/0010_dashboarddata.py
BCerki: Yes, it's the same check we do during releases, either with the dag (I see a draft PR for that, so it might be close) or manually: https://github.com/bcgov/cas-registration/pull/3882/files#:~:text=1.%20If,30,

BCerki (commented):
All tiles no longer link, but I'm still having trouble with this url:

Image
If I go in via the dashboard link I end up at `http://localhost:3000/compliance/compliance-administration/compliance-summaries`

shon-button:
> 10\. http://localhost:3000/compliance/compliance-administration/compliance-summaries

@BCerki 

Note: http://localhost:3000/compliance/my-compliance/compliance-summaries is not a valid link and should not be part of test steps or tile link url.
The link to compliance summaries is now: 

http://localhost:3000/compliance/compliance-administration/compliance-summaries

PR #3895, opened 2025-10-02
feat: consider both payments and adjustments when determining final p…

nataliepiguliak:
…enalty accrual end date


BCerki:
> > * pay the rest with credits
> 
> Hey @BCerki, thanks for a review, just for clarification, you described the scenario where credits bring the obligation to 0. In step “4. pay the rest with credits,” the credits refer to applying compliance units, right?

Yes, I mean apply compliance units

PR #3900, opened 2025-10-03
385/setup compliance e2e

fviduya:
Setup E2E for compliance module
Card: https://github.com/bcgov/cas-compliance/issues/385


bc_obps/reporting/management/commands/load_reporting_fixtures.py
```diff
@@ -56,23 +60,23 @@ def load_reports(self):
...
-                    submit_report_from_fixture(report_version, UUID('ba2ba62a-1218-42e0-942a-ab9e92ce8822'))
+                    for report_version in report_versions:
+                        submit_report_from_fixture(report_version, UUID('ba2ba62a-1218-42e0-942a-ab9e92ce8822'))
 
-            # create supplmentary report
-            for report in Report.objects.filter(operation_id=operation_ids_to_submit[0]):
-                ReportVersionService.create_report_version(report)
+                # create supplementary report
+                for report in Report.objects.filter(operation_id=operation_ids_to_submit[0]):
+                    ReportVersionService.create_report_version(report)
```
fviduya, line 82: This block is not needed for the compliance E2E test, I thought we can save resources if we skip it. We can update this for reporting if needed.

bc_obps/registration/api/e2e_test_setup.py
```diff
@@ -24,12 +32,16 @@ def setup(
             with RlsManager.bypass_rls():
                 if truncate_only:  # only truncate the tables
                     call_command('truncate_dev_data_tables')
+                    apply_specific_compliance_migrations()
```
fviduya, line 35: I wasn't sure if I should include this here or just on line 39?

bc_obps/registration/api/e2e_test_setup.py
```diff
@@ -24,12 +32,16 @@ def setup(
...
                     return HttpResponse("Test setup complete.", status=200)
                 if load_only:  # only load the data
                     call_command('load_fixtures', workflow)
+                    apply_specific_compliance_migrations()
```
fviduya, line 39: I wasn't sure if I should include this here or just on line 44?

bc_obps/registration/api/e2e_test_setup.py
BCerki: I think we shouldn't need this at all. Migrations 2 and 3 will run as part of the backend job already, and since the changes to truncate now preserve the compliance_peiord and charge_rate models' data, we don't have to reload.

bc_obps/reporting/management/commands/load_reporting_fixtures.py
BCerki: Sounds great, thanks for saving resources!

BCerki (commented):
Nice work! I have a couple of inline comments, and I suggest temporarily making a change to a file in compliance to force the job to run (looks like nx-affected skips it now). If it fails, that's fine, we can just make it a non-mandatory check, but if it gets stuck or something and blocks us, we'll want to sort that out before merging.

.github/workflows/main.yaml
```diff
@@ -104,6 +106,17 @@ jobs:
...
+    needs:
+      [build-backend, build-dashboard-e2e, build-compliance, check-nx-affected]
+    uses: ./.github/workflows/test-nx-project-e2e.yaml
+    with:
+      docker_compose_file: "docker-compose-bciers-apps.yaml"
+      image_url: "ghcr.io/bcgov/cas-compliance-frontend"
+      is_nx_affected: ${{contains(needs.check-nx-affected.outputs.NX_AFFECTED_E2E, 'compliance') || github.ref_name == 'develop' || github.ref_name == 'main' || github.head_ref == 'develop' || github.head_ref == 'main'}}
+      nx_app_port: 7000
+      nx_project: compliance
+    secrets: inherit
```
BCerki, line 119: If this is copy/pasted from the other e2e jobs, I have no concerns about the sonarcloud fail

BCerki (approved):
Nice work! 🤸

.github/workflows/main.yaml
fviduya: Yes it was! 😄

PR #3901, opened 2025-10-03
363 change required message

BCerki:
card: https://github.com/bcgov/cas-compliance/issues/363

This PR:
- updates the Action cell to add the new action conditionally
- new alert note component
- alert shows conditionally if the status is changes required
- vitests for ^


bciers/apps/compliance/src/app/components/compliance-summary/request-issuance/request-issuance-of-earned-credits/RequestIssuanceOfEarnedCreditsComponent.tsx
shon-button, line 1: @BCerki 

For your consideration:
- perhaps the alert note visibility reset and form submit should happen only when the user has completed typing a full and "changed" from previous BCCR ID (rather than on initial field typing)?  Instead of hiding on field typing and submitting on the same BCCR ID:
image

PR #3902, opened 2025-10-06
Compliance/Automatic Emails - Notice of Compliance Obligation Due #368

BCerki:
card: https://github.com/bcgov/cas-compliance/issues/368

This PR:
- migration for the compliance due email template
- services to send the email (I called it in process_obligation_integration so that if the invoice generation fails, the emails don't go out)
- pytests for ^
- testing instructions in the card


shon-button (approved):
Extra brilliant! 
✨

PR #3906, opened 2025-10-07
chore: update mock data

BCerki:
card: https://github.com/bcgov/cas-compliance/issues/388


bc_obps/reporting/management/commands/load_reporting_fixtures.py
```diff
@@ -75,8 +75,14 @@ def load_reports(self, workflow):
                     )
 
                     for report_version in report_versions:
+                        report_version.report_operation.operator_legal_name = (
```
shon-button, line 78: @BCerki 

I like the strip()!

Perhaps better to use use `.replace(' - name from admin', '').strip()` instead of `.replace('name from admin', '').strip()` so we do not have dangling dashes?

image



Something like: 
```
...
from reporting.fixtures.utils import submit_report_from_fixture  # adjust if different import

SUFFIX = " - name from admin"
def _strip_admin_suffix(value: str | None) -> str | None:
    if value is None:
        return None
    return value.replace(SUFFIX, "").strip()

class Command(BaseCommand):
...

for report_version in report_versions:
    ro = report_version.report_operation
    ro.operator_legal_name = _strip_admin_suffix(ro.operator_legal_name)
    ro.operation_name = _strip_admin_suffix(ro.operation_name)
    ro.save()
```

shon-button (approved):
Nice catch and refactor! 
👍👍
One comments for your consideration...

PR #3907, opened 2025-10-07
chore: remove compliance example e2e test

fviduya:
Additional work for #3900 to remove the `example.spec.ts` to prevent CI from failing, and happo from taking screenshots.


BCerki (approved):
🤸

PR #3914, opened 2025-10-15
392 credit handlers with multiple supp reports

BCerki:
partially handles card: https://github.com/bcgov/cas-compliance/issues/392

This PR:
- updates the increase credit handler to handle multiple reports when credits are requested
- handler can handle credits not being requested once (multiple times will be handled in https://github.com/bcgov/cas-compliance/issues/399)
- pytests for ^


nataliepiguliak (approved):
LGTM, great work!

PR #3917, opened 2025-10-16
fix for track status of issuance link

BCerki:
Action cell link is fine. I think the problem is in the middleware. The earned credits redirect looks like this:
```
redirect: makeRuleRedirect<RuleContext>(HUB_SUMMARIES_PATH, (id, ctx) => {
      if (ctx.redirectToTrackStatusById[id]) {
        return AppRoutes.RI_TRACK_STATUS;
      }
      if (ctx.redirectToRequestIssuanceCreditsById[id]) {
        return AppRoutes.RI_EARNED_CREDITS;
      }
      return undefined; // → fall back to hub
    }),
``` 
vs `redirect: makeRuleRedirect(HUB_SUMMARIES_PATH),` elsewhere

I got the broken page to load with the changes in this PR, but I haven't done any testing yet to see if I broke any of the other pages. ~maybe the redirect in the next config was working for the other places but not track status because of the resolver?

### DIFF
--- a/bciers/next.config.base.js
+++ b/bciers/next.config.base.js
@@ -30,7 +30,6 @@ const baseConfig = {
         destination: "/",
         permanent: true,
       },
-      // brianna?
       {
         source: "/compliance-administration",
         destination: "/compliance-administration/compliance-summaries",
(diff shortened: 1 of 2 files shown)


bciers/next.config.base.js
shon-button: no change need here, I believe.

shon-button (approved):

PR #3934, opened 2025-10-22
Feat/76 late submission penalty

nataliepiguliak:


bc_obps/compliance/service/penalty_calculation_service.py
```diff
@@ -327,4 +444,189 @@ def create_penalty(cls, obligation: ComplianceObligation) -> CompliancePenalty:
...
+        total_penalty = Decimal('0.00')
+        days_late = (final_accrual_date - accrual_start_date).days + 1
+        invoice = refresh_result.invoice
+
+        # Generate list of monthly accrual dates (1st of each month)
+        monthly_accrual_dates = []
+        current_date = accrual_start_date
+
+        # Add the first accrual date
+        monthly_accrual_dates.append(current_date)
```
shon-button: @nataliepiguliak 

I think ` monthly_accrual_dates.append(current_date)` could record an accrual when the accrual window is "empty" (i.e., accrual_start_date > final_accrual_date):
Also, it seems the loop runs one extra iteration after the last valid date potentially and the conditional append: `if next_month <= final_accrual_date` might exclude the final accrual date's month.

Suggest potential improvement:

```
# Generate dates while within range
while current_date <= final_accrual_date:
    monthly_accrual_dates.append(current_date)
    
    # Move to 1st of next month
    if current_date.month == 12:
        current_date = date(current_date.year + 1, 1, 1)
    else:
        current_date = date(current_date.year, current_date.month + 1, 1)

```

bc_obps/compliance/service/penalty_calculation_service.py
```diff
@@ -327,4 +444,176 @@ def create_penalty(cls, obligation: ComplianceObligation) -> CompliancePenalty:
...
+            "penalty_status": obligation.penalty_status,
+            "penalty_type": CompliancePenalty.PenaltyType.LATE_SUBMISSION,
+            "penalty_charge_rate": (get_rate_for_date(final_accrual_date) * 100) / 12,
+            "days_late": days_late,
+            "accumulated_penalty": accumulated_penalty.quantize(Decimal('0.01'), rounding=ROUND_HALF_UP),
+            "accumulated_compounding": accumulated_compounding.quantize(Decimal('0.01'), rounding=ROUND_HALF_UP),
+            "total_penalty": total_penalty.quantize(Decimal('0.01'), rounding=ROUND_HALF_UP),
+        }
+
+        return result
```
Sepehr-Sobhani: Are we using this return value anywhere, or planning to use it in the future?
If not, we can simply return the newly created penalty when `persist_penalty_data` is True; otherwise, return None. That way, in the `create_late_submission_penalty` function, we don’t need to query like this:

```python
CompliancePenalty.objects.filter(
            compliance_obligation=obligation,
            penalty_type=CompliancePenalty.PenaltyType.LATE_SUBMISSION,
        ).latest('id')
```

bc_obps/compliance/tests/service/test_penalty_calculation_service.py
```diff
@@ -296,3 +297,419 @@ def test_get_automatic_overdue_penalty(self, mock_refresh_data):
...
+            self.obligation.compliance_report_version.compliance_report.compliance_period.compliance_deadline
+            + timedelta(days=1)
+        )
+
+        assert mock_calculate.call_count == 1
+        _, kwargs = mock_calculate.call_args
+        assert kwargs["obligation"] == self.obligation
+        assert kwargs["accrual_start_date"] == expected_start_date
+        assert kwargs["final_accrual_date"] == expected_payment_deadline
+        assert kwargs["persist_penalty_data"] is True
```
Sepehr-Sobhani: Nit: We have a cleaner approach for this:
```python
mock_calculate.assert_called_once_with(
            obligation=self.obligation,
            accrual_start_date=expected_start_date,
            final_accrual_date=expected_payment_deadline,
            persist_penalty_data=True,
        )
```

bc_obps/compliance/service/penalty_calculation_service.py
```diff
@@ -327,4 +450,174 @@ def create_penalty(cls, obligation: ComplianceObligation) -> CompliancePenalty:
...
+                    reference_date = date(y, 6, 15)
+                case 10 | 11 | 12:
+                    reference_date = date(y, 9, 15)
+
+            for rate in interest_rates:
+                if rate.start_date <= reference_date <= rate.end_date:
+                    return rate.interest_rate
+            return ElicensingInterestRate.objects.filter(is_current_rate=True).first().interest_rate  # type: ignore[union-attr]
+
+        return get_rate_for_date
```
dleard: I think we can simplify this function:
I don't think we need the resolver function at all & can just define the get_rate_for_date() function that checks the date of compounding and returns a .get() from the ElicensingInterestRate table for the matching reference_date
```suggestion
      def get_rate_for_date(date_to_check: date) -> Decimal:
          m = date_to_check.month
          y = date_to_check.year
          match m:
              case 1 | 2 | 3:
                  reference_date = date(y - 1, 12, 15)
              case 4 | 5 | 6:
                  reference_date = date(y, 3, 15)
              case 7 | 8 | 9:
                  reference_date = date(y, 6, 15)
              case 10 | 11 | 12:
                  reference_date = date(y, 9, 15)
          return ElicensingInterestRate.objects.get(start_date__lte=reference_date,end_date__gte=reference_date).interest_rate
```
This removes a bunch of unnecessary queries that check for the existence of interest rates before the date matching & still fails appropriately if an interest rate does not exist for the .get() query as there should always be 1 and only 1 interest rate record match for a reference_date

bc_obps/compliance/service/penalty_calculation_service.py
```diff
@@ -74,6 +76,92 @@ def get_automatic_overdue_penalty_data(cls, compliance_report_version_id: int) -
...
+                        elicensing_invoice=obligation.elicensing_invoice,
+                        line_item_type=ElicensingLineItem.LineItemType.FEE,
+                    )
+                )
+            )
+            .order_by('-received_date')
+            .first()
+        )
+
+        # Get interest rate for the penalty period, fallback to current rate
```
dleard: What is this interest rate used for? It's possible that the interest rate is not the same from the accrual start date to the date of final payment

bc_obps/compliance/service/penalty_calculation_service.py
```diff
@@ -211,17 +299,26 @@ def calculate_penalty(
...
             )
 
         last_calculation_day = final_accrual_date if final_accrual_date else cls.TODAY
 
+        submission_date = cls._get_last_supplementary_submission_date(obligation)
+        if submission_date is not None:
+            effective_start_date = submission_date + timedelta(days=31)
+        else:
+            effective_start_date = invoice.due_date + timedelta(days=1)
+
```
dleard: Is this block accounting for the large penalty on an unpaid supplementary version and changing the start date of the large penalty? I'm not sure this logic supports the need here. Let's chat about this one

bc_obps/compliance/service/penalty_calculation_service.py
```diff
@@ -281,6 +379,22 @@ def calculate_penalty(
 
         return result
 
+    @classmethod
+    def _get_last_supplementary_submission_date(cls, obligation: ComplianceObligation) -> date | None:
```
dleard: I don't think this function is necessary, we can just check the date of the obligation

bc_obps/compliance/service/penalty_calculation_service.py
```diff
@@ -329,4 +452,153 @@ def create_penalty(cls, obligation: ComplianceObligation) -> CompliancePenalty:
...
+
+        # Get supplementary submission date and calculate payment deadline (30 days after submission)
+        supplementary_submission_date = cls._get_last_supplementary_submission_date(obligation)
+        payment_deadline = supplementary_submission_date + timedelta(days=30)  # type: ignore[operator]
+
+        # Calculate late submission penalty
+        penalty_record = cls._calculate_late_submission_penalty(
+            obligation=obligation,
+            accrual_start_date=late_submission_start_date,
+            final_accrual_date=payment_deadline,  # type: ignore[arg-type]
```
dleard: The final accrual date is not the payment_deadline, it is the date on which the obligation is paid to zero. Setting this to the payment_deadline of 30 days past the submission date ensures that this penalty stops accruing after 30 days whether it is paid or not.

bc_obps/compliance/service/penalty_calculation_service.py
```diff
@@ -329,4 +452,153 @@ def create_penalty(cls, obligation: ComplianceObligation) -> CompliancePenalty:
...
+            monthly_accrual_dates.append(current_date)
+
+            # Move to 1st of next month
+            if current_date.month == 12:
+                current_date = date(current_date.year + 1, 1, 1)
+            else:
+                current_date = date(current_date.year, current_date.month + 1, 1)
+
+        # Calculate penalty for each monthly accrual date
+        for accrual_date in monthly_accrual_dates:
```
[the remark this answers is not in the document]
nataliepiguliak: thank you! will do

PR #3945, opened 2025-10-29
chore: fix comma separated bug

ayeshmcg:
Card: https://github.com/bcgov/cas-reporting/issues/842
**Issue:**
On the final review page, some sections were not displaying numbers with comma separators because those values were returned as strings from the backend.

**Fix:**
- Updated the Final Review schema.
- Changed fields from int/decimal to float (decimal was returning values as strings).
- Updated the Activity view to use NumberField.
- Added a check to ensure that the year is not displayed with comma separators.

**Testing Steps:**
1. Start a report and enter numeric values with at least 4 digits.
2. Navigate to the Final Review page.
3. Verify that:
     - Numeric values are displayed with comma separators.
     - Initial Compliance Period and Compliance Period in the Compliance Summary section are displayed without commas.


acatchpole (approved):
LGTM.

I think a test that proves the function of the isYear flag could be called for, but I'll leave that up to you

Issue #3590, opened 2025-07-18
Bug: immutable migrations script fails when making a new release and adding a tag


pbastia:
#### Describe the Bug:

On release branches, CI fails because adding a tag makes the script think migrations behind this tag were modified

#### Impact: 3 (false positives can be dangerous and lead devs to ignore real failures. The issue itself can be bypassed easily)

<img src="https://www.getzephyr.com/sites/default/files/riksk%20matrix.jpg" alt="Bug Matrix" width="250" height="250"/>

#### Steps to reproduce the behaviour:

1. Step #1
2. Step #2
3. Step #3

#### Additional information:

Browser: [e.g. chrome, safari]
Version: [e.g. 22]

#### Additional Context

-

PR #3949, opened 2025-10-30
fix: ensure tags already released to main are only immutable

joshgamache:
Addresses #3590. Updates the immutable migration check to run against `main` branch, as we consider any migrations that have made it that far immutable. This should correct the false-failure in ShipIt when doing a release. 

## Changes 🚧

- Check migrations against `main` branch's latest tag rather than current branch's latest tag.


ayeshmcg (approved):
🎉

PR #3952, opened 2025-10-30
chore: make updated_by trigger nullable

Sepehr-Sobhani:


dleard (approved):
Fixes an issue with the update trigger being empty string when processing system driven updates

PR #3966, opened 2025-11-05
Feat:898 dont collect 2024 production data for other years

pbastia:
#### Backend:
- Reporting API V2 
- Standard reporting form response
- Builder object to put together those responses

#### Frontend:
Base pattern to support multiple schemas/uiSchemas for multiple years.
The pattern can probably be iterated over and improved over time, with the use of factory methods

### DIFF
--- a/bciers/apps/reporting/src/data/jsonSchema/productionData.tsx
+++ b/bciers/apps/reporting/src/data/jsonSchema/productionData.tsx
@@ -50,11 +50,7 @@ const buildProductionDataSchemaDefault = (
     definitions: {
       productionDataItem: {
         type: "object",
-        required: [
-          "annual_production",
-          "production_data_apr_dec",
-          "production_methodology",
-        ],
+        required: ["annual_production", "production_methodology"],
         properties: {
           product_id: {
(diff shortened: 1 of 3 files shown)


bciers/apps/reporting/src/data/jsonSchema/productionData.tsx
```diff
@@ -122,7 +121,29 @@ export const buildProductionDataSchema = (
...
 
-export const productionDataUiSchema: UiSchema = {
+export const buildProductionDataSchema = (
+  reporting_year: number,
+  compliance_period_start: string,
+  compliance_period_end: string,
+  product_selection: string[],
+  facility_type: string,
+) => {
+  if (reporting_year == 2024)
```
ayeshmcg: just a nit bit 
reporting_year === 2024

ayeshmcg (changes requested):
@pierre This looks good to me — the 2024 report is working as expected.

For reports from other years, I'm seeing a Form Validation Error 
image
It appears to be caused by production_data_apr_dec being included in the required fields in the default schema.error

PR #3978, opened 2025-11-07
fix: duplicate attachments on storage 

pbastia:
…he storage


andrea-williams:
I guess in theory it shouldn't be an issue because of the hash string that should be applied to filenames, but that seems to have been applied inconsistently...

Issue #372, opened 2023-12-05
e2e tests for all major user workflows


Dianadec:
### Description

We need to make separate tickets for all the things related to this (may be more than what's listed below depending on how many features are complete before this ticket is pulled in):

### Dev Tasks
- Identify User Flows:
Use Figma to identify key user flows and interactions 
- Create Test Scenario Cards:
Based on the identified user flows, create cards for test scenarios that cover critical paths and user interactions.
Include positive and negative test cases to ensure the application behaves as expected.

#### Architectural Design

Using a balanced approach between between testing a page using each role or testing a role for each page.
- Initial tests of critical pages with each user role to ensure comprehensive coverage of essential functionalities. 
- Subsequent tests to target specific roles interacting with specific pages, especially for pages with role-specific features.

1. **Test a Page Using Each Role:**
   - **Pros:**
     - Comprehensive Coverage: This approach ensures that each page is tested with every user role, providing a comprehensive view of how different roles interact with each page.
     - Identifies Cross-Role Issues: Helps identify issues related to role-based permissions and ensures that a user with a specific role can access and interact with the expected content on each page.
   - **Cons:**
     - Repetition: The test scenarios for each role may overlap, leading to some repetition in test code. Maintenance may become more challenging as the number of roles and pages increases.

2. **Test a Role for Each Page:**
   - **Pros:**
     - Clear Test Focus: Each test focuses on a specific role, making it easier to understand and manage the purpose of each test.
     - Isolation of Issues: If a test fails, it is easier to isolate the issue to a specific role and page combination.
   - **Cons:**
     - Limited Cross-Role Coverage: This approach may not provide as comprehensive coverage across all user roles for each page, potentially missing cross-role issues.

### Page by Role Tests: `e2e\pages`

- [x] [home page/login-logout-login](https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=49133752)

### Role by Page Tests: `e2e\workflows`

[Flow of Current Implementation](https://miro.com/app/board/uXjVN6uBj9Y=/)

#### Industry operator flows:

 **bceidbusiness\new user** interactions in the system:
[Ticket 800](https://github.com/orgs/bcgov/projects/122/views/2?pane=issue&itemId=51945351)

 **bceidbusiness\industry_user** interactions in the system:
[Ticket 505](https://github.com/orgs/bcgov/projects/122/views/2?filterQuery=e2e&pane=issue&itemId=49836206)

[...]

PR #3984, opened 2025-11-13
Compliance/Automatic Emails - Supplementary Report Submitted after Deadline with increased emissions #372

BCerki:
card: https://github.com/bcgov/cas-compliance/issues/372

This PR:
- migration to add the email template
- service to send email and retryable
- retryable is called conditionally when processing obligations
- testing instructions in card

### DIFF
--- a/bc_obps/common/migrations/0097_supplementary_report_submitted_after_deadline_template_template.py
+++ b/bc_obps/common/migrations/0097_supplementary_report_submitted_after_deadline_template_template.py
@@ -47,7 +47,7 @@ def reverse_supplementary_report_submitted_after_deadline_template(apps, schema_
 
 class Migration(migrations.Migration):
     dependencies = [
-        ('common', '0095_notice_of_penalty_now_accruing_email_template'),
+        ('common', '0096_V4_4_0'),
     ]
 
     operations = [


bc_obps/compliance/service/elicensing/elicensing_obligation_service.py
```diff
@@ -129,7 +139,13 @@ def process_obligation_integration(cls, obligation_id: int) -> None:
 
                 # If successful, update the compliance status
                 ComplianceReportVersionService.update_compliance_status(obligation.compliance_report_version)
-                retryable_send_notice_of_obligation_due_email.execute(obligation.id)
+                # If the new obligation was created by a supplementary report and the obligation deadline has passed, we send a specific email. Otherwise, we send the general one about the obligation being due.
+                if obligation.compliance_report_version.is_supplementary and cls._has_obligation_deadline_passed(
+                    obligation.obligation_deadline
+                ):
+                    retryable_notice_of_supplementary_report_post_deadline_increases_emissions.execute(obligation.id)
```
Sepehr-Sobhani: Since this is related to a supplementary report, should we call this task somewhere in the `supplementary_version_service.py` file instead?

BCerki: I'm going to answer this question with a question! I put the email logic in `process_obligation_integration` because I wanted to be sure that we only sent the email if the invoice creation and data sync succeeded. If I were to move the email sending function to the supp report service and call it after `ElicensingObligationService.handle_obligation_integration(obligation.id, compliance_report.compliance_period)`, would a failure in `handle_obligation_obligation_integration` always prevent sending of the email? The internet is telling me conflicting things about how errors in retryables work.

Sepehr-Sobhani (approved):
Great work as always! 🤩

PR #3991, opened 2025-11-13
387 - fix sorting/filtering on status column of compliance summary page

gdalcengio:
- **fix: Move status name update to backend before filtering as annotated field display_name to fix filtering and sorting**
- **test: fix tests related to the status column of compliance summaries data grid**

To test:

- create a few different reports
- go to compliance summary page and attempt to sort and filter on the status column

### DIFF
--- a/bc_obps/compliance/enums.py
+++ b/bc_obps/compliance/enums.py
@@ -22,19 +22,3 @@ class ComplianceTableNames(Enum):
 class ComplianceInvoiceTypes(Enum):
     OBLIGATION = "obligation"
     AUTOMATIC_OVERDUE_PENALTY = "automatic overdue penalty"
-
-
-class ComplianceSummaryStatus(Enum):
-    OBLIGATION_NOT_MET = "Obligation not met"
-    OBLIGATION_FULLY_MET = "Obligation fully met"
-    OBLIGATION_PENDING_INVOICE_CREATION = "Obligation pending invoice creation"
-    EARNED_CREDITS = "Earned credits"
-    NO_OBLIGATION_OR_EARNED_CREDITS = "No obligation or earned credits"
-
-
-class IssuanceStatus(Enum):
-    CREDITS_NOT_ISSUED = "Credits Not Issued in BCCR"
-    ISSUANCE_REQUESTED = "Issuance Requested"
-    CHANGES_REQUIRED = "Changes Required"
-    APPROVED = "Approved"
-    DECLINED = "Declined"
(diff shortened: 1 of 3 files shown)


bc_obps/compliance/service/compliance_dashboard_service.py
```diff
@@ -270,3 +272,48 @@ def _annotate_for_schema(
...
+                    status=ComplianceSummaryStatus.EARNED_CREDITS.value,
+                    issuance_status=IssuanceStatus.CHANGES_REQUIRED.value,
+                    then=Value("Earned credits - changes required"),
+                ),
+                When(status=ComplianceSummaryStatus.EARNED_CREDITS.value, then=Value("Earned credits")),
+                default=F("status"),
+                output_field=CharField(),
+            ),
+        )
+        return cast(QuerySet[ComplianceReportVersion], annotated)
```
Sepehr-Sobhani: Since we have some business logic here, I suggest adding a test to cover all the cases in this function, just to make sure we always end up with the correct status.

bc_obps/compliance/enums.py
```diff
@@ -22,3 +22,19 @@ class ComplianceTableNames(Enum):
...
+    EARNED_CREDITS = "Earned credits"
+    NO_OBLIGATION_OR_EARNED_CREDITS = "No obligation or earned credits"
+
+
+class IssuanceStatus(Enum):
+    CREDITS_NOT_ISSUED = "Credits Not Issued in BCCR"
+    ISSUANCE_REQUESTED = "Issuance Requested"
+    CHANGES_REQUIRED = "Changes Required"
+    APPROVED = "Approved"
+    DECLINED = "Declined"
```
Sepehr-Sobhani: I suggest using the model choices from the `ComplianceEarnedCredit` and `ComplianceReportVersion` data models. This way, we won’t have to update multiple places if a status changes later.:
```python
class IssuanceStatus(models.TextChoices):
    CREDITS_NOT_ISSUED = ('Credits Not Issued in BCCR',)
    ISSUANCE_REQUESTED = ('Issuance Requested',)
    CHANGES_REQUIRED = ('Changes Required',)
    APPROVED = ('Approved',)
    DECLINED = ('Declined',)

class ComplianceStatus(models.TextChoices):
    OBLIGATION_NOT_MET = "Obligation not met"
    OBLIGATION_FULLY_MET = "Obligation fully met"
    OBLIGATION_PENDING_INVOICE_CREATION = "Obligation pending invoice creation"
    EARNED_CREDITS = "Earned credits"
    NO_OBLIGATION_OR_EARNED_CREDITS = "No obligation or earned credits"
    SUPERCEDED = "Superceded"
```

PR #3999, opened 2025-11-14
Feat: upgrade Node.js from 20.11.0 to 24.11.1

Sepehr-Sobhani:
- Update .tool-versions to specify Node.js 24.11.1
- Update Dockerfile to use node:24.11.1 base image
- Update @types/node to ^24.10.1 (latest available for Node 24)
- Update yarn.lock with rebuilt native modules
- Update documentation to reflect new Node.js version
- Add `transpilePackages` to `next.config.base.js` based on: https://viclafouch.github.io/mui-tel-input/docs/getting-started/#nextjs-integration

### DIFF
--- a/bciers/next.config.base.js
+++ b/bciers/next.config.base.js
@@ -8,7 +8,7 @@ const baseConfig = {
   experimental: {
     serverActions: {
       bodySizeLimit: "25mb",
-    }
+    },
   },
   transpilePackages: ["mui-tel-input"],
   modularizeImports: {


bciers/yarn.lock
```diff
@@ -6247,7 +6247,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@types/node@npm:*, @types/node@npm:^20.11.25, @types/node@npm:^20.8.4":
+"@types/node@npm:*, @types/node@npm:^20.8.4":
   version: 20.14.2
   resolution: "@types/node@npm:20.14.2"
   dependencies:
```
pbastia: Looks like something still keeps references to node 20 types

Sepehr-Sobhani: We can bump it to version `6.0.1` instead of the latest version, because version 7 has breaking changes and would also require upgrading the MUI package: https://github.com/viclafouch/mui-tel-input/releases/tag/v7.0.0

Sepehr-Sobhani:
> I submitted reports, submitted supp reports, paid invoices, requested credits, and tried to register an operation. Hit a couple errors:
> 
> * when trying to register an operation:
> 
> Image (both my test files were 3.9kb)
> 
> * when clicking back in the operator page:
> 
> Image

@BCerki I wasn’t able to replicate the first issue. @pbastia , could you please test it on your end to see if you can reproduce it?
The second issue is an existing one and will require more effort, as there may be other places affected as well. I’ll create a separate ticket for it.

PR #4001, opened 2025-11-14
Chore/416 outstanding balance refactor

dleard:
Replaces instances where we were using the `outstanding_balance` (which includes FAA interest) value of the invoice incorrectly (ie: when back calculating the outstanding tCO2e value from the $ value). Also adds FAA interest as a separate line item on the invoice & adds it to the total owing amount.

To test the outstanding_balance -> invoice_fee_balance change, any of the Equivalent value in tCO2e values from the summaries page or manage obligation page should not be affected by an invoice that has interest. Changing the outstanding_balance value on the invoice (which includes FAA interest) should have no effect on the tonnes CO2e shown on the summaries page or bottom of the manage obligation page.

Other changes in this PR:

Added an FAA Interest Balance column to the invoices view:
Screenshot from 2025-11-19 14-15-05

Added (Does not include interest) to the track payments page and bottom of the manage obligation page when referencing the outstanding $ amount of an obligation:
Screenshot from 2025-11-19 14-13-13
Screenshot from 2025-11-19 14-12-18

Added "as of invoice date" to the "Excess Emissions" column header for clarity on the compliance summaries page:
Screenshot from 2025-11-19 14-11-02

### DIFF
--- a/bc_obps/compliance/schema/elicensing_invoice.py
+++ b/bc_obps/compliance/schema/elicensing_invoice.py
@@ -69,7 +69,15 @@ class ElicensingInvoiceListOut(ModelSchema):
 
     class Meta:
         model = ElicensingInvoice
-        fields = ["id", "invoice_number", "outstanding_balance", "due_date", "is_void", "last_refreshed"]
+        fields = [
+            "id",
+            "invoice_number",
+            "outstanding_balance",
+            "invoice_interest_balance",
+            "due_date",
+            "is_void",
+            "last_refreshed",
+        ]
 
 
 class ElicensingInvoiceFilterSchema(FilterSchema):
@@ -81,12 +89,3 @@ class ElicensingInvoiceFilterSchema(FilterSchema):
     invoice_total: Optional[str] = Field(None, json_schema_extra={'q': 'invoice_total__icontains'})
     total_payments: Optional[str] = Field(None, json_schema_extra={'q': 'total_payments__icontains'})
     total_adjustments: Optional[str] = Field(None, json_schema_extra={'q': 'total_adjustments__icontains'})
-        fields = [
-            "id",
-            "invoice_number",
-            "outstanding_balance",
-            "invoice_interest_balance",
-            "due_date",
-            "is_void",
-            "last_refreshed",
-        ]


bc_obps/compliance/tests/service/test_compliance_report_version_service.py
```diff
@@ -262,7 +262,8 @@ def test_calculate_outstanding_balance_tco2e_with_existing_invoice(self):
         # Link an ElicensingInvoice with an outstanding balance
         invoice = baker.make_recipe(
             'compliance.tests.utils.elicensing_invoice',
-            outstanding_balance=Decimal("100.00"),
+            outstanding_balance=Decimal("105.00"),
```
BCerki: b

bciers/apps/compliance/src/app/data/jsonSchema/manageObligation/payObligationTrackPaymentsSchema.tsx
```diff
@@ -38,7 +38,9 @@ export const createPayObligationTrackPaymentsSchema = (): RJSFSchema => ({
     ),
     payment_status_note: readOnlyStringField(),
     outstanding_balance: readOnlyNumberField("Outstanding Balance:"),
-    equivalent_value: readOnlyNumberField("Equivalent Value:"),
+    equivalent_value: readOnlyNumberField(
+      "Equivalent Value (Does not include interest):",
```
BCerki: same note as above

BCerki (commented):
Nice work on this! A few suggestions:

- We also use outstanding_balance in some of the emails (part of `_prepare_obligation_context`), and since they usually say something like `a compliance obligation for excess emissions of {{tonnes_of_co2}} at an equivalent amount of  {{outstanding_balance}} is now payable.`, I think we mean the invoice_fee_balance
- could update the model descriptions of outstanding_balance/invoice_fee_balance for extra clarity
- I see outstanding_balance in some of the apply compliance units services and I think here it means something different (e.g. `outstanding_balance = obligation_data.equivalent_value`) than when we get it from elicensing. Maybe we could use a different variable name for clarity?

**Manual testing results**
 I created some interest: 

Image
Correct numbers on the invoice but the formatting is off: 

Image

Correct summary grid but heading looks a bit off (this is probably from the accessibility/zoom settings I use, so may not need fixing)

Image

After I made a payment I lost the interest on the invoice:
Image

Image

If I restarted testing and made an adjustment I also lost the interest:

Image

Image

admcfajn (commented):
Looks good to me, approved thanks @dleard

PR #4007, opened 2025-11-18
chore: fix the field in final review

ayeshmcg:
Card: https://github.com/bcgov/cas-reporting/issues/918
A small PR to fix the reason for change field in final review


andrea-williams (approved):
Screenshot 2025-11-17 at 4 44 16 PM

Ran locally; confirmed that the width is as expected.

Nice!

PR #4013, opened 2025-11-19
Feat/914 add report open date

ayeshmcg:
Card: https://github.com/bcgov/cas-reporting/issues/914

**Changes:**
1. Updated the ReportingYear model to include a new report_open_date field.
2. Added a migration to populate report_open_date for all existing records in the reporting_year table.
3. Updated logic for current annual reports so that if report_open_date is in the future, the UI will display “Available soon” instead of the Start button.
4. Added unit tests for the ReportingYear model.
5. Updated relevant front-end tests to reflect the new logic.

**How to Test:**
1. Run make prepare_backend to apply all migrations.
2. Manually update the 2024 Reporting Year’s report_open_date from 2025 to 2026 in the database.
3. Navigate to the reporting dashboard: http://localhost:3000/reporting/reports
4. You should now see “Available soon” in place of the Start button for the 2024 report.
5. Verify that past reports continue to display normally.

### DIFF
--- a/bc_obps/reporting/migrations/0142_reportingyear_report_open_date.py
+++ b/bc_obps/reporting/migrations/0142_reportingyear_report_open_date.py
@@ -6,7 +6,7 @@ from django.db import migrations, models
 class Migration(migrations.Migration):
 
     dependencies = [
-        ('reporting', '0140_V4_5_0'),
+        ('reporting', '0141_alter_reportproduct_production_data_apr_dec'),
     ]
 
     operations = [
--- a/bc_obps/reporting/migrations/0143_populate_report_open_date.py
+++ b/bc_obps/reporting/migrations/0143_populate_report_open_date.py
@@ -39,7 +39,7 @@ def reverse_populate_report_open_date(apps, schema_editor):
 class Migration(migrations.Migration):
 
     dependencies = [
-        ('reporting', '0141_reportingyear_report_open_date'),
+        ('reporting', '0142_reportingyear_report_open_date'),
     ]
 
     operations = [


bc_obps/reporting/migrations/0141_populate_report_open_date.py
```diff
@@ -0,0 +1,42 @@
...
+    # Bulk update all reporting years
+    ReportingYear.objects.bulk_update(reporting_years, ['report_open_date'])
+
+
+def reverse_populate_report_open_date(apps, schema_editor):
+    """
+    Reverse migration - set report_open_date to None for all reporting years.
+    """
+    ReportingYear = apps.get_model('reporting', 'ReportingYear')
+    ReportingYear.objects.all().update(report_open_date=None)
```
pbastia: This is actually not true. I'd make this migration forward-only, or just not do anything on reverse. We have no way of knowing the state of the `report_open_date` field prior to this migration

bc_obps/reporting/tests/models/test_reporting_year.py
```diff
@@ -1,24 +1,44 @@
...
+        """Test that report_open_date is set to March 1st of the year after the reporting year"""
+        reporting_years = ReportingYear.objects.all()
+
+        for reporting_year in reporting_years:
+            expected_open_date = datetime(reporting_year.reporting_year + 1, 3, 1, 0, 0, 0, tzinfo=ZoneInfo("UTC"))
+            self.assertEqual(
+                reporting_year.report_open_date,
+                expected_open_date,
+                f"Report open date for year {reporting_year.reporting_year} should be March 1st of {reporting_year.reporting_year + 1}",
+            )
```
pbastia: I don't think we'll want to test this at all, unit tests are meant to test code and behaviour, not data

pbastia (changes requested):
Nice work!! I think we should make the API be the referee whether reporting is open or not, so there is a little bit of code to move from the frontend

bciers/apps/reporting/src/app/components/operations/cells/ActionCell.tsx
```diff
@@ -5,10 +5,17 @@ import { createReport } from "@reporting/src/app/utils/createReport";
...
+  BC_GOV_PRIMARY_BRAND_COLOR_BLUE,
+} from "@bciers/styles";
 import { ReportOperationStatus } from "@bciers/utils/src/enums";
 
-const ActionCell = (params: GridRenderCellParams) => {
+interface ActionCellProps extends GridRenderCellParams {
+  isReportingOpen: boolean;
+}
+
+const ActionCell: React.FC<ActionCellProps> = (params) => {
```
pbastia: I suggest destructuring the params object to appease sonarcloud and be consistent with how react FCs are implemented in the rest of the app

```suggestion
const ActionCell: React.FC<ActionCellProps> = ({row, isReportingOpen}) => {
```

bc_obps/reporting/tests/models/test_reporting_year.py
```diff
@@ -1,24 +1,30 @@
...
+from reporting.tests.utils.bakers import reporting_year_baker
+from reporting.models import ReportingYear
+from django.test import TestCase
 
 
-class ReportingYearTest(BaseTestCase):
+class TestInitialData(TestCase):
+    def test_reporting_year_initial_data(self):
+        expected_reporting_years = sorted([2023, 2024, 2025, 2026])
+        existing_reporting_years = sorted(list(ReportingYear.objects.values_list('reporting_year', flat=True)))
```
pbastia: SonarQube is right!
```suggestion
        existing_reporting_years = sorted(ReportingYear.objects.values_list('reporting_year', flat=True))
```

PR #4022, opened 2025-11-20
feat: add penalty status tracking and refactor penalty creation workflow

nataliepiguliak:


bc_obps/compliance/models/compliance_penalty.py
```diff
@@ -14,6 +14,13 @@ class Frequency(models.TextChoices):
         DAILY = 'Daily'
         MONTHLY = 'Monthly'
 
+    class Status(models.TextChoices):
+        NONE = 'NONE', 'None'
+        ACCRUING = 'ACCRUING', 'Accruing'
+        DUE = 'DUE', 'Due'
+        NOT_PAID = 'NOT PAID', 'Not Paid'
+        PAID = 'PAID', 'Paid'
```
dleard: I think we only need PAID & NOT_PAID for statuses on this model.
NONE/ACCRUING will never be set because the penalty does not exist in these states
DUE = NOT_PAID

bc_obps/compliance/service/automated_process/compliance_handlers.py
```diff
@@ -83,6 +89,15 @@ def handle(self, invoice: ElicensingInvoice) -> None:
...
 
+        # Set both penalties for this obligation to ACCRUING when the obligation invoice is overdue
+        CompliancePenalty.objects.filter(
+            compliance_obligation=obligation,
+            penalty_type__in=[
+                CompliancePenalty.PenaltyType.AUTOMATIC_OVERDUE,
+                CompliancePenalty.PenaltyType.LATE_SUBMISSION,
+            ],
+        ).update(status=CompliancePenalty.Status.ACCRUING)
+
```
dleard: CompliancePenalty records will not exist while the penalty is accruing. They are created when the obligation is paid in full

bc_obps/compliance/service/automated_process/compliance_handlers.py
```diff
@@ -91,8 +106,8 @@ def can_handle(self, invoice: ElicensingInvoice) -> bool:
         """Check if the obligation is fully paid and should be updated to OBLIGATION_FULLY_MET."""
 
         # Ensure this is an obligation invoice, not a penalty invoice
-        has_penalty = hasattr(invoice, 'compliance_penalty') and getattr(invoice, 'compliance_penalty', None)
-        if has_penalty:
+        has_penalty_invoice = CompliancePenalty.objects.filter(elicensing_invoice=invoice).exists()
+        if has_penalty_invoice:
```
dleard: I think we can revert this change, the original logic makes more sense here.

bc_obps/compliance/service/automated_process/compliance_handlers.py
```diff
@@ -114,6 +129,15 @@ def handle(self, invoice: ElicensingInvoice) -> None:
...
         logger.info(f"Updated compliance status for obligation {obligation.obligation_id}")
         final_transaction_date = PenaltyCalculationService.determine_last_transaction_date(obligation)
+
+        CompliancePenalty.objects.filter(
+            compliance_obligation=obligation,
+            penalty_type__in=[
+                CompliancePenalty.PenaltyType.AUTOMATIC_OVERDUE,
+                CompliancePenalty.PenaltyType.LATE_SUBMISSION,
+            ],
+        ).update(status=CompliancePenalty.Status.DUE)
```
dleard: This can be removed. CompliancePenalty records will not exist at this point

bc_obps/compliance/service/compliance_obligation_service.py
```diff
@@ -71,6 +72,9 @@ def create_compliance_obligation(
             penalty_status=ComplianceObligation.PenaltyStatus.NONE,
         )
 
+        # Create initial empty penalties for this obligation
+        PenaltyCalculationService.create_initial_penalties_for_obligation(obligation)
+
```
dleard: We don't want our service to be creating penalties here, this should only be done by our automated processes

bc_obps/compliance/service/penalty_calculation_service.py
```diff
@@ -342,6 +349,25 @@ def calculate_penalty(
...
+            penalty_type=CompliancePenalty.PenaltyType.AUTOMATIC_OVERDUE,
+            status=CompliancePenalty.Status.NONE,
+        )
+
+        CompliancePenalty.objects.create(
+            compliance_obligation=obligation,
+            penalty_type=CompliancePenalty.PenaltyType.LATE_SUBMISSION,
+            status=CompliancePenalty.Status.NONE,
+        )
+
```
dleard: Let's chat about the approach here. I don't think we need to be creating penalty records for every obligation. The calculate_penalty functions with persist=False should be enough to show an ephemeral penalty as of a date.

dleard (changes requested):
I've found a couple issues with the frontend:
- The GGEAPAR Interest section is missing the "Download Payment Instructions" & "Pay Penalty/Track Payments" sections (screenshot below)
Screenshot from 2025-11-24 14-21-50

- The GGEAPAR Interest section is displayed before the related obligation is paid to 0
- I wasn't able to look into the internal view at all. I get directed to a page that says there is no obligation when I click view details

bciers/apps/compliance/src/app/components/taskLists/1_manageObligationTaskList.ts
```diff
@@ -103,20 +110,22 @@ export const generateManageObligationTaskList: (
...
   let automaticPenaltySection: TaskListElement[] = [];
 
   if (
-    Number(outstandingBalance) === 0 &&
-    (penaltyStatus === PenaltyStatus.NOT_PAID ||
-      penaltyStatus === PenaltyStatus.PAID)
+    isObligationFullyPaid &&
+    [PenaltyStatus.NOT_PAID, PenaltyStatus.PAID].includes(
+      penaltyStatus as PenaltyStatus,
+    )
```
dleard: This logic is incomplete. It will show the Automatic Overdue penalty section even if one does not exist. This check needs to be similar to what is done above with the hasLateSubmissionPenalty boolean.
I think the schema in `bc_obps/compliance/schema/compliance_report_version.py` needs a resolver for has_overdue_penalty like what was done for has_late_submission_penalty & that value needs to be checked here.

dleard:
Review Interest Summary page is missing the Generate Invoice button (as per wireframe linked in the ticket)
Screenshot:
Screenshot from 2025-11-25 10-50-28
Wireframe:
image

PR #4024, opened 2025-11-21
Mocking date in django

pbastia:
## Work done:

- New `mocks` django app
- Migrations to add a `mocks` schema 
- A middleware that
    - uses time_traveller to freeze the time for the django request process
    - calls a database method that overrides the `now()` sql method to use a mocked version returning a set time (straight up copy-paste from [ciip](https://github.com/bcgov/cas-ciip-portal/tree/develop/mocks_schema/deploy))

## How it works:
1. The datepicker sets a `mock-time` cookie
2. the `actionHandler` (next backend) propagates the value of that cookie to the API
3. the middleware picks up that cookie and freezes time for both django and postgres

### DIFF
--- a/bc_obps/mocks/middleware/mock_time_middleware.py
+++ b/bc_obps/mocks/middleware/mock_time_middleware.py
@@ -1,5 +1,6 @@
 # type: ignore
 from datetime import datetime
+import logging
 from typing import Callable
 
 from django.conf import settings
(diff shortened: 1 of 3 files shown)


bc_obps/common/middleware/mock_time_middleware.py
```diff
@@ -0,0 +1,31 @@
...
+class MockTimeMiddleware:
+    def __init__(self, get_response: Callable[[HttpRequest], HttpResponse]):
+        self.get_response = get_response
+
+    def __call__(self, request: HttpRequest) -> HttpResponse:
+
+        # Mocking time before processing
+        if settings.ENVIRONMENT in ["local", "dev", "test"] and 'mock-time' in request.COOKIES:
+            time_traveller = time_machine.travel(datetime.fromisoformat(request.COOKIES['mock-time']))
+            request.time_traveller = time_traveller
```
pbastia: Necessary to attach the time_traveller instance to the request, otherwise we run into threading issues

bc_obps/mocks/migrations/0003_set_mocked_time_in_transaction.py
```diff
@@ -0,0 +1,37 @@
...
+      parsed_epoch := extract(epoch from mocked_timestamp)::text;
+      perform set_config('mocks.mocked_timestamp', parsed_epoch, false);
+
+    end;
+
+  $function$ language plpgsql volatile;
+
+COMMIT;
+""",
+            reverse_sql="drop function mocks.now();",
```
Sepehr-Sobhani: Should this be `reverse_sql="drop function mocks.set_mocked_time_in_transaction(timestamptz);",`?

bc_obps/mocks/middleware/mock_time_middleware.py
```diff
@@ -0,0 +1,53 @@
...
+
+        self.get_response = get_response
+
+    def __call__(self, request: HttpRequest) -> HttpResponse:
+
+        print(
+            f"""
+-- Time Mock Middleware: found time {request.COOKIES.get("mock-time", "None")}
+"""
+        )
```
Sepehr-Sobhani: nit: Suggest using logger!

bc_obps/mocks/middleware/mock_time_middleware.py
```diff
@@ -0,0 +1,53 @@
...
+            request.time_traveller = None
+
+        # Down the middleware stack
+        response = self.get_response(request)
+
+        # Reverting time mock
+        if request.time_traveller:
+            request.time_traveller.stop()
+
+        return response
```
Sepehr-Sobhani: I suggest using try/finally to ensure time_traveller is always stopped, even if an exception occurs during request processing. Without this, the mocked time could leak into subsequent requests.
```python
try:
    # Mocking time before processing
    if settings.ENVIRONMENT in ["local", "dev", "test"] and 'mock-time' in request.COOKIES:
        datetime_iso_str = request.COOKIES['mock-time']
        self.set_database_time_for_session(datetime_iso_str)

        time_traveller = time_machine.travel(datetime.fromisoformat(datetime_iso_str))
        request.time_traveller = time_traveller
        request.time_traveller.start()
    else:
        request.time_traveller = None

    # Down the middleware stack
    response = self.get_response(request)
finally:
    # Reverting time mock
    if request.time_traveller:
        request.time_traveller.stop()

return response
```

PR #4027, opened 2025-11-22
Chore/76 late submission penalty fixes

dleard:
This PR fixes a couple issues found during QA:

1. No elicensing invoice was being created when creating a late submission penalty
2. The `create_penalty` function was being used to create both late_submission and automatic_overdue penalties, but the call was gated behind logic that only worked for automatic_overdue. The calls to create the different penalties have now been split apart and are now behind their own logic gates

### DIFF
--- a/bc_obps/compliance/tests/service/automated_process/test_compliance_handlers.py
+++ b/bc_obps/compliance/tests/service/automated_process/test_compliance_handlers.py
@@ -254,7 +254,6 @@ class TestObligationPaidHandler:
         mock_create_late_penalty,
     ):
         self.handler.handle(self.invoice)
-        print(self.obligation.created_at)
 
         self.compliance_report_version.refresh_from_db()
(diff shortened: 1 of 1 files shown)


bc_obps/compliance/tests/service/automated_process/test_compliance_handlers.py
```diff
@@ -238,26 +244,31 @@ def test_can_not_handle_with_penalty_invoice(self):
...
     @patch('compliance.service.penalty_calculation_service.PenaltyCalculationService.create_penalty')
     @patch('compliance.tasks.retryable_notice_of_obligation_met_email')
     def test_handle_updates_compliance_status_and_creates_penalty_when_overdue(
         self,
         mock_retryable_notice_of_obligation_met_email,
         mock_create_penalty,
+        mock_create_late_penalty,
     ):
         self.handler.handle(self.invoice)
+        print(self.obligation.created_at)
```
Sepehr-Sobhani: Leftover!

PR #4029, opened 2025-11-24
chore: remove sync functionality from past and transferred reports

ayeshmcg:
Card: https://github.com/bcgov/cas-reporting/issues/845

Changes included in this PR
1. Added a new SyncValidationService with a helper function is_sync_allowed.
2. Included is_sync_allowed in the API response for:
     - Report operation data
     - Facility report data
     - Review facilities
3. Updated the sync (update) functionality to return 403 when sync is not allowed.
4. Updated the UI on:
     - Review Operation Information
     - Review Facility Information
     - Review Facilities
   so that the sync button and info section are shown only when sync is allowed.

How to test
1. Navigate to Past Reports: http://localhost:3000/reporting/reports/previous-years

2. Start any draft report and go to the Review Operation Information page: http://localhost:3000/reporting/reports/9/review-operation-information → The info section at the top and the sync button should not be visible.

3. Go to the Review Facilities (LFO) page: http://localhost:3000/reporting/reports/14/facilities/review-facilities → The info section and sync button should not be visible.

4. Go to the Review Facility Information page (LFO – past report): http://localhost:3000/reporting/reports/14/facilities/f486f2fb-62ed-438d-bb3e-0819b51e3aeb/review-facility-information → The info section and sync button should not be visible.

### DIFF
--- a/bc_obps/reporting/service/report_facilities_service.py
+++ b/bc_obps/reporting/service/report_facilities_service.py
@@ -93,7 +93,6 @@ class ReportFacilitiesService:
         current_facilities: list = []
         past_facilities: list = []
         is_sync_allowed = SyncValidationService.is_sync_allowed(version_id)
-        print('is sync allowed', is_sync_allowed)
 
         # Determine if all current facilities should be considered selected
         all_selected_for_current = not selected_facilities
--- a/bc_obps/reporting/service/sync_validation_service.py
+++ b/bc_obps/reporting/service/sync_validation_service.py
@@ -1,11 +1,46 @@
+from uuid import UUID
+from typing import Optional, Tuple
 from reporting.models import ReportVersion
 from service.reporting_year_service import ReportingYearService
+from registration.models import Facility, Operation
 
 
 class SyncValidationService:
     """
     Service to validate whether sync operations are allowed for a given report version.
     """
+
+    @classmethod
+    def _validate_sync_conditions(cls, version_id: int) -> Tuple[bool, Optional[ReportVersion]]:
+        """
+        Internal helper to validate sync conditions and return the report version.
+
+        Args:
+            version_id: The report version ID
+
+        Returns:
+            Tuple of (is_allowed: bool, report_version: Optional[ReportVersion])
+        """
+        report_version = ReportVersion.objects.select_related(
+            'report__operation', 'report__operator', 'report__reporting_year'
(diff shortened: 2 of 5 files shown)


bc_obps/reporting/service/report_facilities_service.py
```diff
@@ -91,6 +92,8 @@ def get_all_facilities_for_review(version_id: int) -> dict:
 
         current_facilities: list = []
         past_facilities: list = []
+        is_sync_allowed = SyncValidationService.is_sync_allowed(version_id)
+        print('is sync allowed', is_sync_allowed)
```
andrea-williams:
```suggestion
```

bc_obps/reporting/service/sync_validation_service.py
```diff
@@ -0,0 +1,47 @@
...
+    """
+
+    @classmethod
+    def is_sync_allowed(cls, version_id: int) -> bool:
+        """
+        Check if syncing with admin data is allowed for the given report version.
+
+        Sync is NOT allowed if:
+        1. The report is not for the current reporting year
+        2. The operation is not currently owned by the operator who filed the report
```
andrea-williams: I think we need to add a 3rd check here to make sure that a facility (in the case of an LFO) hasn't been transferred (while the operation itself remains with the original operator). This would be relevant for the "Review Facility Info" page (from the ticket)

bc_obps/reporting/service/sync_validation_service.py
```diff
@@ -0,0 +1,47 @@
...
+        Returns:
+            bool: True if sync is allowed, False otherwise
+        """
+        try:
+            report_version = ReportVersion.objects.select_related(
+                'report__operation', 'report__operator', 'report__reporting_year'
+            ).get(id=version_id)
+        except ReportVersion.DoesNotExist:
+            return False
+
```
pbastia: Whoever calls this service with a non-existent report version should be rewarded with a `DoesNotExist` error, making it disappear here only hides a potential issue.
I'd remove the try/except block entirely

```suggestion
        report_version = ReportVersion.objects.select_related(
            'report__operation', 'report__operator', 'report__reporting_year'
        ).get(id=version_id)
```

pbastia (changes requested):
Just one more thing about gobbling errors, this makes troubleshooting unrelated errors harder

PR #4051, opened 2025-11-28
chore: Improve CI by not fetching the whole git history

Sepehr-Sobhani:
- Saving 2 minutes by not fetching the entire Git history when building the apps


joshgamache (approved):
Great quick win!

PR #4080, opened 2025-12-09
374: add e2e earned credits approved flow

shon-button:
Addresses: 
[374](https://github.com/bcgov/cas-compliance/issues/374)
[375](https://github.com/bcgov/cas-compliance/issues/375)

### 🚀 Impact:

- Added E2E helper endpoint used to manipulate DB state using existing services while mocking external services.
- Updated e2e tests to mock calls to endpoints that call external api
- Added POM and workflow test for flow: 
   - `Approved earned credit issuance request`
   - `Declined earned credit issuance request`
- Moved Playwright package to Dev dependencies
- Bumped Playwright version to the latest to fix the issue when running tests in UI mode: https://github.com/microsoft/playwright/pull/37887

---

### 📝Note:
- Requires updating your `bciers/.env.local\E2E STORAGE_STATE AUTH.js V5: found in 1password OBPS FE E2E STORAGESTATES AUTH v5`

---

#### Start Test Environment
1. Start the API server:  
   ```sh
   cd bc_obps
   make prepare_backend
   ```  
   ```sh
   cd bciers && yarn dev-all
   ```  
3. Run compliance e2e tests

   ```sh
   cd bciers && yarn compliance:e2e:ui
   ```  
---

### DIFF
--- a/bciers/apps/compliance-e2e/poms/request-issuance/internal/internal-review-compliance-earned-credits.ts
+++ b/bciers/apps/compliance-e2e/poms/request-issuance/internal/internal-review-compliance-earned-credits.ts
@@ -57,7 +57,7 @@ export class InternalReviewComplianceEarnedCreditsPOM {
         });
       },
 
-      "earned_credits_director_approve",
+      EARNED_CREDITS_DIRECTOR_APPROVE_SCENARIO,
     );
   }
 


bciers/apps/compliance-e2e/utils/constants.ts
```diff
@@ -1,29 +1,71 @@
...
+export const EARNED_CREDITS_PUT_ROUTE_PATTERN =
+  "**/api/compliance/compliance-report-versions/*/earned-credits";
+
+// Regex to extract compliance_report_version_id from the URL
+export const EARNED_CREDITS_CRV_ID_REGEX =
+  /compliance-report-versions\/(\d+)\/earned-credits/;
+
+// Test-only integration stub endpoint
+export const E2E_INTEGRATION_STUB_PATH =
+  "/api/compliance/test/e2e-integration-stub";
```
Sepehr-Sobhani: We have two `E2E_INTEGRATION_STUB_PATH` constants, and this one includes `/test/` in the path, which is not correct.

bc_obps/compliance/api/e2e_integration_stub.py
Sepehr-Sobhani: This looks great; however, it duplicates service logic in the E2E stub, which can drift from the real implementation. Instead, I suggest mocking the external API calls (eLicensing/BCCR) at the HTTP layer so the real services run in the tests. This way, we reuse the actual service logic and only mock the HTTP calls.

bc_obps/compliance/api/e2e_integration_stub.py
```diff
@@ -0,0 +1,565 @@
...
+    ComplianceReport,
+    ComplianceReportVersion,
+    ElicensingInvoice,
+)
+from compliance.models.elicensing_client_operator import ElicensingClientOperator
+from compliance.service.elicensing.elicensing_operator_service import ElicensingOperatorService
+from rls.utils.manager import RlsManager
+
+
+from typing import Union
```
Sepehr-Sobhani: Nit: Consolidate all typing imports on line 16

bciers/apps/compliance/src/app/widgets/BccrHoldingAcountWidget.tsx
Sepehr-Sobhani: If possible, I’d suggest not adding e2e test code to production code, and using Playwright interception or another endpoint stub instead.

bc_obps/compliance/api/e2e_integration_stub.py
Sepehr-Sobhani: Nit: I’d suggest splitting this file into smaller chunks, for example:
```
compliance/api/e2e_integration_stub/
├── __init__.py
├── endpoint.py          # Main endpoint and routing
├── http_mocks.py        # External HTTP mocking
├── service_mocks.py     # Service-layer patches
├── scenarios/
│   ├── submit_report.py
│   ├── earned_credits.py
│   └── __init__.py
└── utils.py            # Helper functions
```

bciers/libs/e2e/src/utils/constants.ts
```diff
@@ -95,3 +95,8 @@ export const appRouteRoles: Record<AppRoute, UserRole[]> = {
   ],
   [AppRoute.USERS]: [UserRole.CAS_ADMIN, UserRole.INDUSTRY_USER_ADMIN],
 };
+
+export const DJANGO_API_BASE_URL = "http://127.0.0.1:8000/api";
```
Sepehr-Sobhani: Since `DJANGO_API_BASE_URL` doesn’t change, can we include it in the `COMPLIANCE_E2E_INTEGRATION_STUB_PATH` constant? That way, we won’t need to import it in other modules.
Another option would be to use `API_URL` from `.env` file.

Sepehr-Sobhani (commented):
Should we also add Happo screenshots for these scenarios?

Issue #4052, opened 2025-11-28
Operation Timeline: start_date shouldn't be now() on operation creation


pbastia:
#### Describe the Bug:

Issue: When a user registers a new operation, its entry in the Operation_Designated_Timeline comes with a start_date of today. This tricks the system into believing the operator didn't own the operation in 2024 if the operation was registered in 2025, preventing users from submitting a report or a supplementary report.

Solution:
- [ ] newly registered operations must have their timeline start on 2024-01-01 (anywhere before the program started)
- [x] (done) there are 20 records in the timeline today with a start date in 2025 and these must be changed, ONLY IF THEY ARE THE FIRST ENTRY FOR THE OPERATION IN THE TIMELINE. A one-time migration.

#### Probability:

(How likely the bug is to happen, scored from 1-5. For example, probability of 5 is something like "it happens to all users every time they log in." whereas probability of 1 "only happens to certain users when a really specific and unlikely path is followed.")

#### Impact:

(How bad the bug is when it does happen, scored from 1-5. For example, effect of 5 is "the entire app crashes and makes it unusable for all users" or "the bug causes the wrong data to be saved, with critical information (e.g. payment) being affected." whereas effect of 1 is "It makes some styling look a little bit weird.")

<img src="https://www.getzephyr.com/sites/default/files/riksk%20matrix.jpg" alt="Bug Matrix" width="250" height="250"/>

#### Steps to reproduce the behaviour:

1. Step #1
2. Step #2
3. Step #3

#### Additional information:

Browser: [e.g. chrome, safari]
Version: [e.g. 22]

#### Additional Context

-

patriciarussellCAS:
Hey @pbastia - does this comment mean this solution has been pushed to Prod already? I wasn't sure. 

- [ ] (done) there are 20 records in the timeline today with a start date in 2025 and these must be changed, ONLY IF THEY ARE THE FIRST ENTRY FOR THE OPERATION IN THE TIMELINE. A one-time migration.

patriciarussellCAS:
Thanks! Moving to accepted. For UAT.

TammieV-BCG:
UAT by Business Areas. Works as expected

PR #4083, opened 2025-12-10
4052 Operation default creation date

acatchpole:
Card: https://github.com/bcgov/cas-registration/issues/4052

The issue: Operations that are registered after the current reporting year ends are not able to start a report in the current reporting year (even though the regs say they are allowed to do that).

The fix: When a new operation is registered, we set the timeline start date to before BCIERS began (2024/01/01) so that users always have the option of creating a report in the current reporting year.

### Changes
 - Created a default start date for operations

### Testing
 1. Create a new operation
 2. Go to the reporting grid page at current-reports
    - Expect to see the newly added operation with the ability to start a report for it


pbastia (approved):
🎉

PR #4101, opened 2025-12-22
Chore/449 nightly ci

joshgamache:
Addresses bcgov/cas-compliance#449. Adds nightly builds of Develop to CI. All images are built and all tests are run. 

## Changes 🚧

- New CI workflow running all builds and tests.
  - Cron scheduled overnight.
- Images get tagged `nightly` when run by scheduled workflow.
- Images *not* tagged `latest` when run by scheduled workflow.
- Added status badge for nightly builds

## Notes for Reviewer 📝

This workflow makes use of the fact that [scheduled workflows use the default branch](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#schedule), which in our case is `develop`.

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -127,6 +127,9 @@ jobs:
     with:
       zap_fail_on_risk_level: 3
     secrets: inherit
+  schemaspy:
+    uses: ./.github/workflows/schemaspy.yaml
+    secrets: inherit
   trivy:
     uses: ./.github/workflows/trivy.yaml
   codeql:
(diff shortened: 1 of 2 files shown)


Sepehr-Sobhani (changes requested):
This is great work, thanks! As part of the ticket, we were hoping to remove the `ZAP OWASP` scan from running on every PR and instead include it only in the nightly build. This would help speed up CI for routine PRs and commits.
Additionally, I think we need a way to communicate failures in nightly builds to the team, as silent job failures can easily be missed.
We could also consider moving `SchemaSpy` to the nightly build, as it currently slows down deployments to TEST.

Issue #4110, opened 2026-01-06
Update activity applicable_to to match regs


acatchpole:
#### Describe the task

Update the applicable_to values to match the values in AC

Description of the Tech Debt
Now that we've built some LFO activities, we should update the list of activities to only show the relevant activities and source types to the SFO/LFO as options. The activity model already has an applicable_to column that we can use to filter on when populating the list of activities shown to the user.

UPDATES:
Due to the way GGERR is written, LFOs and SFOs report on some different activities from Table 1 and Table 2. Because of this, BCIERS should only show operations the relevant activities based on their Operation type (LFO/SFO).

Acceptance Criteria
Given that my Operation type is LFO
Then the list of activities I see matches the LFO-specific list from ([Registration Pt 2. Fields](https://bcgov.sharepoint.com/:x:/r/teams/00608-CPRInternal/Shared%20Documents/Carbon%20Pricing%20Transition%20SMEs/Registration%20App/Registration%20Pt.%202%20Fields.xlsx?d=w8cf871009f8d43c19d1d79c69fa259e8&csf=1&web=1&e=ctoTMH) excel)

Given that my Operation Type is SFO
Then the list of activities I see matches the SFO-specific list from ([Registration Pt 2. Fields](https://bcgov.sharepoint.com/:x:/r/teams/00608-CPRInternal/Shared%20Documents/Carbon%20Pricing%20Transition%20SMEs/Registration%20App/Registration%20Pt.%202%20Fields.xlsx?d=w8cf871009f8d43c19d1d79c69fa259e8&csf=1&web=1&e=ctoTMH) excel)

#### Definition of Done

- [ ] update Activity model so that applicable_to accepts enum[Single Facility Operation, Linear Facilities Operation, all]
- [ ] create a migration to update all applicable_to values to match the lists in the AC
- [ ] third

#### Additional context

- This is preparatory work for #3333 and https://github.com/bcgov/cas-reporting/issues/554

patriciarussellCAS:
Hey @acatchpole - I saw this ticket while doing the release to PROD. I see it didn't go through QA or PO Review. Just want to make sure there wasn't anything I was supposed to review for this? Or was it just backend stuff to prep for the other tickets mentioned above? Thanks.

PR #4112, opened 2026-01-06
Update applicable_to values

acatchpole:
Card: https://github.com/bcgov/cas-registration/issues/4110

NOTE: the applicable_to field is not currently used in any existing logic

### Changes
 - Updated the possible values for ```applicable_to``` from ```['sfo', 'lfo', 'all']``` to ```['Single Facility Operation', 'Linear Facilities Operation', 'all']. This was done to make filtering on operation_type more seamless. The full name is what is actually store in operation_type, so this eliminates the need to map those values.
 - Created a migration that updates all the applicable_to values in the activity table to:
    a) use the new full name values
    b) be current with what activities are applicable to each operation_type (per the AC)


ayeshmcg (approved):
Nice Work.
Tested: Changes align with the [list ](https://bcgov.sharepoint.com/:x:/r/teams/00608-CPRInternal/_layouts/15/Doc.aspx?sourcedoc=%7B8CF87100-9F8D-43C1-9D1D-79C69FA259E8%7D&file=Registration%20Pt.%202%20Fields.xlsx&action=default&mobileredirect=true)

PR #4115, opened 2026-01-07
chore: updated facilities transfer logic

ayeshmcg:
Card: https://github.com/bcgov/cas-reporting/issues/940
Changes:- 
1. Added Model FacilitySnapshot- this is snapshot of facility before it is transferred.
2. Added FacilitySnapshotService which has funtions to create facility snapshot, get facility snaption for give operation id and facility id, get facility snapshot for given operaion id.
3. Update report facilities service to get the facility data from FacilitySnapshot if its an old operator. 
4. Updated tests

To test:-
1. Login as an internal user(cas_analyst) and transfer an operation - for example Banana LFO from Bravo to Alpha operator
     - Login with IDIR
     - change app_role_id to cas_analyst in users table
     - Go to http://localhost:3000/registration/transfers/transfer-entity
image

     - tranfer entity
2. Login with as industry user with bc-cas-dev-secondary and assign operator Alpha
     - `Update erc.user set business_guid=(select id from erc.operator where legal_name like 'Alpha%') where first_name='bc-cas-dev-secondary'`
     - Login as  bc-cas-dev-secondary, select operator Alpha and request access
     - update erc.user_operator and set role to Admin and status Approved
     - Refresh/ clear cache
     - Login again and go to http://localhost:3000/reporting/reports/current-reports(change date to April 2026), Banana Alpha should be visible for Alpha operator. 
     - Go to Review Facilities page - http://localhost:3000/reporting/reports/15/facilities/review-facilities and update and any facility name or add a new facility using http://localhost:3000/administration/operations/002d5a9e-32a6-4191-938c-2c02bfec592d/facilities/add-facility?isNewTab=true. 

3. Login as industry user with bc-cas-dev
    - Go to past reports http://localhost:3000/reporting/reports/previous-years.
    - Delete the existing supplementary report draft for Banana LFO and create new supplementary report
    - Go to Review facilities page
    - You should not see the changes done by the new operator


bc_obps/registration/models/facility_snapshot.py
```diff
@@ -0,0 +1,128 @@
...
+    This model captures the state of a facility when it is transferred from one operation to another,
+    allowing old operators to see the facility data as it existed at the time they owned it.
+    """
+
+    id = models.UUIDField(
+        primary_key=True,
+        default=uuid.uuid4,
+        db_comment="Primary key to identify the facility snapshot",
+        verbose_name="ID",
+    )
```
andrea-williams: nit: I don't think there's any need to explicitly add the `id` field

bc_obps/registration/tests/endpoints/test_facility_snapshots_api.py
```diff
@@ -0,0 +1,129 @@
...
+        response = TestUtils.mock_get_with_auth_role(
+            self,
+            endpoint=endpoint,
+            role_name="industry_user",
+        )
+
+        # Should return 422 (validation error) since operation_id is required
+        assert response.status_code == 422
+
+    def test_get_snapshots_not_associated_with_their_operator(self):
```
andrea-williams: I'd like to see another unit test here to assert that a user only sees the facility snapshots associated with their operator. For example, if "Facility A" gets transferred from Operator 1 to Operator 2 (thus creating a facility snapshot), and then later Facility A gets transferred again from Operator 2 to Operator 3 (thus creating another snapshot for the facility), we should test that an industry user associated with Operator 1 cannot view the snapshot that was created for Operator 2.

andrea-williams (commented):
Saving review progress - still have to run & test locally.

bc_obps/service/facility_snapshot_service.py
```diff
@@ -0,0 +1,103 @@
...
+        Returns:
+            QuerySet of FacilitySnapshot instances
+        """
+        return FacilitySnapshot.objects.filter(
+            operation_id=operation_id,
+            facility_id=facility_id,
+        ).order_by('-snapshot_timestamp')
+
+    @classmethod
+    def get_facility_snapshots_by_operation(
```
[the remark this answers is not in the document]
ayeshmcg: As of now we will just be using create_facility_snapshot, I will keep that and remove the rest. Also I added the endpoint to get the facility_snapshot but thats not required for this. Is okay to remove that as well. We can add it whenever requied

Issue #3750, opened 2025-09-04
Update 'Date of first shipment' logic for new entrant registration


pbastia:
#### Description:

The registration process encodes a "date of first shipment" value that is either before or after Apr 1, 2024, that is only used for new entrants.

- [x] assess whether the new entrant process still needs this date of first shipments for the years moving forward 
- [x] make the necessary changes (either remove the 2024 piece from the options, keeping the "before or after Apr 1 of registration year, or remove that field entirely
    - [ ] backend: model `DateOfFirstShipmentChoices` options and `date_of_first_shipment` field
    - [ ] backend: API endpoint (see test_new_entrant_application.py) and ninja schema
    - [ ] frontend: `administrationRegistrationInformation.ts`, `newEntrantOperation.ts`
    - [ ] tests (e.g. test_operation_service.py checks the date of first shipment according to the 2024 rules)

#### Acceptance Criteria:

Given I am registering a new entrant operation
When I fill out my registration info for year 2025
Then I can't see 2024 fields

New entrant application page is now out of date and needs to be updated by removing the 2 application form/dates
For context, the deadline to apply for new entrant designation based on having a DOFS that was before March 31, 2025 was on May 31, 2025, so it is no longer an applicable option for any operators 

[Wireframe](https://www.figma.com/design/LsDmLDCdnJqI8UahyqvOD8/BCIERS-Designs?node-id=13653-1955&t=SUJVAuMjOqGcswXK-4)

<img width="753" height="251" alt="Image" src="https://github.com/user-attachments/assets/7349ac24-a158-49ff-ac07-2f9deba18c55" />

#### Development Checklist:

- [ ] Replace current wording with new wording from wireframe
     - [ ] move existing (2024) page to a "2024" folder, make this new wireframe the "default" (see reporting module's `data/jsonSchema/<year>` folders for an example) 
- [ ] Keep the April 1 2024 application form template as the "application form template"
- [ ] Remove the March application form template
- [ ] update backend tests (e.g., test_operation_service.py, test_new_entrant_application.py) to reflect new date of first shipment rules
- [ ] Meets the DOD

#### Definition of Ready (Note: If any of these points are not applicable, mark N/A)

- [ ] User story is included
- [ ] User role and type are identified
- [ ] Acceptance criteria are included
- [ ] Wireframes are included (if required)
- [ ] Design / Solution is accepted by Product Owner
- [ ] Dependencies are identified (technical, business, regulatory/policy)
[...]

PR #4118, opened 2026-01-07
chore: fixed ui issues for new entrant operation

ayeshmcg:
Card: https://github.com/bcgov/cas-registration/issues/3750
Addressed the ui issues mentioned in https://github.com/bcgov/cas-registration/issues/3750#issuecomment-3719744985

### DIFF
--- a/bciers/apps/administration-e2e/utils/enums.ts
+++ b/bciers/apps/administration-e2e/utils/enums.ts
@@ -225,8 +225,6 @@ export enum OperationFields { //All of these fields are also used for New Entran
   BOUNDARY_MAP = "Boundary Map",
   REPORTING_ACTIVITIES = "Reporting Activities*",
   REGULATED_PRODUCT_NAMES = "Regulated Product Name(s)*",
-  NEW_ENTRANT_APPLICATION = "New Entrant Application and Statutory Declaration*",
-  FIRST_SHIPMENT_DATE = "When is this operation's date of First Shipment?*",
 }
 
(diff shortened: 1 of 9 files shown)


bciers/apps/administration-e2e/utils/enums.ts
```diff
@@ -225,8 +225,6 @@ export enum OperationFields { //All of these fields are also used for New Entran
   BOUNDARY_MAP = "Boundary Map",
   REPORTING_ACTIVITIES = "Reporting Activities*",
   REGULATED_PRODUCT_NAMES = "Regulated Product Name(s)*",
-  NEW_ENTRANT_APPLICATION = "New Entrant Application and Statutory Declaration*",
```
andrea-williams: why did you remove this label from E2E? I think this is still relevant

andrea-williams (commented):
Ran locally - confirmed that the 2024 values no longer appear in the administration module's operation details page, and that the error message doesn't overlap anymore.

I am skeptical about removing the E2E check for the new entrant application - as far as I can tell, that should stay in.

bciers/apps/administration-e2e/utils/enums.ts
ayeshmcg: @andrea-williams  this is no longer used , As we removed that as per the new design

andrea-williams: ahhh I see. sorry, I didn't realize the labels in the wireframe had changed too.

andrea-williams (approved):

PR #4120, opened 2026-01-07
945 Activity name updates

acatchpole:
Card: https://github.com/bcgov/cas-reporting/issues/945

~Probably not going to implement this PR. See https://github.com/bcgov/cas-registration/pull/4125~

#### Changes
 - Migration added that:
    - Adds ```regulated_name``` field
    - populates ```regulated_name``` field with all the current values in the ```name``` field
    - Where applicable, updates ```name``` to an easier-to-understand version, per the ticket
 - Updates the schemas for the affected activity forms so the title shows the new names
 - Migration to update activity_json_schema titles
 - Removed the custom save() from activity_json_schema, activity_source_type_json_schema, and configuration_element models as they were never used because custom save overrides don't run during migrations. See [here](https://github.com/bcgov/cas-registration/issues/4194) for follow-up work.
 - All other file changes are updating tests with the new names
 
 Note: While no data should be affected by this change, users looking at or updating old reports will see the new names instead of the old ones. This was deemed acceptable by the BA. There is a follow up [ticket](https://github.com/bcgov/cas-reporting/issues/953) to make the names as they appear in the regulations available.
 
#### Testing
This change will be effective on every page that displays the names of activities  
Registration: Register an Operation form
Administration: Operation Information page
Reporting: Operation Information, Review Facility page, and the task list and activity pages themselves


ayeshmcg (approved):
Nice Work. Looks good

PR #4135, opened 2026-01-15
942 start report after transfer

acatchpole:
Card: https://github.com/bcgov/cas-reporting/issues/942

This PR addresses not being able to start a report for an operation after that operation has been transferred to a different operator.

~The only thing that was preventing this was the authorization check on the endpoint for create-report.~ The auth check included a call to check_operation_ownership, which checked if the user was an approved user_operator for the operator that CURRENTLY owns the operation. This check was updated to include the reporting_year for the report being created, and gets associated operator from the operation_designated_operator_timeline.

Creating a new report_version used the same auth check as creating a new report, but the change broke this. So I also made a new auth check just for report_version.

The actual create_report method has also been updated to get operator from timeline. A special handler for creating report_versions when the operation has been transferred has been created. The only thing different it does right now is populate the operation_representatives differently, but it stands as a placeholder for when further special handling for this case is developed.

#### Changes
 - New function in operation_designated_operator_timeline_service: ```get_operation_designated_operator_for_reporting_year``` which returns the timeline for the end of the given reporting year (which includes who the designated operator was at that point in time), with an extra property has_been_transferred which is a boolean that indicates whether an operation has been transferred from the operator. This check is based on today()'s date, just in case end_date is ever set to a future date.
 - Updated the auth check for ```create-report``` endpoint to use this method to get the designated operator
 - Updated the auth check for ```create-report-version``` endpoint to use a new auth check, which gets the designated operator from the report itself.
 - added create_transferred_operation_report_version. This method is for the special handling that creating a report requires after an operation has been transferred. Right now, the only special handling implemented is how the operation_representatives are populated. Rather than getting them from operation_contacts, which are tied to the transferred operation, we get them from contacts, which are associated with the operator.

   Note: Code Analysis is failing due to duplication of code in this function. This is due to the fact that we have not implemented a strategy for the other special handling required for a report with transferred operation like operation_name. But we will need to implement them at some point. This function stands mostly as a placeholder until that happens.

- Updated the filtering in get_operations_for_reporting_dashboard. There was a possible bug that was unrelated to this ticket. The logic on the filter for an operation designated timeline WAS:
``` if (end_date is in the year after the reporting year) OR (end_date == null and start_date is before the reporting_year)```
   The fact that the check for start_date could be missed doesn't make sense. start_date being <= the reporting_year should be checked no matter what. And then checking if end_date is reporting_year + 1 is needlessly constrained. All we should care is if the start_date <= reporting_year and end_date == null || end_date > reporting_year. This update may be more relevant when we have a reporting dashboard that allows for starting past reports.

#### Testing
Set up for testing requires two things: a transferred operation and a change of date. While there are several ways you could do this, I will describe one.

1. Promote your IDIR user to 'cas_analyst' or 'cas_director' so that you are authorized to transfer.
2. Transfer 'Bat LFO' operation from Bravo to Alpha effective date 2026-01-18. I picked 'Bat LFO' because it does not have a started 2025 report in the fixtures. 
3. Login as 'bc-cas-dev', navigate to the 'Annual Reports' page and set the date to 2025-12-12 (or anytime in the latter half of 2025)
4. Next to 'Bat LFO', hit the ```Start``` button
   - Expect to be routed review-operation-information page, the first page of a new report. 
   Note: due to some issues with the recent frontend upgrades, this redirect might not occur. You may need to refresh the page, then expect to see the 'Status' for 'Bat LFO' to be in 'Draft' and click 'Continue' to go into the report.
5. On review-operation-information page, expect no Operation representative to be selected. Normally this would already have the same operation representatives selected as were selected in Administration. As these are no longer accessible, there will be no rep pre-selected, but all the operator contacts with a business role of "Operation Representative" will be available to select.
6. Continue the report as you like to and see that it will submit.
7. Create a supplementary report
8. Complete supplementary report and see that it will submit.

### DIFF
--- a/bc_obps/reporting/api/permissions.py
+++ b/bc_obps/reporting/api/permissions.py
@@ -67,7 +67,7 @@ def _validate_operator_ownership_in_url(
     return is_access_granted(user_operator)
 
 
-def _validate_report_version_ownership_in_url(request: HttpRequest, version_id_param: str) -> bool:
+def _validate_version_ownership_in_url(request: HttpRequest, version_id_param: str) -> bool:
     return _validate_operator_ownership_in_url(request, version_id_param, get_operator_from_version_id)
 
 
@@ -114,7 +114,7 @@ def check_version_ownership_in_url(
         if request.current_user.is_irc_user():  # type: ignore
             return True
 
-        return _validate_report_version_ownership_in_url(request, version_id_param)
+        return _validate_version_ownership_in_url(request, version_id_param)
 
     return validate_func
 

### CODE REVIEW

File: bc_obps/reporting/api/permissions.py [hunk 1]
```diff
@@ -41,21 +43,55 @@ def _validate_version_ownership_in_url(request: HttpRequest, version_id_param: s
     return is_access_granted(user_operator)
 
 
+def _validate_report_ownership_in_url(request: HttpRequest, report_id_param: str) -> bool:
+    if not request.resolver_match:
```
pbastia: Report ownership *to an operator* doesn't change with transfers, and this is what this function checks.
- add a comment to make this explicit
- rename function to `_validate_report_belongs_to_operator_in_url`

bc_obps/reporting/api/permissions.py
```diff
@@ -41,21 +43,55 @@ def _validate_version_ownership_in_url(request: HttpRequest, version_id_param: s
...
+    if not report:
+        logger.warning("No report found.")
+        return False
+
+    user_operator = UserOperator.objects.filter(
+        user=request.current_user,  # type: ignore
+        operator=report.operator,
+    ).first()
+
+    return is_access_granted(user_operator)
```
pbastia: I suggest refactoring the code between this method and the `_validate_version_ownership_in_url` above to remove code duplication

pbastia (changes requested):
Nice work! We discussed a few opportunities to refactor some code duplication

bc_obps/service/report_version_service.py
```diff
@@ -67,6 +68,54 @@ def create_report_version(
...
+            ),
+            report_version=report_version,
+            registration_purpose=operation.registration_purpose or 'OBPS Regulated Operation',
+        )
+        for contact in Contact.objects.filter(operator=operator, business_role="Operation Representative"):
+            ReportOperationRepresentative.objects.create(
+                report_version=report_version,
+                representative_name=contact.get_full_name(),
+                selected_for_report=False,
+            )
```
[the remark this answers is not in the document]
acatchpole: removed create_report_version_transferred_operation and put special handling in create_report_version

bc_obps/reporting/api/permissions.py
acatchpole: arguably, this check is validating that the operator (through the user) has ownership of the report[_version] and is accurate, though I see the argument. 

But there are auth checks in Compliance that follow this naming convention and I think it is more important that they match, so I have not changed these names.

bc_obps/reporting/api/permissions.py
acatchpole: The bulk of these two functions is error handling around the specific entity that they relate to (checking if specific param is present and if it matches an object in the db). The actual logic in both is a single db query, one that has a join that the other doesn't. I wrote version that minimized duplicated code, but it had more LOC and was less readable (in my opinion). I don't think refactoring is warranted here.

PR #4154, opened 2026-01-17
445 Retry elicensing adjustment when applying compliance units

Sepehr-Sobhani:
[ISSUE 445](https://github.com/bcgov/cas-compliance/issues/445)


bc_obps/compliance/tests/service/test_compliance_adjustment_service.py
shon-button: @Sepehr-Sobhani 

Question: Is there a reason not to keep using `create_adjustment_for_current_version` here?

PR #4162, opened 2026-01-22
chore: fixed report start error

ayeshmcg:
Card: https://github.com/bcgov/cas-reporting/issues/956

When a user clicked the "Start" button in the ActionCell component, it would send a request to create a new report. However, if the user clicked the button multiple times before the navigation completed, it would send duplicate requests to create reports for the same operation and reporting year. This caused an error:
`We couldn't create a report for operation ID 'ef9044dd-2a27-4d26-86fe-02e51e0755f7' and reporting year '2025': A report already exists for this operation and year, unable to create a new one.`

**Root Cause:**
The button's onClick handler was setting hasClicked to true to disable the button, but then resetting it back to false after the button action completed. This meant that during the async operation (creating the report and navigating), if a user clicked the button again quickly, it could trigger duplicate report creation requests.

**Solution:**
- Added setHasClicked(true) at the beginning of handleStartClick to immediately disable the button when the Start action is triggered
- Modified the button's onClick handler to only reset hasClicked to false for "Continue" and "View Details" actions, but keep it true for the "Start" action
- This ensures the "Start" button remains disabled after the first click, preventing duplicate report creation requests

### DIFF
--- a/bciers/apps/reporting/src/app/components/changeReview/templates/ReasonForChange.tsx
+++ b/bciers/apps/reporting/src/app/components/changeReview/templates/ReasonForChange.tsx
@@ -12,14 +12,13 @@ const ReasonForChangeForm: React.FC<ReasonForChangeProps> = ({
   reasonForChange,
   onReasonChange,
 }) => {
-  const borderColor = DARK_GREY_BG_COLOR;
   const styles: React.CSSProperties = {
     width: "100%",
     padding: "8px",
     fontSize: "16px",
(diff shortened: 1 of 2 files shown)


gdalcengio (approved):
Fixes the bug as intended and the changes appear clean to me, approved. I did however find that going back from the newly created report screen to the reports grid renders the button as start once more and clicking it makes the bug come back, so I'm tagging @fviduya as well. However, I think that might be a separate ticket to cleanup later?

PR #4163, opened 2026-01-22
chore: fixed person responsible page issue

ayeshmcg:
Card: https://github.com/bcgov/cas-reporting/issues/957

When a user initially lands on the Person Responsible page with no person selected and then selects a contact for the first time, the contact details would not display immediately. The details only appear after changing to a different contact.

The issue was caused by React's asynchronous state updates:
1. In the **handleContactSelect** function, after fetching contact data and calculating the addressError, the code called:
    - setSelectedContactAddressError(addressError) - updates state asynchronously
    - updateContactShown(newSelectedContactId, newContactFormData) - called immediately
2. Inside **updateContactShown**, when creating the schema via createPersonResponsibleSchema, it was using:
     - **selectedContactId** and **contactFormData** - old state values (still undefined on first selection)
     - **selectedContactAddressError** - old state value (still undefined on first selection)
3. Because React state updates are asynchronous, the updateContactShown function was using stale state values instead of the fresh data that had just been fetched.

Solution:-
- Updated updateContactShown signature to accept addressError as a parameter
- Fixed the schema creation to use the fresh parameter values instead of stale state
- Updated the condition from newContactId !== null to newContactId !== undefined to match the TypeScript type.

### DIFF
--- a/bciers/apps/reporting/src/app/components/operations/personResponsible/PersonResponsibleForm.tsx
+++ b/bciers/apps/reporting/src/app/components/operations/personResponsible/PersonResponsibleForm.tsx
@@ -1,8 +1,8 @@
 "use client";
-import { useState } from "react";
+import { useState, useRef } from "react";
 import { RJSFSchema } from "@rjsf/utils";
+import { IChangeEvent } from "@rjsf/core";
 import { getContacts, getContact } from "@bciers/actions/api";
-import debounce from "lodash.debounce";
 import {
   personResponsibleSchema,
   personResponsibleUiSchema,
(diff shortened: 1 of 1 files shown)


gdalcengio (approved):
Nice, fixes the bug, approved. 

Related bug found as well, not sure if there's a ticket for it or not yet but I will tag @fviduya in this as well again. If the user is quickly switching between persons responsible while one is loading, the asynchronous loading gets stuck in a little bit of a loop, cycling between people clicked over and over again.

gdalcengio (approved):
Changes fixed the additional bug, nicely done!

PR #4165, opened 2026-01-22
fix: create manual handling supp decrease report if prev report increased emission with void + adjustment

gdalcengio:
Essentially if an invoice is void, it doesn't grab an obligation during the fallback stage of building the adjustment strategy, checking if the previous invoice has cash, then no cash would be found since no previous invoice would be attached to the supplementary report's anchor id.

The fix is to iteratively check previous CRV's linked IDs to grab the closest obligation invoices.


shon-button (approved):
🥇

PR #4178, opened 2026-01-28
chore: add favicon to each app

gdalcengio:


gdalcengio:
Turns out it actually does work

PR #4179, opened 2026-01-28
chore: added validation

ayeshmcg:
Card: https://github.com/bcgov/cas-reporting/issues/932
Changes:-
1. Updated form builder service to add gastype and fuel type/name as required
2. updated report activity save service to throw user error.
3. Added custom validator for appying methodlogy validations
4. updting customTransformErrors.

To test:
1. Start a report
2. Go to activity page and click on submit. 
3. You should see all the validation errors.


ayeshmcg:
Thanks @andrea-williams 
Yes thats not a part of this, We may have some ticket or may need to create for other vaidations. This was for the select  fields like gasType, Methodoloy, Fuel name

PR #4192, opened 2026-01-30
Fix nightly build schemaspy and happo

Sepehr-Sobhani:


joshgamache (approved):
Great to have those fixed in the nightly! The docs are a solid addition too, I've used the same methods before too. You even caught the `event_name` and `schedule:cron`=`develop` branch limitations I was thinking as I was reading the rest!

dleard (approved):
🏅

PR #4198, opened 2026-02-03
Feat: New pulp and paper lime recovery kiln product

pbastia:


bc_obps/registration/migrations/0157_add_pulp_and_paper_product.py
```diff
@@ -0,0 +1,31 @@
...
+from django.db import migrations
+
+
+def add_product(apps, schema_monitor):
+    '''
+    Add the new Pulp & Paper: Lime Revovery Kiln
+    '''
+    RegulatedProduct = apps.get_model('registration', 'RegulatedProduct')
+    RegulatedProduct.objects.create(
+        name="Pulp and paper: lime recovery kiln",
```
andrea-williams: being pedantic: does it matter that this capitalization doesn't match the capitalization in the ticket? `Pulp & Paper: Lime Recovery Kiln`

andrea-williams (approved):
Ran the code locally and confirmed that the help text appears when one or the other Pulp & paper lime products are selected in Reporting, but doesn't appear when both are already selected.

I noticed that the help text doesn't appear when one of those products is selected in the Operation Information in the Admin module, but I thought there was a bit of ambiguity in the ticket AC regarding that, so I just flagged it there too.

Apart from a couple of instances in the component test where `toBeInTheDocument()` should be replaced with `toBeVisible()` per a recent Dev Team Agreements discussion, everything looks good!

bc_obps/registration/migrations/0157_add_pulp_and_paper_product.py
pbastia: It doesn't, I chose to keep it consistent with the other p&p products instead

PR #4223, opened 2026-02-09
Fix: Production Data Page error in DEV

andrea-williams:
chore: fixed key for operation_opted_out_final_reporting_year in Reporting's Production Data Page; added more unit tests to confirm correct behaviour

### DIFF
--- a/bciers/apps/reporting/src/app/components/products/ProductionDataPage.tsx
+++ b/bciers/apps/reporting/src/app/components/products/ProductionDataPage.tsx
@@ -26,9 +26,10 @@ export default async function ProductionDataPage({
   const facilityType = response.facility_data.facility_type;
 
   const reportingYear = response.report_data.reporting_year;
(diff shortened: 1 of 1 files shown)


bciers/apps/reporting/src/app/components/products/ProductionDataPage.tsx
```diff
@@ -27,8 +27,8 @@ export default async function ProductionDataPage({
 
   const reportingYear = response.report_data.reporting_year;
   const isOptedOut =
-    (response.report_operation.operation_opted_out_final_reporting_year ?? 0) <=
-    reportingYear;
+    response.payload.operation_opted_out_final_reporting_year != null &&
```
pbastia: This could backfire between null and undefined, I suggest leaving it to the null-coalesce operator:


```suggestion
    response.payload.operation_opted_out_final_reporting_year &&
```

PR #4225, opened 2026-02-09
Feat: 911 regulatory values overrides in compliance summary

pbastia:
Work done:

#### 2. industrial process calculations now take in account biogenic emission split

- a method to retrieve this split from the activity form (pending [#cas-reporting/965](https://github.com/bcgov/cas-reporting/issues/965))
- update to the calculations to include the split in the emissions overlap between industrial process and excluded biomass
   
#### 3. a new table to store Naics regulatory overrides

- migrations
- and some logic to retrieve the override or default to the industry regulatory values
    
#### 4. Tests

### DIFF
--- a/bc_obps/reporting/schema/compliance_data.py
+++ b/bc_obps/reporting/schema/compliance_data.py
@@ -1,12 +1,13 @@
-from typing import List, Optional
+import dataclasses
+from typing import List
 from ninja import Schema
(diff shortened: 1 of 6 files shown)


ayeshmcg (approved):
@pbastia  
LGTM
I tested the Compliance Summary page, and both RF and TR appear correctly under all products.
Note: There are some SonarQube failures.

PR #4226, opened 2026-02-10
Feat/compliance calculations jan mar 25 production/929

andrea-williams:
Resolves [reporting/#929](https://github.com/bcgov/cas-reporting/issues/929)

Changes made:

- actually saves the Jan-Mar production data to the database (failed to do that in previous PR!)
- added new function `get_production_period` to ComplianceService class: determines whether "jan_mar", "apr_dec", or "annual" production amounts should be used to calculate compliance obligations
- refactored some code in ComplianceService to remove 2024-specific references, instead uses generic "prorated" limits
- added new RJSF schema for 2025 ComplianceSummary (hence the SonarCloud failure because of code duplication)
- added a bunch of unit tests

Recommended Testing Steps:

1. as user `bc-cas-dev`, create and register an operation as Opted-in
2. as cas-director, mark the operation as Opted-out, with final reporting year of 2025
3. as `bc-cas-dev`, start a new annual report for the opted-out operation for reporting year 2025. Confirm that when filling out the facility product, you are required to fill out production data for both annual data and Jan 1 - Mar 31 2025. Populate the operation's emissions data so that it exceeds the allocated emissions
5. in the Final Review section of the report, confirm that both annual and Jan 1 - Mar 31 production data fields are displayed. Also confirm that the Apr 1 - Dec 31 2024 production field is not displayed.

Screenshot from Final Review page:

Screenshot 2026-03-02 at 5 49 03 PM

6.


bc_obps/reporting/tests/service/test_compliance_service/test_compliance_service_class.py
```diff
@@ -116,6 +119,46 @@ def test_apr_dec_product_level_behavior_zero_production(self):
...
+        )
+        Report.objects.filter(pk=build_data.report_1.id).update(reporting_year=2025)
+
+        result = ComplianceService.get_calculated_compliance_data(build_data.report_version_1.id)
+
+        # For annual period, allocated_compliance_emissions should equal full allocated
+        for product in result.products:
+            # Annual production should be used, so no prorating
+            self.assertEqual(product.allocated_compliance_emissions > 0, True)
+
```
andrea-williams: if anyone has suggestions for a better assertion, please let me know! Without having a spreadsheet to use (like we have for the Apr-Dec calculations), I'm not totally sure what I should be testing against.

bc_obps/reporting/service/compliance_service_parameters.py
```diff
@@ -1,33 +1,50 @@
...
+        prorated_allocated = Decimal(0) if annual == 0 else (allocated_for_compliance / annual) * apr_dec
+    elif production_period == "jan_mar":
+        jan_mar = Decimal(production_totals.get("jan_mar") or 0)
+        production_for_limit = jan_mar
+        prorated_allocated = Decimal(0) if annual == 0 else (allocated_for_compliance / annual) * jan_mar
     else:
-        allocated_for_compliance_2024 = Decimal(0)
-        production_for_limit = annual
-        allocated_compliance_emissions_value = _round4(allocated_for_compliance)
+        raise ValueError(f"Invalid production period: {production_period}")
```
pbastia: This default case won't be needed if we use a real enum

pbastia (dismissed):
Looks good to me! We just need to fix the tests to make sure we have the expected compliance data, happy to help pair on that if you want

bciers/apps/reporting/src/data/jsonSchema/2025/complianceSummary.ts
```diff
@@ -0,0 +1,170 @@
...
+            type: "number",
+            title: "Allocated Emissions attributable to compliance",
+          },
+        },
+      },
+    },
+  },
+};
+
+export const complianceSummaryUiSchema2024 = {
```
ayeshmcg: We can change this to 2025

ayeshmcg (changes requested):

bc_obps/reporting/service/compliance_service/parameters.py
```diff
@@ -18,27 +25,42 @@ def round(value: Decimal | float) -> Decimal:
...
         """
         annual = Decimal(cast(Decimal, production_totals.get("annual_amount")))
-        apr_dec = Decimal(production_totals.get("apr_dec") or 0)
 
-        if use_apr_dec:
-            # If no annual production, prorated allocation is zero to avoid division-by-zero
-            allocated_for_compliance_2024 = Decimal(0) if annual == 0 else (allocated_for_compliance / annual) * apr_dec
+        if production_period == ProductionPeriod.ANNUAL:
+            production_for_limit = annual
+            prorated_allocated = Decimal(0)
```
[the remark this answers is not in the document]
andrea-williams: outcome from chat with Pierre: keeping this as-is, because elsewhere our code assumes that prorated_allocated equals 0 when the annual production period is being used. Instead simplified L60 by including variable assignment within case statements.

PR #4233, opened 2026-02-11
chore: 966 renaming product and adding pwaei

pbastia:
Work done:
- Reworked migration to include the right product name
- Added ProductEmissionIntensity record 
- Fixed test that checked for regulated products without pwaei defined
- Fixed help text template

### DIFF
--- a/bciers/apps/reporting/src/app/components/facility/FacilityEmissionAllocationForm.tsx
+++ b/bciers/apps/reporting/src/app/components/facility/FacilityEmissionAllocationForm.tsx
@@ -131,17 +131,29 @@ const validateFormData = (
       industrialEmissionAllocations?.products?.find(
         (p) => p.product_name === "Pulp and paper: chemical pulp",
       );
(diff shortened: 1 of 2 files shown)


ayeshmcg (commented):
@pbastia 
Conditionally approving, assuming the following isn’t an issue.

I’m not sure if this is part of this PR, but when I select “Lime recovered by Kiln”, nothing appears below the multiselect. However, when I add both products, I still see the note (see screenshot below).

Image

Also, a question about this existing logic (from
https://github.com/bcgov/cas-registration/blob/develop/bciers/apps/reporting/src/app/components/facility/FacilityEmissionAllocationForm.tsx
):

```
if (
  chemicalPulpAllocation.allocated_quantity -
    overlappingIndustrialProcessEmissions <
  0
)
  newErrors.push(
    "Invalid allocation: Industrial Process quantity allocated to product 'Pulp and paper: chemical pulp' is too low",
  );
```


Will this logic still be valid with the current changes?

Issue #4234, opened 2026-02-11
Bug: Changing registration purpose affects past reports


fviduya:
#### AC for Ticket:

Given that an operation has submitted report/s for the previous year/s,
When industry user updates the registration purpose for the current RY, 
And user creates a supplementary report for the past year,
Then the registration purpose update should not apply to the past year report 

Given that an operation has a draft report in progress for a previous year/s
when industry user updates the registration purpose for the current RY
Then the registration purpose update should not apply to the past year report 

A real-life example is LNG Canada. They reported as a Reporting Operation in 2024, but now need to report in 2025 as a New Entrant. We need to ensure their 2024 report remains as a 'reporting operation' report (no compliance obligation) and only their 2025 report includes the new entrant reporting forms. 

Given that I am an industry user
when I change registration purpose in administrtaion 
Then I see the updated [confirmation modal.](https://www.figma.com/design/LsDmLDCdnJqI8UahyqvOD8/BCIERS-Designs?node-id=37219-12784&t=sXJpGyxPKYhNqt3h-4) 

[TRANSFERS]
Given that an operation has been transferred to a new operator,
And the previous operator has draft and/or submitted reports for current or past reporting year(s),
When the new operator updates the registration purpose,
Then the updated registration purpose must NOT be reflected in:
current RY: the previous operator’s draft reports
past RYs: the previous operator’s draft reports
any supplementary reports created by the previous operator
Given that new operator has updated the registration purpose,
Then the previous operator's draft report does not get deleted for both the current RY and the past RYs

#### Probability:

(How likely the bug is to happen, scored from 1-5. For example, probability of 5 is something like "it happens to all users every time they log in." whereas probability of 1 "only happens to certain users when a really specific and unlikely path is followed.")

#### Impact:

(How bad the bug is when it does happen, scored from 1-5. For example, effect of 5 is "the entire app crashes and makes it unusable for all users" or "the bug causes the wrong data to be saved, with critical information (e.g. payment) being affected." whereas effect of 1 is "It makes some styling look a little bit weird.")

<img src="https://www.getzephyr.com/sites/default/files/riksk%20matrix.jpg" alt="Bug Matrix" width="250" height="250"/>

#### Steps to reproduce the behaviour:

1. Log in as bc-cas-dev
2. Click `View past reports`
3. Click `Continue` for Compliance SFO - No obligation or earned credits
4. Change the url to `/sign-off`
5. Sign off and submit the report
6. Go to the main dashboard
[...]

PR #4250, opened 2026-02-13
chore: updated supplementary report create logic

ayeshmcg:
**Card:** https://github.com/bcgov/cas-registration/issues/4234
**Issue:**
When creating a supplementary report version, the current logic checks whether the registration purpose in Reporting and Operation differ:
- **If different:** A new report version is created with the updated registration purpose, and all other data in the supplementary report version is reset.

- **If the same:** A supplementary report is created as a replica of the submitted report version.

This logic did not account for past reporting years or reports belonging to a previous operator, causing unintended data loss and incorrect registration purpose propagation.

**Fix**
- Updated create supplementary report save service to also check whether the report belongs to a past reporting year or a previous operator before applying the registration purpose diff logic.

- Updated signal handle_registration_purpose_changed to skip deletion of draft report versions when the draft belongs to a past reporting year or was created under a previous operator.

**Testing Steps**
1. Navigate to http://localhost:3000/reporting/reports/previous-years and start the report for Bugle SFO - Registered so that it appears as a draft on the reporting dashboard.
2. Navigate to http://localhost:3000/reporting/reports/current-reports and start the report for Bugle SFO - Registered.
3. On the Review Operation Information page (current reports), click [Update Operation Information](http://localhost:3000/administration/operations/3b5b95ea-2a1a-450d-8e2e-2e15feed96c9?operations_title=Bugle%20SFO%20-%20Registered%20-%20name%20from%20admin&isNewTab=true) and change the registration purpose.
4. Once updated, return to the Review Operation Information page and click Sync Latest Data from Administration. 5. Confirm the updated registration purpose is reflected.
5. Navigate back to Previous Years and open Bugle SFO - Registered (Draft). Verify that:
    - The draft has not been deleted.
    - The registration purpose has not changed.
6. For Bangles SFO - Registered - has Multiple Operators (current year), update the registration purpose and click Sync.
7. Navigate to Past Reports and create a supplementary report for Bangles SFO - Registered - has Multiple Operators. Verify that:
   - The supplementary report version uses the same registration purpose as the previously submitted report, not the updated registration purpose from the current report version.

### DIFF
--- a/bc_obps/reporting/service/report_supplementary_version_service/report_supplementary_cloning.py
+++ b/bc_obps/reporting/service/report_supplementary_version_service/report_supplementary_cloning.py
@@ -2,7 +2,9 @@ import copy
 from typing import Optional
 
 from django.core.files.storage import default_storage
+from django.forms import model_to_dict
 
+from reporting.service.emission_category_mapping_service import EmissionCategoryMappingService
 from reporting.models import (
     FacilityReport,
     ReportActivity,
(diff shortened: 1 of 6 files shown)


bc_obps/reporting/service/report_supplementary_version_service.py
```diff
@@ -31,40 +32,111 @@
...
+        report_version = ReportVersion.objects.select_related(
+            "report__operation", "report__reporting_year", "report_operation"
+        ).get(id=report_version_id)
+
+        operation = report_version.report.operation
+        report_operation = report_version.report_operation
+
+        operator_changed = operation.operator_id != report_version.report.operator_id
+        registration_purpose_changed = operation.registration_purpose != report_operation.registration_purpose
+        is_past_year = report_version.report.reporting_year.reporting_year < timezone.now().year
```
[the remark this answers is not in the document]
andrea-williams: we have existing functionality to fetch the current reporting year - that should be used instead

bc_obps/reporting/service/report_supplementary_version_service.py
```diff
@@ -31,40 +32,128 @@
...
+            return ReportSupplementaryVersionService.create_report_supplementary_version(
+                report_version_id, is_past_year=is_past_year
+            )
+
+        # Otherwise just make a new blank one
+        return ReportVersionService.create_report_version(report_version.report)
+
+    @staticmethod
+    @transaction.atomic
+    def create_report_supplementary_version(report_version_id: int, is_past_year: bool = False) -> ReportVersion:
```
andrea-williams: this method would also benefit from being broken down into smaller parts per SRP

bc_obps/reporting/service/report_supplementary_version_service.py
```diff
@@ -31,40 +32,128 @@
...
+                .order_by('-created_at')
+                .first()
+            )
+
+            if latest_submitted_report_version:
+                # Clone from the matching version instead
+                version_to_clone_id = latest_submitted_report_version.id
+
+        # Get the version to clone (might be the original or the matching version)
+        source_version = ReportVersion.objects.get(id=version_to_clone_id)
```
andrea-williams: this may be making more database calls than are necessary - e.g., if we have a `latest_submitted_report_version`, we already have the ReportVersion object we need

bc_obps/reporting/signals/consumers.py
```diff
@@ -4,32 +4,56 @@
...
         return
 
-    draft_version = ReportVersion.objects.filter(
-        report__operation_id=operation_id, status=ReportVersion.ReportVersionStatus.Draft
-    ).first()
+    draft_version = (
+        ReportVersion.objects.select_related("report__operation", "report__reporting_year", "report__operator")
+        .filter(report__operation_id=operation_id, status=ReportVersion.ReportVersionStatus.Draft)
+        .first()
+    )
```
andrea-williams: I don't think we should only be taking the first() result anymore - from RY2026 onwards, it could be the case that there are multiple drafts for an operation (e.g., for RY24, RY25, ...).

andrea-williams: actually what I think we should be doing is first determining what the current_reporting_year is, and then filtering ReportVersion objects on only the current reporting year.

bc_obps/reporting/tests/service/test_report_supplementary_version_service.py
```diff
@@ -173,6 +177,115 @@ def test_create_report_supplementary_version(self):
...
+        self.new_report_version.status = ReportVersion.ReportVersionStatus.Submitted
+        self.new_report_version.save()
+
+        # Act
+        new_version = ReportSupplementaryVersionService.create_or_clone_report_version(self.old_report_version.id)
+
+        # Assert: Should clone data (supplementary version)
+        self.assertEqual(new_version.report, self.old_report_version.report)
+        self.assertIsNotNone(ReportOperation.objects.filter(report_version=new_version).first())
+
```
andrea-williams: I don't really understand what the purpose of this assertion is. Why not something like `self.assertEqual(new_version.report.report_operation, self.old_report_version.report.report_operation)`?

bc_obps/reporting/service/report_supplementary_version_service/report_supplementary_cloning.py
```diff
@@ -26,48 +27,10 @@
...
-        ReportSupplementaryVersionService.clone_electricity_import_data(report_version, new_report_version)
-        ReportSupplementaryVersionService.clone_report_version_additional_data(report_version, new_report_version)
-        ReportSupplementaryVersionService.clone_report_version_new_entrant_data(report_version, new_report_version)
-        ReportSupplementaryVersionService.clone_report_version_verification(report_version, new_report_version)
-        ReportSupplementaryVersionService.clone_report_version_attachments(report_version, new_report_version)
-        ReportSupplementaryVersionService.clone_report_version_facilities(report_version, new_report_version)
-        ReportSupplementaryVersionService.reapply_emission_categories(new_report_version)
-        # Return the newly created report version
-        return new_report_version
+class ReportSupplementaryCloning:
```
pbastia: Now that we've refactored the "service" part out of the `ReportSupplementaryVersionService` class, we don't need to artificially add a class with only static members to the namespace.

I suggest having all these methods declared in the `report_supplementary_cloning.py` file directly, this is the pythonic way of declaring a module

pbastia (approved):
🎉

PR #4273, opened 2026-02-23
Chore/474 Verify backup used to check DB restoration is less than 24 hours old

joshgamache:
Addresses bcgov/cas-compliance#474. Adds check to the migrations test to ensure age of DB backup is less than a day.

## Changes 🚧

- Add a cronjob to the helm chart deployed for the DB. This cronjob runs a shell script to validate the age of the last full backup is less than 24 hours old.
- Add a task to the migration test Dag that runs the cronjob, in parallel with the check that checks the database restoration.

## Notes 📝

The cronjob is currently triggered by Airflow, but could easily be adapted to run as a heartbeat check as well.

### DIFF
--- a/helm/migration-test/cas-obps-postgres-migration-test/templates/cronJobs/test-pgbackup-age.yaml
+++ b/helm/migration-test/cas-obps-postgres-migration-test/templates/cronJobs/test-pgbackup-age.yaml
@@ -65,13 +65,11 @@ spec:
                   CURRENT_EPOCH=$(date +%s)
                   echo "Current epoch: $CURRENT_EPOCH"
                   LATEST_BACKUP_EPOCH=$(
-                    pgbackrest info |
+                    pgbackrest info --type=full |
                     awk '
-                        /full backup:/ {found=1}
-                        found && /timestamp start\/stop:/ {
+                        /timestamp start\/stop:/ {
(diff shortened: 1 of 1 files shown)


helm/migration-test/cas-obps-postgres-migration-test/templates/cronJobs/test-pgbackup-age.yaml
```diff
@@ -0,0 +1,91 @@
...
+                  LATEST_BACKUP_EPOCH=$(
+                    pgbackrest info |
+                    awk '
+                        /full backup:/ {found=1}
+                        found && /timestamp start\/stop:/ {
+                            split($0,a,"/ ")
+                            print a[2]
+                            found=0
+                        }
+                    ' |
```
dleard: Very minor optmisation, totally up to you if you want to implement it. It just makes the awk command a little simpler. You can pass `pgbackrest info` a `--type=full` argument and it will only return full backups removing the need to find `/full backup:/` & conditionally parsing the stop time of the backup.
https://pgbackrest.org/command.html#command-info
```suggestion
                    pgbackrest info --type=full |
                    awk '
                        /timestamp start\/stop:/ {
                            split($0,a,"/ ")
                            print a[2]
                        }
                    ' |
```

PR #4283, opened 2026-02-26
chore: update report open date for reporting year 2025

ayeshmcg:
Card: https://github.com/bcgov/cas-reporting/issues/984
A small PR to update report_open_date for reporting year 2025.

### DIFF
--- a/bc_obps/reporting/migrations/0161_update_report_open_date.py
+++ b/bc_obps/reporting/migrations/0161_update_report_open_date.py
@@ -12,17 +12,10 @@ def update_report_open_date(apps, schema_editor):
     reporting_year_2025.save()
 
 
-def reverse_update_report_open_date(apps, schema_editor):
-    ReportingYear = apps.get_model('reporting', 'ReportingYear')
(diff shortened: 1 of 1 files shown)


bc_obps/reporting/migrations/0161_update_report_open_date.py
```diff
@@ -0,0 +1,28 @@
...
+    reporting_year_2025.save()
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('reporting', '0160_compliance_summary_product_override_values'),
+    ]
+
+    operations = [migrations.RunPython(update_report_open_date, reverse_update_report_open_date)]
```
pbastia: I don't think we need to reverse this one, it would only save an invalid state. We can also mark it as [elidable](https://docs.djangoproject.com/en/6.0/ref/migration-operations/#runpython) for when @Sepehr-Sobhani does magic with migrations again


```suggestion
    operations = [migrations.RunPython(update_report_open_date, elidable=True)]
```

PR #4326, opened 2026-03-11
Feat: 693 internal report history view

pbastia:
Making the report history view available to all internal users as well.

- adjusting next pages to display the right page component
- backend API route now check the report ownership
- tests


andrea-williams (approved):
Ran the code locally and confirmed that results are filtered so that internal users only see submitted reports. And I can view operation reports & facility reports without the app crashing. 

One nit-picky question regarding the code, otherwise all looks good!

PR #4358, opened 2026-03-17
983 Submission errors include links

acatchpole:
Card: https://github.com/bcgov/cas-reporting/issues/843

This PR adds links to the errors generated in the report_validation validators. It also adds a ReportValidationException to carry an array of ReportValidationErrors, which allows all validation errors to be shown on the sign-off page at once.

#### Changes not specified in the card:
 - Supporting multiple error messages showing at once was not specified in the card, but makes for a much better user experience when multiple validations fail.
 - Removed `fuelType` and `gasType` as error keys supported by the validation utils `as these fields have been made required in order to save Activity data, and so further validation shouldn't be needed.
 
#### Changes
 - Add optional `context` to `ReportValidationError` and implemented in relevant validators
 - Created `ReportValidationException` which takes an array of `ReportValidationErrors`
 - updated `report_submission_service` to use `ReportValidationException` to raise validation errors
 - Updated front-end validation error messages to match those found in [this wireframe](https://www.figma.com/design/LsDmLDCdnJqI8UahyqvOD8/BCIERS-Designs?node-id=44401-8603&t=rfiwOwvZt37vJHpP-4).
 - Took `ReportValidationSummary` and supporting files from https://github.com/bcgov/cas-registration/pull/4434 which provides an easily configurable method for rendering errors and warnings with or without links.
 
 ##### Unstructured error handling
 I tried redirecting non-structured errors by throwing them in the `postSubmitReports` action handler as it is done in the `postSupplementaryReportVersion` action handler, but this caused submission to hang when the response was an unstructured error.
 
 So instead I added a `generic_error` key in the config and create a structured error out of any error response that isn't already structured.
 
 #### Testing

Use the 2024 report for Banana LFO with status == 'Draft Supplementary Report'
1. Progress through the report to the activity page for 'Facility 1'
2. Input some valid emissions data and save
3. Navigate directly to `/reporting/reports/13/sign-off` page
4. Fill in the required checkboxes and fields and click `Submit Report`
 expect to see a set of errors with in-line hyperlinks
image
5. In order to see the other two possible errors with links, you must go to the `attachments` page.

6. 'Upload' a verification statement so that you can save the form
7. Select the first checkbox and save.
8. Navigate back to sign-off page and attempt to submit report
expect to no longer see "You must confirm that all required supplementary attachments have been uploaded and existing attchments are still relevant to the the supplementary submission on the Attachments page."
expect to see "You must confirm that all required supplementary attachments have been uploaded on the Attachments page." with working link.
9. Navigate back to the `attachments` page.
10. Unselect the first checkbox and select the second and save.
11. Navidate back to the sign-off page and attempt to submit report
expect to no longer see "You must confirm that all required supplementary attachments have been uploaded on the Attachments page."
expect to see "You must confirm that all existing attachments are still relevant to the supplementary submission on the Attachments page." with working link

**Optional**
- Return a generic backend exception:
**Expected Results**
image


bc_obps/reporting/service/report_submission_service.py
```diff
@@ -25,10 +25,8 @@ def submit_report(version_id: int, user_guid: UUID, sign_off_data: ReportSignOff
 
         is_regulated_operation = report_version.report.operation.is_regulated_operation
 
-        # The validation service now returns errors, but to not change the system behaviour,
-        # we raise an error for now.
         if validation_result:
-            raise ValidationError({key: val.message for key, val in validation_result.items()})
+            raise ReportValidationException(validation_result)
```
pbastia, line 29: Nice!!!

bciers/apps/reporting/src/app/utils/reportValidationMessages.ts
```diff
@@ -55,33 +94,81 @@ function cleanErrorString(error: unknown): string {
...
+      React.Fragment,
+      null,
+      message + " ",
+      React.createElement(
+        "a",
+        { href: `/${error.fix_url}` },
+        "Click here to fix this issue.",
+      ),
+    );
+  }
```
pbastia: These should be react components instead

bciers/apps/reporting/src/app/components/signOff/SignOffForm.tsx
```diff
@@ -28,7 +28,7 @@ export default function SignOffForm({
     date: "",
     supplementary: {},
   });
-  const [errors, setErrors] = useState<string[]>();
+  const [errors, setErrors] = useState<(string | ReactNode)[]>();
```
pbastia: We should declare a structured model for the error (similar to what we have on the backend) and let a react component render that, instead of storing DOM elements in the state.

What I mean is:
- state contains an error class (or dict or object or something)

the `render()` method renders it with an ad hoc component

```tsx
const errors, setErrors = useState(errorObj)

const errorComponent = buildValidationErrors(errors)

return <MultiStepFormWithTaskList errors={errorComponent} ... />
```


And we can write a tech debt card so we can pass these error structures down to the Form objects, having the props contain react nodes is a bad pattern for this

bc_obps/reporting/service/report_validation/validators/mandatory_verification_statement.py
```diff
@@ -28,7 +28,9 @@ def validate(report_version: ReportVersion) -> dict[str, ReportValidationError]:
             ReportVerification.objects.get(report_version_id=report_version.id)  # attempt to get the object.
         except ObjectDoesNotExist:
             errors["missing_report_verification"] = ReportValidationError(
-                Severity.ERROR, "Report verification form not found in the report."
+                Severity.ERROR,
+                "Report verification form not found in the report.",
+                fix_url=f"reporting/reports/{report_version.id}/verification",
```
pbastia: the page URLs are a frontend concern, the backend shouldn't depend on how the frontend organises the pages. (otherwise this would mean that a frontend change has impact on this code, the associated tests, etc)

The error needs to contain:
- an error key identifier for the frontend (it does already, `missing_report_verification` here)
- metadata like a report version id and a facility id so the frontend can build the page URL

bciers/apps/reporting/src/app/utils/reportValidationMessages.ts
```diff
@@ -1,28 +1,67 @@
...
-    fuelType: "Fuel is missing Fuel Name.",
-    gasType: "Emission is missing Gas Type.",
+      "A verification statement must be uploaded with this report on the Attachments page.",
     missing_required_attachment_confirmation:
       "You must confirm that all required supplementary attachments have been uploaded on the Attachments page.",
     missing_existing_attachment_confirmation:
       "You must confirm that all existing attachments are still relevant to the supplementary submission on the Attachments page.",
     missing_supplementary_report_attachment_confirmation:
-      "No attachment confirmation found for this report version. Please confirm the attachments on the Attachments page.", // Added the message here
+      "You must confirm that all required supplementary attachments have been uploaded and existing attchments are still relecant to the the supplementary submission on the Attachments page.",
```
shon-button: @acatchpole 

typos:
`attchments`
 `relecant`

duplicated:
`the the`

bc_obps/reporting/service/report_validation/report_validation_error.py
```diff
@@ -1,20 +1,54 @@
...
+        alias_generator=to_camel,
+        populate_by_name=True,
+    )
+
+    report_version_id: int
+    facility_id: Optional[UUID] = None
+    facility_name: Optional[str] = None
+    emission_category_id: Optional[int] = None
+    emission_category_name: Optional[str] = None
+    activity_id: Optional[int] = None
```
pbastia, line 37: This is a fantastic base for building a common error framework for individual reporting pages!

bc_obps/reporting/service/report_validation/report_validation_error.py
```diff
@@ -1,20 +1,54 @@
...
+from pydantic import BaseModel, ConfigDict
+from pydantic.alias_generators import to_camel
 
 
 class Severity(Enum):
     WARNING = "Warning"
     ERROR = "Error"
 
 
+class ReportValidationErrorKey(Enum):
```
pbastia: I suggest [`StrEnum`](https://docs.python.org/3/library/enum.html#enum.StrEnum), this way the serialization in the payload comes for free, without type checks


```suggestion
class ReportValidationErrorKey(StrEnum):
```

bc_obps/service/error_service/handle_exception.py
```diff
@@ -44,6 +47,27 @@ class ExceptionHandler:
...
             "Internal Server Error.", 500, "database_error"
         ),
+        (ReportValidationException,): ExceptionResponse(
+            None,
+            422,
+            payload_builder=lambda exc: {
+                "errors": [
+                    {
+                        "key": (
+                            e.key.value if isinstance(e.key, ReportValidationErrorKey) else e.key
```
pbastia, line 57: I think if we use StrEnum above this can just be `"key":e.key`

bciers/apps/reporting/src/app/components/shared/validation/types.ts
```diff
@@ -0,0 +1,82 @@
...
+  fuelType?: string;
+  fieldName?: string;
+  expectedRange?: string;
+  userInput?: string | number | null;
+
+  facilityName?: string;
+  emissionCategoryName?: string;
+  emissionCategoryId?: number;
+  sourceActivityName?: string;
+  targetProductName?: string;
```
pbastia: I'm not entirely sure this makes sense to define. The objects described by that interface never exist, so we could either make it a a union of actual error context types, or just define a general spec like:

```ts
interface ReportValidationErrorContext {
    reportVersionId: number;
    [key:string] : string | number | undefined;
}
```

(probably preferred, the value of maintaining the collection of types is pretty small and adds friction when adding more validation errors)

bciers/apps/reporting/src/app/components/shared/validation/utils.ts
```diff
@@ -0,0 +1,69 @@
...
+  const config = validationUIConfig[key];
+  const label = getValidationErrorLabel(key, error);
+  const message = getValidationErrorMessage(key, error);
+
+  if (config?.formatMessage) {
+    return config.formatMessage({ label, message, error });
+  }
+
+  return message;
+}
```
pbastia: These methods could be implemented on the ValidationUiConfig type direclty

bciers/apps/reporting/src/app/components/signOff/SignOffForm.tsx
```diff
@@ -28,9 +30,20 @@ export default function SignOffForm({
...
 
+  const isStructuredValidationError = (
+    error: unknown,
+  ): error is { errors: ReportValidationErrors } => {
+    return (
+      !!error &&
+      typeof error === "object" &&
+      Array.isArray((error as { errors?: unknown }).errors)
+    );
+  };
```
pbastia: This is really brittle. I understand it might be necessary to accomodate validation errors vs server errors but maybe we could instead:
- either have the payload pass validation errors in a different key, instead of reusing "response.error"?
- or at least make a tech debt card to address this

PR #4359, opened 2026-03-17
Chore/545 penalty calc execution transaction

dleard:
### Summary
These changes address the possibility that an invoice could be created in elicensing, and if the execution was interrupted or failed part way through afterwards, there could be an invoice our system does not know about. Which would cause the system to create a new, duplicate invoice.

Also addressed an issue found during this work where the compliance handler throwing a blanket atomic transaction on everything it called, which included external API calls. The use of transaction.atomic is now more targeted.

Making the create_penalty() call retryable & adding a function that idempotently creates the CompliancePenalty record / corresponding elicensing invoice makes this call stack more robust.

### In this PR:
- A few small fixes around dates:
  - remove static TODAY=date.today() from top of class (evaluated at instantiation)
  - Properly set fee dates / invoice due dates to use the day the invoice was generated as reference (rather than the day of the final transaction which could have caused invoices to be due before they were issued in some edge cases)
- Refactors the `calculate..()` functions to have a single responsibility (just calculate rather than calculate & persist data)
- Refactors the `create_penalty()` function to:
  -  handle both types of penalties (Overdue & Late Submission), which allowed removal of the create_late_submission_penalty() function
  - idempotently create penalty records & corresponding elicensing invoices
  - ensure that immediately following a successful invoice creation in elicensing, the invoice number is persisted to the CompliancePenalty record
  - create the accrual records for each day the penalty accrued inside a transaction
- Refactors the `automated_process_service` to not throw everything insde a blanket atomic transaction. It now determines further down in the execution which parts should or should not be wrapped in a transaction
- Refactors the `create_penalty()` calls in `compliance_handlers.ObligationPaidHandler` to be retryable functions

### DIFF
--- a/bc_obps/compliance/service/penalty_calculation_service.py
+++ b/bc_obps/compliance/service/penalty_calculation_service.py
@@ -365,7 +365,7 @@ class PenaltyCalculationService:
             final_transaction_date: The date of the last transaction that paid off the obligation invoice
 
         Returns:
-            CalculatedPenaltyData Dataclass
+            CompliancePenalty
         """
 
         compounding_frequency = (
@@ -440,6 +440,8 @@ class PenaltyCalculationService:
                 accrual_start_date=penalty_accrual_start_date,
                 final_accrual_date=final_transaction_date,
             )
+        else:
+            raise ValueError('Unrecognized PenaltyType. Must be one of CompliancePenalty.PenaltyType options')
 
         # Create the CompliancePenalty object if it does not exist & create invoice in elicensing if not previously created
         compliance_penalty_record = PenaltyCalculationService.perform_idempotent_penalty_creation(
@@ -478,7 +480,11 @@ class PenaltyCalculationService:
                     accumulated_penalty=acc.accumulated_penalty,
                     accumulated_compounded=acc.accumulated_compounded,
                 )
+        from compliance.tasks import (
+            retryable_notice_of_obligation_met_penalty_due_email,
+        )
 
+        retryable_notice_of_obligation_met_penalty_due_email.execute(obligation.id)
         return compliance_penalty_record
 
     @classmethod
@@ -505,7 +511,6 @@ class PenaltyCalculationService:
         return annual_rate / days_in_year
 
     @classmethod
-    @transaction.atomic
     def calculate_late_submission_penalty(
         cls,
         obligation: ComplianceObligation,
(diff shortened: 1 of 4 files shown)


bc_obps/compliance/service/penalty_calculation_service.py
```diff
@@ -363,39 +344,142 @@ def determine_last_transaction_date(obligation: ComplianceObligation) -> date |
...
+                obligation=obligation,
+                accrual_start_date=penalty_accrual_start_date,
+                final_accrual_date=final_transaction_date,
+            )
+        elif penalty_type == CompliancePenalty.PenaltyType.LATE_SUBMISSION:
+            penalty_data = cls.calculate_late_submission_penalty(
+                obligation=obligation,
+                accrual_start_date=penalty_accrual_start_date,
+                final_accrual_date=final_transaction_date,
+            )
```
Sepehr-Sobhani: nit: Not possible to hit but just for code quality, it would be good to include the `else` statement as well.

bc_obps/compliance/service/penalty_calculation_service.py
```diff
@@ -422,13 +506,12 @@ def _get_rate_for_date(cls, date_to_check: date) -> Decimal:
 
     @classmethod
     @transaction.atomic
```
Sepehr-Sobhani: This function no longer requires `@transaction.atomic`

bc_obps/compliance/service/penalty_calculation_service.py
```diff
@@ -363,39 +344,142 @@ def determine_last_transaction_date(obligation: ComplianceObligation) -> date |
...
+        compliance_penalty_record.elicensing_invoice = invoice_record
+        compliance_penalty_record.save()
+        # Set obligation penalty status
+        obligation.penalty_status = ComplianceObligation.PenaltyStatus.NOT_PAID
+        obligation.save(update_fields=['penalty_status'])
+
+        with transaction.atomic():
+            # Create compliance_penalty_accrual records for audit purposes
+            for acc in penalty_data.daily_accumulated_list:
+                CompliancePenaltyAccrual.objects.create(
```
Sepehr-Sobhani: Do you think we should make this a `get_or_create` function as well?

Sepehr-Sobhani (approved):
🥇

PR #4422, opened 2026-04-01
Enhance QA experience using django admin

Sepehr-Sobhani:


bc_obps/compliance/admin.py
```diff
@@ -1,14 +1,15 @@
...
-        'report_compliance_summary',
-        'excess_emissions_delta_from_previous',
-        'credited_emissions_delta_from_previous',
-        'status',
+        'reporting_year',
+        'start_date',
+        'end_date',
+        'compliance_deadline',
+        'invoice_generation_date',
+        'max_credit_usage_percentage',
```
andrea-williams, line 14: I'm a bit confused as to how this change is relevant to django-admin, but I'm assuming it's not a big risk since it won't affect prod

andrea-williams (approved):
One failing backend test that seems like it's relevant to the changes in this PR.

I haven't reviewed this in detail but I'm sure it's good 😄  and it's not going to prod so the risk is lower.

bc_obps/compliance/admin.py
Sepehr-Sobhani: Everything in `admin.py` is related to the Django Admin. In this PR, I mostly removed existing admin models and kept only what’s needed for QA testers. Happy to give a quick demo if you’re interested.

andrea-williams: nah it's all good, I trust ya 😉

Issue #4425, opened 2026-04-02
Delete code related to NAICS code gate


andrea-williams:
#### Description of the Tech Debt

Delete the code that we were using for our chemical pulp gate. Keyword is `RESTRICTED_NAICS_CODE_FOR_REPORTING`.

The environment variable has already been deleted from our prod Openshift namespace, we just want to delete all references to this variable in our codebase.

#### Tech Debt Triage

The purpose of our technical debt triage process is to analyze technical debt to determine risk level of the technical debt and the value in tackling that technical debt.

#### Risk Value Scoring:

| Level  | Value                 |
| ------ | --------------------- |
| High   | <div align="center">3 |
| Medium | <div align="center">2 |
| Low    | <div align="center">1 |

| Technical Debt - Risk Types                                                                                                   | Level | Value |
| ----------------------------------------------------------------------------------------------------------------------------- | ----- | ----- |
| Business Area Risk - Risk of business area visibility / damage to user experience                                             | 0     | 0     |
| Developer Fault Risk - How likely will this tech debt cause a future error related to coding on top of it                     | 0     | 0     |
| System Fault Risk - Risk of system errors or application downtime                                                             | 0     | 0     |
| Time Scale Risk - Compound risk effect if left alone. How much more difficult to fix or dangerous will this become over time? | 0     | 0     |
| Time Sink Risk - How much will this tech debt slow the development process down                                               | 0     | 0     |
| <div align="right">**TOTAL SCORE:**                                                                                           | 0     | 0     |

#### Development Checklist:

- [ ] Checklist item
- [ ] Checklist item
- [ ] Checklist item

joshgamache:
I have tested this in the DEV environment, here is the [test document for reference](https://bcgov.sharepoint.com/:w:/r/teams/00608-ScrumTeam/Shared%20Documents/Scrum%20Team/OBPS/QA%20Test/4425_Test_Document.docx?d=wce0059c289804ddb9dbc2fc64dea05d4&csf=1&web=1&e=pYEah5).

I can verify that:

- [x] When I use the previously restricted NAICS code, no errors are flagged where they should be removed.

Moving to Pending PO for approval. Thank you!

cc: @patriciarussellCAS

patriciarussellCAS:
@TammieV-BCG should be a quick UAT

patriciarussellCAS:
Working in Test

PR #4427, opened 2026-04-02
chore: removed all references to RESTRICTED_NAICS_CODES_FOR_REPORTING

andrea-williams:
Ticket #4425 

### Changes:
- removed frontend & backend usage of NAICS code gate
- deleted backend tests that were specifically about testing the gate
- updated backend tests that made reference to the gate (but where tests were still relevant)


Sepehr-Sobhani (approved):
🧹 Nice!

PR #4437, opened 2026-04-07
chore: printing stack trace if pytest is running

pbastia:


shon-button (approved):
Thanks very much for this!

PR #4445, opened 2026-04-08
chore: update wording on allocation of emissions missing product alert

### LINKED ISSUE #954: Wording Update: Alert in Allocation of Emissions page
#### Describe the Bug:

~~Given that a regulated operation has a mix of both regulated and unregulated products,
When user has filled out the information for all regulated products in the Production Data page,
Then the alert in the Allocation of Emissions page should not be visible~~

Update the alert on allocation of emissions page, so that users who select a mix of regulated and unregulated products are not confused if they do not see their unregulated products on this page

Image

- [x] To allocate emissions to a product that isn’t shown below, return to the [production data page] and select the product first. Note: unregulated products will not be shown below.
- [x] production data page is hyperlinked

#### Probability:

#### Impact:

Bug Matrix

1. Step bcgov/cas-reporting#1
2. Step bcgov/cas-reporting#2
3. Step bcgov/cas-reporting#3

#### Screenshots:

Image

This is the alert that is displayed in the Allocation of Emissions page
#### Additional information:

Version: [e.g. 22]

#### Additional Context

-

gdalcengio:
resolves https://github.com/bcgov/cas-reporting/issues/954

### DIFF
--- a/bciers/apps/reporting/src/data/jsonSchema/facility/MissingProductAlertFieldTemplate.tsx
+++ b/bciers/apps/reporting/src/data/jsonSchema/facility/MissingProductAlertFieldTemplate.tsx
@@ -18,7 +18,8 @@ const MissingProductAlertContent: React.FC<Props> = (props) => {
   return (
     <div>
       To allocate emissions to a product that isn&apos;t shown below, return to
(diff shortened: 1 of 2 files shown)


Sepehr-Sobhani (approved):
🥇

PR #4482, opened 2026-04-14
Reporting e2e: new SFO/LFO report

Sepehr-Sobhani:
[ISSUE 1000](https://github.com/bcgov/cas-reporting/issues/1000)

### DIFF
--- a/bc_obps/registration/fixtures/mock/operation.json
+++ b/bc_obps/registration/fixtures/mock/operation.json
@@ -55,6 +55,7 @@
       "submission_date": "2023-12-16T15:27:00.000Z",
       "type": "Single Facility Operation",
       "naics_code": 21,
+      "bcghg_id": "23219990028",
       "regulated_products": [2, 6, 7, 8],
       "status": "Registered",
       "bc_obps_regulated_operation": "23-0001",
--- a/bc_obps/registration/fixtures/mock/operation.json
+++ b/bc_obps/registration/fixtures/mock/operation.json
@@ -59,7 +59,7 @@
       "regulated_products": [2, 6, 7, 8],
       "status": "Registered",
       "bc_obps_regulated_operation": "23-0001",
-      "activities": [1, 2],
+      "activities": [1],
       "registration_purpose": "OBPS Regulated Operation",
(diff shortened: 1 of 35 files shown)


bc_obps/registration/fixtures/mock/operation.json
```diff
@@ -55,6 +55,7 @@
       "submission_date": "2023-12-16T15:27:00.000Z",
       "type": "Single Facility Operation",
       "naics_code": 21,
+      "bcghg_id": "23219990028",
```
pbastia, line 58: Do we have other tests relying on this bcghgid being empty in the system?

bciers/apps/reporting-e2e/poms/current-reports.ts
```diff
@@ -108,6 +109,46 @@ export class CurrentReportsPOM {
...
+   * the new report version ID extracted from the URL.
+   */
+  async startNewReportForOperation(operationName: string): Promise<number> {
+    await waitForGridReady(this.page);
+
+    const row = this.page
+      .getByRole("row")
+      .filter({ hasText: operationName })
+      .first();
+    await expect(row).toBeVisible();
```
pbastia, line 126: What happens if the row is on the second page?
should we mock the search? or probably have an option in the constructor to pass the operation search string, like
`.../reporting/reports/current-reports?operation_name=Banana`

bc_obps/registration/fixtures/mock/operation.json
Sepehr-Sobhani: That’s why I pushed this code earlier than the rest to see if it breaks any other tests 😄

bciers/apps/reporting-e2e/poms/current-reports.ts
Sepehr-Sobhani: We could. I thought that might be better as its own test for searching; For this test just going directly to that operation to keep the test fast and simple.

shon-button (approved):
Inspirational!  💫 🤩 

Thanks for all the foundational work here, it will be very helpful for subsequent e2e tests.

PR #4483, opened 2026-04-14
fix: remove unevaluated methodology fields in activity raw fixture + re-enable report_validation for e2e tests

### LINKED ISSUE #941: Reporting/Bug: json schema validation fails on report submission, methodology was unexpected
#### Update March 31, 2026:
Issue seems to be
> our mock methodology fixtures in reporting/fixtures/mock/report_methodology.json have empty json_data fields {}. These fixtures are used when calling submitReportById() to submit reports in order to test the compliance e2e scenarios. This empty {} for the methodology json data is not valid and would not be possible through regular flow through the frontend as it would have blocked saving of that with an error showing the methodology as required.

To do:
- [ ] ~~fill out the empty `json_data` fields in `reporting/fixtures/mock/report_methodology.json` with realistic data~~
- [x] fix data in `reporting/fixtures/mock/report_raw_activity_data.json ` with proper data
- [x] re-enable the validator in `report_activity_json_validation.py`
- [x] Assess that the error doesn't happen anymore with the compliance e2e for example
- [x] Submit a small and happy PR

If the error persists, tag the issue with exploration and refinement again, to be discussed at refinement.

#### Describe the Bug:

```bash
"Validation error: Unevaluated properties are not allowed ('methodology' was unexpected) at: sourceTypes >  
gscWithProductionOfUsefulEnergy > units > 0 > fuels > 0 > emissions > 0"
```

the (disabled) validator in `report_activity_json_validation.py` fails randomly and especially during e2e compliance tests (why?)
I have a suspicion that if something is entered and saved from the actual frontend, things work fine, but otherwise there are mistakes around the "emissions" schema that's missing a methodology field.

The 'unevaluatedProperties' inserted during the validation highlights this.
- An avenue to explore should be to change the "dependencies" to a "oneOf" statement that plays better with that one?
- Or just add the missing property? 
- There is something odd about the methodology field declaring dependent schemas when already being an enum:
```
"methodology": {
                                        "type": "object",
                                        "properties": {
                                          "methodology": {
                                            "title": "Methodology",
                                            "type": "string",
                                            "enum": [
                                              "CEMS",
                                              "Replacement Methodology",
                                              "Alternative Parameter Measurement Methodology",
                                              "Measured Steam/Measured EF",
                                              "Measured CC",
                                              "Measured Steam/Default EF",
                                              "Measured HHV/Default EF",
                                              "Default EF",
                                              "Default HHV/Default EF"
                                            ]
                                          }
                                        },
                                        "unevaluatedProperties": false,
                                        "dependentSchemas": {
                                          "methodology": {
                                            "oneOf": [
                                              {
                                                "properties": {
                                                  "methodology": {
                                                    "enum": [
                                                      "CEMS"
                                                    ]
                                                  }
                                                }
                                              },
                                              {
                                                "properties": {
                                                  "methodology": {
                                                    "enum": [
                                                      "Replacement Methodology"
                                                    ]
                                                  }
                                                }
                                              },
                                           ...
```

Anyway: investigate what's up.
A good way of reproducing the error is to run the compliance e2e tests: `CI=true yarn run compliance:e2e`
Understanding [dependentSchemas](https://json-schema.org/understanding-json-schema/reference/conditionals#dependentSchemas) better 

#### Probability:

#### Impact:

Bug Matrix

1. Step #1
2. Step #2
3. Step #3

#### Additional information:

Version: [e.g. 22]

#### Additional Context

-

gdalcengio:
fixes [941](https://github.com/bcgov/cas-reporting/issues/941)

### DIFF
--- a/bc_obps/reporting/service/report_validation/validators/report_activity_json_validation.py
+++ b/bc_obps/reporting/service/report_validation/validators/report_activity_json_validation.py
@@ -40,11 +40,6 @@ def enable_jsonschema_draft_2020_validation(schema: Any) -> None:
 
 
 def validate(report_version: ReportVersion) -> dict[str, ReportValidationError]:
-    logger.warning(
-        "Report activity json validation has been suspended until the jsonschema draft2020 can be fully implemented."
-    )
-    return {}
-
     errors = {}
 
     for facility_report in report_version.facility_reports.all():
--- a/bc_obps/reporting/service/report_validation/validators/report_activity_json_validation.py
+++ b/bc_obps/reporting/service/report_validation/validators/report_activity_json_validation.py
@@ -6,6 +6,7 @@ from reporting.models.report_version import ReportVersion
 from reporting.models.source_type import SourceType
 from reporting.service.report_validation.report_validation_error import (
     ReportValidationError,
+    ReportValidationErrorKey,
     Severity,
 )
 from service.form_builder_service import FormBuilderService
@@ -71,6 +72,7 @@ def validate(report_version: ReportVersion) -> dict[str, ReportValidationError]:
                     ReportValidationError(
                         severity=Severity.ERROR,
                         message=f"Validation error: {e.message} at: {' > '.join(str(elt) for elt in e.path)}",
+                        key=ReportValidationErrorKey.ACTIVITY_JSON_SCHEMA_VALIDATION_ERROR,
                     )
                 )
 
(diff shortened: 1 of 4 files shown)


bc_obps/reporting/service/report_validation/validators/report_activity_json_validation.py
shon-button, line 1: @gdalcengio 

FYI: this file will now need to to use a validation tag implemented in PR: https://github.com/bcgov/cas-registration/pull/4501


```

from reporting.service.report_validation.report_validation_tags import ValidationTags

TAGS = [ValidationTags.ON_SUBMIT]
```

gdalcengio: file has been rebased and errors have been fixed now

pbastia (approved):
amazing!

PR #4501, opened 2026-04-16
chore: define tags so each validator knows where to run

dleard:


bc_obps/reporting/service/report_validation/report_validation_service.py
```diff
@@ -35,13 +36,14 @@ class ReportValidationService:
     validation_plugins = collect_validation_plugins()
 
     @staticmethod
-    def validate_report_version(version_id: int) -> dict[str, ReportValidationError]:
+    def validate_report_version(version_id: int, tag: ValidationTags) -> dict[str, ReportValidationError]:
```
pbastia: I'd make this `tag` parameter optional, so that we run all validation plugins if not supplied

pbastia (approved):
Looks good! I think one scenario is not defined: what if the `TAGS` property is not defined on the validator? we could either throw an error at import time, or have a check to see if it's defined before calling it

Issue #4455, opened 2025-05-26
Reporting/Internal Users: Add Reporting Year to Report Attachments page


fviduya:
#### Describe the task

Following bcgov/cas-reporting#750, there are some changes to be made for the Report Attachments page
1. Add reporting year to the grid as the first column
2. Display reports by most recent reporting year 

[Wireframe](https://www.figma.com/design/LsDmLDCdnJqI8UahyqvOD8/BCIERS-Designs?node-id=38448-12750&t=iWwvPFfDwXJbgTcB-4)

#### Acceptance Criteria

- [x] Given I'm an Internal User (all user types), when I am on the Report Attachments page, then I have a "Reporting Year" column with a search bar, so that I can know which reporting year the attachment is from
- [x]  Given I'm an Internal User (all user types), when I am on the Report Attachments page, then the "Reporting Year" column is set to show the current reporting year at the top, descending down from there, so that I can access the most relevant documents easiest. 
- [x] Given that there are supplementary reports, then the report attachments grid will display all version's attachments in order of latest report submission

#### Development checklist

- [ ] confirm that this works now so users can see old versions of attachments even when a new one is uploaded in a supplementary report
- [x] Add reporting year to API response and filtering schema
- [x] Add reporting year to column definition on the grid

#### Definition of Done

- [x] The Reporting year column is visible
- [x] The search bar for the Reporting Year column works
- [x] The default setting for the Reporting Year column is latest reporting year descending to earliest reporting year. 

#### Additional context

- Add any other context about the task here.
- Or here

patriciarussellCAS:
Question for devs: If an industry report submits more than one attachment for the same reporting year, do we still see both? And what order do they appear in (i.e. should we default to show current RY + latest reporting version ID)?

pbastia:
@patriciarussellCAS it's a bit confusing at the moment. The table shows ALL the attachments from all submitted versions, and there is no filtering by year that was implemented.
so:
> If an industry report submits more than one attachment for the same reporting year, do we still see both? 

yes

> And what order do they appear in (i.e. should we default to show current RY + latest reporting version ID)?

Ordering is done by the table, but we should definitely show that default order.

fviduya:
[...]

PR #4514, opened 2026-04-20
Task/add reporting year attachments page/4455

andrea-williams:
Card: #4455 

#### Changes made:

- added Reporting Year column (with search bar) to grid for Report Attachments Download (CAS internal users only)
- changed default sorting behaviour of grid so that it's in descending order by report_version_id (because report versions are number sequentially, this automatically means that more recent reporting years will be displayed first, and that supplementary versions of a report will appear before their originals)
- replaced a chain of if/else statements to a match case to make Pierre happier 😺 
- refactored the relevant API endpoint to instead use a new function I created in the ReportAttachmentService - better code practices

#### Changes not made:

- didn't add E2E tests because we don't have any pre-created poms/E2E infrastructure for this page. Made new ticket #4516 instead
- didn't update user access permissions, so despite what ticket #4455  says, cas-view-only users will not be able to access this page. I figured that changing user permissions was beyond the scope of this ticket

### DIFF
--- a/bc_obps/reporting/api/report_attachments.py
+++ b/bc_obps/reporting/api/report_attachments.py
@@ -116,7 +116,7 @@ def get_all_attachments(
             mapped_sort_field = "report_version__report__operator__legal_name"
         case "operation":
             mapped_sort_field = "report_version__report__operation__name"
-        case "reporting_year":
+        case "reporting_year_id":
             mapped_sort_field = "report_version__report__reporting_year_id"
         case _:
             mapped_sort_field = sort_field or "id"
--- a/bc_obps/reporting/api/report_attachments.py
+++ b/bc_obps/reporting/api/report_attachments.py
@@ -1,4 +1,5 @@
 from typing import List, Literal, Optional, Tuple
+from reporting.models.report_attachment import ReportAttachment
 from common.api.utils.current_user_utils import get_current_user_guid
 from common.permissions import authorize
 from django.db import transaction
@@ -8,8 +9,6 @@ from ninja import File, Form, Query, UploadedFile
 from ninja.pagination import paginate
 from registration.utils import CustomPagination
 from reporting.constants import EMISSIONS_REPORT_TAGS
-from reporting.models.report_attachment import ReportAttachment
-from reporting.models.report_version import ReportVersion
 from reporting.schema.generic import Message
 from reporting.schema.report_attachment import (
     AttachmentsWithConfirmationOut,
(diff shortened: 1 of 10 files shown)


bc_obps/reporting/api/report_attachments.py
```diff
@@ -106,24 +105,9 @@ def get_report_attachment_url(request: HttpRequest, version_id: int, file_id: in
 def get_all_attachments(
     request: HttpRequest,
     filters: InternalReportAttachmentFilterSchema = Query(...),
-    sort_field: Optional[str] = "id",
+    sort_field: Optional[str] | Optional[(str)] = "report_version_id",
```
Sepehr-Sobhani: I think both `Optional[str]` and `Optional[(str)]` are the same.

bciers/apps/reporting/src/app/components/attachments/attachmentsList/AttachmentsListPage.tsx
```diff
@@ -5,6 +5,7 @@ import AttachmentsListGrid from "./AttachmentsListGrid";
 
 export type Attachment = {
   id: number;
+  reporting_year: number;
```
Sepehr-Sobhani, line 8: In the API schema, we’re returning `reporting_year_id`, so we should either update this type or change that field in the schema.

bciers/apps/reporting/src/tests/components/attachments/attachmentsList/AttachmentListGrid.test.tsx
```diff
@@ -37,19 +37,18 @@ describe("The AttachmentsListGrid component", () => {
...
       "Operator",
       "Operation",
       "Version ID",
       "Type",
       "Download",
     ]) {
-      expect(
-        screen.getByRole("columnheader", { name: column }),
-      ).toBeInTheDocument();
+      expect(screen.getByRole("columnheader", { name: column })).toBeVisible();
```
Sepehr-Sobhani, line 47: 😍

bc_obps/reporting/schema/report_attachment.py
```diff
@@ -43,5 +44,6 @@ class InternalReportAttachmentFilterSchema(FilterSchema):
     operator: Annotated[str | None, Field(q="report_version__report__operator__legal_name__icontains")] = None
     operation: Annotated[str | None, Field(q="report_version__report__operation__name__icontains")] = None
     report_version_id: Annotated[int | None, Field(q="report_version__id")] = None
+    reporting_year_id: Annotated[int | None, Field(q="report_version__report__reporting_year__reporting_year")] = None
```
Sepehr-Sobhani: Should we add `__icontains` so we cover both `2022` and `22` when searching for the year?

Sepehr-Sobhani (approved):
🤩

PR #4521, opened 2026-04-21
Fix/increment chart with release

joshgamache:
Addresses release bug. Increments the Backend Migration Test Chart's `version` when running _release-it_. It previously only updated the `appVersion` along with the `cas-bciers` chart, causing the _chart release_ action to fail due to the "new" chart.

## Changes 🚧

- Update the chart version bump (on release) to also bump the `version` by a patch.


dleard (approved):
:shipit:

PR #4534, opened 2026-04-23
4466 preserve boolean capture_emissions in transformReportAdditionalData

shon-button:
resolves [4466](https://github.com/bcgov/cas-reporting/issues/4466)


andrea-williams (approved):
Image

confirmed that when I select Yes in the form, it shows Yes (with my inputted values) on the final review page. Also confirmed that selected No renders as No.

Nice!

PR #4538, opened 2026-04-23
3198 - playwright accessibility issues

### LINKED ISSUE #3198: Registration/Tech Debt: Fix playwright accessibility issues in Registration workflow e2e tests
#### Description of the Tech Debt

In bciers/apps/registration-e2e/tests/register-an-operation.spec.ts, we commented out the `analyzeAccessibility` check where it was failing. This ticket is about fixing the failures and re-enabling the check.

**In the Register an Existing LFO test:**
```
Array [
+   Object {
+     "description": "Ensures every ARIA progressbar node has an accessible name",
+     "help": "ARIA progressbar nodes must have an accessible name",
+     "helpUrl": "https://dequeuniversity.com/rules/axe/4.9/aria-progressbar-name?application=playwright",
+     "id": "aria-progressbar-name",
+     "impact": "serious",
+     "nodes": Array [
+       Object {
+         "all": Array [],
+         "any": Array [
+           Object {
+             "data": null,
+             "id": "aria-label",
+             "impact": "serious",
+             "message": "aria-label attribute does not exist or is empty",
+             "relatedNodes": Array [],
+           },
+           Object {
+             "data": null,
+             "id": "aria-labelledby",
+             "impact": "serious",
+             "message": "aria-labelledby attribute does not exist, references elements that do not exist or references elements that are empty",
+             "relatedNodes": Array [],
+           },
+           Object {
+             "data": Object {
+               "messageKey": "noAttr",
+             },
+             "id": "non-empty-title",
+             "impact": "serious",
+             "message": "Element has no title attribute",
+             "relatedNodes": Array [],
+           },
+         ],
+         "failureSummary": "Fix any of the following:
+   aria-label attribute does not exist or is empty
+   aria-labelledby attribute does not exist, references elements that do not exist or references elements that are empty
+   Element has no title attribute",
+         "html": "<span class=\"MuiCircularProgress-root MuiCircularProgress-indeterminate MuiCircularProgress-colorPrimary mui-1sgvys3-MuiCircularProgress-root\" role=\"progressbar\" style=\"width: 40px; height: 40px;\">",
+         "impact": "serious",
+         "none": Array [],
+         "target": Array [
+           ".MuiCircularProgress-root",
+         ],
+       },
+     ],
+     "tags": Array [
+       "cat.aria",
+       "wcag2a",
+       "wcag111",
+       "EN-301-549",
+       "EN-9.1.1.1",
+     ],
+   },
+   Object {
+     "description": "Ensures elements with an ARIA role that require child roles contain them",
+     "help": "Certain ARIA roles must contain particular children",
+     "helpUrl": "https://dequeuniversity.com/rules/axe/4.9/aria-required-children?application=playwright",
+     "id": "aria-required-children",
+     "impact": "critical",
+     "nodes": Array [
+       Object {
+         "all": Array [],
+         "any": Array [
+           Object {
+             "data": Object {
+               "messageKey": "unallowed",
+               "values": "[role=progressbar]",
+             },
+             "id": "aria-required-children",
+             "impact": "critical",
+             "message": "Element has children which are not allowed: [role=progressbar]",
+             "relatedNodes": Array [
+               Object {
+                 "html": "<span class=\"MuiCircularProgress-root MuiCircularProgress-indeterminate MuiCircularProgress-colorPrimary mui-1sgvys3-MuiCircularProgress-root\" role=\"progressbar\" style=\"width: 40px; height: 40px;\">",
+                 "target": Array [
+                   ".MuiCircularProgress-root",
+                 ],
+               },
+             ],
+           },
+         ],
+         "failureSummary": "Fix any of the following:
+   Element has children which are not allowed: [role=progressbar]",
+         "html": "<div class=\"MuiDataGrid-main mui-204u17-MuiDataGrid-main\" role=\"grid\" aria-colcount=\"4\" aria-rowcount=\"3\" aria-multiselectable=\"false\">",
+         "impact": "critical",
+         "none": Array [],
+         "target": Array [
+           ".MuiDataGrid-main",
+         ],
+       },
+     ],
+     "tags": Array [
+       "cat.aria",
+       "wcag2a",
+       "wcag131",
+       "EN-301-549",
+       "EN-9.1.3.1",
+     ],
+   },
+ ]
```

**In Register a New SFO** (this only happens in CI and playwright flagged as flaky, so it may be testing before the page is fully loaded and in fact there are no violations):
```
Array [
    +   Object {
    +     "description": "Ensures the contrast between foreground and background colors meets WCAG 2 AA minimum contrast ratio thresholds",
    +     "help": "Elements must meet minimum color contrast ratio thresholds",
    +     "helpUrl": "https://dequeuniversity.com/rules/axe/4.9/color-contrast?application=playwright",
    +     "id": "color-contrast",
    +     "impact": "serious",
    +     "nodes": Array [
    +       Object {
    +         "all": Array [],
    +         "any": Array [
    +           Object {
    +             "data": Object {
    +               "bgColor": "#105058",
    +               "contrastRatio": 2.55,
    +               "expectedContrastRatio": "4.5:1",
    +               "fgColor": "#658f94",
    +               "fontSize": "10.5pt (14px)",
    +               "fontWeight": "normal",
    +               "messageKey": null,
    +             },
    +             "id": "color-contrast",
    +             "impact": "serious",
    +             "message": "Element has insufficient color contrast of 2.55 (foreground color: #658f94, background color: #105058, font size: 10.5pt (14px), font weight: normal). Expected contrast ratio of 4.5:1",
    +             "relatedNodes": Array [
    +               Object {
    +                 "html": "<div class=\"MuiPaper-root MuiPaper-elevation MuiPaper-elevation6 MuiSnackbarContent-root mui-1actny7\" role=\"alert\" direction=\"up\" style=\"opacity: 1; transform: scale(1, 1); transition: opacity 225ms cubic-bezier(0.4, 0, 0.2, 1) 0ms, transform 150ms cubic-bezier(0.4, 0, 0.2, 1) 0ms;\">",
    +                 "target": Array [
    +                   ".MuiPaper-root",
    +                 ],
    +               },
    +               Object {
    +                 "html": "<div class=\"bg-bc-primary-blue relative left-1/2 transform -translate-x-1/2 w-screen max-w-none\">",
    +                 "target": Array [
    +                   "footer > .bg-bc-primary-blue.w-screen.max-w-none",
    +                 ],
    +               },
    +             ],
    +           },
    +         ],
    +         "failureSummary": "Fix any of the following:
    +   Element has insufficient color contrast of 2.55 (foreground color: #658f94, background color: #105058, font size: 10.5pt (14px), font weight: normal). Expected contrast ratio of 4.5:1",
    +         "html": "<div class=\"MuiSnackbarContent-message mui-1w0ym84\">Operation Representative saved successfully</div>",
    +         "impact": "serious",
    +         "none": Array [],
    +         "target": Array [
    +           ".MuiSnackbarContent-message",
    +         ],
    +       },
    +     ],
    +     "tags": Array [
    +       "cat.color",
    +       "wcag2aa",
    +       "wcag143",
    +       "TTv5",
    +       "TT13.c",
    +       "EN-301-549",
    +       "EN-9.1.4.3",
    +       "ACT",
    +     ],
    +   },
    + ]
```

#### Tech Debt Triage

The purpose of our technical debt triage process is to analyze technical debt to determine risk level of the technical debt and the value in tackling that technical debt.

#### Risk Value Scoring:

| Level  | Value                 |
| ------ | --------------------- |
| High   | <div align="center">3 |
| Medium | <div align="center">2 |
| Low    | <div align="center">1 |

| Technical Debt - Risk Types                                                                                                   | Level | Value |
| ----------------------------------------------------------------------------------------------------------------------------- | ----- | ----- |
| Business Area Risk - Risk of business area visibility / damage to user experience                                             | 0     | 0     |
| Developer Fault Risk - How likely will this tech debt cause a future error related to coding on top of it                     | 0     | 0     |
| System Fault Risk - Risk of system errors or application downtime                                                             | 0     | 0     |
| Time Scale Risk - Compound risk effect if left alone. How much more difficult to fix or dangerous will this become over time? | 0     | 0     |
| Time Sink Risk - How much will this tech debt slow the development process down                                               | 0     | 0     |
| <div align="right">**TOTAL SCORE:**                                                                                           | 0     | 0     |

#### Development Checklist:

- [ ] Checklist item
- [ ] Checklist item
- [ ] Checklist item

gdalcengio:
Resolves: [3198](https://github.com/bcgov/cas-registration/issues/3198)

- Uncommented the analyzeAccessibility() lines that were causing errors. One error was no longer there, the color-contrast was still there.
- I've added the tags 'wcag2aa', 'wcag22aa', and 'best-practice' according to @fviduya 's suggestion in the original ticket
        - This is still reporting the issues of empty-table-header, color-contrast, 

- The empty-table-header alert described here:
`https://dequeuniversity.com/rules/axe/4.9/empty-table-header?application=playwright`
        - Our action column header leaf cell is empty purposefully, so this will be excluded

- The color-contrast error is to do with the snackbar disappearing I'm assuming. Adding a function to wait for it to disappear seems to have resolved this issue

### DIFF
--- a/bciers/libs/e2e/src/utils/helpers.ts
+++ b/bciers/libs/e2e/src/utils/helpers.ts
@@ -17,14 +17,16 @@ import path from "node:path";
 export async function analyzeAccessibility(
   page: Page,
   description: string = "",
+  disableRules?: string[],
 ) {
-  const accessibilityScanResults = await new AxeBuilder({
+  const builder = new AxeBuilder({
     page,
-    // ignore empty-table-headers since our search bar action cell is intenionally empty
-  })
-    .withTags(["wcag2aa", "wcag22aa", "best-practice"])
-    .disableRules(["empty-table-header"])
-    .analyze();
+  });
+
+  if (disableRules && disableRules.length > 0)
+    builder.disableRules(disableRules);
+
+  const accessibilityScanResults = await builder.analyze();
 
   if (accessibilityScanResults.violations.length > 0) {
     // eslint-disable-next-line no-console
--- a/bciers/libs/e2e/src/utils/helpers.ts
+++ b/bciers/libs/e2e/src/utils/helpers.ts
@@ -20,7 +20,11 @@ export async function analyzeAccessibility(
 ) {
   const accessibilityScanResults = await new AxeBuilder({
     page,
-  }).analyze();
+    // ignore empty-table-headers since our search bar action cell is intenionally empty
+  })
+    .withTags(["wcag2aa", "wcag22aa", "best-practice"])
+    .disableRules(["empty-table-header"])
+    .analyze();
 
   if (accessibilityScanResults.violations.length > 0) {
     // eslint-disable-next-line no-console
@@ -447,6 +451,15 @@ export async function assertSuccessfulSnackbar(
     .getByText(message);
   await snackbarLocator.waitFor();
   await expect(snackbarLocator).toBeVisible();
+}
+
+export async function waitForSnackbarToHide(
+  page: Page,
+  message: string | RegExp,
+) {
+  const snackbar = page.locator(".MuiSnackbarContent-root").getByText(message);
+
+  await snackbar.waitFor({ state: "detached" });
 }
 
 export async function clickWithRetry(
--- a/bciers/apps/administration-e2e/tests/workflows/contacts.spec.ts
+++ b/bciers/apps/administration-e2e/tests/workflows/contacts.spec.ts
@@ -63,7 +63,7 @@ test.describe("Test add/edit contact", () => {
       component: "Contacts grid",
       variant: "filled",
     });
-    await analyzeAccessibility(page);
+    await analyzeAccessibility(page, "Contacts grid", ["empty-table-header"]);
 
     await viewDetails.click();
     // Check for presence of heading
--- a/bciers/apps/administration-e2e/tests/workflows/contacts.spec.ts
+++ b/bciers/apps/administration-e2e/tests/workflows/contacts.spec.ts
@@ -63,8 +63,7 @@ test.describe("Test add/edit contact", () => {
       component: "Contacts grid",
       variant: "filled",
     });
-    // TODO: investigate accessibility failing
-    // await analyzeAccessibility(page);
+    await analyzeAccessibility(page);
 
     await viewDetails.click();
     // Check for presence of heading
--- a/bciers/apps/administration-e2e/tests/workflows/issue-boro-bcghg-id.spec.ts
+++ b/bciers/apps/administration-e2e/tests/workflows/issue-boro-bcghg-id.spec.ts
@@ -13,6 +13,7 @@ import {
   assertSuccessfulSnackbar,
   clickButton,
   urlIsCorrect,
+  waitForSnackbarToHide,
 } from "@bciers/e2e/utils/helpers";
 
 const test = setupBeforeAllTest(UserRole.CAS_DIRECTOR);
@@ -58,6 +59,9 @@ test.describe("Issue BCGHG and BORO ID", () => {
       component: "BORO ID and BCGHG ID issued",
       variant: "default",
     });
+
+    await waitForSnackbarToHide(page, SnackbarMessages.ISSUED_BCGHG_ID);
+    await waitForSnackbarToHide(page, SnackbarMessages.ISSUED_BORO_ID);
     await analyzeAccessibility(page);
   });
 });
--- a/bciers/apps/registration-e2e/tests/register-an-operation.spec.ts
+++ b/bciers/apps/registration-e2e/tests/register-an-operation.spec.ts
@@ -78,8 +78,7 @@ test.describe("Test register operations", () => {
       variant: "filled",
     });
     await clickButton(registrationPage.page, /save operation representative/i);
-    // TODO fix the accessibility errors and uncomment:  https://github.com/bcgov/cas-registration/issues/3198
-    // await analyzeAccessibility(page, componentName);
+    await analyzeAccessibility(page, componentName);
     await clickButton(registrationPage.page, /continue/i); // button on this form is `Continue` instead of `Save and Continue`
     await registrationPage.waitForRegistrationUrl(5);
 
@@ -134,8 +133,7 @@ test.describe("Test register operations", () => {
       component: componentName,
       variant: "filled",
     });
-    // TODO fix the accessibility errors and uncomment:  https://github.com/bcgov/cas-registration/issues/3198
-    // await analyzeAccessibility(page, componentName);
+    await analyzeAccessibility(page, componentName);
     await clickButton(registrationPage.page, /continue/i); // button on this form is `Continue` instead of `Save and Continue`
     await registrationPage.waitForRegistrationUrl(3);
 


bciers/libs/e2e/src/utils/helpers.ts
```diff
@@ -20,7 +20,11 @@ export async function analyzeAccessibility(
 ) {
   const accessibilityScanResults = await new AxeBuilder({
     page,
-  }).analyze();
+    // ignore empty-table-headers since our search bar action cell is intenionally empty
+  })
+    .withTags(["wcag2aa", "wcag22aa", "best-practice"])
```
Sepehr-Sobhani: I’d suggest not using specific tags and running the tests with the default. Using these tags means we’d lose the following rules:

### WCAG 2.x Level A — Critical

| Rule | Description |
|---|---|
| `area-alt` | `<area>` elements of image maps have alternate text |
| `aria-allowed-attr` | Element role supports its ARIA attributes |
| `aria-braille-equivalent` | `aria-braillelabel` and `aria-brailleroledescription` have a non-braille equivalent |
| `aria-command-name` | Every ARIA button, link and menuitem has an accessible name |
| `aria-conditional-attr` | ARIA attributes are used as described in the specification of the element's role |
| `aria-deprecated-role` | Elements do not use deprecated roles |
| `aria-hidden-body` | `aria-hidden="true"` is not present on the document body |
| `aria-hidden-focus` | aria-hidden elements are not focusable nor contain focusable elements |
| `aria-input-field-name` | Every ARIA input field has an accessible name |
| `aria-meter-name` | Every ARIA meter node has an accessible name |
| `aria-progressbar-name` | Every ARIA progressbar node has an accessible name |
| `aria-prohibited-attr` | ARIA attributes are not prohibited for an element's role |
| `aria-required-attr` | Elements with ARIA roles have all required ARIA attributes |
| `aria-required-children` | Elements with an ARIA role that require child roles contain them |
| `aria-required-parent` | Elements with an ARIA role that require parent roles are contained by them |
| `aria-roledescription` | `aria-roledescription` is only used on elements with an implicit or explicit role |
| `aria-roles` | All elements with a role attribute use a valid value |
| `aria-toggle-field-name` | Every ARIA toggle field has an accessible name |
| `aria-tooltip-name` | Every ARIA tooltip node has an accessible name |
| `aria-valid-attr-value` | All ARIA attributes have valid values |
| `aria-valid-attr` | Attributes that begin with `aria-` are valid ARIA attributes |
| `audio-caption` | `<audio>` elements have captions |
| `blink` | `<blink>` elements are not used |
| `button-name` | Buttons have discernible text |
| `bypass` | Page has at least one mechanism to bypass navigation and jump to content |
| `definition-list` | `<dl>` elements are structured correctly |
| `dlitem` | `<dt>` and `<dd>` elements are contained by a `<dl>` |
| `document-title` | Each HTML document contains a non-empty `<title>` element |
| `duplicate-id-aria` | Every id used in ARIA and in labels is unique |
| `form-field-multiple-labels` | Form field does not have multiple label elements |
| `frame-focusable-content` | `<frame>` and `<iframe>` elements with focusable content do not have `tabindex=-1` |
| `frame-title-unique` | `<iframe>` and `<frame>` elements contain a unique title attribute |
| `frame-title` | `<iframe>` and `<frame>` elements have an accessible name |
| `html-has-lang` | Every HTML document has a lang attribute |
| `html-lang-valid` | The lang attribute of the `<html>` element has a valid value |
| `html-xml-lang-mismatch` | HTML elements with both valid lang and xml:lang attributes agree on the base language |
| `image-alt` | `` elements have alternate text or a role of none or presentation |
| `input-button-name` | Input buttons have discernible text |
| `input-image-alt` | `<input type="image">` elements have alternate text |
| `label` | Every form element has a label |
| `link-in-text-block` | Links are distinguished from surrounding text in a way that does not rely on color |
| `link-name` | Links have discernible text |
| `list` | Lists are structured correctly |
| `listitem` | `<li>` elements are used semantically |
| `marquee` | `<marquee>` elements are not used |
| `meta-refresh` | `<meta http-equiv="refresh">` is not used for delayed refresh |
| `nested-interactive` | Interactive controls are not nested inside each other |
| `no-autoplay-audio` | `<video>` or `<audio>` elements do not autoplay audio for more than 3 seconds without a control mechanism |
| `object-alt` | `<object>` elements have alternate text |
| `p-as-heading` | Bold, italic text and font-size are not used to style `<p>` elements as a heading |
| `role-img-alt` | `[role="img"]` elements have alternate text |
| `scrollable-region-focusable` | Elements with scrollable content are accessible by keyboard |
| `select-name` | `<select>` element has an accessible name |
| `server-side-image-map` | Server-side image maps are not used |
| `svg-img-alt` | `<svg>` elements with an img/graphics-document/graphics-symbol role have accessible text |
| `table-fake-caption` | Tables with a caption use the `<caption>` element |
| `td-has-header` | Each non-empty data cell in a table larger than 3×3 has one or more table headers |
| `td-headers-attr` | Each cell using the headers attribute refers only to other cells in that table |
| `th-has-data-cells` | `<th>` elements and elements with role=columnheader/rowheader have data cells they describe |
| `video-caption` | `<video>` elements have captions |

### WCAG 2.1 Level AA

| Rule | Description |
|---|---|
| `autocomplete-valid` | The autocomplete attribute is correct and suitable for the form field |
| `avoid-inline-spacing` | Text spacing set through style attributes can be adjusted with custom stylesheets |
| `css-orientation-lock` | Content is not locked to any specific display orientation |

### WCAG 2.1 Level A

| Rule | Description |
|---|---|
| `label-content-name-mismatch` | Elements labelled through their content have their visible text as part of their accessible name |

### WCAG Level AAA

| Rule | Description |
|---|---|
| `color-contrast-enhanced` | Foreground/background colors meet WCAG AAA enhanced contrast ratio thresholds |
| `identical-links-same-purpose` | Links with the same accessible name serve a similar purpose |
| `meta-refresh-no-exceptions` | `<meta http-equiv="refresh">` is not used for delayed refresh |

bciers/libs/e2e/src/utils/helpers.ts
```diff
@@ -20,7 +20,11 @@ export async function analyzeAccessibility(
...
+    .disableRules(["empty-table-header"])
```
Sepehr-Sobhani: This would disable the rule for all tables, but some tables don’t have an action column. We should make `analyzeAccessibility` flexible enough to bypass rules only where needed, not by default.

Sepehr-Sobhani (changes requested):

bciers/libs/e2e/src/utils/helpers.ts
gdalcengio: Just pushed a commit that makes it more flexible, thanks

Sepehr-Sobhani (approved):
🤩 Great work!

PR #4552, opened 2026-04-27
fix: Correct ProductEmissionIntensity for RY2025, Chemical Pulp product

andrea-williams:
added reporting migration to update PWAEI for Chemical Pulp

To Check:
- before running migration, note the Regulated Product.id for "Pulp and paper: chemical pulp" (in my local it's ID 16, so probably yours too) and then query the Product Emission Intensity table for that product_id (in my local it's product_emission_intensity.id 18) - you'll see the PWAEI is 0.3177, with valid_to="9999-12-31")
- after running migration, you should see that product_emission_intensity.id=18 now has a valid_to of "2024-12-31" (for reporting year 2024), and there's a new record (probably ID 39?) for product_id=16, pwaei=0.3095, valid_from="2025-01-01" (for reporting year 2025), valid_to="9999-12-31"
- should also confirm that the PWAEI for regulated product "Pulp and paper: lime recovered by kiln" is 0.3822


bc_obps/reporting/tests/service/test_compliance_service/test_compliance_service_class.py
```diff
@@ -277,8 +277,8 @@ def test_compliance_summary_with_pulp_and_paper_regulatory_override(self):
             'credited_emissions': Decimal('0.0000'),
             'emissions_attributable_for_compliance': Decimal('30000.9989'),
             'emissions_attributable_for_reporting': Decimal('40501.0489'),
-            'emissions_limit': Decimal('7346.6017'),
-            'excess_emissions': Decimal('22654.3972'),
+            'emissions_limit': Decimal('7293.5067'),
+            'excess_emissions': Decimal('22707.4922'),
```
andrea-williams, line 281: if you're wondering how I calculated these new values - I didn't. I just took the values that the test itself was spitting out. Super risky, but I don't know how to calculate the correct values otherwise (now that the PWAEI has changed)

joshgamache (approved):

PR #4559, opened 2026-04-28
Chore/4529 protect slugs trigger

dleard:
Adds a trigger to the tables that have a "slug" column to ensure those fields do not change.
There was some _registration_ data (Activity) updated in a recent _reporting_ migration to fix the slugs, so I had to create a dependency on that migration and update it to ignore the trigger.
This means that the check-immutable-migrations CI check will fail until this gets into main.


Sepehr-Sobhani (approved):
🥇

PR #4561, opened 2026-04-28
chore: release v5.9.0

gdalcengio:


andrea-williams (changes requested):
something's gone wrong - there's 2 migration files for v5.9 for each of the directories. Could you please delete the second one for each?

PR #4573, opened 2026-04-29
fix error message handling for validation errors

Sepehr-Sobhani:
## Summary

- `generate_useful_error` in `registration/utils.py` raised `AttributeError: 'ValidationError' object has no attribute 'error_dict'` when a Django `ValidationError` was constructed with a plain string or list instead of a dictionary. Only dict-form `ValidationError`s have `message_dict`; string/list forms do not.
- Fixed by guarding with `hasattr(error, 'message_dict')` and adding an `else` branch that falls back to `error.messages`, which is always available regardless of how the error was raised.
- Refactored the function using early returns to flatten nested `if` blocks, reducing Cognitive Complexity from 17 to 15.

### DIFF
--- a/bc_obps/registration/tests/test_utils.py
+++ b/bc_obps/registration/tests/test_utils.py
@@ -123,6 +123,16 @@ class TestGenerateUsefulError:
         expected_error = "Field1: Error message 1 for field1"
         assert useful_error == expected_error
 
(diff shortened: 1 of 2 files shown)


pbastia (approved):
Thank you so much!

Issue #4600, opened 2026-05-05
Reporting/Task: Order of months and quarters for Cement Production in the Review Changes page


fviduya:
#### Describe the task

In the Cement Production activity, when Methodology = `Calcination Emissions`, months and quarters fields are displayed. 

Given that industry user has submitted a report that includes data for Cement Production activity, then creates a supplementary report that changes values in the said activity, the months and quarters are not displayed in chronological order in the Review Changes page.

<img width="362" height="682" alt="Image" src="https://github.com/user-attachments/assets/b517f8b7-213a-4099-bb72-76b120141f24" />

<img width="286" height="307" alt="Image" src="https://github.com/user-attachments/assets/ceabc4cb-0492-4cce-af70-bbb037c7ef37" />

#### Acceptance Criteria

- [ ] The months and quarters are displayed in chronological order in the Review Changes page (e.g. January, February, ..., December, Quarter 1, Quarter 2, ...)

#### Definition of Done

- [ ] first
- [ ] second
- [ ] third

#### Additional context

- Final Review and Submitted page are already in chronological order, done under #4489

fviduya:
I have tested this in the DEV environment, [here is the test document for reference](https://buttoninc-my.sharepoint.com/:w:/g/personal/francesca_button_is/IQC8NLl8GGdISqugTin3zgjxAQ9lgqgh-FVPudF4AY4rPwM?e=wlYRB6).

I can verify that:
- [x] The months and quarters are displayed in chronological order in the Final Review page (e.g. January, February, ..., December, Quarter 1, Quarter 2, ...)
- [x] The months and quarters are displayed in chronological order in the Submitted page (e.g. January, February, ..., December, Quarter 1, Quarter 2, ...)
- [x] The months and quarters are displayed in chronological order in the Review Changes page (e.g. January, February, ..., December, Quarter 1, Quarter 2, ...)

Moving to Pending PO for review of test. Thank you!

patriciarussellCAS:
Moving to accepted. @TammieV-BCG for UAT.

TammieV-BCG:
Tested in TEST. works as expected

PR #4603, opened 2026-05-05
4600 sort methodologies chronologically in review changes page

Sepehr-Sobhani:
[ISSUE 4600](https://github.com/bcgov/cas-registration/issues/4600)


andrea-williams (approved):
👑

PR #4605, opened 2026-05-06
feat: o&g NP NC emission allocation validation

pbastia:


bc_obps/reporting/service/report_validation/validators/report_emission_allocation_other_excluded_category.py
```diff
@@ -0,0 +1,83 @@
...
+    for cat in basic_categories:
+        other_excluded_emission_total = (
+            ReportEmission.objects_with_decimal_emissions.select_related("report_source_type__report_activity")
+            .filter(report_source_type__report_activity__facility_report_id=facility_report.id)
+            .filter(emission_categories=cat)
+            .filter(emission_categories__category_type="other_excluded")
+            .aggregate(Sum("emission"))['emission__sum']
+        ) or Decimal("0")
+
+        allocated_to_non_reg_products = ReportProductEmissionAllocation.objects.filter(
```
dleard: Should we filter out the FOG product here? It will _always_ trigger this alert because it's handled in a special way that means there will be no excluded emissions to compare it to

bc_obps/reporting/service/report_validation/validators/report_emission_allocation_other_excluded_category.py
```diff
@@ -0,0 +1,83 @@
...
+            emission_category=cat,
+            report_emission_allocation__facility_report_id=facility_report.id,
+        ).aggregate(Sum("allocated_quantity"))['allocated_quantity__sum'] or Decimal("0")
+
+        if other_excluded_emission_total < allocated_to_non_reg_products:
+            facility_report_errors[
+                f"og_np_nc_allocation_mismatch_{facility_report.facility_id}_{cat.category_name}"
+            ] = ReportValidationError(
+                key=ReportValidationErrorKey.OG_NP_NC_ALLOCATION_MISMATCH,
+                message=f"For the emission category '{cat.category_name}', emissions allocated to the O&G non-processing, non-combustion product exceed the emissions reported under the 'other_excluded' category.",
```
dleard: Is this message too specific? The validator will also pick up the "Refineries Line Tracing" unregulated product (and any others that get added in the future)

bciers/apps/reporting/src/app/components/shared/validation/config.ts
```diff
@@ -241,6 +241,21 @@ If the value is accurate, you may save & continue.`;
...
+    renderMode: "inline_link",
+    getHref: (ctx) =>
+      facilityRoutes.allocationOfEmissions(
+        String(ctx?.report_version_id),
+        String(ctx?.facility_id),
+      ),
+    formatMessage: ({
+      label,
+      error: { context },
+    }) => `Facility ${context?.facility_name}: Please review the ${label} and ensure that emissions from non-processing
```
dleard: same question about the message being too specific here.

dleard (approved):
Looks great! Works as expected, but the message may be too specific?

PR #4616, opened 2026-05-07
Fix false positive changes when creating supplementary report with no changes

### LINKED ISSUE #4542: Reporting/Bug: Review Changes shows difference in Allocation of Emissions when there are no changes made in emissions
#### Describe the Bug:

Given that the industry user creates a supplementary report that does not change the emissions / activity forms (therefore no changes in the Emissions Summary page / Allocation of Emissions page), 
When industry user gets to the Review Changes page,
Then it shows some difference in the Allocation of Emissions page

Note: Does not affect compliance summary. There is also no difference in the values. See screenshot below.

#### Probability:

#### Impact:

Bug Matrix

1. Log in as bc-cas-dev
2. Click `View Annual Reports` under the Reporting tile
3. Click `Start` for Bugle SFO
4. Fill out all required fields until submission
5. Sign off and submit the report
6. Go back to the reporting grid
7. Click `Create Supplementary Report` for Bugle SFO
8. Click `Save & Continue` until you reach the Review Changes page

Expected Result: No changes visible in the page
Actual Result: Allocation of Emission section is visible, shows that there is a change made but the crossed out value and new value is similar

#### Screenshots:

Image

#### Additional information:

Version: [e.g. 22]

#### Additional Context

-

gdalcengio:
fixes [4542](https://github.com/bcgov/cas-registration/issues/4542)

Problem:
- On Creating of supplementary report, new `report_product`s are created. When checking diffs in each facility, deep diffs sees something like `{"allocated_quantity": 32.0, "report_product_id": 9},` vs `{"allocated_quantity": 32.0, "report_product_id": 7}`. 

Solution:
- I've added product_id to the schema to keep the comparison of product_id allocations consistent
- Excluded report_product_id ONLY on the review-changes page because the allocation schema is used on the final review page as well

### DIFF
--- a/bc_obps/reporting/api/report_review_changes.py
+++ b/bc_obps/reporting/api/report_review_changes.py
@@ -10,6 +10,16 @@ from .permissions import approved_industry_user_report_version_composite_auth
 from service.report_version_service import ReportVersionService
 from reporting.service.review_changes_service.report_review_changes_service import ReportReviewChangesService
 
+EXCLUDE_REPORT_PRODUCT_ID = {
+    "facility_reports": {
+        "__all__": {
+            "report_emission_allocation": {
+                "report_product_emission_allocations": {"__all__": {"products": {"__all__": {"report_product_id"}}}},
+            }
+        }
+    }
+}
+
 
 @router.get(
     "/report-version/{version_id}/diff-data",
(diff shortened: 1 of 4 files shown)


andrea-williams (approved):
Ideally we'd have unit tests for the review changes functionality so that we could test this fix, but we don't...  ☹️ 

Submitted a supplementary report locally - confirmed that no product changes are shown in Review Changes page when production data hasn't changed. Also confirmed that when I submit a supplementary where production data HAS changed, the diff shows up in the Review Changes page.

Nicely done!

Issue #4592, opened 2026-05-04
Reporting/Bug: Error in the Production Data page when there are no regulated products


fviduya:
#### Describe the Bug:

Error message introduced in #4261 is visible in the Production Data page when an operation has no regulated product and user is filling out a report for a Small Aggregate/Medium Facility

#### Probability:

(How likely the bug is to happen, scored from 1-5. For example, probability of 5 is something like "it happens to all users every time they log in." whereas probability of 1 "only happens to certain users when a really specific and unlikely path is followed.")

#### Impact:

(How bad the bug is when it does happen, scored from 1-5. For example, effect of 5 is "the entire app crashes and makes it unusable for all users" or "the bug causes the wrong data to be saved, with critical information (e.g. payment) being affected." whereas effect of 1 is "It makes some styling look a little bit weird.")

<img src="https://www.getzephyr.com/sites/default/files/riksk%20matrix.jpg" alt="Bug Matrix" width="250" height="250"/>

#### Steps to reproduce the behaviour:

1. Log in as bc-cas-dev
2. Click `View annual reports` under the Reporting tile
3. Click `Start` for Banana LFO
4. Remove the selected products
5. Add the following products: Refineries line tracing, Oil and gas non-processing non-compression
6. Click `Save & Continue` until you reach the Review Facilities page
7. Select only the following facilities: Facility 4 (Small Aggregate), Facility 3 (Medium Facility), Facility 1 (Large Facility)
8. Click `Save & Continue`
9. Click `Continue` for Facility 1
10. Click `Production Data` from the task list
11. Click `Save` or `Save and Continue`

Expected Result: No error message
Actual Result: Error message `This form can't be saved yet. Please fix the errors above.` is visible

12. Repeat steps 9-11 for the other facilities

Expected Result: No error message displayed
Actual Result: Error message `This form can't be saved yet. Please fix the errors above.` is visible

 #### Screenshots:

<img width="764" height="237" alt="Image" src="https://github.com/user-attachments/assets/aa619b06-3e81-4286-8e1e-ef251e4bc5b0" />

#### Additional information:

Browser: [e.g. chrome, safari]
Version: [e.g. 22]

#### Additional Context

-

dleard:
Sounds like we just need a short circuit in that Production Data Page where if there are no regulated products and the operation type is LFO. That we just display a message that says "No Regulated Products Selected" and a Continue button.

fviduya:
[...]

PR #4628, opened 2026-05-08
Fix/4592 allow lfo facility to continue without selecting a regulated product

dleard:
#4592 LFO facilities can now continue past the ProductionDataForm without selecting any regulated products.

To test:

Scenario 1: A regulated product is available to select

- Start a report for an LFO operation
- Ensure both a regulated and non-regulated product is selected on the operation info page
- Go to the production data form
- Do not select the regulated product
- Click "Save & Continue"
- The page should not error & continue successfully to the next section

Scenario 2: no regulated product is available to select

- Start a report for an LFO operation
- Ensure only a non-regulated product is selected on the operation info page (ie: Refineries line tracing)
- Go to the production data form
- There should be no products to select & a message that says "No Regulated Products to select"
- Click "Continue"
- The page should not error & continue successfully to the next section

### DIFF
--- a/bciers/apps/reporting/src/app/components/products/ProductionDataForm.tsx
+++ b/bciers/apps/reporting/src/app/components/products/ProductionDataForm.tsx
@@ -1,11 +1,15 @@
 "use client";
 import MultiStepFormWithTaskList from "@bciers/components/form/MultiStepFormWithTaskList";
 import { useState } from "react";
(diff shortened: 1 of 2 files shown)


andrea-williams (approved):
Tested locally with both scenarios described in PR - both work as expected.

Note that there is Sonar complaint about using === true in a test.

Issue #4582, opened 2026-04-30
Reporting/Bug: Required fields not highlighted when user says yes on Non attributable emissions page.


TammieV-BCG:
#### Describe the Bug:
Required fields not highlighted when users attempts to proceed with out filling all required fields on Non attributable emissions page.

#### Steps to reproduce the behaviour:

1. log in as external user 
2. start annual report for any registered SFO 
3. fill all required fields till  you reach the Non-attributable emissions page.
4. select yes on "Did your non-attributable emissions exceed 100 tCO2e?'
5. select any Emission Category
6. click save 
7. Observe that an error message appear at the bottom but the relevant fields aren't highlighted

#### Screenshots:

<img width="1041" height="707" alt="Image" src="https://github.com/user-attachments/assets/02d5eb2b-b6d3-4bdc-9ced-873eebb87b1c" />

#### Additional information:
another minor discrepancy: The existing error message states the missing field (albeit one at a time), but note that the referenced field name in the error message is different from actual field name eg the "Activity name" field is just called activity in error message

dleard:
The issue where we're only showing one error at a time _likely_ comes from `customTransformErrors()`

fviduya:
I have tested this in the DEV environment, [here is the test document for reference](https://buttoninc-my.sharepoint.com/:w:/g/personal/francesca_button_is/IQDFKka8nUvNR6VmwX47P_gGAQgL5LDJcA9oIg1Mk89PN8M?e=Ah7BCi).

I can verify that:
- Given that industry user is filling out the Non-attributable Emissions page, when any of the following fields are not entered, the following errors appear:
   - [x] `Activity name is required` - when Activity name is left blank
   - [x] `Source type is required` - when Source type is left blank
   - [x] `Emission Category is required` - when user has not selected an emission category 
   - [x] `Select at least one option` - when user has not selected any gas type
- [x] When user runs into errors, a generic error message is displayed visible at the top of the Save buttons: `This form can't be saved yet. Please fix the errors above`

Moving to Pending PO for review of test. Thank you!

patriciarussellCAS:
Moving to accepted. @TammieV-BCG for UAT

patriciarussellCAS:
@fviduya a bit of a random question, but are you able to confirm whether this issue (see reporting ticket 531) was resolved as part of 4582, or is it still an issue? https://github.com/bcgov/cas-reporting/issues/531

fviduya:
[...]

PR #4630, opened 2026-05-08
4582 Fix Non attributable emissions page required fields

Sepehr-Sobhani:
[ISSUE 4582](https://github.com/bcgov/cas-registration/issues/4582)

## Summary

- **Bug fix**: Required fields (`activity`, `source_type`, `emission_category`) on the non-attributable emissions form were not highlighted when left blank. Added `required` to the activity items in the frontend JSON schema so RJSF highlights missing fields client-side before the form reaches the backend.
- **`gas_type` now required**: Added `gas_type` as a required field with `minItems: 1` in the frontend JSON schema so RJSF fires a "Select at least one option" validation error when no gas type is selected. Added `Field(min_length=1)` to `ReportNonAttributableIn` so the backend independently rejects empty gas type lists with a 422. Fixed `CheckboxGroupWidget` to not pass `required` down to individual `<Checkbox>` elements (which caused asterisks to appear on each option instead of just the field label).
- **Backend GET response**: Refactored the `GET` endpoint to return `{ emissions_exceeded, activities }` instead of a flat list, so the frontend can correctly initialise `emissions_exceeded` from the server rather than inferring it from array length.
- **`customTransformErrors` fix**: The regex matching for array-level required errors (e.g. `.activities`) was too broad — it also matched sub-field errors (e.g. `.activities.0.activity`), causing them to display "Select at least one option" instead of a field-level message. Tightened the regex to match only the array field itself.
- **Schema cleanup**: Removed redundant `to_camel`/`to_snake` wrapper functions across multiple schema files and replaced with `alias_generators` directly. Cleaned up `ReportNonAttributableIn` to remove manually re-declared fields that `ModelSchema` already provides.
- **Frontend refactor**: Moved `gasTypeMap`/`emissionCategoryMap` computation from `NonAttributableEmissionsPage` into `NonAttributableEmissionsForm` (they are only consumed there), added explicit types (`ActivityFormItem`, `NonAttributableFormData`), and used a lazy `useState` initializer so map computation only runs once on mount. Simplified the Page to use `Promise.all` for parallel data fetching.

## Test plan

Follow steps in the issue.

### DIFF
--- a/bciers/apps/reporting/src/app/components/reportInformation/nonAttributableEmissions/NonAttributableEmissionsPage.tsx
+++ b/bciers/apps/reporting/src/app/components/reportInformation/nonAttributableEmissions/NonAttributableEmissionsPage.tsx
@@ -12,18 +12,19 @@ export default async function NonAttributableEmissionsPage({
   version_id,
   facility_id,
 }: HasFacilityId) {
-  const gasTypes = await getBasicGasTypes();
-  const emissionCategories = await getAllEmissionCategories();
-  const emissionFormData = await getNonAttributableEmissionsData(
-    version_id,
-    facility_id,
-  );
-
-  const tasklistData = await getReportInformationTasklist(
-    version_id,
-    facility_id,
-  );
-  const orderedActivities = await getOrderedActivities(version_id, facility_id);
+  const [
+    gasTypes,
+    emissionCategories,
+    emissionFormData,
+    tasklistData,
+    orderedActivities,
+  ] = await Promise.all([
+    getBasicGasTypes(),
+    getAllEmissionCategories(),
+    getNonAttributableEmissionsData(version_id, facility_id),
+    getReportInformationTasklist(version_id, facility_id),
+    getOrderedActivities(version_id, facility_id),
+  ]);
 
   const navigationInformation = await getNavigationInformation(
     HeaderStep.ReportInformation,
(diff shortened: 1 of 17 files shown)


bciers/apps/reporting/src/app/components/reportInformation/nonAttributableEmissions/NonAttributableEmissionsPage.tsx
```diff
@@ -12,18 +12,19 @@ export default async function NonAttributableEmissionsPage({
...
-    version_id,
-    facility_id,
-  );
-
-  const tasklistData = await getReportInformationTasklist(
-    version_id,
-    facility_id,
-  );
-  const orderedActivities = await getOrderedActivities(version_id, facility_id);
+  const [
```
shon-button, line 15: 🚀 😍

bciers/apps/reporting/src/data/jsonSchema/nonAttributableEmissions/nonAttributableEmissions.ts
```diff
@@ -75,6 +75,7 @@ export const generateUpdatedSchema = (
               type: "array",
               items: {
                 type: "object",
+                required: ["activity", "source_type", "emission_category"],
```
shon-button: @Sepehr-Sobhani 

Question:
Is it a valid state to have a non-attributable emission record exist without any gas types selected?

If, yes it can then we need to update the required field validator `bc_obps/reporting/service/report_validation/validators/required_fields/required_fields_report_operation_information.py` 

Image

Sepehr-Sobhani: I didn’t make that field required because it’s an M2M field, so technically it can contain some, all, or no selections. But I’m going to ping @dleard and @pbastia to get their insight on it.

shon-button (approved):
@Sepehr-Sobhani 

Fantastic fixes — this is a really solid example of thoughtful refactoring decisions, and a great guideline for the kind of patterns and cleanup I’ll try to apply in the future as well.

Sepehr-Sobhani:
> @Sepehr-Sobhani
> 
> Fantastic fixes — this is a really solid example of thoughtful refactoring decisions, and a great guideline for the kind of patterns and cleanup I’ll try to apply in the future as well.

Thank you so much, Shon! 🙏 
I’ve noticed some code smells in Reporting, especially around how we use `Schema` vs `ModelSchema`, how we can better take advantage of `ModelSchema`, and how the logic is split between services, endpoints, and frontend schemas. As we work on new features or bug fixes, we can gradually clean these up and refactor things to make the overall structure clearer and easier to maintain.

Issue #4526, opened 2026-04-22
Switch to Docker Hardened Images


joshgamache:
#### Tech Debt Triage

We currently use standard, widely available images as our base images for our containers. For increased security, we could switch to [Docker Hardened Images](https://www.docker.com/products/hardened-images/). Exploration work has shown that this will require some changes to images beyond just updating the `FROM` source. This includes adding health checks to all the containers and ensuring some particular packages are update.

#### Risk Value Scoring:

Not critical, but could be an easy win as long as we can keep up with it.

#### Development Checklist:

- [ ] Update all containers to use `dhi.io` base images
- [ ] Update packages to align with minimums for hardened images
- [ ] Add health checks to containers
  - [ ] Respond to health checks in charts

joshgamache:
Exploration was done on the [chore/harden](https://github.com/bcgov/cas-registration/tree/chore/harden) branch.

PR #4640, opened 2026-05-11
Chore/harden

joshgamache:
Addresses #4526. Makes use of [Docker Hardened Images](https://www.docker.com/products/hardened-images/) for image bases. 

## Changes 🚧

- Replace source images in backend pods with [dhi/python image](https://hub.docker.com/hardened-images/catalog/dhi/python/guides)
- Replace source images in frontend pods with [dhi/node image](https://hub.docker.com/hardened-images/catalog/dhi/node/guides)
- Refactored Dockerfiles to accommodate changes needed for DHI to run
- Add step to CI to extract Python's version from `/bc_obps/.tool-versions` to pass as an ARG to the backend Dockerfile

## To test 🔬

- Prior to merging, these images can be swapped out as replacements in local deployments and remotely in OpenShift by using this PR's commit hash after being built. 

## Notes 📝

We're using the `-dev` images from DHI for Node and Python so that we can still use shell scripting. These images are still more secured than our existing images. A future pass could remove any shell usage and switch to the non`-dev` DHI for the runner layers.

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,3 +1,3 @@
 nodejs 24.15.0
 gcloud 450.0.0
-python 3.12.11
+python 3.12.13
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,3 +1,3 @@
-nodejs 24.14.1
+nodejs 24.15.0
 gcloud 450.0.0
 python 3.12.11
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,3 +1,3 @@
-nodejs 24.15.0
+nodejs 24.16.0
 gcloud 450.0.0
 python 3.12.13
--- a/bc_obps/Dockerfile
+++ b/bc_obps/Dockerfile
@@ -1,7 +1,10 @@
+# In CI, the Python version is specified in the build args, see .github/workflows/build-backend.yaml
+ARG PYTHON_VERSION
+
 ####################
 # Base Stage
 ####################
-FROM dhi.io/python:3.12.13-debian13-dev AS base
+FROM dhi.io/python:${PYTHON_VERSION}-debian13-dev AS base
 
 # Prevent Python from writing .pyc files to disk
 ENV PYTHONDONTWRITEBYTECODE=1 \
--- a/bc_obps/Dockerfile
+++ b/bc_obps/Dockerfile
@@ -29,8 +29,6 @@ FROM base AS builder
 
 # Refresh package lists from repositories
 RUN apt-get update && \
-# Upgrade the installed system packages from apt-get update, making sure security is up to date
-    apt-get upgrade -y && \
 # Install minimal build tools required for asdf (git), curl for Poetry plugin, build-essential for compiling Python, and additional libs for Python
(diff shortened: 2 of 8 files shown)


Sepehr-Sobhani:
@joshgamache We can close PR [4453](https://github.com/bcgov/cas-registration/pull/4453) once this PR is merged.

bc_obps/Dockerfile
```diff
@@ -1,7 +1,10 @@
+# In CI, the Python version is specified in the build args, see .github/workflows/build-backend.yaml
+ARG PYTHON_VERSION
+
 ####################
 # Base Stage
 ####################
-FROM python:3.12.11-slim AS base
+FROM dhi.io/python:${PYTHON_VERSION}-debian13-dev AS base
```
pbastia, line 7: I think it's fine to fail a build if we don't supply the python version

bciers/Dockerfile
```diff
@@ -1,47 +1,57 @@
...
+COPY ./.next/static ./dist/${application_path}/.next/static
+
+ADD https://github.com/Yelp/dumb-init/releases/download/v1.2.5/dumb-init_1.2.5_x86_64 /usr/local/bin/dumb-init
+RUN chmod +x /usr/local/bin/dumb-init
+####################
+# Runner Stage
+# - Uses DHI.io runtime image which does not include shell
+# - Production image, copy all the files and run next
+####################
+FROM dhi.io/node:24.15.0-alpine3.23-dev AS runner
```
pbastia: 24.16 since a week or so ago! Maybe we could use the same pattern as from the backend image and grab this from the .tool-versions file, if it's not too complicated with Nx?

docs/developer-environment-setup-linux-os.md
```diff
@@ -182,7 +182,7 @@ This guide will help you set up the frontend development environment for the CAS
 ### Required Tools
 
 1. **Node.js**
-   - Version: 24.11.1 (as specified in `.tool-versions`)
+   - Version: 24.15.0 (as specified in `.tool-versions`)
```
pbastia: 24.16.0

.tool-versions
```diff
@@ -1,3 +1,3 @@
-nodejs 24.14.1
+nodejs 24.15.0
```
pbastia: 24.16.0

pbastia (approved):
Looks good! This simplifies the backend image a lot, that's sweet 

I left a few suggestions, mostly just grabbing the latest node that was released after the PR got put together

.tool-versions
joshgamache: Bumped with 92d3d15548bd202c1e0d4a9021b7e0447d176543.

docs/developer-environment-setup-linux-os.md
joshgamache: Bumped with 92d3d15548bd202c1e0d4a9021b7e0447d176543.

bciers/Dockerfile
joshgamache: Used .tool-versions with in 473025f and bumped with 92d3d15548bd202c1e0d4a9021b7e0447d176543.

PR #4647, opened 2026-05-13
Replace nx lint logic with direct ESLint execution to avoid PTY panic in CI

Sepehr-Sobhani:


joshgamache (approved):
Great workaround. Still get the advantages of `nx affected`, but not having to deal with the headache. Awesome to find the root cause of this and get around it!

PR #4657, opened 2026-05-13
Fix/reporting only validator ignore regulated products

dleard:
The required_fields validator `reporting/service/report_validation/validators/required_fields/required_fields_report_operation_information.py` was incorrectly firing when a reporting-only operation had no regulated products. This PR makes.a small edit to the validator to ignore the m2m regulated_products field as a requirement & adds a test for the logic for the reporting-only flow passing without regulated_products.

To test this:

- Create a report for a reporting-only operation
- Go to the validation page
- There should be no validation error

#### Before fix:
image

#### After fix:
image


pbastia (approved):
🥇

Issue #4653, opened 2026-05-13
Reporting/Bug: Annual Fuel Amount is optional on the front end/required on the backend


patriciarussellCAS:
#### Describe the Bug:
Based on a user error (Teams Thread [here](https://teams.microsoft.com/l/message/19:57fe54e9af5e4fefbf34e89fc3936e1f@thread.tacv2/1778686885753?tenantId=6fdb5200-3d0d-4a8a-b036-d3685e359adc&groupId=29821293-3244-455d-beae-e01d27762cfe&parentMessageId=1778686885753&teamName=External%3A%20CAS%20-%20Clean%20Growth%20Digital%20Service%20Team&channelName=Scrum%20Team&createdTime=1778686885753&ngc=true)) and while testing #4652 it was discovered that we have a bug where the annual fuel amount field is optional on the front end but required for validation. 

#### Impact: 3
Highly impactful only when users leave that field blank. Hopefully most will not do this, but it will block anyone who does from being able to complete reporting. 

#### Steps to reproduce the behaviour:

1. 
2. 
3. 

#### Screenshots:
------------------------------------------------ERROR START------------------------------------------------
Traceback (most recent call last):
  File "/Users/awilliam/Documents/cas-registration/bc_obps/.venv/lib/python3.12/site-packages/ninja/operation.py", line 134, in run
    result = self.view_func(request, **values)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/awilliam/Documents/cas-registration/bc_obps/reporting/api_v2/forms/report_validation_data.py", line 35, in get_report_validation_data
    errors = ReportValidationService.validate_report_version(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/awilliam/Documents/cas-registration/bc_obps/reporting/service/report_validation/report_validation_service.py", line 54, in validate_report_version
    errors.update(validation_plugin.validate(report_version))
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/awilliam/Documents/cas-registration/bc_obps/reporting/service/report_validation/validators/report_data_by_fuel_type_validator.py", line 125, in validate
    validate_fuel_amount(
[...]

PR #4660, opened 2026-05-13
fix: make annualFuelAmount required on the front-end forms (to match …

andrea-williams:
ISSUE #4653 

### Changes Made

- updated MANY RJSF schemas to make `fuelName`, `annualFuelAmount`, `emission`, and `gasType` required fields (where they appear in a schema). See [this comment](https://github.com/bcgov/cas-registration/issues/4653#issuecomment-4558740266) for full list of affected files
- updated `form_builder_service.py` so that it's not injecting required fields into the schemas. Instead the required fields are defined by the RJSF schemas
- added a `_helpers.py` file in `/migrations` that reloads activity and source_type json files into the database. Has been written to be reuseable, although will need updating as we add new schemas for future reporting years
- added a Django migration to re-load all schemas (using the helper functions)
- added logic to insert an error message in the UI when `fuelName` is left empty (previously there was no error message)

### Screenshots
- can no longer save an activity form if Annual Fuel Amount field is empty (users could before). Still can't save if Fuel Name, Gas Type, or Emissions fields are empty (same behaviour as before, but implemented differently):
Screenshot 2026-05-28 at 4 07 37 PM

- can no longer save Mobile Fuel Combustion activity form if quarterly fuel amount fields are empty:
Screenshot 2026-05-28 at 4 09 39 PM

#### Manual Testing setup when running locally

1. Run `cd ./bc_obps && make migrate` to update your local database
2. Because the `form_builder_service` caches results, you'll probably need to restart your backend server otherwise the reporting forms you see won't match the schemas actually in your database.

#### Suggested Manual Testing

1. Log in with `bc-cas-dev`

3. Start a new report (either SFO/LFO, Regulated/Reporting/NewEntrant - you choose)
4. Select multiple activities, including Mobile Fuel Combustion (has quarterly fuel amount fields, no unit), GSC excluding line tracing (has fuel and unit), Cement production (has no fuel fields), and whatever else you want. The RJSF schemas for every single activity and source type have been updated, but manually testing all of them isn't realllllly feasible (unless you're ambitious...)
5. Partially fill out the activity forms but leave some combination of these fields blank: [fuelName, annualFuelAmount, emission, gasType], then try to Save the form. **Expected behaviour: you should get an error message at the bottom of the form indicating there are errors, and you should see error messages next to the empty fields that are required**. Repeat this for all activity forms.
6. Populate your activity forms with the missing required fields wherever they appear. Save the forms.
7. Proceed through report to Report Validation page. **Expected behaviour: green message indicating no automatic validation errors detected.**
8. Proceed to end, submit the report. **Expected behaviour: report submits successfully.**


bc_obps/reporting/migrations/0188_update_sourcetype_required_quarterly_fuel_amounts.py
```diff
@@ -0,0 +1,39 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("reporting", "0187_reportpersonresponsible_contact"),
+    ]
+
+    operations = [
+        migrations.RunPython(
+            code=reload_mobile_combustion_source_type_schema,
+            reverse_code=migrations.RunPython.noop,
```
pbastia:
```suggestion
            reverse_code=migrations.RunPython.noop,
            elidable=True
```

dleard (dismissed):
I don't think we should be relying on injections from the form_builder service to apply required properties to json schema fields. These properties should be part of the json_schema definitions themselves and the updated schemas should be reloaded into the database.

I see that this kind of injection was done the first time for emission fields a couple months ago in a [different PR](https://github.com/bcgov/cas-registration/pull/4179). Here is my suggestion:

- Remove all instances where we're setting the "required" property in the form_builder service (looks like gas_type, emission & annualFuelAmount have been set this way).
- Update all the json schemas to set the "required" property for these fields
- Create a reusable reload_schemas_from_json() utility function to be used in this migration and future migrations where we need to update the database schemas from the json files.
- Reload all the json schemas in to the database

Let me know if you have questions!

bc_obps/reporting/migrations/_helpers.py
```diff
@@ -0,0 +1,1402 @@
...
+
+# Schemas for 2025 reporting year - only applies to pulp and paper activity
+ACTIVITY_SCHEMA_MAPPING_2025 = [('Pulp and paper production', 'pulp_and_paper_production')]
+
+ACTIVITY_SOURCE_TYPE_SCHEMA_MAPPING = [
+    # each tuple is structured as (activity_name, directory_path, schema_dir, file_name, source_type_name, has_unit, has_fuel)
+    # GSC excluding line tracing
+    (
+        'General stationary combustion excluding line tracing',
+        "reporting/json_schemas/2024",
```
andrea-williams, line 42: random aside - I asked GH copilot to insert the "reporting/json_schemas/2024" string into every tuple in this list, and it did it, but it chewed through the majority of my monthly allocated credits just doing this one simple, repetitive task. And copilot also complained about having to do it too.

bc_obps/reporting/json_schemas/2024/fuel_combustion_mobile/combustion_by_equipment.json
```diff
@@ -8,6 +8,12 @@
       "type": "array",
       "items": {
         "type": "object",
+        "required": [
+          "q1FuelAmount",
+          "q2FuelAmount",
+          "q3FuelAmount",
+          "q4FuelAmount"
+        ],
```
pbastia, line 17: Would we want to make the annualFuelAmount required too?

bc_obps/reporting/json_schemas/2024/lng_activities/10_centrifugal_compressor_venting.json
```diff
@@ -44,7 +44,8 @@
               }
             }
           }
-        }
+        },
+        "required": ["gasType", "emission"]
```
pbastia: This should apply to the emission items, a couple of levels down the tree

bc_obps/reporting/migrations/0192_reload_json_schemas_updated_required_fields.py
```diff
@@ -0,0 +1,16 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('reporting', '0191_add_report_version_to_report_models_missing_it'),
+    ]
+
+    operations = [
+        migrations.RunPython(load_activity_schemas_from_json, migrations.RunPython.noop),
+        migrations.RunPython(load_source_type_schemas_from_json, migrations.RunPython.noop),
```
pbastia: These migration steps could be made elidable, the old json schemas will have been loaded previously

bc_obps/reporting/tests/migration/test_helpers.py
```diff
@@ -0,0 +1,100 @@
...
+    for (_, directory_path, schema_dir, file_name, *_), path in zip(ACTIVITY_SOURCE_TYPE_SCHEMA_MAPPING, opened_paths):
+        expected_suffix = f"/{directory_path}/{schema_dir}/{file_name}.json"
+        assert path.endswith(expected_suffix), f"unexpected schema path: {path}"
+
+    # Spot-check kwargs passed to update_or_create() for the first tuple
+    first_call_kwargs = fake_model.objects.update_or_create.call_args_list[0].kwargs
+    activity_slug, _, _, _, source_type_json_key, has_unit, has_fuel = ACTIVITY_SOURCE_TYPE_SCHEMA_MAPPING[0]
+    assert first_call_kwargs["has_unit"] == has_unit
+    assert first_call_kwargs["has_fuel"] == has_fuel
+    assert first_call_kwargs["defaults"]["json_schema"] == FAKE_SCHEMA
```
pbastia: Why spot-check when...we can check them all :D

bc_obps/reporting/tests/migration/test_helpers.py
```diff
@@ -0,0 +1,100 @@
...
+    load_source_type_schemas_from_json(fake_apps, schema_editor=None)
+
+    expected = len(ACTIVITY_SOURCE_TYPE_SCHEMA_MAPPING)
+
+    # One open() + json.load() per mapping entry.
+    assert mock_open_fn.call_count == expected
+    assert mock_json_load.call_count == expected
+
+    # One ActivitySourceTypeJsonSchema.objects.update_or_create() per mapping entry.
+    assert fake_model.objects.update_or_create.call_count == expected
```
pbastia: I would hard-code these expectations, otherwise we're just testing that "x = x" with more steps involved

pbastia (changes requested):
Just a few minor things. Nice work!!

bc_obps/reporting/json_schemas/2024/fuel_combustion_mobile/combustion_by_equipment.json
andrea-williams: In this activity form annualFuelAmount is a read-only field that's calculated based on q1-4 fields, but I can def make annual required too if you want!

bc_obps/reporting/json_schemas/2024/lng_activities/10_centrifugal_compressor_venting.json
andrea-williams: excellent catch! I can't believe you noticed that in amongst the 150 other json files with changes!

pbastia: @andrea-williams I wrote a script 😁

bc_obps/reporting/tests/migration/test_helpers.py
andrea-williams: I'm not clear whether you want any changes here. Do you mean check every file in ACTIVITY_SOURCE_TYPE_SCHEMA_MAPPING?

pbastia: Yep!
intead of checking "call_args_list[0] == MAPPING[0]"
we can check it for every item:
```py
for index, kwargs in fake_model.objects.update_or_create.call_args_list:
    activity_slug, _, _ ... = MAPPING[index]
    assert kwargs["has_unit"] == has_unit
    # ... etc
```

bc_obps/reporting/json_schemas/2024/fuel_combustion_mobile/combustion_by_equipment.json
pbastia: It will be required by the backend on its way back so we should make it so, as a safety

pbastia: We still rely on the client to set it

pbastia (approved):
Approved with requested changes!

bc_obps/reporting/service/report_activity_save_service.py
```diff
@@ -210,11 +210,22 @@ def save_fuel(
...
+        fuel_type_data = fuel_data.get("fuelType")
+        if not isinstance(fuel_type_data, dict):
+            raise UserError("Fuel Name is required")
+
+        fuel_name = fuel_type_data.get("fuelName")
+        if not fuel_name:
+            raise UserError("Fuel Name is required")
+
+        if "annualFuelAmount" not in fuel_data:
+            raise UserError("Annual Fuel Amount is required")
```
andrea-williams, line 222: this is new post- @pbastia 's approval

Issue #4652, opened 2026-05-13
Reporting/LFO: Empty Activity form causing error and NOT refreshing after activity removed


patriciarussellCAS:
#### Describe the Bug:
If an LFO fills out a Facility Report that includes a blank activity form, they will get an error on the Report Validation page. 
If the user goes back to the Facility Information page, and deselects that activity. The error is still displayed on the Report Validation page. 

There are 2 different issues:
1. ~If there is a blank fuel record in their data, they will get an internal server error when they get to Report Validation page~ split into issue #4653 
2. In some cases (we are still investigating) If there is a blank activity record, it will show a warning on the Report Validation page. *If the user goes back and removed that activity form from the Facility Information page (and now they no longer see that activity form in their report) they still see the same warning message on the report validation page

<img width="2148" height="1044" alt="Image" src="https://github.com/user-attachments/assets/c1319d9c-e920-4489-8417-9daab4ef5a1d" />

#### Steps to reproduce the behaviour:

1. Start a report for an LFO 
2. When filling out a Facility Report, leave one of the activity forms blank
3. Continue to the Report Validation page
4. See a warning (or Report Validation page breaks?)
5. Go back to the facility information page and deselect the blank activity
6. Continue through the facility report again
7. Get to the Report Validation page and see that the warning is still there
Expected behaviour: Once the activity is removed from the facility report, the warning should no longer be visible (or the report validation page should display properly)

**Chesca's steps to reproduce**
1. Log in as bc-cas-dev
2. Click `View annual reports` under the Reporting tile
3. Click `Start` for Bees LFO (any regulated LFO)
4. Add another reporting activity - in this case, added Alumina
5. Click `Save & Continue` until you get to the Review Facility Information page
6. Select Alumina 
7. Click `Save & Continue`
8. Report only for GSC excluding line tracing activity
9. Click `Save & Continue`, fill out the required pages until you reach the Report Validation 
#### Additional information:

Browser: [e.g. chrome, safari]
Version: [e.g. 22]

#### Additional Context

-

fviduya:
I have tested this in the DEV environment, [here is the test document for reference](https://buttoninc-my.sharepoint.com/:w:/g/personal/francesca_button_is/IQDKRGAYptGLQY2b4zDsR5CBARLgQxqE3p3avTrbz9gTv_4?e=CV2hPw).

I can verify that:
- [x] When industry user selects an activity but leaves it empty, no error is encountered in the Report Validation page
[...]

PR #4662, opened 2026-05-13
fix: remove validator blocking expected LFO facility behaviour

dleard:
Issue #4652 

I've removed the `activity_data_coverage_validator`. It was checking that all facility reports report every activity defined in the operation information page. For LFOs this is not correct, as we **want** users to remove activities from each facility report that the facility does not do. For SFOs I don't think this is actually checking anything, since they don't have an option to remove activities after the operation information page.

Removing this validator will fix the issue where LFOs are incorrectly seeing the "activities are missing" validation error when their facility reports only report on a subset of the activities defiined by the operation.

This is the error that was being shown.
image


shon-button (approved):
@dleard 

Thanks for this correction.

For the SFO activity coverage validation, assuming we must trust that `create_report_version` successfully creates the associated report operation activity records.


**Note**: Test `bciers/apps/reporting/src/tests/components/shared/validation/config.test.ts\has the expected number of configs` needs updating to the new config number.

PR #4669, opened 2026-05-14
Chore/4667 pdb anti affinity

dleard:
Adds Pod Disruption Budgets and Node Anti-Affinity rules for bciers deployments for better reliability during cluster maintenance.


pbastia (approved):
🥇

Issue #4512, opened 2025-02-07
Reporting/Bug: missing report_version field on some report elements


pbastia:
#### Describe the Bug:

Not really a bug, but an invariant in our database design (that isn't documented sadly) to facilitate metabase lookups:
> Every data model part of a report version should have a foreign key to that report version

To do:
- Find missing key and add `report_version` fkey to those models (`ReportNonAttributableEmissions` is one of them)
- document this principle of database design 

#### Impact: 2

<img src="https://www.getzephyr.com/sites/default/files/riksk%20matrix.jpg" alt="Bug Matrix" width="250" height="250"/>

#### Steps to reproduce the behaviour:

1. Step bcgov/cas-reporting#1
2. Step bcgov/cas-reporting#2
3. Step bcgov/cas-reporting#3

#### Additional information:

Browser: [e.g. chrome, safari]
Version: [e.g. 22]

#### Additional Context

-

patriciarussellCAS:
@acatchpole or @fviduya did this ticket skip QA?

pbastia:
@patriciarussellCAS there isn't much to QA in this except to verify in Metabase that there is a new "report_version_id" column in the models adam identified:
- report_raw_activity_data
- report_new_entrant_emission
- report_new_entrant_production

...Which I just checked, all good :)

PR #4671, opened 2026-05-14
4512 Ensure all report-related models have report version

acatchpole:
Card: https://github.com/bcgov/cas-registration/issues/4512

### Changes
- Updated report-related models missing report_version foreign keys:
 1. report_raw_activity_data
 2. report_new_entrant_emission
 3. report_new_entrant_production
- Migration to add the report_version field to these models, and populate that field
- Updated services and tests related to these models to include report_version
- Updated data_models doc to include conventions for report-related models

### Changes not made
for the sake of keeping the migration simple, there were some possible updates that I decided not to implement, but could be considered as possible tech debt.
 - A ReportVersionBaseModel mixin that would include the `report_version` foreign key and the `immutable_report_version_trigger`. We could remove the field from the report-related models and composite this mixin with TimeStampedModel and ReportDataBaseModel
 - Update the `immutable_report_version_trigger` in the updated models to simplify the query to report_version (one less join)
 
 ### Testing (devs only. No QA testing required)
I have included 2 tests in TestPopulateReportVersionMigration. 

The second test is meant to be run with prod data to confirm that the report_version is populated as expected with real data. 

The first test tests actually running the migration but uses mock data. Due to the operations in the migration, this test can be flaky. To run this test successfully, you must set the migration to not run atomically (add ```atomic = False``` at line 52 in the migration).

This test file will be deleted before merging. The Reverse method in the migration is also only in there for the sake of the test (and probably would not work on actual submitted reports). That will also be removed before merge


bc_obps/reporting/migrations/0191_add_report_version_to_report_models_missing_it.py
```diff
@@ -0,0 +1,127 @@
...
+
+def reverse_populate_report_version_fields(apps, schema_editor):
+    """Reverse the population of report_version fields by clearing them."""
+    ReportNewEntrantEmission = apps.get_model('reporting', 'ReportNewEntrantEmission')
+    ReportNewEntrantProduction = apps.get_model('reporting', 'ReportNewEntrantProduction')
+    ReportRawActivityData = apps.get_model('reporting', 'ReportRawActivityData')
+
+    ReportNewEntrantEmission.objects.all().update(report_version_id=None)
+    ReportNewEntrantProduction.objects.all().update(report_version_id=None)
+    ReportRawActivityData.objects.all().update(report_version_id=None)
```
pbastia: This is not needed - reverting the field will erase the data already

pbastia (approved):
Looks good, thank you! We can just delete that manual revert step, removing the column will erase its contents anyways

bc_obps/reporting/migrations/0191_add_report_version_to_report_models_missing_it.py
acatchpole: yeah, that was just there because the migration test needed the option to reverse. removing this and the test.

Issue #4658, opened 2026-05-13
Reporting/Validation for Fuel Type - Update Diesel & NG Safe Ranges


patriciarussellCAS:
#### Describe the task

A follow-up to ticket https://github.com/bcgov/cas-reporting/issues/948
The BA would like to update the following: 
1. Fuel Type = Diesel
Field = Annual Fuel Amount (kl)
**"safe" value = >0 - 200,000**     (update from current safe value: >0 - 5,000)

2. Fuel Type = Natural Gas
Field = Annual Fuel Amount (Sm3)
**"Safe" value = >0 - 200,000,000** (update from current safe value >0 – 5,000,000)

#### Acceptance Criteria

- [x] a report that includes either of these fuels with a fuel amount with the safe range should not display any warnings on the Validation page
- [x] a report that includes one or more of these fuels with a fuel amount outside the safe range should display a warning on the Validation page

#### Dev Checklist

- [x] create a new Django migration that updates the existing `ExpectedValueRangeFuelAmount` records for Diesel and Natural Gas to match the new upper bounds

#### Definition of Done

- [ ] first
- [ ] second
- [ ] third

#### Additional context

- Add any other context about the task here.
- Or here

fviduya:
I have tested this in the DEV environment, [here is the test document for reference](https://buttoninc-my.sharepoint.com/:w:/g/personal/francesca_button_is/IQCLfXjwegUsQ6wYT_vWAiLuARyhIXUpBN59f_px_iDgQfI?e=bh9Bww).

I can verify that, given that an industry user is filling out a report:
- [x] When fuel name = Diesel, and the value entered by user in the `Annual Fuel Amount` field is **greater than or equal to 0 but less than 200,001**, no warning is encountered in the Report Validation page
- [x] When fuel name = Diesel, and the value entered by user in the `Annual Fuel Amount` field is **greater than 200,000**, a warning is encountered in the Report Validation page
- [x] When fuel name = Natural Gas, and the value entered by user in the `Annual Fuel Amount` field is **greater than or equal to 0 but less than 200,000,001**, no warning is encountered in the Report Validation page
- [x] When fuel name = Natural Gas, and the value entered by user in the `Annual Fuel Amount` field is **greater than 200,000,000**, a warning is encountered in the Report Validation page
- [x] When fuel name = Diesel/Natural Gas, and the value entered by user in the `Annual Fuel Amount` field is less than 0, an error is encountered in the activity page blocking the user from saving the value
[...]

PR #4672, opened 2026-05-14
4658 Update upper bounds for Diesel and Natural Gas fuel amount ranges

Sepehr-Sobhani:
[ISSUE 4658](https://github.com/bcgov/cas-registration/issues/4658)


shon-button (approved):
⭐

PR #4675, opened 2026-05-15
temp: 4558 add reporting field display titles

### LINKED ISSUE #4558: Reporting/Activity labels in Final Review & Review Changes pages
#### Describe the task

In the Final Review & Review Changes pages in Reporting, the `field_name` of each ReportingField is always displayed - this should be overridden by `field_display_title` wherever they're non-null.

#### Acceptance Criteria

- [x] for each ReportingField (see the database model), its label as it appears in the Reporting workflow should be the same in the Final Review and Review Changes pages

#### Dev Checklist
- [x] for any ReportingField, the frontend should first check whether a `field_display_title` exists for the record, and use that. If the `field_display_title` is null, then the frontend should use `field_name` for the label instead.

#### Additional context

- [ ] see `handle_methodologies()` in `form_builder_service.py` - for each methodology object, the `reporting_field.field_display_title` (stored in the database as an optional field) is prioritized for display, but if it's null, then `reporting_field.field_name` is used. This is because the `field_display_title` is intended to be the more user-friendly label name that can change over time, whereas the ReportingField's field_name is used as an index in some places (although we shouldn't be, we should actually use the field's slug instead...) and so shouldn't change over time. `field_display_title` will be null for a ReportingField if there hasn't been a need to display something other than the required `field_name` value
- [ ] because of form_builder_service's handle_methodologies() function, every ReportingField's label is displayed as desired in the Reporting workflow (where reporters are filling out their emissions data)
- [x] see `bciers/apps/reporting/src/app/components/finalReview/templates/FinalReviewReportSections.tsx` and `bciers/apps/reporting/src/app/components/shared/FacilityReportSection.tsx` - this is where the contents of every report are mapped out into sections for both the Final Review and (if needed) the Review Changes page. AW suspects that somewhere in here a change is required that checks for `field_display_title` first instead of `field_name`, although further investigation is required as part of the work for this ticket

#### Example of the Problem
- [x] fill out a report for any operation (in my example, I'm using "Compliance SFO - Obligation not met" from our dev fixtures). In the activities section, use General stationary combustion excluding line tracing. Under Emission, select CO2 gas type, and "Default HHV/Default EF" as the methodology. Note that here, the label reads "Default High Heating Value" (the . Save the data on this page.

Image

- [x] Navigate to the /final-review page for the report. Under the Report Information section for the activity, the field label is now "Fuel Default High Heating Value" - we want it to say "Default High Heating Value"

Image

See below screenshot for the ReportingField in prod db. For slug `fuelDefaultHighHeatingValue`, the Field Display Title is used when a reporter is filling out the report, but in the Final Review page it's the Field Name that's displayed instead. For supplementary reports, this will also be the case on the Review Changes page.

Image

shon-button:
resolves: [4558](https://github.com/bcgov/cas-registration/issues/4558)

### 🚀 Impact
* Added `ReportingField` display titles to activity data rendering in both `Review Changes` and `Final Review`

  * **Review Changes:** attached `field_display_title` directly to each diff item so the resolved display label travels with the exact changed field
  * **Final Review:** injected `_field_display_titles` metadata into methodology blocks for dynamic field label rendering
  * Preserved existing frontend key formatting as a fallback when no display title exists

* Refactored `report_review_changes_service.py` to reduce review-change noise by recursively removing system-managed fields before diffing (e.g. `report_product_id`, `report_version`, `status`, `is_latest_submitted`, `is_supplementary_report`)

  * Prevents false-positive “changes” caused by regenerated IDs/version metadata rather than actual user-edited report data:
          before: `changed: Array(24)`
          after: `changed: Array(7)`

### Screenshots:

image

image

### DIFF
--- a/bc_obps/reporting/schema/report_final_review.py
+++ b/bc_obps/reporting/schema/report_final_review.py
@@ -23,6 +23,7 @@ from reporting.models import (
     SourceType,
     ReportNewEntrantEmission,
     ReportNewEntrantProduction,
+    ReportingField,
 )
 from reporting.schema.compliance_data import ComplianceDataSchemaOut
 from reporting.schema.emission_category import EmissionSummarySchemaOut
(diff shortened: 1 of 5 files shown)


gdalcengio (approved):
Fantastic, much cleaner solution and works as intended

Issue #4587, opened 2026-05-01
Reporting/ Rename facility report download file name


zoeyli-46:
#### Describe the task

When a user downloads facility reports,  all facility reports are named the same way (CAS OBPS_Reporting. This will be a poor user experience for LFOs and verifiers receiving LFO reports. An LFO user must rename all of their reports downloads so they can keep track of the facility reports, and if they don't rename it according to the facility name, it will be difficult for verifiers to keep track of facility reports as well.

<img width="800" height="67" alt="Image" src="https://github.com/user-attachments/assets/8b1b4080-bbb5-48d8-a78d-1bd53bfd9241" />

#### Acceptance Criteria

- [x] Given I am downloading a facility report, the file name appends the facility name in this format: CAS OBPS_Reporting_FacilityName
For example:

<img width="617" height="84" alt="Image" src="https://github.com/user-attachments/assets/e210d298-6876-44d6-afd9-bb0e47fdd2a8" />

#### Definition of Done

- [ ] first
- [ ] second
- [ ] third

#### Steps to get to that page

- Log in as `bc-cas-dev`
- Click `View annual reports` under the Reporting tile
- Click Start for `Banana LFO` (this has 20 facilities), or you can choose a smaller one first `Bees LFO` with 1 facility
- Fill out the report until you get to the `Final Review page`
- Scroll down to the Report Information section 
- All facilities listed will View Details -> Click the link to open the facility report itself
- In the facility report page, there's a `Save as PDF` button

zoeyli-46:
@patriciarussellCAS created a ticket for renaming facility reports PDF file downloads

joshgamache:
If there isn't anything specific to NextJS/RJSF or otherwise, the below might achieve what we need:

```javascript
window.addEventListener("beforeprint", (event) => {
  document.title='My desired save filename';
});

window.addEventListener("afterprint", (event) => {
  document.title='The original page title';
});
```

shon-button:
> If there isn't anything specific to NextJS/RJSF or otherwise, the below might achieve what we need:
> 
> window.addEventListener("beforeprint", (event) => {
>   document.title='My desired save filename';
> });
> 
> window.addEventListener("afterprint", (event) => {
>   document.title='The original page title';
> });

Perhaps using this approach would best scope the document.title mutation to the print lifecycle rather than the full component lifecycle.  Otherwise the browser tab title is updated for the entire time the page is mounted, even when the user is not printing/downloading the report.

[...]

PR #4678, opened 2026-05-15
Rename facility report download file name

Sepehr-Sobhani:
[ISSUE 4587](https://github.com/bcgov/cas-registration/issues/4587)

### Approach: setting `document.title` on mount

The browser derives the PDF filename from `document.title` when printing. The cleanest approach is to set the page title to the correct filename while the user is on the facility report page.

`FacilityReportFinalReviewContent` now uses a `useEffect` to set `document.title` on mount and restore the original title on unmount. This ensures the correct filename is used for both the print button and `Cmd/Ctrl + P`, without adding any special handling inside `DownloadPdfButton`.

As a result, `DownloadPdfButton` remains a simple `window.print()` call with no awareness of filenames.

### Refactor: server/client component split

`FacilityReportFinalReview` was a client component that fetched data via `useEffect` + `useState`, showing a loading skeleton while waiting. It has been split into:

- **`FacilityReportFinalReview`** (server component) — `async` function that awaits the data fetch directly. Loading is handled by the `Suspense` boundary already provided by `defaultPageFactory`.
- **`FacilityReportFinalReviewContent`** (client component) — receives `data` and `backUrl` as props and owns all interactivity: the task list, PDF button, `document.title` effect, and back navigation.

### DIFF
--- a/bciers/apps/reporting/src/app/components/finalReview/FacilityReportFinalReviewContent.tsx
+++ b/bciers/apps/reporting/src/app/components/finalReview/FacilityReportFinalReviewContent.tsx
@@ -32,12 +32,12 @@ export default function FacilityReportFinalReviewContent({
       document.title = originalTitle;
     };
 
-    window.addEventListener("beforeprint", handleBeforePrint);
-    window.addEventListener("afterprint", handleAfterPrint);
+    globalThis.addEventListener("beforeprint", handleBeforePrint);
+    globalThis.addEventListener("afterprint", handleAfterPrint);
 
     return () => {
-      window.removeEventListener("beforeprint", handleBeforePrint);
-      window.removeEventListener("afterprint", handleAfterPrint);
+      globalThis.removeEventListener("beforeprint", handleBeforePrint);
+      globalThis.removeEventListener("afterprint", handleAfterPrint);
     };
   }, [data.facility_name]);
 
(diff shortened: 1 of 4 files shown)


joshgamache (approved):
Simple way to handle the filename. The refactor is straightforward too, and tests cover the new code. Great 🎊!

Low level flags from SonarQube on maintainability, but not blocking.

bciers/apps/reporting/src/app/components/finalReview/FacilityReportFinalReviewContent.tsx
```diff
@@ -0,0 +1,68 @@
...
+}
+
+export default function FacilityReportFinalReviewContent({
+  data,
+  backUrl,
+}: Props) {
+  const router = useRouter();
+
+  // Controls the PDF filename for both the Save as PDF button and Cmd/Ctrl+P.
+  useEffect(() => {
```
shon-button, line 24: Perhaps using the alternative approach suggested by @joshgamache — leveraging the browser `beforeprint` / `afterprint` events — would better scope the document.title mutation to the print lifecycle rather than the full component lifecycle.
Otherwise the browser tab title is updated for the entire time the page is mounted, even when the user is not printing/downloading the report.

Image

bciers/apps/reporting/src/app/components/finalReview/FacilityReportFinalReviewContent.tsx
```diff
@@ -0,0 +1,79 @@
...
+      document.title = originalTitle;
+    };
+
+    globalThis.addEventListener("beforeprint", handleBeforePrint);
+    globalThis.addEventListener("afterprint", handleAfterPrint);
+
+    return () => {
+      globalThis.removeEventListener("beforeprint", handleBeforePrint);
+      globalThis.removeEventListener("afterprint", handleAfterPrint);
+    };
```
joshgamache, line 41: Great quick win to narrow the scope!

joshgamache (approved):

PR #4659, opened 2026-05-13
4599 add naics code to report operation table

Sepehr-Sobhani:
[ISSUE 4599](https://github.com/bcgov/cas-registration/issues/4599)

PR #4681, opened 2026-05-19
chore: add ignore trigger to migration

acatchpole:
Related PR: https://github.com/bcgov/cas-registration/pull/4659

A fix that adds pgtrigger.ignore when populating the NAICS code to avoid immutable report version errors


pbastia (approved):
🎉

Issue #4599, opened 2026-05-05
Reporting/Add naics_code to report_operation table


dleard:
#### Describe the task

The report doesn't have a discrete record of the naics_code of the operation at the time the report was created. This _could_ cause issues later if an operation's naics_code changes down the line.

Follow-up ticket is #4679 

Task: Add a naics_code fkey to the report_operation table & run a migration to populate the current records.

**Moved from 4679: task related to compliance summary**
`naics_code_id` has been added in the `report_operation` table under #4599  where users can now see the NAICS Code in the Review Operation Information page.

One of the scenarios for 4599 is to initially use NAICS Code 322112 (Chemical pulp mill) with lime recovered by kiln as part of the products.

If NAICS Code == 322112, RF for lime recovered by kiln should be 0.90
If NAICS Code != 322112, RF for lime recovered by kiln should be 0.65

In the scenario, user updated to 322111 but did not sync yet in Review Operation Information page, but it looks like the RF has already updated to 0.65 in the Compliance Summary page. [See this part of the test document](https://buttoninc-my.sharepoint.com/:w:/g/personal/francesca_button_is/IQDA4uZ09-jnS63dWKmCMC84AUnzTq91D33XRhuWkJy2akM?e=8Llg4X&nav=eyJoIjoiMTIzNzA5MTU4MyJ9)

#### Developer Checklist

- [x] Add a `naics_code` fkey to the `report_operation` table
- [x] Run a migration to update the existing `report_operation` records with the proper naics codes
- [x] Display the `naics_code` on the Operation Information page as a read-only field
- [x] The `naics_code` is synced from registration when a sync request is triggered
- [ ] Update the compliance summary calculations to use the report_operation naics value when resolving regulatory values
- [ ] Update get_naics_code_by_version_id to return report_operation.naics_code
- [ ] Notify IRC via the BCIERS internal channel that this field is now available to them for querying

Sepehr-Sobhani:
I assume we also need to add this new field to the Final Review page and the Submitted page. Also, we’ll probably need to include the operation NAICS code when the report gets created — just confirming here.

fviduya:
I have tested this in the DEV environment, [here is the test document for reference](https://buttoninc-my.sharepoint.com/:w:/g/personal/francesca_button_is/IQDA4uZ09-jnS63dWKmCMC84AUnzTq91D33XRhuWkJy2akM?e=nmdjv2).

I can verify that:
[...]

PR #4690, opened 2026-05-21
4599 use report naics in compliance

acatchpole:
Card: https://github.com/bcgov/cas-registration/issues/4599

### Changes
1. Updated services to get NAICS code from report_operation:
   - industrial_process
   - regulatory_values
   - form_reponse_builder
2. Updated tests with this new relationship
3. Deleted NaicsCodeService as it wasn't being used anywhere


bc_obps/reporting/api_v2/forms/form_response_builder.py
```diff
@@ -72,14 +71,10 @@ def operation_data(self) -> Self:
         """
         Builds general operation data into the response
         """
-        report_operation = ReportOperation.objects.select_related("report_version__report__operation__naics_code").get(
-            report_version_id=self.report_version_id
-        )
-
-        naics_code = NaicsCodeService.get_naics_code_by_version_id(self.report_version_id)
+        report_operation = ReportOperation.objects.get(report_version_id=self.report_version_id)
```
acatchpole: should add select_related back to this query

bc_obps/reporting/service/compliance_service/industrial_process.py
```diff
@@ -79,10 +80,11 @@ def compute_industrial_process_emissions(rp: ReportProduct) -> Decimal:
     # Handle Pulp & Paper specific edge case:
     # Subtract the sum of emissions that were categorized as industrial_process & (woody_biomass or other_excluded_biomass) from
     # the industrial_process emission total attributed to the product "Pulp and paper: chemical pulp".
-
+    ro = ReportOperation.objects.get(report_version_id=rp.report_version_id)
+    naics_code = ro.naics_code.naics_code if ro.naics_code else None
     if (
-        rp.report_version.report.operation.naics_code
-        and rp.report_version.report.operation.naics_code.naics_code.startswith("322112")
+        naics_code is not None
```
dleard: Is there a scenario where `report_operation.naics_code` is None? Should we be guarding against this scenario at all? I see that in the model, `report_operation.naics_code` is nullable, but I'm thinking it probably shouldn't be.

acatchpole: the reason i added the guard is because it is nullable. it is also nullable in the admin operation model

dleard: I did some investigating and we do indeed have scenarios where that can be null. Resolving this comment.

bc_obps/reporting/service/compliance_service/industrial_process.py
```diff
@@ -79,10 +80,11 @@ def compute_industrial_process_emissions(rp: ReportProduct) -> Decimal:
     # Handle Pulp & Paper specific edge case:
     # Subtract the sum of emissions that were categorized as industrial_process & (woody_biomass or other_excluded_biomass) from
     # the industrial_process emission total attributed to the product "Pulp and paper: chemical pulp".
-
+    ro = ReportOperation.objects.get(report_version_id=rp.report_version_id)
+    naics_code = ro.naics_code.naics_code if ro.naics_code else None
```
dleard: I think here, we should throw an error if there is no NAICS code. This function should not be hit without it.

bc_obps/reporting/api_v2/forms/form_response_builder.py
```diff
@@ -72,14 +71,12 @@ def operation_data(self) -> Self:
...
-        report_operation = ReportOperation.objects.select_related("report_version__report__operation__naics_code").get(
+        report_operation = ReportOperation.objects.select_related("naics_code").get(
             report_version_id=self.report_version_id
         )
 
-        naics_code = NaicsCodeService.get_naics_code_by_version_id(self.report_version_id)
-
         operation_data = OperationData(
-            naics_code=naics_code,
+            naics_code=report_operation.naics_code.naics_code if report_operation.naics_code else None,
```
dleard, line 79: Here we can leave the guard against `None` as is. This response builder will be reached by EIOs who do not have NAICS codes

dleard (approved):
Looks good to me. I did a cursory search through the codebase for any other hanging places where we need to fetch naics_code from report_operation and didn't find any.
There's just one place where I think we should change the guard against `None` to a `ValueError`. That industrial_process function should never be hit without a NAICS

Issue #4464, opened 2026-04-10
Tech Debt: Straighten out the error handling in the reporting app


Sepehr-Sobhani:
keywords: refactor errors reporting v2 api

Follow-up to #4463 

#### Description of the Tech Debt

> Pierre: Right now every page and API endpoint has its own custom way of feeding errors back, and we should really transition to having common patterns and common code handling that instead 

#### Tech Debt Triage

The purpose of our technical debt triage process is to analyze technical debt to determine risk level of the technical debt and the value in tackling that technical debt.

#### Risk Value Scoring:

| Level  | Value                 |
| ------ | --------------------- |
| High   | <div align="center">3 |
| Medium | <div align="center">2 |
| Low    | <div align="center">1 |

| Technical Debt - Risk Types                                                                                                   | Level | Value |
| ----------------------------------------------------------------------------------------------------------------------------- | ----- | ----- |
| Business Area Risk - Risk of business area visibility / damage to user experience                                             | 0     | 0     |
| Developer Fault Risk - How likely will this tech debt cause a future error related to coding on top of it                     | 0     | 0     |
| System Fault Risk - Risk of system errors or application downtime                                                             | 0     | 0     |
| Time Scale Risk - Compound risk effect if left alone. How much more difficult to fix or dangerous will this become over time? | 0     | 0     |
| Time Sink Risk - How much will this tech debt slow the development process down                                               | 0     | 0     |
| <div align="right">**TOTAL SCORE:**                                                                                           | 0     | 0     |

#### Development Checklist:

- [ ] Checklist item
- [ ] Checklist item
- [ ] Checklist item

patrickisaac:
Refinement Notes:
- @pbastia to add additional notes

shon-button:
@fviduya 

For testing, I suspect any test document you have where a "Save" error is displayed on the front-end could be part of your QA testing...on my side, I mocked raised back-end errors to validate front-end display: https://github.com/bcgov/cas-registration/pull/4720

fviduya:
I have tested this in the DEV environment, [here is the test document for reference](https://buttoninc-my.sharepoint.com/:w:/g/personal/francesca_button_is/IQDftn_CLpwbT7rUvz_owmHvAVAR43Ii8Pq_CqBZGdLIU7o?e=tkLS8F).

I can verify that:
- [x] Expected error messages are still being displayed in reporting. Tested on the following pages:
   - [x] Review Operation Information 
   - [x] Person Responsible 
[...]

PR #4692, opened 2026-05-22
4464 - Straighten out the error handling in the reporting app

shon-button:
Ticket: [4464](https://github.com/bcgov/cas-registration/issues/4464)

### 🚀 Impact

- Refactor service-level user-facing errors to flow through the same shared validation UI as report validation errors for endpoints:
   - `bc_obps/reporting/api/submit.py`
   - `bc_obps/reporting/api/report_activity.py`

---

### 🚧 Changes:

**Backend**

* Refactored `UserError` to support structured user-facing errors with `severity`, `message`, `key`, and optional `context`
* Added `serialize_user_error(error)` so service errors can use the same frontend shape as report validation errors
* Added `build_user_error_response(version_id, error)` to wrap `UserError` in the standard `FormResponseBuilder(...).payload({ errors })` response
* Added `generic_error_codes_4xx` to separate generic API errors from structured validation responses
* Updated endpoints to catch/return `UserError` as structured `ReportingFormSchema[ReportValidationPayloadSchema]` validation responses

**Frontend**
* Updated frontend `actionHandler` error parsing to recognize `response.payload.errors` as `validation`
* Updated forms to pass structured validation errors into `<ReportValidationSummary errors={validationErrors} />`
* **Note:** Moved sign-off acknowledgement validation from frontend-only submit-button management into backend validation logic, so the backend returns validation error for "All required acknowledgements must be checked to submit the report."

---

### 🧰 To Do:

* Update/add tests
* Refactor service-level user-facing errors to flow through the same shared validation UI as report validation errors for endpoints:
   - TBD

---

#### Start Test Environment
1. Start the API server:  
   ```sh
   cd bc_obps
   make prepare_backend
   ```  
   ```sh
   cd bciers && yarn dev-all
   ```  

---

#### Test: Sign-off checkbox validation:

1. Log in using `bc-cas-dev`
2. Navigate to reporting
3. Navigate to a report's `sign-off` route
4. Enter signature
5. Click `Submit`
**Expected Results:**
- Validation error displays for missing checkbox values
image


bc_obps/common/exceptions.py
```diff
@@ -1,7 +1,45 @@
...
+        message: Human-readable explanation for the user
+        key: Stable frontend identifier used for rendering/handling
+        context: Optional metadata related to the error
     """
 
-    pass
+    severity: Severity
+    message: str
+    key: UserErrorKey
+    context: dict[str, Any] | None
```
pbastia: providing default values here will make them optional in the constructor generated, so we can keep the dataclass' default behaviour

```suggestion
    severity: Severity = Severity.ERROR
    message: str
    key: UserErrorKey = UserErrorKey.GENERIC_ERROR
    context: dict[str, Any] | None = None
```

bc_obps/reporting/api/report_activity.py
```diff
@@ -1,20 +1,28 @@
...
 
 
 @router.post(
     "report-version/{version_id}/facilities/{facility_id}/activity/{activity_id}/report-activity",
-    response={200: dict, custom_codes_4xx: Message},
+    response={
+        200: dict,
+        422: ReportingFormSchema[ReportValidationPayloadSchema],
+        generic_error_codes_4xx: Message,
+    },
```
pbastia, line 24: We could probably build a generic type for this (not 100% if ninja would like this):
```
response=ReportingResponse[dict]
```

bc_obps/reporting/api_v2/utils/build_user_error_response.py
```diff
@@ -0,0 +1,16 @@
...
+def build_user_error_response(
+    version_id: int,
+    error: UserError,
+) -> dict[str, Any]:
+    error.context = {
+        "report_version_id": version_id,
+        **(error.context or {}),
+    }
+
+    return FormResponseBuilder(version_id).operation_data().payload({"errors": [serialize_user_error(error)]}).build()
```
pbastia: I think that instead of putting this in an another method, we could bake this into the ResponseBuilder itself

the API would then be directly accessible from the api methods:
```py
FormResponseBuilder(version_id).whatever().we().want_to_build().errors(error1, error2).build()
```

bc_obps/reporting/api_v2/utils/validation_error_serializer.py
```diff
@@ -0,0 +1,37 @@
...
+        "key": getattr(error.key, "value", error.key),
+        "error": _build_error_data(error),
+    }
+
+
+def serialize_user_error(error: UserError) -> dict[str, Any]:
+    return {
+        "key": getattr(error.key, "value", error.key),
+        "error": _build_error_data(error),
+    }
```
pbastia: OOP would put these methods directly on the class so we can call

```
error.serialize()
```

We could put this in a mixin to avoid the code reuse

```py
class SerializableError:
    def serialize(self):
        return {
            "key": ...
            # ...
         }

def UserError(SerializableError):
    # current implementation

def ReportValiadationError(SerializableError):
    # same
```

bc_obps/reporting/api_v2/utils/build_user_error_response.py
pbastia: the errors field could be side by side with the payload, so there we can keep typing clean and concerns separate
the response would be:
{
  payload: ...
  errors: ...
}

PR #4695, opened 2026-05-25
4542 - undo pr 4616 and filter false positives

### LINKED ISSUE #4542: Reporting/Bug: Review Changes shows difference in Allocation of Emissions when there are no changes made in emissions

gdalcengio:
resolves: [4542](https://github.com/bcgov/cas-registration/issues/4542)
undoes: [4616](https://github.com/bcgov/cas-registration/pull/4616)

changes:

- undo 4616
- removed `report_product_id` from noisy diff keys
- new solution to filter false positive changes based on matching the report product name

### DIFF
--- a/bc_obps/reporting/service/review_changes_service/report_review_changes_service.py
+++ b/bc_obps/reporting/service/review_changes_service/report_review_changes_service.py
@@ -8,7 +8,6 @@ from .diff_helpers import _DIFF_KWARGS, _CHANGE_TYPE_MAP, detect_renames, diff_s
 
 
 NOISY_DIFF_KEYS = {
-    "report_product_id",
     "report_version",
     "is_supplementary_report",
     "is_latest_submitted",
--- a/bc_obps/reporting/service/review_changes_service/report_review_changes_service.py
+++ b/bc_obps/reporting/service/review_changes_service/report_review_changes_service.py
@@ -46,9 +46,31 @@ def _get_field_display_title(field_path: str, field_display_titles: Dict[str, st
     return field_display_titles.get(field_name)
 
 
+def _normalize_products(products: List[Any]) -> List[tuple[Any, Any]] | None:
+    """
+    Extracts core business identifiers (product name and allocated quantity) from a list
+    of raw product records and returns a predictably sorted, comparable structure.
+
+    Returns:
+        - A sorted list of (product_name, allocated_quantity) tuples.
+        - None if any individual product item is not a dictionary structure.
+    """
+    normalized_products = []
+    for product in products:
+        if not isinstance(product, dict):
+            return None
+        normalized_products.append(
+            (
+                product.get("product_name"),
+                product.get("allocated_quantity"),
+            )
+        )
+    return sorted(normalized_products, key=lambda x: str(x[0]))
+
+
 def _is_false_positive_product_change(change: Dict[str, Any]) -> bool:
     """
-    Determine if a chsnge item is a false positive that should be ignored.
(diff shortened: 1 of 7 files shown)


bciers/apps/reporting/src/app/components/changeReview/templates/EmissionAllocationChangeView.tsx
```diff
@@ -340,6 +340,30 @@ export const EmissionAllocationChangeView: React.FC<
     return result;
   }
 
+  // Helper to determine if change is a false positive due to supplementary report report_product_id
```
shon-button: @gdalcengio 

Do you think it would be possible to move filtering logic to the backend service?

gdalcengio: I think it may be possible, I initially tried that which led to the other errors introduced: e.g. allocations of emissions diffs not showing up or showing up under the wrong name. But, I think these were because I tried to filter by erasing the product_id which appears to be necessary on the front end. Maybe right before we pass it through the api as a list of changes we can filter it in the `get_report_version_diff_changes` function?

gdalcengio: Moving the filter to the backend has been pushed now

bc_obps/reporting/service/review_changes_service/report_review_changes_service.py
```diff
@@ -47,6 +46,41 @@ def _get_field_display_title(field_path: str, field_display_titles: Dict[str, st
...
+
+    old_products = old_value.get('products', [])
+    new_products = new_value.get('products', [])
+
+    # Check for equal amount of products
+    if len(old_products) != len(new_products):
+        return False
+
+    # Check for matching product names and quantities
+    for i in range(len(new_products)):
```
shon-button: @gdalcengio 

Suggestion: Sort lists before comparing them to ensure consistent matching regardless of array index order.

See example refactor here: https://github.com/bcgov/cas-registration/pull/4707/changes#diff-2cb31bf0922ba606493047f968f25cc19db5f5a14904919f74f18bf73ccf8d27


See also related tests: https://github.com/bcgov/cas-registration/pull/4707/changes#diff-e72d15573b769fadb92fa8cadb29ac46e5c5c6897abcfb13925d54acdd54c781

gdalcengio: Suggested refactor and tests have been added, thank you!

shon-button (approved):
⭐⭐

PR #4708, opened 2026-05-27
chore: transfers e2e

fviduya:
Card: https://github.com/bcgov/cas-registration/issues/3447

- Role: CAS Analyst
   - [x] `Make a Transfer` button is visible in the Transfers grid
   - [x] transfer an operation > effective date in the future > status = to be transferred
   - [x] transfer an operation > effective date in the past > status = transferred
   - [x] edit a pending transfer 
   - [x] cancel a pending transfer
- Role: CAS Director
   - [x] `Make a Transfer` button is visible in the Transfers grid
   - [x] transfer an operation > effective date in the future > status = to be transferred
   - [x] transfer an operation > effective date in the past > status = transferred
   ~- [ ] edit a pending transfer - skipped for now~
   ~- [ ] cancel a pending transfer - skipped for now~

Notes:
- Created the transfer-facility and transfer-operation helpers to reduce code duplication and keep it cleaner. Ideally, what CAS Analyst can do, CAS Director can also do.
~- I couldn't find the ticket related to the CAS Director not being able to edit or cancel a transfers request, but in the test for CAS Director those two are currently being skipped. Tried to use `test.fail()` but it looks like that still runs the test and then expects it to fail, I thought to use `test.skip()` to save time as we already know that edit/cancel isn't built yet.~ - removed tests related to this based on review
~- atm, I couldn't seem to address the code duplication issue noted by sonarqube~


bc_obps/registration/fixtures/mock/transfer_event.json
```diff
@@ -7,7 +7,7 @@
       "operation": "b65a3fbc-c81a-49c0-a43a-67bd3a0b488e",
       "from_operator": "4242ea9d-b917-4129-93c2-db00b7451051",
       "to_operator": "4a792f0f-cf9d-48c8-9a95-f504c5f84b12",
-      "effective_date": "2025-01-01T09:00:00Z",
+      "effective_date": "2099-12-31T09:00:00Z",
```
fviduya, line 10: Updated this to 2099 so we always have a record for `To be transferred` status

bc_obps/common/management/commands/load_fixtures.py
```diff
@@ -56,3 +59,13 @@ def handle(self, *args, **options):
...
+
+        if 'transfer_event.json' in ' '.join(fixtures):
+            self._fix_transfer_event_submission_timestamps()
+
+    def _fix_transfer_event_submission_timestamps(self):
+        from common.lib import pgtrigger
+        from registration.models.event.transfer_event import TransferEvent
+
+        with pgtrigger.ignore("registration.TransferEvent:set_updated_audit_columns"):
+            TransferEvent.objects.update(created_at=FIXED_TRANSFER_SUBMISSION_TIMESTAMP)
```
fviduya, line 71: The submission date field in the grid is derived from the `created_at` field. Although the `created_at` value is set in the fixtures, the TimeStampedModel overwrites the value to `now()`. This would ideally make the submission date fixed so that happo doesn't always detect a change.

Sepehr-Sobhani (commented):
This is great, thank you, Chesca!

I have two suggestions:

* We have a way to use the same test file with different user types for these scenarios, which could simplify this PR quite a bit. Here's an example: https://github.com/bcgov/cas-registration/blob/bc507324107597dd89f7375d0695374e359e58ab/bciers/apps/reporting-e2e/tests/workflows/SFO/view-report.spec.ts#L9-L10
* I suggest avoiding adding tests that are marked with `.skip`. Since this feature hasn't been implemented yet and we don't know when it will be, I'd prefer not to add tests for functionality that doesn't exist yet, as it can unnecessarily clutter the codebase.

bc_obps/common/management/commands/load_fixtures.py
```diff
@@ -56,3 +59,13 @@ def handle(self, *args, **options):
         for fixture in fixtures:
             self.stdout.write(self.style.SUCCESS(f"Loading: {fixture}"))
             call_command('loaddata', fixture)
+
+        if 'transfer_event.json' in ' '.join(fixtures):
```
Sepehr-Sobhani: nit: We could do the loop instead of joining all fixture names
```py
if any('transfer_event.json' in f for f in fixtures):
...
```

Sepehr-Sobhani (approved):
🤩

PR #4769, opened 2026-06-04
chore: move deprecated enumNames from schema to uiSchema (rjsf6 pattern)

dleard:
This rjsf schema was missed when handling the deprecated "enumNames" property in the rjsf upgrade PR. Facilities in the transfers menu dropdown were showing as UUIDs rather than by their names.

Fixed:
image


Sepehr-Sobhani (approved):
🥇

PR #4777, opened 2026-06-09
chore: registration file upload

pbastia:
#### Backend
- Refactored the registration endpoints containing data-url files to accept regular, "multipart/form-data" data containing both a JSON payload and file entries.
- Updated existing ninja schemas and service calls, and deleted data-url transformation steps
- GET endpoints now serve a serialized JSON structure to display the file name and the 

#### Frontend
- Updated schemas to only contain only `string` type for the file schema properties. For RJSF, any file upload will look like a string
- Updated FileWidget, simplified code, and expose a hook to keep track of uploaded files until submission
- FileWidget will set the value to a string containing a JSON-serialized render of the file name and scan status

See [updated documentation](https://github.com/bcgov/cas-registration/blob/chore/2123-refactor-registration-file-uploads/docs/file_uploads.md)

Work left to do:
- File preview for uploaded files. #4931
- Update `Document` model to store the original file name, so that we can display it without the django name disambiguation string. #4932


bciers/libs/components/src/form/widgets/FileWidget.tsx
```diff
@@ -101,153 +43,136 @@ const showScanStatus = (status: FileScanStatus | undefined | null) => {
...
+      const anchorTag = document.createElement("a");
+      Object.assign(anchorTag, {
+        target: "_blank",
+        rel: "noopener noreferrer",
+        href: URL.createObjectURL(localFile),
+        download: localFile.name,
+      });
+      anchorTag.click();
+      anchorTag.remove();
+      return;
```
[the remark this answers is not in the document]
pbastia: oh dang, good catch

bciers/libs/components/src/form/widgets/FileWidget.tsx
```diff
@@ -101,153 +43,136 @@ const showScanStatus = (status: FileScanStatus | undefined | null) => {
...
-  // 🥷 Prevent resetting the value to null when user switch tabs
-  useEffect(() => {
-    if (localValue && !value) {
-      onChange(localValue);
-    }
-  }, [localValue, onChange, value]);
+  const [localFile, setLocalFile] = useState<File | undefined>(undefined);
+  const parsedValue: FileInfo | undefined = value
+    ? JSON.parse(value)
+    : undefined;
```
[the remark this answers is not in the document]
pbastia: Yes this is by design and would be the sign of tampering

bciers/libs/components/src/form/widgets/readOnly/ReadOnlyFileWidget.tsx
```diff
@@ -1,37 +1,15 @@
...
-
-  useEffect(() => {
-    // Fix for window not being defined on page load
-    if (fileValue.length) {
-      setFileInfo(fileValue);
-    }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []);
+const ReadOnlyFileWidget: React.FC<WidgetProps> = ({ id, options, value }) => {
+  const parsedValue = value ? JSON.parse(value) : undefined;
```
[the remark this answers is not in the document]
pbastia: This is by design. The tests shouldn't pass old data urls

bciers/libs/components/src/form/widgets/FileWidget.tsx
```diff
@@ -101,153 +43,136 @@ const showScanStatus = (status: FileScanStatus | undefined | null) => {
...
-  const { translateString } = registry;
-  const { dataURL, name } = fileInfo;
-  if (!dataURL) {
-    return null;
-  }
+const useFileUploadWidget = (): [
+  typeof formContext,
+  typeof submitWithFiles,
+] => {
+  const files: Record<string, File> = {};
```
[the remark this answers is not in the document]
pbastia: good point

bc_obps/registration/api/_operations/_operation_id/_registration/new_entrant_application.py
```diff
@@ -32,16 +31,18 @@ def get_operation_new_entrant_application(request: HttpRequest, operation_id: UU
     return 200, OperationService.get_if_authorized(get_current_user_guid(request), operation_id, ['id', 'operator_id'])
 
 
-@router.put(
+@router.post(
```
Sepehr-Sobhani, line 34: I'm pretty sure you've seen [this](https://django-ninja.dev/guides/input/file-params/#:~:text=(user)-,Handling%20request.FILES%20in%20PUT/PATCH%20Requests,-Problem) but decided to change the verb instead of adding that middleware.

bc_obps/registration/schema/operation.py
```diff
@@ -1,18 +1,27 @@
...
+
+def serialize_document(doc: Document | None) -> Optional[str]:
+    # Something similar to dataURL to allow passing metadata along with the filename.
+    # Required because RJSF expects a string for a file field.
+    if doc:
+        name = doc.file.name.split('/')[-1]
+        return json.dumps({"name": name, "id": doc.id, "status": doc.status})
+    return None
+
+
```
Sepehr-Sobhani, line 24: This is a great refactor 🤩 We could move this to `bc_obps/registration/utils.py` to make it reusable.

bc_obps/service/data_types/operation_service.py
Sepehr-Sobhani, line 1: nit: We can omit `_data_types` from the file name since it is already under the `data_types` folder.

bc_obps/registration/schema/operation.py
```diff
@@ -77,41 +80,17 @@ class Meta:
         fields = ['id']
 
 
-class OperationInformationIn(ModelSchema):
+class OperationInformationIn(Schema):
```
Sepehr-Sobhani: Just out of curiosity, why did we switch to a simple Schema?

bc_obps/common/tests/utils/test_files.py
Sepehr-Sobhani, line 1: This is a great utility. I suggest moving it to the common app so we can reuse it in all similar instances in `test_report_attachment_service.py`.

bc_obps/registration/tests/endpoints/_operations/_operation_id/_registration/test_new_entrant_application.py
```diff
@@ -64,18 +62,17 @@ def test_put_register_operation_new_entrant_application_submission_endpoint_user
...
         )
         TestUtils.authorize_current_user_as_operator_user(self, operator_baker())
-        response = TestUtils.mock_put_with_auth_role(
-            self,
-            "industry_user",
-            self.content_type,
-            {
-                "operation_id": operation.id,
-                "new_entrant_application": MOCK_DATA_URL,
+        response = TestUtils.client.post(
```
Sepehr-Sobhani, line 65: Didn't use `mock_post_with_auth_role` because of the `format` argument? You could update `mock_post_with_auth_role` to accept `*args` and `**kwargs`, which would let you pass additional arguments without breaking the existing functionality.

bc_obps/registration/api/operations.py
```diff
@@ -75,6 +75,18 @@ def get_registration_purposes(request: HttpRequest) -> Tuple[Literal[200], List[
...
+    request: HttpRequest,
+    payload: OperationInformationIn,
+    boundary_map: File[UploadedFile] = None,  # type: ignore
+    process_flow_diagram: File[UploadedFile] = None,  # type: ignore
+    new_entrant_application: File[UploadedFile] = None,  # type: ignore
 ) -> Tuple[Literal[201], Operation]:
-    return 201, OperationService.register_operation_information(get_current_user_guid(request), None, payload)
+
+    operation_data = OperationData(
+        **payload.model_dump(),
```
Sepehr-Sobhani: Saving an operation with multiple operators throws `TypeError: asdict() should be called on dataclass instances` here. `payload.model_dump()` recursively converts `multiple_operators_array` into a list of **dicts**, and since `OperationData` is a plain dataclass it stores them as-is — so `upsert_multiple_operators` ends up calling `asdict()` on a dict.
This is going to affect `update_operation` and `register_edit_operation_information` endpoints too.

Sepehr-Sobhani (changes requested):
Amazing work 🤩

Sepehr-Sobhani (commented):
Not directly related to this PR, but do we want to add some backend validation to enforce allowed file extensions? I noticed `accept: ".pdf"` in the new entrant schema, but that is only a frontend UI hint and users can bypass it.

bc_obps/registration/schema/operation.py
pbastia: It's because the service called by the API method now exposes a data class for its input data as an intermediary. Generating this schema from the model would keep a tight coupling between database columns and form schema, and we want to try to avoid that

bc_obps/common/tests/utils/test_files.py
pbastia: Wrote https://github.com/bcgov/cas-registration/issues/4959

bc_obps/registration/schema/operation.py
pbastia: I'd rather leave it here until we need it somewhere else and can find a place that makes sense for it, it will make it easier to discover than in a giant bag of utils

bc_obps/registration/tests/endpoints/_operations/_operation_id/_registration/test_new_entrant_application.py
pbastia: I'm going to skip refactoring the TestUtils methods, because of the order of the arguments, this would mean bigger update than I'd like, and it's not saving any lines of code from this syntax. We could make it a tech debt if we want

Sepehr-Sobhani (approved):
🥇

PR #4785, opened 2026-06-10
4693 Handle EC record for supplementary reports where previous credited emissions are between 0 and 1

Sepehr-Sobhani:
[ISSUE 4693](https://github.com/bcgov/cas-registration/issues/4693)

## Summary

- **Bug:** When a previous report had `0 < credited_emissions < 1`, submitting a supplementary report that crosses the `>= 1` threshold raised a "no handler found" exception. `IncreasedCreditHandler` requires `previous.credited_emissions >= 1`, and `NewEarnedCreditsHandler` required `previous.credited_emissions == 0` — leaving the sub-one range unhandled.
- **Fix:** Relaxed `NewEarnedCreditsHandler.can_handle()` from `== ZERO_DECIMAL` to `< ONE_DECIMAL`. Credits below the 1 threshold are never issued, so both `0` and `0 < x < 1` represent the same state: no prior `ComplianceEarnedCredit` record exists. The fix correctly routes these cases to `NewEarnedCreditsHandler`.


dleard (approved):
Totally makes sense! :shipit:

PR #4786, opened 2026-06-11
4754: Aluminum/Alumina Production activity bug fix

gdalcengio:
Resolves: 4754

- Dynamic schemas with allOf/anyOf/oneOf + if/then dependancies are not recommended to have unevaluatedProperties = False. https://rjsf-team.github.io/react-jsonschema-form/docs/advanced-customization/internals/#json-schema-supporting-status
- AllOf evaluates on each object branch individually, leading to errors as one branch does not contain the properties of the other. This PR makes sure the validator does not add unevaluatedProperties if the activity is aluminum_production
- also cleans up an extra test found in the same related test file

### DIFF
--- a/bc_obps/reporting/service/report_validation/validators/report_activity_json_validation.py
+++ b/bc_obps/reporting/service/report_validation/validators/report_activity_json_validation.py
@@ -19,7 +19,7 @@ logger = logging.getLogger(__name__)
 TAGS = [ValidationTags.REPORT_VALIDATION, ValidationTags.ON_SUBMIT]
 
 
-def enable_jsonschema_draft_2020_validation(schema: Any, validateUnevaluatedProperties: bool = True) -> None:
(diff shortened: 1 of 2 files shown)


bc_obps/reporting/service/report_validation/validators/report_activity_json_validation.py
```diff
@@ -18,7 +19,7 @@
 TAGS = [ValidationTags.REPORT_VALIDATION, ValidationTags.ON_SUBMIT]
 
 
-def enable_jsonschema_draft_2020_validation(schema: Any) -> None:
+def enable_jsonschema_draft_2020_validation(schema: Any, validateUnevaluatedProperties: bool = True) -> None:
```
dleard: The sonarcloud warning here is valid. Let's change this variable to snake_case to align with Python variable naming convention.

dleard (approved):
Looks good to me! Let's just fix that variable name & then this is good to go.

Issue #4714, opened 2026-05-27
Cluster Resource Tuning 2026


joshgamache:
#### Description of the Tech Debt

Check application/pod/resource usage history and tune our requests/limits and HPAs to be consistent with our actual needs.

PR #4790, opened 2026-06-11
chore: tune pod resources

joshgamache:
Addresses #4714 . Tunes prod resources based on our peak around the deadline.


dleard (approved):
Thanks for taking a look at this!

PR #4792, opened 2026-06-11
1459 Columns in the common tables have meaningful metadata

Sepehr-Sobhani:
[ISSUE 1459](https://github.com/bcgov/cas-registration/issues/1459)


bc_obps/common/migrations/0128_alter_emailnotification_message_id_and_more.py
```diff
@@ -0,0 +1,31 @@
+# Generated by Django 5.1.15 on 2026-06-11 22:38
```
andrea-williams: nit: could rename this file to be more specific, e.g., `_update_emailnotification_metadata`

andrea-williams (approved):

Issue #4795, opened 2026-06-12
Reporting/Bug: Emission subheading missing in some activity forms


TammieV-BCG:
#### Describe the Bug:
Emission subheading missing in activity forms for the following activities;
•	Electricity generation
•	General stationary combustion excluding line tracing (at SFO)
•	Natural gas transmission (compression and processing)
•	Natural gas transmission (other than compression and processing)
•	Oil & Gas extraction (compression and processing)
•	Oil & Gas extraction (other than compression and processing)

 
#### Steps to reproduce the behaviour:
1. log in as external user
2. start a report for any operation
3. select the following activities:
Electricity generation
General stationary combustion excluding line tracing (at SFO)
Natural gas transmission (compression and processing)
Natural gas transmission (other than compression and processing)
Oil & Gas extraction (compression and processing)
Oil & Gas extraction (other than compression and processing)
4. fill out report till you get to activity forms
5.  open each activity form 
6. Observe that Emission subheading is missing on each of the activity forms

#### Screenshots:
Expected result
<img width="513" height="245" alt="Image" src="https://github.com/user-attachments/assets/ef9e6b0e-0fb5-4577-b579-6c7529c8ea32" />

Actual result
<img width="502" height="300" alt="Image" src="https://github.com/user-attachments/assets/a1c26619-ce51-4f72-b9b3-bdc4a743bc0d" />

#### Additional information:

Browser: [e.g. chrome, safari]
Version: [e.g. 22]

#### Additional Context

-

fviduya:
I have tested this in the DEV environment, [here is the test document for reference](https://buttoninc-my.sharepoint.com/:w:/g/personal/francesca_button_is/IQCFg5_dhDiNTYp97--LD1eWAWc84GCFfYSDxxAfAVQiDic?e=PjwiU0).

I can verify that:
- [x] The `Emission` subheading is visible in the activity forms where emissions can be entered. 
- [x] When industry user clicks `Add Emission`, a new Emission section is displayed where user can enter data
- [x] Clicking the garbage icon next to the Emission subheading deletes the whole emission section

Moving to Pending PO for review of test. Thank you!

patriciarussellCAS:
Thanks! Moving to accepted. @TammieV-BCG for UAT.

TammieV-BCG:
[...]

PR #4811, opened 2026-06-16
Ensure 'title' is present in emissions uischemas

acatchpole:
Card: https://github.com/bcgov/cas-registration/issues/4795

### Changes
Several Activity UISchemas, as well as the shared `emissionsFieldsUiSchema`, did not have the "title" defined for the emissions blocks. These have been updated.

I also explored utilizing the `emissionsFieldsUiSchema` more, as the emissions sections all tend to share most of the what is defined there. But I found that with the fact that most schemas do have _something_ different, the necessitated spreading and adding and overriding of values made the schemas less readable.


shon-button (approved):
✨

Issue #4758, opened 2026-06-03
Compliance/Ignore unpaid FAA interest when deciding to generate a penalty invoice


dleard:
#### Describe the task
Currently the system considers the Compliance obligation as "met" when both the compliance obligation and FAA have been paid. An automatic penalty is triggered when the obligation is considered "met". So if a user only pays the compliance obligation and does not pay the FAA, the system currently will not consider the obligation met and will not generate the automatically penalty invoice. 

As per the decision note from IRC, unpaid FAA interest should not block BCIERS from considering an obligation "met" & generating a penalty invoice.

The part of the app where this condition is check is in the Obligation Paid handler in
`bc_obps/compliance/service/automated_process/compliance_handlers.py`
This checks that `invoice.outstanding_balance == 0`. The outstanding_balance includes any unpaid interest. This should be changed to check that `invoice.fee_amount_balance == 0` to only consider the balance of the obligation.

~Question: Should we add a new status & handler for `Obligation Met - Interest Owing` or something to be clear that while the **obligation** from tCO2e has been met, the invoice has not been fully paid?~ #4787 

#### Acceptance Criteria

- [x] Penalties are generated by the Obligation Paid handler regardless of FAA interest still owing

#### Developer Checklist

- [x] bc_obps/compliance/service/automated_process/compliance_handlers.py` has been changed to consider `invoice.fee_amount_balance == 0` when deciding if an obligation is met
- [x] Any additional tests are added to ensure that penalties are generated as per the requirements above

Note: #4759 should be done shortly after this ticket - still needs more investigation on the CSNR side and needs BA approval 
- note may be difficult to test given that FAA comes from e-licensing

patriciarussellCAS:
Refinement note: @dleard will split out this question into a new ticket: 

- [ ] Question: Should we add a new status & handler for Obligation Met - Interest Owing or something to be clear that while the obligation from tCO2e has been met, the invoice has not been fully paid?

Otherwise I think this is good to remove the refinement tag.

fviduya:
I have tested this in the DEV environment, [here is the test document for reference](https://buttoninc-my.sharepoint.com/:w:/g/personal/francesca_button_is/IQAYUXE5ZyyMT7NTpFAD40YXAZAH8nKv759JX8B7Si-7veU?e=ysTYxl).

I can verify that:
[...]

PR #4818, opened 2026-06-17
4758 Ignore unpaid interest when deciding if an obligation is met

Sepehr-Sobhani:
[ISSUE 4758](https://github.com/bcgov/cas-registration/issues/4758)

### DIFF
--- a/bc_obps/compliance/service/automated_process/compliance_handlers.py
+++ b/bc_obps/compliance/service/automated_process/compliance_handlers.py
@@ -112,10 +112,13 @@ class ObligationPaidHandler(ComplianceUpdateHandler):
         has_penalty = hasattr(invoice, 'compliance_penalty') and getattr(invoice, 'compliance_penalty', None)
         if has_penalty:
             return False
+        # Only consider the fee balance (the tCO2e obligation) when deciding if the obligation is met.
+        # Unpaid FAA interest (part of outstanding_balance) must not block the obligation from being
+        # considered met or block automatic penalty generation.
         return (
             invoice.compliance_obligation.compliance_report_version.status
             == ComplianceReportVersion.ComplianceStatus.OBLIGATION_NOT_MET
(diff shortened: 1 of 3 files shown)


bc_obps/compliance/service/automated_process/compliance_handlers.py
```diff
@@ -112,10 +112,13 @@ def can_handle(self, invoice: ElicensingInvoice) -> bool:
...
         if has_penalty:
             return False
+        # Only consider the fee balance (the tCO2e obligation) when deciding if the obligation is met.
+        # Unpaid FAA interest (part of outstanding_balance) must not block the obligation from being
+        # considered met or block automatic penalty generation.
         return (
             invoice.compliance_obligation.compliance_report_version.status
             == ComplianceReportVersion.ComplianceStatus.OBLIGATION_NOT_MET
-            and invoice.outstanding_balance == Decimal('0.00')
+            and (invoice.invoice_fee_balance or Decimal('0.00')) == Decimal('0.00')
```
dleard: Is the extra `or` condition necessary here? If for whatever reason `invoice.fee_amount_balance` comes back as `None`, I don't think we want this part of the conditional to default to truthy and continue execution.

Sepehr-Sobhani: @dleard I wasn't entirely sure either. Since this field is nullable, I thought we might want to handle the case where invoice.fee_amount_balance is None. I'm not sure what value we expect once the obligation is paid, does it become Decimal('0'), or is it set to null? That was the reason I included the extra condition.

dleard: It will always have a decimal value (0.00 when paid), so in that case I'd remove this `or Decimal('0.00')` condition

dleard (approved):
Looks good to me! :shipit:

PR #4820, opened 2026-06-17
Less warnings in backend python tests

Sepehr-Sobhani:
## Summary

Running the backend test suite (`make pythontests_parallel`) emitted ~167 warnings. This PR removes most of them (22 remains which requires thorough testing) by fixing the underlying causes at the source rather than suppressing them. The warnings fell into three root causes:

1. A deprecated Django `CheckConstraint` API (`RemovedInDjango60Warning`)
2. Naive (timezone-unaware) datetimes assigned to `DateTimeField`s in tests (`RuntimeWarning`)
3. A `model_bakery` `seq()` misuse (`UserWarning`)

No production behavior changes are intended.

### 1. Django `CheckConstraint.check` → `condition` deprecation

`RemovedInDjango60Warning: CheckConstraint.check is deprecated in favor of '.condition'.`

Verified with `manage.py makemigrations --check` that this produces **no new migration** (the constraint definition is unchanged; only the kwarg name differs).

### 2. Naive datetimes in test fixtures

`RuntimeWarning: DateTimeField <Model>.<field> received a naive datetime (...) while time zone support is active.`

Many test fixtures assigned naive date/datetime values (date-only strings, `datetime(...)` objects, or `.date()` calls) to `DateTimeField`s. Made these timezone-aware while preserving the exact value:

- Date/datetime **strings** → appended a UTC marker (e.g. `"2023-05-01"` → `"2023-05-01T00:00:00Z"`, `"2024-01-01 01:46:20.789146"` → `"...+00:00"`).
- **`datetime(...)` objects** → made aware via `tzinfo=timezone.utc` or `timezone.make_aware(...)`.
- **`.date()` / `timezone.now().date()`** assignments to `DateTimeField`s → replaced with aware datetimes.

Only lines targeting `DateTimeField`s were changed; assignments to `DateField`s (e.g. `CompliancePeriod.start_date`) were intentionally left alone since they don't trigger the warning.

#### Related assertion update — `reporting/tests/endpoints/test_report_new_entrant_endpoint.py`

Making the new-entrant date fixtures (`authorization_date`, `first_shipment_date`, `new_entrant_period_start`) timezone-aware changed the serialized API response from the naive form (`'2022-01-01T00:00:00'`) to the realistic aware form that production actually returns under `USE_TZ` (`'2022-01-01T00:00:00Z'`). Updated the expected values in the two affected `get_new_entrant_data` tests to match. (Verified the exact serialized output by running the response schema through Pydantic's JSON encoder.)

### 3. `model_bakery` `seq()` misuse — `reporting/tests/utils/baker_recipes.py`

`UserWarning: start parameter is ignored when using seq with date, time or datetime objects`

The `regulated_product` recipe passed both a positional start value **and** a `start=` kwarg to `seq()` with `date` objects. `model_bakery` ignores `start=` for date/time/datetime sequences (the positional value already seeds the sequence), so the kwarg was dead and emitted a warning. Removed the redundant `start=` kwarg — behavior is unchanged.

## Out of scope

A separate group of Pydantic serialization warnings (`PydanticSerializationUnexpectedValue` for `business_structure` / `app_role`) is **not** addressed here. Those stem from input schemas whose field validators intentionally coerce a name string into a Django model instance, and fixing them cleanly is a larger schema-design change best handled on its own.

### DIFF
--- a/bc_obps/compliance/tests/api/test_permissions.py
+++ b/bc_obps/compliance/tests/api/test_permissions.py
@@ -65,21 +65,21 @@ class TestVersionOwnershipFromUrl:
             "registration.tests.utils.operation_designated_operator_timeline",
             operation=operation,
             operator=old_user_operator.operator,
-            start_date="2023-05-01",
-            end_date="2023-12-31",
+            start_date="2023-05-01T00:00:00Z",
+            end_date="2023-12-31T00:00:00Z",
         )
         make_recipe(
             "registration.tests.utils.operation_designated_operator_timeline",
             operation=operation,
             operator=new_user_operator.operator,
-            start_date="2024-01-01",
+            start_date="2024-01-01T00:00:00Z",
             end_date=None,
         )
 
         compliance_report_version = make_recipe(
             "compliance.tests.utils.compliance_report_version",
             report__operator=new_user_operator.operator,
-            compliance_report__created_at='2024-06-15',
+            compliance_report__created_at='2024-06-15T00:00:00Z',
             compliance_report__report__operation=operation,
         )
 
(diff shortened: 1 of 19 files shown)


dleard (approved):
Thanks for this! So much cleaner 👍

Issue #3253, opened 2025-04-21
Facility form front-end validation doesn't work [3]


BCerki:
#### Describe the Bug:

https://github.com/user-attachments/assets/09a1785a-5daa-4df4-8d67-d74d2c007d1a
When you try to submit a facility form without a name, the form allows you to submit with no validation errors even though required fields are missing. The backend error then prevents you from proceeding

#### Probability: 2

(How likely the bug is to happen, scored from 1-5. For example, probability of 5 is something like "it happens to all users every time they log in." whereas probability of 1 "only happens to certain users when a really specific and unlikely path is followed.")

#### Impact: 2

(How bad the bug is when it does happen, scored from 1-5. For example, effect of 5 is "the entire app crashes and makes it unusable for all users" or "the bug causes the wrong data to be saved, with critical information (e.g. payment) being affected." whereas effect of 1 is "It makes some styling look a little bit weird.")

<img src="https://www.getzephyr.com/sites/default/files/riksk%20matrix.jpg" alt="Bug Matrix" width="250" height="250"/>

#### Steps to reproduce the behaviour:

1. Step #1
2. Step #2
3. Step #3

#### Additional information:

Browser: [e.g. chrome, safari]
Version: [e.g. 22]

#### Additional Context

-

patriciarussellCAS:
Confirmed this is still an issue.

fviduya:
I have tested this in the DEV environment, [here is the test document for reference](https://buttoninc-my.sharepoint.com/:w:/g/personal/francesca_button_is/IQCbXDQ9Ph-oQ6r7PJZDpfgrAR7ZQ6KTSQY4BvJN5ktAqfM?e=L305pE).

I can verify that:
- [x] When industry user is adding a facility, and user leaves any of the required fields blank and attempts to save the data, then an inline error message appears denoting that the field is required
- [x] When industry user is adding a facility, and user leaves any of the required fields blank and attempts to save the data, a generic error message is visible on top of the `Save` button that reads: `This form can't be saved yet. Please fix the errors above.`
- [x] When industry user is adding a facility, and user fills out all required information and saves the data, no error is encountered and a new facility is created

Moving to Pending PO for review of test. Thank you!

patriciarussellCAS:
Thanks! Moving to accepted. @TammieV-BCG for UAT.

TammieV-BCG:
[...]

PR #4720, opened 2026-05-29
4464 refactor error handling

shon-button:
Ticket: [4464](https://github.com/bcgov/cas-registration/issues/4464)

### 🚀 Impact

Standardized user-facing **4xx endpoint responses** into a consistent error payload while preserving backward compatibility through the existing top-level `message` field.

The new response format includes both:

```json
{
  "message": "...",
  "errors": [
    {
      "key": "generic_error",
      "error": {
        "severity": "Error",
        "message": "..."
      }
    }
  ]
}
```

This allows the frontend to continue consuming `message` for existing flows while leveraging the structured `errors` array to drive the shared validation UI and enable error-specific handling through stable error keys.

---

### 🚧 Changes:

**Backend**

* In `bc_obps/service/error_service/handle_exception.py` updated mapped exceptions (`UserError`, `ValidationError`, `PermissionError`, etc.) to return a standardized error payload with `message` and `errors[]` instead of a simple `{ "message": ... }` response, enabling generic service errors to render through the shared validation UI.
* Preserved the existing custom payload format for `ReportValidationException`

**Frontend**
* Updated `actionHandler` to distinguish between recoverable form validation errors and unrecoverable application errors:
   * `GET`, `DELETE` requests, from page components, and unrecoverable errors are thrown to be handled by the Next.js error boundary
   * `POST`, `PUT`, and `PATCH` requests from form components return structured validation errors for display through the shared validation UI, 
* Updated forms to pass structured validation errors to `<ReportValidationSummary />`

---

### 🧰 Possible To Do:
- Update all app forms to pass structured validation errors to `<ReportValidationSummary />`
- Refactor proxies which redirect to onboarding on API failures, i.e. `bciers/apps/compliance/src/proxies/withRuleHasComplianceRouteAccessBCeID.ts`

---

#### Start Test Environment
1. Start the API server:  
   ```sh
   cd bc_obps
   make prepare_backend
   ```  
2. Start the app development server:  
   ```sh
   cd bciers && yarn dev-all
   ```  

---
#### Test: report validation errors
1. Log in using `bc-cas-dev`
2. Navigate to `report-validation`
**Expected Results:**
- Route does not go to the Next.js error boundary
- Validation errors render normally
- pecific validation keys and contexts should be preserved
- Response should use the validation wrapper
image

#### Test: generic errors - POST, PUT, PATCH endpoint 
1. Log in using `bc-cas-dev`
2. Temporarily force a backend endpoint used by a form to raise an error, e.g.
[...]

PR #4824, opened 2026-06-18
3253 fix inline validation on empty facility form

Sepehr-Sobhani:
[ISSUE 3253](https://github.com/bcgov/cas-registration/issues/3253)

### Problem
Submitting the LFO facility form empty bypassed client-side validation — users saw a backend error instead of inline field errors.

### Root cause
`SingleStepTaskListForm` passed `{}` to RJSF when `rawFormData` was empty. The top-level schema has no `required` at the root, so RJSF passed validation silently. Nesting as `{ section1: {} }` makes RJSF validate the section's required fields inline.

### Changes

**`SingleStepTaskListForm.tsx`** — always call `createNestedFormData` (even for empty data)

**`SingleStepTaskListForm.test.tsx`** — fix "snackbar on create" test to use sections with `required: []` since empty submit now correctly triggers validation

**`facilities/types.ts`** — add `FacilityFormData` (from `FacilityOut` backend schema) and `Operation` interfaces

**`FacilityForm.tsx`** — replace `any` types (`schema`, `uiSchema`, `onSubmit` param, error state); remove dead `?? {}`; extract `isCasDirector`/`canEdit` variables; merge duplicate enum imports; remove redundant comments

**`FacilityPage.tsx`** — cast `getOperation` to `Operation`; use `"error" in facilityFormData` for proper union narrowing; move error check before `isCreating` derivation

### DIFF
--- a/bciers/apps/administration/app/components/facilities/FacilityForm.tsx
+++ b/bciers/apps/administration/app/components/facilities/FacilityForm.tsx
@@ -2,20 +2,22 @@
 
 import { useState } from "react";
 import { useParams, useRouter, useSearchParams } from "next/navigation";
+import { IChangeEvent } from "@rjsf/core";
+import { RJSFSchema, UiSchema } from "@rjsf/utils";
 import SingleStepTaskListForm from "@bciers/components/form/SingleStepTaskListForm";
 import { actionHandler } from "@bciers/actions";
 import serializeSearchParams from "@bciers/utils/src/serializeSearchParams";
-import { FacilityTypes, FrontEndRoles } from "@bciers/utils/src/enums";
-import { FormMode } from "@bciers/utils/src/enums";
+import {
+  FacilityTypes,
+  FrontEndRoles,
+  FormMode,
+} from "@bciers/utils/src/enums";
 import { useSessionRole } from "@bciers/utils/src/sessionUtils";
-
-export interface FacilityFormData {
-  [key: string]: any;
-}
+import { FacilityFormData } from "./types";
 
 interface Props {
-  schema: any;
-  uiSchema: any;
+  schema: RJSFSchema;
+  uiSchema: UiSchema;
   formData: FacilityFormData;
   isCreating?: boolean;
 }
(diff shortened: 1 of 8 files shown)


bciers/apps/administration/app/components/facilities/FacilityPage.tsx
```diff
@@ -8,42 +8,42 @@ import {
...
 }) {
-  let facilityFormData: { [key: string]: any } | { error: string } = {};
+  let facilityFormData: FacilityFormData | { error: string } = {};
 
-  const operation = await getOperation(operationId);
-  if (operation.error) {
+  const operation = (await getOperation(operationId)) as
+    | Operation
+    | { error: string };
+  if ("error" in operation) {
```
shon-button: @Sepehr-Sobhani 

I believe since [PR 597 – Update error boundary message](https://github.com/bcgov/cas-registration/pull/2973), error messages thrown from components are masked by our error boundary; so, custom thrown messages are never displayed to the user.

Additionally, following [PR 4464 – Refactor error handling](https://github.com/bcgov/cas-registration/pull/4720), GET request failures are now routed to the error boundary. 

As a result, throwing a custom error here does not provide a different user experience than the underlying application error; as such, ` if (operation.error) ` could be removed and users will still see the same error if ` await getOperation(operationId)` fails:

Suggestion: remove the if error\throw
```
 if (operation.error) {
      throw new Error(
        "We couldn't find your operation information. Please ensure you have been approved for access to this operation.",
      );
    }
```

Image

bciers/apps/administration/app/components/facilities/FacilityForm.tsx
```diff
@@ -26,16 +28,17 @@ export default function FacilityForm({
   uiSchema,
   isCreating,
 }: Readonly<Props>) {
-  // To get the user's role from the session
   const role = useSessionRole();
-  const [error, setError] = useState(undefined);
-  const [formState, setFormState] = useState(formData ?? {});
+  const [error, setError] = useState<string | undefined>(undefined);
```
shon-button, line 32: @Sepehr-Sobhani 

A thought for your consideration, perhaps this would be an opportunity to start implementing [standardized form validation error  across the apps](https://github.com/bcgov/cas-registration/issues/4827)

shon-button (approved):

bciers/apps/administration/app/components/facilities/FacilityPage.tsx
Sepehr-Sobhani: Thanks for sharing this. I was a bit hesitant about whether to remove that line or not. I actually removed it at first, but then I noticed a test that depended on that error message and thought maybe I should put it back. 😄

bciers/apps/administration/app/components/facilities/FacilityForm.tsx
Sepehr-Sobhani: That's a brilliant idea. All forms should follow the same pattern for displaying errors. We just need to make sure those errors are surfaced consistently across all HTTP methods.

Issue #4749, opened 2026-06-03
Reporting/Filter valid regulated products when creating a report


andrea-williams:
#### Describe the task

~We'll need to check that in a report, the only products selected are those that are valid for the report's reporting year.~
Ticket #4748 makes it so we CAN'T select invalid products by filtering them out from all product selectors. A path for invalid products to get into reports still exists. We need to prevent invalid products being transferred from Admin on report creation (mostly a concern when creating reports in past years).

#### Dev Checklist

- [ ] ~update `reporting/service/report_product_service`: `get_allowed_products()` should assert that the only products allowed in a report are those whose `valid_from` and `valid_to` date range encapsulated the report's reporting year~
- [ ] update `/service/report_version_service` to filter operation.regulated_products for validity before being added to report_operation.regulated_products
- [ ] write unit tests to assert products are being filtered correctly

#### Definition of Done

- [ ] first
- [ ] second
- [ ] third

#### Additional context

- Add any other context about the task here.
- Or here

fviduya:
I have tested this in the DEV environment, [here is the test document for reference](https://buttoninc-my.sharepoint.com/:w:/g/personal/francesca_button_is/IQB5qqcz6dbYQZQST77cATiZAbqT3hFLl-eKbeCM9lPMdkU?e=ZOyhoj).

I can verify that, given that an operation has two products: `Pulp and paper: lime recovered by kiln` (valid from 2025-01-01) and `Pulp and paper: chemical pulp` (valid from 2023-01-01):
- [x] When user starts a report for RY 2025, then both products `Pulp and paper: lime recovered by kiln` and `Pulp and paper: chemical pulp` are selected
- [x] When user starts a report for RY 2025, then the product `Pulp and paper: lime recovered by kiln` can be selected under the Regulated Products in the Review Operation Information page
- [x] When user starts a report for RY 2024, then only the product `Pulp and paper: chemical pulp` is selected 
- [ ] When user starts a report for RY 2024, then the product `Pulp and paper: lime recovered by kiln` CANNOT be selected under the Regulated Products in the Review Operation Information page - **NOTE: User can select the product**
- [x] User can successfully submit a report for RY 2025
- [x] User can successfully submit a report for RY 2024

[...]

PR #4831, opened 2026-06-19
4749 filter valid products during report creation

acatchpole:
Card: https://github.com/bcgov/cas-registration/issues/4749

This work blocks regulated products being copied from Administration data to a report that is outside of their valid timeframes.
### Changes
 - `reporting_year_service` : added `get_date_for_reporting_year`, a method that returns a `date` of May 1 of the given reporting year. This would be used to compare a reporting_year against date fields (like valid_from and valid_to). 
 - 'operation_service` : added `get_valid_operation_regulated_products` that returns the operation.regulated_products and filters the products for validity based on a given reporting year.
 - `report_version_service` :  modified `create_report_version` so that it utilizes `get_valid_operation_regulated_products` to populate the _report_operation's_ regulated products
 
 #### Things I'm not sure about
`get_date_for_reporting_year` is only appropriate if valid date ranges continue to use 01-01-YEAR and 31-12-YEAR. As we configure things to be effective for a reporting year, this seems like a reasonable assumption, but isn't the cleanest. A better solution might be to refactor our valid_from and valid_to fields from `dates` into `ints` for whole years or maybe even `reporting_year` FKs.

Unit tests: I have included two tests, but I'm not sure it makes sense for both to exist as they are pretty much identical. One directly tests `get_valid_operation_regulated_products`. The other is indirectly testing the same function, but as it is run in `create_report_version`.
 
 ### Testing
Testing requires creating a past report, which is not currently possible in the UI.

1. In Administration, edit the 'Bees LFO' operation and add 2 new regulated prodcuts: `Pulp and paper: lime recovered by kiln` and one other product (dealer's choice).
2. Access the backend terminal (or from `/cas-registration/bc_obps`)
3. run 
```bash 
poetry run python manage.py shell
```
4. then
```python
from service.report_service import ReportService
ReportService.create_report('6d07d02a-1ad2-46ed-ad56-2f84313e98bf', 2024)
```
   This creates a 2024 report for the Bees LFO operation

5. Navigate to `http://localhost:3000/reporting/reports/previous-years`
6. Find the row for the 2024 Report for 'Bees LFO' and click `Continue`
7. On the `Review Operation Information` page, scroll down to the `Regulated products` field
   EXPECT TO SEE these products selected: 
    - `Compression, positive displacement - consumed energy` (preexisting product)
    - The second product you selected in step 1.

   EXPECT TO NOT SEE this product selected:
    - `Pulp and paper: lime recovered by kiln` (as it is not a valid product for 2024)


bc_obps/service/tests/test_operation_service.py
```diff
@@ -0,0 +1,34 @@
...
+        # Create an operation with both valid and invalid products
+        operation = baker.make_recipe(
+            "registration.tests.utils.operation",
+            regulated_products=valid_products + invalid_products,
+        )
+        # Call the service method to get valid regulated products for the operation
+        valid_regulated_products = OperationService.get_valid_operation_regulated_products(operation, report_year)
+
+        # Assert that all valid products are included and all invalid products are excluded from the report
+        self.assertCountEqual(valid_regulated_products, valid_products)
```
andrea-williams: IMO this test isn't comprehensive enough. Since there's also 2 invalid_products being created, get_valid_operation_regulated_products() might be returning the invalid_products and this test would still pass.

andrea-williams: should add assertion that the IDs of valid_regulated_products == the IDs of valid_products

bc_obps/service/operation_service.py
```diff
@@ -749,3 +751,15 @@ def handle_change_of_registration_purpose(
...
+    @classmethod
+    def get_valid_operation_regulated_products(
+        cls, operation: Operation, reporting_year: int
+    ) -> QuerySet[RegulatedProduct]:
+        """
+        Get the operation's regulated products that are valid for a specific reporting year.
+        """
+        reporting_year_date = ReportingYearService.get_date_for_reporting_year(reporting_year)
+        return operation.regulated_products.filter(
+            valid_from__lte=reporting_year_date, valid_to__gte=reporting_year_date
```
andrea-williams, line 762:
```suggestion
        return operation.regulated_products.filter(
            valid_from__lte=date(reporting_year, 5, 1), valid_to__gte=date(reporting_year, 12, 31)
```

bc_obps/service/operation_service.py
```diff
@@ -749,3 +751,15 @@ def handle_change_of_registration_purpose(
...
         return payload
+
+    @classmethod
+    def get_valid_operation_regulated_products(
+        cls, operation: Operation, reporting_year: int
+    ) -> QuerySet[RegulatedProduct]:
+        """
+        Get the operation's regulated products that are valid for a specific reporting year.
+        """
+        reporting_year_date = ReportingYearService.get_date_for_reporting_year(reporting_year)
```
andrea-williams: I'd recommend deleting the `get_date_for_reporting_year` function - I don't think it's necessary. See suggestion below.

andrea-williams (changes requested):
Ran locally - confirmed that behaviour works as expected.

Some small changes requested to remove unnecessary code and to improve testing - also needs a rebase against develop.

bc_obps/service/tests/test_operation_service.py
acatchpole: yeah. not sure how i managed that. test corrected to actually test for identity of products

bc_obps/service/operation_service.py
acatchpole: I thought that some standard function for a date in the middle of the reporting year would be good for comparing to all these valid_to/from fields as they currently all start Jan 1 and end Dec 31. But it should probably be part of a refactor and not just thrown up in some random PR. removed

andrea-williams: it's definitely a not-ideal scenario that in setting the validity of Regulated Products, we have to specify a month and day too, not just a reporting year, which would be ideal. It might be something to discuss at our next Dev Team Agreements

andrea-williams (approved):

PR #4857, opened 2026-06-23
BCCR load balancer session propagation lag

Sepehr-Sobhani:
## Problem

The BCCR API was returning intermittent 401 Unauthorized errors even when the cached JWT token was valid. The errors were non-deterministic — the same request would succeed on retry.

## Root Cause

**Load balancer session propagation lag** — the BCCR API uses stateful session-based authentication on POST search endpoints. Sessions are not instantly shared across all load-balanced nodes. A valid token could still receive a 401 if the request landed on a node that had not yet received the session.

## Changes

**`bc_carbon_registry_api_client.py`**

- **Persistent `requests.Session`** — added a `requests.Session` instance to the singleton in `__new__`. All data requests now go through `self._session.request(...)` instead of `requests.request(...)`. This reuses the underlying TCP connection across calls, pinning the client to the same backend node and eliminating the session propagation 401s.

- **60-second pre-expiry buffer** — `_ensure_authenticated` now re-authenticates when `timezone.now() >= token_expiry - 60s`, preventing requests from being sent with a token that is about to expire.

- **Retry-on-401 with `_retried` guard** — if `_make_request` receives a 401, it clears the cached token, forces re-authentication, and retries the request once. The `_retried` flag prevents infinite recursion. This handles the case where the persistent connection drops (server timeout, network blip) and the next connection lands on a different node mid-session.

## Test Plan

- [x] Manually trigger the BCCR-dependent endpoint multiple times in a row and confirm no 401 errors surface

### DIFF
--- a/bc_obps/compliance/service/bc_carbon_registry/bc_carbon_registry_api_client.py
+++ b/bc_obps/compliance/service/bc_carbon_registry/bc_carbon_registry_api_client.py
@@ -39,6 +39,7 @@ class BCCarbonRegistryAPIClient:
     api_url: Optional[str]
     client_id: Optional[str]
     client_secret: Optional[str]
+    _session: requests.Session
     token_expiry: Optional[datetime] = None
     COMPLIANCE_ACCOUNT_TYPE_ID = 14
     OPERATOR_OF_REGULATED_OPERATION_TYPE_ID = 11  # Only master accounts with this ID are allowed to request units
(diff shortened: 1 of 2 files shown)


bc_obps/compliance/service/bc_carbon_registry/bc_carbon_registry_api_client.py
```diff
@@ -110,7 +112,10 @@ def _ensure_authenticated(self) -> None:
         """
         token = getattr(self, "token", None)
         token_expiry = getattr(self, "token_expiry", None)
-        if token is None or token_expiry is None or timezone.now() >= token_expiry:
+        # Re-authenticate 60 seconds before expiry to avoid the race where the
+        # client considers the token valid but the server rejects it as expired.
+        buffer = timedelta(seconds=60)
```
pbastia: nit: a buffer is a chunk of memory to store transient data, I suggest naming this "sixty_seconds" or "reauth_advance_delay" or something like that

pbastia (approved):
Just a nit! This is great, thanks Sepehr

PR #4877, opened 2026-06-26
chore: create flattened views for erc data & a new schema to house them

dleard:
Innovation Week work: Create some views to flatten out some of the more commonly queried erc data.
The easiest way to see what these views do is to copy the `select` statement inside the `create or replace view..` and copy it into an sql query in metabase prod.

Creates:
- erc_views (schema)
- latest_submitted_emission_view
- latest_submitted_fuel_view
- latest_submitted_production_allocation_view
- compliance_obligation_view
- compliance_earned_credit_view

Needs:
- Metadata for the views & their columns
- Question: Is the `common` app the correct place for this migration to run? 
  - It creates a dependency on all other apps, but that will be true of any app where these views are created
- Explore the benefits and downsides of generating these views into the database vs just creating sql questions in metabase
  - ie: benefit: visibility for IRC in metabase. A distinct new schema with the views will exist on the schema select page. downside: altering the table columns that the views depend on in any way will require dropping and recreating the views (otherwise you get a postgres error that the column cannot be changed because a view depends on it. I only noticed this because of the recent metadata update work that did some `alter column...` statements on a lot of the erc tables).


dleard:
This was a spike week PR. Closing. Could be used for reference at a later date

PR #4901, opened 2026-07-03
Bump release-it package to 20.2.1

Sepehr-Sobhani:
Resolves: https://github.com/bcgov/cas-registration/security/dependabot/398 https://github.com/bcgov/cas-registration/security/dependabot/387


joshgamache (approved):
👍

Issue #4897, opened 2026-07-02
Compliance/Task: Action name for obligations when FAA interest is unpaid


fviduya:
#### Describe the task

Context:
#4787 will introduce a new Compliance Status `Obligation met - Interest Unpaid` that will be used when obligation has been met but it has incurred FAA interest that the user has not paid yet

Given that the user has an obligation with Compliance Status = Obligation met - Interest Unpaid, and obligation also has incurred a penalty (Automatic Overdue / Late Submission aka GGEAPAR),
When user has paid the penalty, but left FAA interest unpaid,
Then the action name should be `Manage Obligation` 

#### Acceptance Criteria

~- [ ] Given that the user has paid the obligation but not the FAA interest and has not incurred any penalty, then the Compliance Status should read `Obligation met - Interest Unpaid`, Penalty Status should read `NONE`, and Action should be `Manage Obligation`~ - invalid scenario, when FAA starts accruing then it means it's past the deadline
- [ ] Given that the user has paid the obligation but not the FAA interest and has incurred a penalty, then the Compliance Status should read `Obligation met - Interest Unpaid`, Penalty Status should read `NOT PAID`, and Action should read `Manage Obligation`
- [ ] Given that the user has paid the obligation but not the FAA interest and has incurred a penalty that has already been paid, then the Compliance Status should read `Obligation met - Interest Unpaid`, Penalty Status should read `PAID`, and Action should read `Manage Obligation`
- [ ] Given that the user has paid the obligation and its FAA interest and has incurred a penalty that has already been paid as well, then the Compliance Status should read `Obligation met`, Penalty Status should read `PAID`, and Action should read `View Details`

#### Definition of Done

- [ ] first
- [ ] second
- [ ] third

#### Additional context

- Add any other context about the task here.
- Or here

fviduya:
Hi @patriciarussellCAS, I created this ticket specifically for the action name column in the Compliance Summaries grid. Since we now allow the penalties to be displayed and paid even if obligation's FAA Interest is unpaid, I think there also work that needs to be done with making sure that the Action name remains to be `Manage Obligation` if the user hasn't paid all monetary obligations. 

This is what it currently looks like when user has paid the obligation and penalty but has left FAA interest unpaid:

<img width="1159" height="176" alt="Image" src="https://github.com/user-attachments/assets/d361731c-943c-4246-af8e-963a1f43fa2d" />

[...]

Issue #4787, opened 2026-06-11
Compliance/Status for when an obligation is met, but FAA interest has not yet been paid


dleard:
#### Describe the task

After #4758, it will not be clear to users when viewing their obligations on the compliance summary dashboard that there is still something to pay in the case where their base obligation has been paid, but FAA interest has not.

Suggestion is to add a new status along the lines of `Obligation Met - Interest Owing` to be explicit as to the remaining action required when the invoice is in that state.

#### Acceptance Criteria

As an industry user
When my base obligation is paid but there is remaining FAA interest accumulating on the invoice
Then it is clear to me at a glance on the compliance summary dashboard that I still have an amount to pay
- [x] The status is `Obligation Met - Interest Not Paid` when faa interest is unpaid
- [x] The status is `Obligation Met` when faa interest is paid

As an internal use
When a base obligation is paid but there is remaining FAA interest accumulating on the invoice
Then it is clear to me at a glance on the compliance summary dashboard that there is still an unpaid amount that the operation must handle

#### Development Checklist

- [x] Add a new status with an appropriate name that captures this state
- [x] Update the ComplianceHandlers to update to/from this state appropriately (may need a new handler)
- [x] Ensure the new status can be clearly seen on the compliance summaries page grid (does not get cut-off from length)
- [x] Add tests for the new flows in & out of this status

patriciarussellCAS:
@dleard would it make sense to include the work to show when FAA has been paid - or should be a separate ticket that follows this one?

dleard:
@patriciarussellCAS That's covered by the second point in the DC - moving obligations into / and out of this status as appropriate, which likely requires a new handler

patriciarussellCAS:
Confirming that the issue of displaying when FAA has been paid is not included in this ticket and needs a separate ticket.

fviduya:
I have tested this in the DEV environment, [here is the test document for reference](https://buttoninc-my.sharepoint.com/:w:/g/personal/francesca_button_is/IQDVlR4AIJSeQ6rtmYmradQMAeba4YoGvGkZAUZlH5xMi-s?e=W5yYVy).

I can verify that:
- [x] Given that the industry user has paid the obligation but not the FAA interest, when user gets to the Compliance Summaries grid, then the Compliance Status is set to `Obligation met - Interest Not Paid`
[...]

PR #4903, opened 2026-07-03
4787 status for when obligation is met but not faa interest

Sepehr-Sobhani:
[ISSUE 4787](https://github.com/bcgov/cas-registration/issues/4787) && [ISSUE 4897](https://github.com/bcgov/cas-registration/issues/4897)

[Docs helpful for testing](https://buttoninc-my.sharepoint.com/:w:/g/personal/francesca_button_is/IQAYUXE5ZyyMT7NTpFAD40YXAZAH8nKv759JX8B7Si-7veU?e=jefbVg)

## Changes

- New `ComplianceReportVersion.ComplianceStatus.OBLIGATION_MET_INTEREST_NOT_PAID` enum value + migration.
- `ObligationPaidHandler` now branches to the new status (instead of always going to fully-met) when `invoice_interest_balance > 0`; new `InterestPaidHandler` promotes it to fully-met once interest is paid off.
- `DecreasedObligationHandler` made interest-aware so it sets the correct status too.
- Dashboard display label, frontend `ComplianceSummaryStatus` enum, route-guard access list, and `ActionCell` updated so industry users see "Manage Obligation" and CAS users see "View Details" for the new status — even if a related penalty has already been paid, since interest is still owed.
- Widened the "Compliance Status" grid column so the new label isn't truncated.

## Django Admin changes (QA/testing convenience only)

Added admin registrations for `ComplianceReportVersion`, `ComplianceObligation`, `ElicensingInvoice`, and `ElicensingInterestRate` purely to make manual testing easier in DEV/TEST. Two new functionalities:

1. **Refresh invoices manually** — an action on `ElicensingInvoice` that refreshes selected invoice(s) from eLicensing and runs the real compliance handler logic.
Screenshot 2026-07-08 at 9 17 02 AM

2. **Fetch interest rates** — a button on `ElicensingInterestRate` to sync rates from eLicensing on demand, since DEV/TEST reset their DB and lose this table until the nightly cron job runs again.
Screenshot 2026-07-08 at 9 16 52 AM


bc_obps/compliance/service/automated_process/compliance_handlers.py
```diff
@@ -170,6 +182,33 @@ def handle(self, invoice: ElicensingInvoice) -> None:
...
 
+class InterestPaidHandler(ComplianceUpdateHandler):
+    """Handler for transitioning obligations from interest not paid to fully met once FAA interest is paid off."""
+
+    def can_handle(self, invoice: ElicensingInvoice) -> bool:
+        """Check if the obligation is OBLIGATION_MET_INTEREST_NOT_PAID and FAA interest has now been paid."""
+
+        has_penalty = _invoice_has_penalty(invoice)
+        if has_penalty:
+            return False
```
dleard: We should remove this part. An obligation in this state can have unpaid interest _and_ a penalty. The existence of a penalty should not stop this handler from executing.

bc_obps/compliance/service/automated_process/compliance_handlers.py
```diff
@@ -170,6 +182,33 @@ def handle(self, invoice: ElicensingInvoice) -> None:
...
+
+        return (
+            invoice.compliance_obligation.compliance_report_version.status
+            == ComplianceReportVersion.ComplianceStatus.OBLIGATION_MET_INTEREST_NOT_PAID
+            and (invoice.invoice_interest_balance or ZERO_DECIMAL) == ZERO_DECIMAL
```
dleard:
```suggestion
            and (invoice.invoice_interest_balance) == ZERO_DECIMAL
```
I don't think we should guard against invoice.invoice_interest_balance in this way. This value should never be `None` and if it is, something has gone wrong and we shouldn't assume that we should proceed.

bc_obps/compliance/service/automated_process/compliance_handlers.py
```diff
@@ -12,6 +12,13 @@
 
 logger = logging.getLogger(__name__)
 
+ZERO_DECIMAL = Decimal('0.00')
+
+
+def _invoice_has_penalty(invoice: ElicensingInvoice) -> bool:
+    """Check if the invoice is a penalty invoice"""
+    return bool(hasattr(invoice, 'compliance_penalty') and getattr(invoice, 'compliance_penalty', None))
+
```
pbastia: This is a code smell, I'm not sure why there should be a "compliance_penalty" attribute on a type that doesn't expose it. 

It was there before so not part of this PR, but we should probably refactor that, for our future selves

bc_obps/compliance/service/supplementary_version_service/decreased_obligation_handler.py
```diff
@@ -634,11 +634,20 @@ def _void_unpaid_invoices(previous_version_id: int) -> None:
...
+        Set the previous CRV's status to OBLIGATION_FULLY_MET or OBLIGATION_MET_INTEREST_NOT_PAID if FAA
+        interest is still outstanding, once its invoice's fee balance reaches $0.
+
+        Reads the invoice fresh from the DB: by this point `create_adjustment_for_target_version` has
+        already posted the fee adjustment and force-refreshed the invoice's balances from eLicensing.
         """
-        ComplianceReportVersion.objects.filter(id=previous_version_id).update(
-            status=ComplianceReportVersion.ComplianceStatus.OBLIGATION_FULLY_MET
+        invoice = ElicensingInvoice.objects.get(compliance_obligation__compliance_report_version_id=previous_version_id)
+        interest_not_paid = (invoice.invoice_interest_balance or ZERO_DECIMAL) > ZERO_DECIMAL
```
dleard:
```suggestion
        interest_not_paid = (invoice.invoice_interest_balance) > ZERO_DECIMAL
```
Similar thing. It's less of a problem here, but it's unnecessary. Either `None` is not going to evaluate to not greater than 0 or 0 will evaluate to not greater than 0.

bc_obps/compliance/service/automated_process/compliance_handlers.py
```diff
@@ -133,7 +140,12 @@ def handle(self, invoice: ElicensingInvoice) -> None:
 
         obligation = invoice.compliance_obligation
         compliance_report_version = obligation.compliance_report_version
-        compliance_report_version.status = ComplianceReportVersion.ComplianceStatus.OBLIGATION_FULLY_MET
+        interest_not_paid = (invoice.invoice_interest_balance or ZERO_DECIMAL) > ZERO_DECIMAL
```
dleard:
```suggestion
        interest_not_paid = (invoice.invoice_interest_balance) > ZERO_DECIMAL
```
Same thing here

dleard (changes requested):
This status makes me feel better! 😆 
Looks good for the most part, I've left a few comments for your consideration. Let me know if you want to chat about anything!

pbastia (commented):
One little unrelated thing, but looks good otherwise! Very nice!

bc_obps/compliance/service/automated_process/compliance_handlers.py
Sepehr-Sobhani: I'm not fully following what you mean. 🤔 compliance_penalty has a one-to-one relationship with elicensing_invoice, just like the other models that have a one-to-one relationship with elicensing_invoice.

pbastia: Aaah I see, it didn't occur to me to check the reverse relationship. So wouldn't `hasattr(invoice, 'compliance_penalty')` always return True then?

bc_obps/compliance/service/automated_process/compliance_handlers.py
Sepehr-Sobhani: This is not a very accurate name for the function, since it mainly determines whether an invoice is a penalty or obligation invoice, rather than checking whether the invoice has a penalty. I'm going to change it to `_is_penalty_invoice`

bc_obps/compliance/service/automated_process/compliance_handlers.py
Sepehr-Sobhani: Nope, `hasattr` catches Django's `RelatedObjectDoesNotExist`, so it correctly returns `False` when no penalty is linked and `True` only when one exists. Now that I think about it more, the second part (using `getattr`) is actually redundant since we're not using its value.

dleard (approved):
Looks good! Thanks, this distinct status makes the state of the obligation much clearer. when interest is still due

Issue #4636, opened 2026-05-08
Tech Debt/E2E: Add real attachments to the test fixtures


acatchpole:
#### Description of the Tech Debt
While there are currently reporting `attachments` included the fixtures, they exist only as db references, and not actual files. Further, these references aren't even loaded outside the local env. In order for E2E tests to confirm that attachments are accessible, we need actual files to be used as attachments that can be viewed.

Once this is done, the `view-report.spec.ts` e2e tests should be updated to include accessing these attachments.

#### Tech Debt Triage

The purpose of our technical debt triage process is to analyze technical debt to determine risk level of the technical debt and the value in tackling that technical debt.

#### Risk Value Scoring:

| Level  | Value                 |
| ------ | --------------------- |
| High   | <div align="center">3 |
| Medium | <div align="center">2 |
| Low    | <div align="center">1 |

| Technical Debt - Risk Types                                                                                                   | Level | Value |
| ----------------------------------------------------------------------------------------------------------------------------- | ----- | ----- |
| Business Area Risk - Risk of business area visibility / damage to user experience                                             | 0     | 0     |
| Developer Fault Risk - How likely will this tech debt cause a future error related to coding on top of it                     | 0     | 0     |
| System Fault Risk - Risk of system errors or application downtime                                                             | 0     | 0     |
| Time Scale Risk - Compound risk effect if left alone. How much more difficult to fix or dangerous will this become over time? | 0     | 0     |
| Time Sink Risk - How much will this tech debt slow the development process down                                               | 0     | 0     |
| <div align="right">**TOTAL SCORE:**                                                                                           | 0     | 0     |

#### Development Checklist:

App is running in a special docker-compose environment for e2e. We'll need to:
- [ ] Validate that the storage looks for local storage in CI environment (See `storage_backends.py`)
- [ ] Mount a volume setup with the files matching the fixtures to that location (`settings.MEDIA_ROOT`)
- [ ] Check that it works

acatchpole:
The changes for this card only affect local environment. There is nothing to test in DEV/TEST

fviduya:
Moving to Accepted as there is nothing to test in DEV/TEST based on comment above. cc: @patriciarussellCAS

PR #4935, opened 2026-07-14
chore: define url for accessing attachments in local

acatchpole:
Card: https://github.com/bcgov/cas-registration/issues/4636

The card stated to add real files to act as attachments. We actually already have actual files associated with the attachments in the fixtures located in `bc_obps/test_media/report_attachments`. The problem was that the code just didn't support accessing from that location anymore.

The changes included so far correct that bug.

This bug was relevant to confirming behaviour in the E2E tests. Before, when hitting the `download` button for a test attachment, we would get a `resource not found` error. Now, a PDF preview tab is opened where the PDF can be downloaded. As the test files are not actual PDFs, this page displays a `could not load` error.

The point of this ticket was to allow for the full testing of downloading attachments in the E2E tests. Currently, the report-attachments E2E spec tests this functionality by confirming the server call. In order to be able to test downloading an actual file, we would have to make the test files valid PDFs, otherwise the download button on the PDF preview tab doesn't show (at least in the browser I was using)

### DIFF
--- a/bc_obps/bc_obps/storage_backends.py
+++ b/bc_obps/bc_obps/storage_backends.py
@@ -41,6 +41,9 @@ class SimpleLocal(FileSystemStorage):
 
     location = os.path.join(settings.MEDIA_ROOT)
 
+    def url(self, name: str | None = None) -> str:
+        return f"http://localhost:8000{super().url(name)}"
+
     def get_file_bucket(self, name: str) -> str | None:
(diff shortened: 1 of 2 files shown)


andrea-williams (approved):
I don't really understand how all this works, but have confirmed that attachment-specific E2E tests pass locally and in CI.

Issue #4759, opened 2026-06-03
Compliance/Penalty calculation should recalculate if obligation decreases from supp. report


dleard:
#### Describe the task

As per the IRC decision note, An unpaid obligation that is accruing a penalty should consider any decreases in the obligation due to a supplementary report **before** calculating the penalty accruals _and_ ~~the FAA amount~~ owed should also recalculate on the lower amount.

Consider:
An obligation of $10,000 has gone unpaid past the compliance deadline
A supplementary report that decreases that obligation to $8000 is submitted on Feb 1 (~ 2 months after the deadline)
The FAA interest that began accruing on Dec 1, should recalculate to use $8000 as the base amount
The penalty calculation should also use $8000 as the base for the penalty once the obligation is paid

It is unclear if eLicensing does this recalculation for adjustments made in the past.
BCIERS will have to back-date adjustments from supplementary reports to Compliance Deadline - 1 day.

We should also check if the DecreaseObligationHandler properly flags any reports where the amount of credits used is now over the allowed percentage after the decrease.

As per conversation with IRC, this recalculation only applies to supplementary reports that **decrease** an obligation. Current behaviour for **increased** obligations will remain the same to align with the regulation.

#### Acceptance Criteria

- [ ] It is tested and confirmed with CSNR that FAA interest will recalculate if you push a back-dated adjustment [note: currently it is not recalculated as we expected so their team is looking into it]
- [x] BCIERS back-dates adjustments made from supplementary reports if they are submitted past the deadline to before the deadline

#### Developer checklist

- [ ] FAA behaviour tested manually & confirmed with CSNR
- [ ] Adjustments from supp reports submitted after the deadline are back-dated to before the deadline (compliance deadline - 1day)
- [x] Behaviour of DecreasedObligationHandler when the % of credits used is over the allowed amount is confirmed
- [x] Tests are written for the new behaviour

patriciarussellCAS:
Refinement notes: Working with CSNR to confirm how FAA recalculates in the system currently so we know what needs to be done for this ticket. 
This work should get done along side or right after #4758

Asobu01:
Quick follow-up on `bcgov/cas-registration#4759`. I am still ready to start immediately and can post the first repro/checklist note before the draft PR if helpful. If you want me on this one, I can begin now. (@Asobu01)

patriciarussellCAS:
@Asobu01 this project is not looking for outside contributors. While our code is public, the project is not. Thanks.

[...]

PR #4940, opened 2026-07-14
4759 Add back-dating logic for decreased compliance obligations

Sepehr-Sobhani:
[ISSUE 4759](https://github.com/bcgov/cas-registration/issues/4759)

### 1. Dating decrease adjustments to the fee date

- `DecreasedObligationHandler` now dates every decrease adjustment to **the target obligation's own fee date** (`ElicensingLineItem.fee_date`) instead of today.
- This applies the reduction from the start of that fee's accrual window, so eLicensing's FAA interest recalculates on the lower obligation. It works for **both regular and supplementary/late obligations**, and eLicensing accepts it because the adjustment is never dated *before* the fee it adjusts.

- Implemented via a `date_adjustment_to_fee_date` flag threaded through `ComplianceAdjustmentService.create_adjustment` / `create_adjustment_for_target_version`. `create_adjustment` already loads the FEE `elicensing_line_item`, so it reads `fee_date` directly.
- The flag is scoped to **decrease adjustments only**. The "Compliance Units Applied" adjustments (`ApplyComplianceUnitsService`) keep their existing "dated today" behaviour on purpose — their reduction is effective *when applied*, not retroactively, so dating them to `fee_date` would understate penalties when units are applied after the deadline.

### 2. Over-cap compliance units → manual handling

- New model context `ComplianceReportVersionManualHandling.Context.CREDITS_OVER_ALLOWED_PERCENTAGE`.
- New `ApplyComplianceUnitsService.is_credit_usage_over_cap(...)` detects when applied units exceed the post-decrease cap.
- `_calculate_apply_units_cap` / `is_credit_usage_over_cap` gained an optional `pending_supplementary_adjustment`, so the cap can be computed for the **post-decrease** obligation *before* the decrease row is posted.

**When this new context actually applies:** it's only recorded when the obligation was **not fully paid in cash** — i.e. the operator used compliance units up to the cap to cover part of the obligation, and a later supplementary decrease shrinks that cap below what's already applied. In that case the CRV is flagged `REQUIRES_MANUAL_HANDLING` with `CREDITS_OVER_ALLOWED_PERCENTAGE`.

If the obligation was instead **fully paid off in cash** and *then* decreased by a supplementary report, the existing `OBLIGATION_REFUND_POOL_CASH` context is recorded instead— even if that same decrease also pushes applied credits over the (now smaller) cap. This overlap case is rare and is called out below as a known limitation rather than fixed in this PR.

### 3. Refactor: single decision point

- All outcomes are now decided in `_build_adjustment_strategy` and returned on the `AdjustmentStrategy`; `should_record_manual_handling: bool` is replaced by `manual_handling_context: Optional[str]`.
- `_process_adjustment_after_commit` is now a thin executor: post the adjustments, then record manual handling (with the strategy's context) **or** create earned credits.
- Precedence (mutually exclusive): **refund-pool cash** → **earned credits** → **credits over cap**.

## Known limitations / follow-ups

- **No penalty recalculation/display** This PR only dates the adjustment to the fee date for eLicensing's FAA recalc. BCIERS does not recalculate or show users an updated penalty amount — that would be a separate feature.
- **FAA interest is blocked on CSNR** FAA recalculation depends on eLicensing honouring the fee-dated adjustment.
- **Manual-handling overlap** A CRV can only hold one manual-handling context. If the obligation was fully paid in cash and a later decrease both leaves a **refundable cash pool** *and* pushes credits **over the cap**, only the cash context is recorded and the over-cap reason is not surfaced. Rare (rare ∩ rare); the CRV is still flagged for manual handling either way. To be handled later if needed (e.g. capture both reasons or add a log).
- **3× penalty cap** still uses the gross `fee_amount_dollars`, not the decreased amount (`penalty_calculation_service.py`). Left as-is; handle manually if it ever does.


bc_obps/compliance/service/supplementary_version_service/decreased_obligation_handler.py
```diff
@@ -58,6 +60,14 @@ def handle(
         # Charge rate for the applicable reporting year (used for $ conversions)
         charge_rate = ComplianceChargeRateService.get_rate_for_year(new_summary.report_version.report.reporting_year)
 
+        # Back-date the decrease adjustment to the day before the deadline so the penalty recalculates
+        # on the lower obligation. Only when the deadline has passed, otherwise back-dating would stamp
+        # a future date.
+        compliance_deadline = compliance_report.compliance_period.compliance_deadline
+        back_dated_adjustment_date: Optional[str] = None
+        if compliance_deadline < timezone.now().date():
+            back_dated_adjustment_date = (compliance_deadline - timedelta(days=1)).strftime("%Y-%m-%d")
```
dleard: I hadn't considered this when I wrote up the card, but this won't work for supplementary obligations. It'll set an adjustment date _before_ the fee date and the call to elicensing will break because it doesn't like that.

I think we can simplify this to just date all adjustments that decrease an obligation to use the original fee's date (elicensing_line_item.fee_date) & then it'll work for all scenarios. Shouldn't even need to check if it's past the deadline or not in this case.

dleard (changes requested):
One change around what we should use as the adjustment date that I hadn't originally considered. Let me know if you have questions!

bc_obps/compliance/service/supplementary_version_service/decreased_obligation_handler.py
Sepehr-Sobhani: @dleard How about the [`create_adjustment`](https://github.com/bcgov/cas-registration/blob/318d57f454317944767fd261799f74163187e1db/bc_obps/compliance/service/compliance_adjustment_service.py#L78) function? We currently always pass today's date. Should we make it conditional, or should we always pass `elicensing_line_item.fee_date` instead?

dleard: I think it's fine to just always pass `elicensing_line_item.fee_date`.

dleard (approved):
Nicely done! Code looks great & everything worked out as expected after walking through the penalty creation workflow locally 💯

Issue #4906, opened 2026-07-06
Fixtures in TEST environment are the same as DEV


fviduya:
#### Describe the task

Update the fixtures in the TEST environment to make sure that they are the same ones as for dev

1. Remove reports in past reports grid with reporting year < 2024
2. Create fixtures for compliance similar to what we have in DEV
See ticket #4830 

#### Acceptance Criteria

- [ ] Dev and Test have the same fixtures
- [ ] Access is still consistent (bc-cas-dev and bc-cas-test have their respective memberships, and the existing preset access

#### Definition of Done

- [ ] Test has the same fixtures as dev after a fresh deploy
- [ ] Test access is unchanged and accounts still have access to the fixtures by default. Including UAT accounts like @TammieV-BCG 

#### Development Checklist

- [ ] Fixture command changes
- [ ] Exception code for access UUIDs

#### Additional context

- Add any other context about the task here.
- Or here

TammieV-BCG:
Refinement note:

- do we need to have 2 set of fixtures for each environment? TEST and DEV
- Decision: reword ticket to duplicate dev fixtures in test. 
**Note:** keep an eye on access ensuring that default users in each environment are defaulted to the right environment

fviduya:
Moving this to Pending PO, nothing to test for this one in DEV. cc: @patriciarussellCAS

PR #4951, opened 2026-07-15
chore: consolidate dev and test data

fviduya:
Ticket: #4906 

Changes made:
1. Removed `check-test-data`, and updated `check-dev-data` to `check-env-data`
2. Removed unused `load_test_data` and `load_reporting_test_data` from the Makefile
3. Test and dev now loads the same set of fixtures in custom_migrate
4. New files created specifically for identifying users for DEV & TEST: 
   1. `user_dev` and `user_operator_dev` - houses the dev-specific logins. Operator points to Bravo.
   2. `user_test` and `user_operator_test` - houses the test-specific logins. Operator points to Bravo.
5.  Since in TEST we don't load bc-cas-dev's user information, I updated all instances where bc-cas-dev's guid is used (mostly in `created_by` field). The selected user is Blair Balloons which is user operator for Bravo. 
6. The `user_operator` file is retained, these are the user operators that have no logins.
7. Removed test fixtures directories

Need approval as well for: 
- I updated the `check-dev-data` to `check-env-data`. I'll put it back if we want to keep the original name
- I haven't removed the `check-test-data` and `check-dev-data` job from the branch rules. Not sure if that should be done now or when we're ready to merge.


bc_obps/common/management/commands/load_fixtures.py
```diff
@@ -56,6 +57,18 @@ def handle(self, *args, **options):
 
             fixtures = fixtures_base
 
+        environment = settings.ENVIRONMENT
+        env_user = (
+            f'{fixture_base_dir}/user_dev.json' if environment == 'local' else f'{fixture_base_dir}/user_test.json'
```
pbastia: this will load the "user_test.json" in the dev environment, is it what we want?

bc_obps/common/management/commands/load_fixtures.py
```diff
@@ -56,6 +57,18 @@ def handle(self, *args, **options):
...
+        )
+        env_user_operator = (
+            f'{fixture_base_dir}/user_operator_dev.json'
+            if environment == 'local'
+            else f'{fixture_base_dir}/user_operator_test.json'
```
pbastia, line 69: same here, "user_operator_test.json" will be loaded when environment = "dev"

pbastia (commented):
🥇

bc_obps/common/management/commands/load_fixtures.py
fviduya: Thanks, Pierre! No that's not intended. I didn't notice I missed adding 'dev'. Thanks for pointing that out!

pbastia (approved):
🎉

PR #4953, opened 2026-07-15
chore: fix migration policy issue before release to test

gdalcengio:


acatchpole (approved):

dleard (changes requested):
We need to re-enable the policy after the alter field statement is complete

dleard (approved):
My mistake, custom_migrate will re-apply this policy automatically!

Issue #4798, opened 2026-06-15
Terraform Chart reliability improvements


joshgamache:
#### Description of the Tech Debt

When the Terraform Registry was out, it identified a hiccup in our deployment pipeline.

- Env value flagged in ShipIt, which hooks to a Helm value to enable/disable our Terraform pod scripts. 
- Use GCS to query our storage deploys, then conditionally run the terraform apply scripts.
~~- Pull Terraform charts out of BCIERS Helm completely, run terraform scripts manually~~
~~- Pull Terraform charts out of BCIERS Helm main chart and run on their own chart. Either run it manually, or add to the deploy script with failover if they don't work.~~
- ~~Add option to allow job pass when terraform chart fails for certain errors (i.e. `the request failed after 2 attempts, please try again later: 502 Bad Gateway returned from registry.terraform.io` that flagged this issue)~~

#### Tech Debt Triage

The purpose of our technical debt triage process is to analyze technical debt to determine risk level of the technical debt and the value in tackling that technical debt.

#### Risk Value Scoring:

| Level  | Value                 |
| ------ | --------------------- |
| High   | <div align="center">3 |
| Medium | <div align="center">2 |
| Low    | <div align="center">1 |

| Technical Debt - Risk Types                                                                                                   | Level | Value |
| ----------------------------------------------------------------------------------------------------------------------------- | ----- | ----- |
| Business Area Risk - Risk of business area visibility / damage to user experience                                             | 0     | 0     |
| Developer Fault Risk - How likely will this tech debt cause a future error related to coding on top of it                     | 0     | 0     |
| System Fault Risk - Risk of system errors or application downtime                                                             | 0     | 0     |
| Time Scale Risk - Compound risk effect if left alone. How much more difficult to fix or dangerous will this become over time? | 0     | 0     |
| Time Sink Risk - How much will this tech debt slow the development process down                                               | 0     | 0     |
| <div align="right">**TOTAL SCORE:**                                                                                           | 0     | 0     |

#### Development Checklist:

- [ ] Checklist item
- [ ] Checklist item
- [ ] Checklist item

TammieV-BCG:
refinement note:

- need further discussion in dev channel on what option to choose

PR #4976, opened 2026-07-21
Chore/chart terraform revisions

joshgamache:
Addresses #4798. Since our infrastructure deployed via Terraform is rarely changed, this PR makes the chart deployed version of it optional plus adds checks to see if the storage buckets already exist

## Changes 🚧

- Make `terraform-bucket-provision` optional, disabled by default.
  - Use an env variable in ShipIt or manually to enable.
- Add a pre-install,pre-upgrade container to check for the existence of the storage buckets. Runs *after* the terraform hook.

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,3 +1,3 @@
 nodejs 24.18.0
-gcloud 450.0.0
+gcloud 574.0.0
 python 3.14.6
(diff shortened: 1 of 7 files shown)


dleard (changes requested):
There are some valid issues that sonarcloud has raised around specifying request/limit values. I think we can safely add the `automountServiceAccountToken: false` suggestion as well.

Looks good otherwise!

dleard (approved):
:shipit:

PR #4983, opened 2026-07-22
fix: reason for an elicensing adjustment now allows any message

pbastia:


dleard (approved):
Looks good! I did a quick test and a custom adjustment reason now successfully saves on sync request. :shipit:

Issue #4911, opened 2026-07-07
Airflow warning `runtime-variable values in Dag construction` for `cas_bciers_test_migrations.py` dag


joshgamache:
#### Description of the Tech Debt

Airflow's latest update to v3.2.2 is surfacing the following warning on the `cas_bciers_test_migrations` dag:

```
This Dag uses runtime-variable values in Dag construction.
It causes the Dag version to increase as values change on every Dag parse.

Line 47, Col 1
Code: dag(dag_id=TEST_MIGRATIONS_DAG_NAME, default_args=default_args, schedule=None, catchup=False, is_paused_upon_creation=False, doc_md=DAG_DOC, tags=['bciers'])
Issue: Don't use runtime-varying value as argument in Dag constructor

️Don't use the variables as arguments in Dag/Task constructors:
  Line 11: 'TWO_DAYS_AGO' related 'datetime.now()'
  Line 31: 'default_args' related 'datetime.now()'
```

We should update the Dag configuration in order to resolve the warning.

#### Tech Debt Triage

The purpose of our technical debt triage process is to analyze technical debt to determine risk level of the technical debt and the value in tackling that technical debt.

#### Risk Value Scoring:

| Level  | Value                 |
| ------ | --------------------- |
| High   | <div align="center">3 |
| Medium | <div align="center">2 |
| Low    | <div align="center">1 |

| Technical Debt - Risk Types                                                                                                   | Level | Value |
| ----------------------------------------------------------------------------------------------------------------------------- | ----- | ----- |
| Business Area Risk - Risk of business area visibility / damage to user experience                                             | 0     | 0     |
| Developer Fault Risk - How likely will this tech debt cause a future error related to coding on top of it                     | 0     | 0     |
| System Fault Risk - Risk of system errors or application downtime                                                             | 0     | 0     |
| Time Scale Risk - Compound risk effect if left alone. How much more difficult to fix or dangerous will this become over time? | 0     | 0     |
| Time Sink Risk - How much will this tech debt slow the development process down                                               | 0     | 0     |
| <div align="right">**TOTAL SCORE:**                                                                                           | 0     | 0     |

#### Development Checklist:

- [ ] Checklist item
- [ ] Checklist item
- [ ] Checklist item

PR #5004, opened 2026-07-24
Fix/airflow dag time

joshgamache:
Addresses #4911. Sets the start time of Dags to a static time. As Dags are versioned as of Airflow 3, this prevents the version from increasing unnecessarily. 

## Changes 🚧

- Change Dags to use static time (based on initial adding of the Dag, since they are triggered so it's not particularly relevant).
- Run a Ruff check using [AIR ](https://docs.astral.sh/ruff/rules/#airflow-air) linter for Airflow 3 issues and correct them.
- Add Prek check for Dag Ruff check.

## Notes 📝

The issue flagged by 4911 is relevant to most of our Dags in other cas repos. Cloned that issue into Cas-Airflow to cover the rest of the repos as bcgov/cas-airflow#232.

### DIFF
--- a/dags/bc_obps_reset_data.py
+++ b/dags/bc_obps_reset_data.py
@@ -2,7 +2,7 @@ from dag_configuration import default_dag_args
 from trigger_k8s_cronjob import trigger_k8s_cronjob
 from airflow.providers.cncf.kubernetes.operators.job import KubernetesJobOperator
 from airflow.providers.standard.operators.trigger_dagrun import TriggerDagRunOperator
-from datetime import datetime, timedelta, timezone
+from datetime import datetime, timezone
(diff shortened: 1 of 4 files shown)


dags/bc_obps_reset_data.py
```diff
@@ -27,6 +27,7 @@
     dag_id=RESET_DATA_DAG_NAME,
     schedule=None,  # This dag is intended to be run manually
     default_args=default_args,
+    catchup=False,
```
dleard, line 30: 💯

dleard (approved):
Looks good! I had to go do a little reading to make sure I understood what the catchup arg did. Thanks! :shipit:

PR #5033, opened 2026-07-30
chore(ci): raise backend healthcheck budget and unify compose files

Sepehr-Sobhani:
## Why

`test-registration-e2e` flaked with `container bciers-backend-1 is unhealthy`. The backend
runs `custom_migrate` before gunicorn and takes 193–295s on CI, but
`docker-compose-nx-app.yaml` allowed only ~260s (`start_period 60s` + `20 × 10s`). The
worst passing run finished in 255s — a 5s margin.

Fixing that exposed a transfer-form flaky issue that had been hiding behind it.

## Changes

**CI**
- Backend healthcheck budget raised to 480s (`start_period 180s` + `30 × 10s`).
- Added the dashboard/app healthchecks `nx-app.yaml` never had — `up --wait` was returning
  before Next.js was listening.
- Merged the two compose files into `bciers/docker-compose.yaml`, with one profile per app.
  Each job now declares what it needs (`compliance,reporting` instead of the full stack),
  and `docker_compose_file` / `image_url` / `nx_app_port` are gone.
- Fixed each job's `needs` — compliance and reporting pulled all five app images while
  depending only on their own build.
- `DB_PASSWORD` is generated per run and masked, replacing the hardcoded value SonarQube
  flagged.

**Transfer form** — two issues, same cause: async handlers rewriting state after their
fetches resolve.
- Operator changes rebuilt state from a stale snapshot, discarding anything entered during
  the fetches (the e2e flake, and lost input for real users).
- The same-operator error was overwritten by the rebuilt uiSchema, so it flashed and
  vanished.
- `disabled` is now derived instead of stored; out-of-order responses are ignored; all 12
  `any` types replaced.

## ⚠️ Happo baselines need re-approval

App containers now set `CI: "true"` (matching what `nx-app` already did), which hides the
dev-only MockTimePicker in the footer.

- **Will diff:** compliance, reporting
- **Unchanged:** registration, administration, dashboard

No test logic depends on it — baseline approval, not failures.


bciers/apps/registration/app/components/transfers/TransferForm.tsx
```diff
@@ -171,20 +202,23 @@ export default function TransferForm({
...
     transferFormData: TransferFormData,
   ) => {
+    const changeId = ++latestOperationChange.current;
+
     setError(undefined);
 
     const facilitiesRes = await fetchFacilities(
       transferFormData?.from_operation,
     );
+    if (changeId !== latestOperationChange.current) return;
```
pbastia: This feels a little bit like a code smell, react should be tracking the changes so that we don't have to. Are we engineering a race condition?

bciers/docker-compose.yaml
joshgamache, line 1: ❤️ Using compose profiles is such a good solution! I didn't even know they existed until recently and I hadn't put the thought together to use them, great work!

bciers/apps/registration/app/components/transfers/TransferForm.tsx
Sepehr-Sobhani: We did have a race condition when fetching the information, but it was mostly showing up in CI since e2e tests click much faster than real users, causing some flakiness. I initially added the flag to make it more predictable, but after looking at it again today with a fresh mind, I changed the approach and made it synchronous instead of relying on the flag.

Issue #4825, opened 2026-06-18
Reporting/Bug: No error displayed


fviduya:
#### Describe the Bug:

Some pages in reporting does not display an error. These pages are:
1. Person Responsible page - when user tries to save without selecting one. Clicking `Save` does nothing and no error is displayed
2. Review Changes page - when user tries to save without filling out the reason for change field, clicking `Save` does nothing and no error is displayed
3. New Entrant Information page - when user has provided the required dates but has not checked the Assertion statement box, no error or information is displayed and `Save & Continue` button remains greyed out until user confirms (to confirm if this is a bug or expected behaviour)
4. Additional Reporting Data page - when user selects `Yes` to `Did you capture emissions?` and selects more than 1 Capture type, upon clicking `Save` only the first emissions field for the first capture type has an inline error

These were captured during regression testing for #4464 and checked in TEST environment that the same behaviour already exists. [Test document for reference](https://buttoninc-my.sharepoint.com/:w:/g/personal/francesca_button_is/IQDftn_CLpwbT7rUvz_owmHvAVAR43Ii8Pq_CqBZGdLIU7o?e=8luuOQ). 

#### Probability:

(How likely the bug is to happen, scored from 1-5. For example, probability of 5 is something like "it happens to all users every time they log in." whereas probability of 1 "only happens to certain users when a really specific and unlikely path is followed.")

#### Impact:

(How bad the bug is when it does happen, scored from 1-5. For example, effect of 5 is "the entire app crashes and makes it unusable for all users" or "the bug causes the wrong data to be saved, with critical information (e.g. payment) being affected." whereas effect of 1 is "It makes some styling look a little bit weird.")

<img src="https://www.getzephyr.com/sites/default/files/riksk%20matrix.jpg" alt="Bug Matrix" width="250" height="250"/>

#### Steps to reproduce the behaviour:

1. Step #1
2. Step #2
3. Step #3

#### Dev checklist

- [x] Person Responsible Page
- [x] Review Changes Page
- [x] New Entrant Info Page
- [ ] Additional Reporting Data Page 

#### Additional information:

Browser: [e.g. chrome, safari]
Version: [e.g. 22]

#### Additional Context

-

fviduya:
Update: For the Person Responsible page, clicking `Save` results in an error, but the error might still need to be fixed:

<img width="894" height="287" alt="Image" src="https://github.com/user-attachments/assets/a5befeba-37db-4835-b3d3-7d854ecfcc39" />

[...]

PR #5037, opened 2026-07-30
fix: display broken errors on responsible, review, and additional pages

gdalcengio:
resolves: 4825

I got 3 out of the 4 of these errors displayed, however with the additional reporting page, there exists the allOf/if/then rjsf issue we dealt with before. See comment [here](https://github.com/bcgov/cas-registration/issues/4825#issuecomment-5137402298)


bciers/apps/reporting/src/app/components/changeReview/ChangeReviewForm.tsx
```diff
@@ -55,6 +56,12 @@ export default function ChangeReviewForm({
   }, [versionId, displayChanges]);
 
   const handleSubmit = async (canContinue: boolean) => {
+    if (!reasonForChange || reasonForChange === "") {
```
pbastia: empty string is already falsy, we can remove the second check

```suggestion
    if (!reasonForChange) {
```

pbastia (approved):
Looks great! One small nit

